Issue #2145, opened 2014-05-17
Feature Request: XCode Group Support


a34729t:
I'm putting together a cocoapod with around 20 files, some of them extensions, others config files, others real source code. I have them in subfolders, but this distinction doesn't make it's way to XCode when I import the pod. Is there a way to make a folder an XCode group?

What my Cocoapod looks like:

```
/src
  /classes
     Foo.h
     Foo.m
  /extensions
     NSData+Ext.h
     NSData+Ext.m
```

What it looks like in XCode now (I want it to look like the above):

```
 /MyPod (Group)
   Foo.h
   Foo.m
   NSData+Ext.h
   NSData+Ext.m
```

Issue #648, opened 2012-11-13
Folder structure in xcode


Michenux:
I have used header_mappings_dir to keep the folder structure.
If i look inside the Pods/Headers/MyLib folder, the structure is well kept.

But in XCode, the headers structure is still flattened in the group Pods/Local Pods/MyLib.

Is it possible to fix that ?

PR #2647, opened 2014-10-11
[installer] Group support added for Development Pods mirroring the folde...

imihaly:
Hopefully fixes https://github.com/CocoaPods/CocoaPods/issues/2145 by adding a group support for development pods.
It reflects the filesystem structure using subgroups under Development Pods group, thus eases the pod development in tandem with a host application.

Similar complain here: http://stackoverflow.com/questions/21168826/preserve-folder-structure-cocoa-pods and here: https://github.com/CocoaPods/CocoaPods/issues/648

### DIFF
--- a/lib/cocoapods/installer/file_references_installer.rb
+++ b/lib/cocoapods/installer/file_references_installer.rb
@@ -152,20 +152,20 @@ module Pod
       # @param  [Symbol] group_key
       #         The key of the group of the Pods project.
       #
-      # @param  [Bool] group_if_development
+      # @param  [Bool] reflect_file_system_structure_for_development
       #         Wether organizing the a local pod's files in subgroups inside
       #         the pod's group is allowed.
       #
       # @return [void]
       #
-      def add_file_accessors_paths_to_pods_group(file_accessor_key, group_key = nil, group_if_development = false)
+      def add_file_accessors_paths_to_pods_group(file_accessor_key, group_key = nil, reflect_file_system_structure_for_development = false)
         file_accessors.each do |file_accessor|
           pod_name = file_accessor.spec.name
           local = sandbox.local?(pod_name)
           paths = file_accessor.send(file_accessor_key)
           paths.each do |path|
             group = pods_project.group_for_spec(file_accessor.spec.name, group_key)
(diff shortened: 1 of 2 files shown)


iosdev-republicofapps:
Thanks @orta .  Does `pod lib create` now include this group in the template Xcode project created?

orta:
It would, if you were using the 0.36 beta version.

danflynn76:
This is a great feature and I installed 0.36 beta, so I could use it with a pod I'm working on.  Is there a good reason to not offer this feature as an option for installing pods post-development?  When I'm using an open source project I often dig into the source code to understand what's going on and preserving the author's folder hierarchy would be very valuable for large projects.  Would anyone else see value in a flag like :preserve_hierarchy => true ?

onmyway133:
Yeah, 0.36 works ^^

mhuusko5:
Could we please have a flag to disable this new behaviour? It's not universally desirable...

jpshelley:
@imihaly @alloy This seems awesome, but its not documented anywhere, and took forever to find out about it. And I still don't know how to use it. Is there a flag I need to add to my podspec to enforce `groups`?

Every time I run a `pod update` my groups get flattened and its getting annoying.

orta:
it has to be a `:path` pod, but that's about it, it's been the default now for a few versions.

jpshelley:
@orta I have that in my Example Podfile..that doesn't persist the `groups` created in xcode though.

orta:
they are based off the folder structure, not your internal Xcode settings

jpshelley:
Thanks @orta!

segiddins:
Please don't comment on a PR that's been closed for half a year.

jamessmith-airside:
There seems to be a bug where this feature only works in a dev pod if there is a file at the root dev pod directory. If the root directory contains nothing but the podspec file and sub-folders, pod update flattens the hierarchy in Xcode. But if I place just one other file in the root dev pod directory and run pod update, my disk's file structure is properly reflected in Xcode.

I'm just going to work around the issue, but if anyone wants to file a bug report there's your steps to reproduce.

Issue #2402, opened 2014-08-19
Cannot analyze since Xcode 6 beta 5, get "error reading 'pic'" and "no analyzer checkers are associated with '-mrelocation-model'"


dburr:
Now that iOS 8 and OS X 10.10 releases are right around the corner, I thought I'd better start testing my various apps against the new stuff.  Unfortunately I've run into a problem.  Building projects works fine, however starting with Beta 5, I am no longer able to analyze projects with Cocoapods in them.  When analyzing, the process stops with the following errors:

```
Analyze AFNetworking/AFNetworking/AFHTTPClient.m
    cd /Users/dburr/Developer/iOS/CrApp.net/Pods
    export LANG=en_US.US-ASCII
    export PATH="/Applications/Xcode6-Beta6.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/usr/bin:/Applications/Xcode6-Beta6.app/Contents/Developer/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin"
[...]

PR #2816, opened 2014-11-12
Use `-analyzer-disable-all-checks` for disabling static analysis.

kommen:
`-analyzer-disable-checker` without arguments causes an error in the
static analyzer, but due to a bug in Xcode < 6.0 it was silently
ignored. As of Xcode 6.1 `-analyzer-disable-all-checks` can be used
to disable static analysis for specfic files.

rdar://18907351
https://github.com/CocoaPods/CocoaPods/issues/2402

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ## Master
 
+##### Enhancements
+
+* Use `-analyzer-disable-all-checks` to disable static analyzer for
+  for pods with inhibits_warnings on. (requires Xcode >= 6.1)  
+  [Dieter Komendera](https://github.com/kommen)
+  [#2402](https://github.com/CocoaPods/CocoaPods/issues/2402)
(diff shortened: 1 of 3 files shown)


scottknight:
Any update on this? It's now 6 months later, I would imagine a new big Xcode update will be coming with WWDC. Right now after every `pod install` I have to remember to go switch the files the need to use `-analyzer-disable-all-checks`

allewun:
Seconded, the older Xcode versions are becoming increasingly obsolete and the lack of the new `-Xanalyzer -analyzer-disable-all-checks` flags is a setback for current Xcode versions. Would appreciate another look at this.

For anyone stumbling across this, the workaround until this PR is merged is to manually add the flags to the files in the `post_install` hook. Personally, I found and modified [this](http://www.egeek.me/2013/05/25/podfiles-post-install-hook-example/) and found it sufficient, but it looks like the scripts mentioned [here](https://github.com/CocoaPods/CocoaPods/issues/2402) will also work.

PR #2892, opened 2014-11-29
[CHANGELOG] Add blog links

AliSoftware:
Add links to each detailed blog article for each "major release" entry in the `CHANGELOG.md`.
_(as approved by @alloy on Slack)_

This also allows users to discover the blog for those who don't know about it yet.


AliSoftware:
It seems that the GitHub README renderer does not take :memo: smiley into account?!
(Totally defeats its emphasis purpose)

https://github.com/CocoaPods/CocoaPods/blob/master/CHANGELOG.md#0350

BTW, we forgot to add that link for that 0.36 entry

Issue #2914, opened 2014-12-07
Linter should check for .m files in the public_headers


orta:
E.g. https://github.com/CocoaPods/Specs/blob/master/Specs/ActionSheetPicker/1.0.6/ActionSheetPicker.podspec.json

```
  "public_header_files": [
    "ActionSheetPicker.h",
    "Pickers/*.{h,m}"
  ],
```

It's not been noticeable in library CocoaPods but in Swift pods this breaks the library.

PR #2917, opened 2014-12-07
Validate that a specification's `public_header_files` and `private_header_files` file patterns only match header files.

segiddins:
Also, validate that all file patterns, if given, match at least one file.

Closes https://github.com/CocoaPods/CocoaPods/issues/2914.


reidmain:
So I would contest that ensuring that at least one file is matched is a bad decision because at my company we maintain a plethora of private pods and we make sure that the private_header_files attribute is defined for every module so that if any developer adds a file that ends in +Internal.h it is automatically flagged as a private header file.

After this change it is now the responsibility of the developer to realize they are adding the first +Internal.h header to a pod and that they must go and figure out what the value of the private_header_files attribute should be.

Issue #2949, opened 2014-12-17
Update will always download git pod, regardless of changes. 


markst:
If I have a pod specified as:
`pod 'RestKit', :git => 'git@github.com:Papercloud/RestKit.git', branch: 'master'`

`pod update` will always pre-download, regardless of whether there have been changes on the master, even if I specify a particular commit.

Can cocoapods be more intelligent & identify whether the commit id has changed & only download if necessary?

krzyzanowskim:
I have somehow related issue, but `pod update` don't fetch fresh version and keep using the old one.
Sequence is:
1. pod install
2. update git repo
3. pod update (changes not fetched)

`pod 'ObjectivePGP', :git => 'https://github.com/krzyzanowskim/ObjectivePGP.git', :branch => :master`

Fetching external sources
-> Pre-downloading: `ObjectivePGP` from `https://github.com/krzyzanowskim/ObjectivePGP.git`, commit `c4652b63f5c9ac998c785590f09e697bb89a2ac8`

while there is newer commit available.

$ pod --version
0.35.0

kylef:
@krzyzanowskim That sounds like a bug introduced in 0.35, could you file a separate issue for this?

krzyzanowskim:
#2997

kylef:
@krzyzanowskim thank you

segiddins:
We can use `git ls-remote URL #{options[:branch] || 'HEAD'}` to figure out whether the SHA has actually changed

segiddins:
I'm going to close this as not worth changing. A path implementing this would be much appreciated, but the difficulty of getting this right probably isn't worth the slight UX improvement.

markst:
this is supposedly resolved with https://github.com/CocoaPods/CocoaPods/pull/3236
But this is still a major issue for me. Admittedly my internet isn't the greatest, but my pods take forever to update when there is no actual change on the git repo.

segiddins:
@markst as I said, I'd love to see a patch implementing this behavior, but unless it's contributed by a member of the community, I just don't think it's worth the core team's time.

danielgalasko:
Pity this was closed, would love to submit a PR if I could get pointed in the right direction. We use SHAs for our internal pods and it 'pre-downloads' them everytime we run pod update.

segiddins:
@danielgalasko we'd still love to see a PR for this! As I posted above, using `git ls-remote` and comparing that SHA with what's in the lockfile would be a reasonable way to go

PR #2959, opened 2014-12-20
Don't refresh externals deps when pinned to a commit we already have

tspacek:
First draft at addressing https://github.com/CocoaPods/CocoaPods/issues/2949

### DIFF
--- a/lib/cocoapods/installer/analyzer.rb
+++ b/lib/cocoapods/installer/analyzer.rb
@@ -1,3 +1,5 @@
+require 'pry-byebug'
+
 module Pod
   class Installer
     # Analyzes the Podfile, the Lockfile, and the sandbox manifest to generate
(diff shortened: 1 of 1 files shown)


kylef:
@tspacek Would you be able to add specs for this?

lib/cocoapods/installer/analyzer.rb
```diff
@@ -304,6 +304,7 @@ def dependencies_to_fetch
 
           if update_mode == :all
             deps_to_fetch = deps_with_external_source
+            deps_to_fetch -= deps_to_fetch.select{ |dep| !dep.external_source[:commit].nil? && !checkout_requires_update?(dep) }
           else
             deps_to_fetch = deps_with_external_source.select { |dep| pods_to_fetch.include?(dep.name) }
```
segiddins: the logic also needs to be duplicated here

tspacek:
Yup I'll add some specs

orta:
^ ping

orta:
( don't want to be _too_ annoying, but would rather this didn't become a hung PR :) )

segiddins:
I think this won't really be necessary once we have pod caching in place.

kylef:
Thanks for your pull request and contribution @tspacek, I'm going to close this since it will be no longer necessary with #3236.

PR #2979, opened 2014-12-26
[PodspecSource] Rescue HTTPError exceptions and handle them

k0nserv:
This is might suggestion for how to handle this. I am not sure about what the correct exit procedure is, but this does the job. Please correct me if there is a specific way to do it.

I haven't added tests or a changelog entry just yet in case this is not the way we wanna deal with this issue

Closes: https://github.com/CocoaPods/CocoaPods/issues/2823

### DIFF
--- a/lib/cocoapods/external_sources/podspec_source.rb
+++ b/lib/cocoapods/external_sources/podspec_source.rb
@@ -15,7 +15,7 @@ module Pod
             open(podspec_uri) { |io| store_podspec(sandbox, io.read, is_json) }
           rescue OpenURI::HTTPError => e
             status = e.io.status.join(' ')
-            raise Informative, "Failed to fetch podspec for `#{name}` at #{podspec_uri}. Error: #{status}"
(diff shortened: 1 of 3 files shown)


k0nserv:
Updated with test.

lib/cocoapods/external_sources/podspec_source.rb
```diff
@@ -11,7 +11,12 @@ def fetch(sandbox)
         UI.titled_section(title,  :verbose_prefix => '-> ') do
           is_json = podspec_uri.split('.').last == 'json'
           require 'open-uri'
-          open(podspec_uri) { |io| store_podspec(sandbox, io.read, is_json) }
+          begin
+            open(podspec_uri) { |io| store_podspec(sandbox, io.read, is_json) }
+          rescue OpenURI::HTTPError => e
+            status = e.io.status.join(' ')
+            raise Informative, "Failed to fetch podspec for `#{name}` at #{podspec_uri}. Error: #{status}"
```
segiddins: Podspec uri in back ticks. Error on new line.

kylef:
Merged via 6f547d2

Issue #2946, opened 2014-12-15
Run `pod repo setup` if no explicit sources are defined.


DanielTomlinson:
As implicit sources have been undeprecated, we should run `pod repo setup` if no explicit sources are defined to ensure that people using existing projects aren't forced to update podfiles on new machines.

related: https://github.com/CocoaPods/CocoaPods/pull/2945

PR #2998, opened 2015-01-03
[Podfile] Change the default spec repository to the master

kylef:
Along with fixing #2946, I think it makes more sense if by default. That only the master spec repository is used. If a user wants a custom repository they should have to explicitly specified this in their Podfile.

This change means that a Podfile on one system will have the exact same dependencies on another. Currently you could consume another Podfile and have different dependencies since by default, the Podfile would pick up all sources which might change the dependencies.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ## Master
 
+##### Breaking
+
+* Changes the default spec repositories used from all configured spec
+  repositories, to the master spec repository when no spec repositories
+  are explicitly configured in a Podfile.  
+  [Kyle Fuller](https://github.com/kylef)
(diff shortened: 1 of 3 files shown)


kylef:
Note: Integration specs need regenerating because the output included warnings from using all repos:

```diff
-[!] Found multiple specifications for `JSONKit (1.4)`:
-- ROOT/spec/fixtures/spec-repos/master/Specs/JSONKit/1.4/JSONKit.podspec.json
-- ROOT/spec/fixtures/spec-repos/test_repo/JSONKit/1.4/JSONKit.podspec
```

segiddins:
@alloy, this one's for you.
Everyone knows my position, which was for requiring specification of sources at all times.

segiddins:
@kylef please rebase, @orta gave his :+1:

PR #2971, opened 2014-12-24
added '$(inherited)' to OTHER_LDFLAG (tested) - ...

Speakus:
resolve part of request https://github.com/CocoaPods/CocoaPods/issues/546 which was missed in the https://github.com/CocoaPods/CocoaPods/issues/2467

Issue #2467, opened 2014-09-15
Allow custom xcconfigs to extend definitions in generated files


fabiopelosin:
## Goal

Allow users to use custom xcconfig files, which import the generated xcconfigs and make further definitions based on the values contained in those, as injecting manual integrated dependencies.

This is based on a previous disucssion in https://github.com/CocoaPods/CocoaPods/issues/1212.
## Tasks
- [ ] namespace the build settings in user / aggregate target xcconfigs to allow to easily override them
- [ ] define a base xcconfig to inherit the common settings of the installation to reduce SCM noise
  - [ ] `PODS_ROOT`
  - [ ] `GCC_PREPROCESSOR_DEFINITIONS`, `OTHER_SWIFT_FLAGS`
  - [ ] `LD_RUNPATH_SEARCH_PATHS`, `EMBEDDED_CONTENT_CONTAINS_SWIFT`
- [x] append `$(inherited)` to the value of each setting
## Example

```
HEADER_SEARCH_PATHS  = "${PODS_ROOT}/Headers"
```

should be:

```
PODS_HEADER_SEARCH_PATHS = "${PODS_ROOT}/Headers"
HEADER_SEARCH_PATHS = ${PODS_HEADER_SEARCH_PATHS}
```

segiddins:
Note that we can't add `$(inherited)` to singular build settings -- only the plural ones

nevir:
FWIW, here's a bit of a hacktastic workaround:

```rb
# CocoaPods-managed variables that you should support.
#
# If CocoaPods adds one not in this list, an error will be thrown to let you know.
EXPECTED_KEYS = ['FRAMEWORK_SEARCH_PATHS', 'GCC_PREPROCESSOR_DEFINITIONS', 'HEADER_SEARCH_PATHS', 'OTHER_CFLAGS', 'OTHER_LDFLAGS', 'PODS_ROOT'].sort

post_install do |installer|
  driver_pod = installer.aggregate_targets.find { |t| t.label == 'Pods-Driver' }
  driver_pod.native_target.build_configurations.each do |config|
    config_path = driver_pod.xcconfig_path(config.name)
    config_data = config_path.read

    keys = config_data.each_line.map {|l| l.sub(/(^.*?)\s*=.*$/, '\1').strip }
    unknown_keys = keys - EXPECTED_KEYS
    if unknown_keys.size > 0
[...]

Issue #546, opened 2012-09-22
Pods.xcconfig should include $(inherited)


nschum:
Is there a reason Pods.xcconfig doesn't use $(inherited) for properties like OTHER_CFLAGS, OTHER_LDFLAGS, FRAMEWORK_SEARCH_PATHS or LIBRARY_SEARCH_PATHS?

Currently it overwrites settings made on the project level. That seems wrong, as the user might want to add settings for all targets there.

PR #3027, opened 2015-01-10
prepended $(inherited) to OTHER_LDFLAGS and HEADER_SEARCH_PATHS to the generated xcconfig

patcheng:
that way, settings from Project level in xcodeproj could be inherited.

### DIFF
--- a/lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
+++ b/lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
@@ -49,7 +49,7 @@ module Pod
         def generate
           pod_targets = target.pod_targets_for_build_configuration(@configuration_name)
           config = {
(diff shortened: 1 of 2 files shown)


patcheng:
it has the same change for OTHER_LDFLAGS as PR https://github.com/CocoaPods/CocoaPods/pull/2971

Looks like the issue is mentioned in issue https://github.com/CocoaPods/CocoaPods/issues/546 and https://github.com/CocoaPods/CocoaPods/issues/2467

segiddins:
@samdmarshall can you think of any reason not to add `$(inherited)` to `OTHER_LDFLAGS` / `HEADER_SEARCH_PATHS`?

segiddins:
Can confirm that the only failures are expected changes to the integration specs.

samdmarshall:
I don't think there is any reason not to add it.

hypotamtam:
When do you think that it will be release?

Issue #2558, opened 2014-09-30
Local (non-git) private repository not working on 0.34.X


sabadow:
I have a private repository on my local machine (it's a enterprise SVN repo) defined on ~/.cocoapods/repos.

````
pod repo list

arcadia-repo
- Type: local copy
- Path: /Users/cperez/.cocoapods/repos/arcadia-repo

master
- Type: git (origin)
- URL:  https://github.com/CocoaPods/Specs.git
- Path: /Users/cperez/.cocoapods/repos/master
````
It works correctly on all versions including 0.33.1, but not in 0.34.1.

When I try to make a `pod install` this error appears:

````
pod install
Analyzing dependencies
fatal: No remote configured to list refs from.
[!] The `arcadia-repo` repo is not a git repo.
````

Thats right, the repo isn't a git repo. It's a new requirement for local repositories?
Thanks in advance.

kylef:
https://github.com/CocoaPods/CocoaPods/pull/2973 does just that @neonichu.

clarkda:
@kylef i'm still having the same issue with `pod install` from non-git repositories -- can we reopen this issue?

i'll do a PR in just a few

kylef:
@clarkda Do you know if this has been resolved?

PR #3051, opened 2015-01-20
[Analyzer] Only update git sourced spec repos

clarkda:
Updated Analyzer::update_repositories_if_needed to only update a spec repository if it is a git sourced repository.

This is a proposed fix for #2558

/cc: @neonichu @kylef

### DIFF
--- a/lib/cocoapods/installer/analyzer.rb
+++ b/lib/cocoapods/installer/analyzer.rb
@@ -163,12 +163,12 @@ module Pod
         end
       end
 
-      # Updates the source repositories unless the config indicates to skip it.
+      # Updates the git source repositories unless the config indicates to skip it.
       #
       def update_repositories_if_needed
         unless config.skip_repo_update?
           UI.section 'Updating spec repositories' do
-            sources.each { |source| SourcesManager.update(source.name) }
+            sources.each { |source| SourcesManager.update(source.name) if SourcesManager.git_repo?(source.repo) }
           end
         end
       end
--- a/lib/cocoapods/installer/analyzer.rb
+++ b/lib/cocoapods/installer/analyzer.rb
@@ -172,7 +172,7 @@ module Pod
               if SourcesManager.git_repo?(source.repo)
                 SourcesManager.update(source.name)
               else
-                UI.warn "Skipping `#{source.name}` update because the repository is not a git source repository."
+                UI.message "Skipping `#{source.name}` update because the repository is not a git source repository."
               end
             end
           end
--- a/lib/cocoapods/installer/analyzer.rb
+++ b/lib/cocoapods/installer/analyzer.rb
@@ -168,7 +168,13 @@ module Pod
       def update_repositories_if_needed
         unless config.skip_repo_update?
           UI.section 'Updating spec repositories' do
-            sources.each { |source| SourcesManager.update(source.name) if SourcesManager.git_repo?(source.repo) }
+            sources.each do |source|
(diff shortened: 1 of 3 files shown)


segiddins:
:+1:

clarkda:
@segiddins RE: your comment in my other PR. 

> Just needs test coverage. Otherwise, :+1: maybe a verbose message when skipping an update would be nice too?

I feel like that might be too verbose? But at the same time a huge drawback of the hg/svn plugins is having to manually keep the spec-repos updated -- and I probably need to make that point more clear

segiddins:
No, I agree with my past comment :)

-Samuel E. Giddins

On Jan 19, 2015, at 11:13 PM, Dustin Clark <notifications@github.com> wrote:

@segiddins RE: your comment in my other PR.

Just needs test coverage. Otherwise,  maybe a verbose message when skipping an update would be nice too?

I feel like that might be too verbose? But at the same time a huge drawback of the hg/svn plugins is having to manually keep the spec-repos updated -- and I probably need to make that point more clear

clarkda:
@segiddins :wink: I've added a little bit of verbosity, but a few of my tests are failing..

Looking at recent commits this one could actually be a problem: https://github.com/CocoaPods/CocoaPods/commit/0994612ebb65ad1aa221f384235e4702560df7d2#diff-f5c3fee776e864ada6541431a1532b00L468

At first glance it looks like this breaks local spec repositories --  as we can't specify a `source` in the `podspec` if the `source` has no associated URL?

@kylef Does that sound right? What are your thoughts?

clarkda:
Opened https://github.com/CocoaPods/CocoaPods/issues/3061

segiddins:
@clarkda `source` goes in the Podfile, not in the Podspec :)

clarkda:
Doh, typo. So many pods, specs, and sources. I'm easily confused. :cake:

clarkda:
@segiddins verbosity!

```
chunk:pod-tests dusty$ pod install
Analyzing dependencies
Downloading dependencies
Using secretsquirrel-local (0.3)
Generating Pods project
Integrating client project

[!] Skipping `master-local` update because the repository is not a git source repository.
```

lib/cocoapods/installer/analyzer.rb
```diff
@@ -163,12 +163,18 @@ def generate_podfile_state
...
       #
       def update_repositories_if_needed
         unless config.skip_repo_update?
           UI.section 'Updating spec repositories' do
-            sources.each { |source| SourcesManager.update(source.name) }
+            sources.each do |source|
+              if SourcesManager.git_repo?(source.repo)
+                SourcesManager.update(source.name)
+              else
+                UI.warn "Skipping `#{source.name}` update because the repository is not a git source repository."
```
segiddins: I'd personally use a `UI.message`

clarkda: Are those only shown with --verbose?

segiddins: Yes, which I believe is what we want in this case. @kylef ?

clarkda: Yeah, that's way better. Thanks @segiddins!

segiddins:
One last thing, and I promise we can merge -- add a test that the message is shown in verbose mode?

clarkda:
@segiddins This look okay?

segiddins:
Yup! @kylef giving you a few hours to take a look before I merge. Thanks so much @clarkda!

PR #3069, opened 2015-01-22
[ErrorReport] Show full command in the error report

segiddins:


orta:
I like this idea :+1:

segiddins:
@kylef let me know if this looks good

lib/cocoapods/user_interface/error_report.rb
```diff
@@ -137,6 +143,10 @@ def repo_information
             end
           end
         end
+
+        def original_command
+          "#{$0} #{ARGV.join(' ')}"
```
kylef: Can we switch `$0` to `$PROGRAM_NAME` instead

kylef:
Good to merge once feedback is addressed :+1:

PR #3072, opened 2015-01-25
implemented Xcode license check for `pod setup`

### LINKED ISSUE #3002: Won't running pod setup with sudo create some root-owned stuff in my user-dir?
```
$ pod setup

"Agreeing to the Xcode/iOS license requires admin privileges, please re-run as root via sudo."
```

This seems like a bad suggestion to me. Won't it create a ~/.cocoapods that is owned by root?

In my case, I just launched Xcode and agreed to the ULA, and then re-ran pod setup.

CalQL8ed-K-OS:
fixes #3002 

reviews
- [ ] @kylef 
- [ ] @neonichu 
- [ ] @grav

it's my first PR, so please offer any advice and suggestions you might have.
although functional, this probably could use some refactoring.
i also had no idea how to write a test for this, especially since i don't currently have a machine where i haven't accepted the license agreement.

### DIFF
--- a/lib/cocoapods/command.rb
+++ b/lib/cocoapods/command.rb
@@ -41,6 +41,7 @@ module Pod
 
     def self.run(argv)
       help! 'You cannot run CocoaPods as root.' if Process.uid == 0
+      verify_xcode_license_approved!
 
       super(argv)
       UI.print_warnings
@@ -120,6 +121,25 @@ module Pod
       unless config.lockfile
         raise Informative, "No `Podfile.lock' found in the project directory, run `pod install'."
       end
+    end
+
+    def self.verify_xcode_license_approved!
+      unless xcode_license_approved?
+        raise Informative, xcode_license_message
+      end
+    end
+
+    def self.xcode_license_approved?
+      !(`/usr/bin/xcrun clang 2>&1` =~ /license/ && !$?.success?)
+    end
+
+    def self.xcode_license_message
+      <<-EOS.strip_heredoc
+          You have not agreed to the Xcode license.
+          The setup command requires you do so.
+          Agree to the license by running:
+          `xcodebuild -license`
+      EOS
     end
   end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -134,6 +134,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
+* Check for Xcode License Agreement before running commands.
+  [Xavi Matos](https://github.com/CalQL8ed-K-OS)
+  [#3002](https://github.com/CocoaPods/CocoaPods/issues/3002)
+
 * Do not generate targets for Pods without sources.  
   [Boris Bügling](https://github.com/neonichu)
   [#2918](https://github.com/CocoaPods/CocoaPods/issues/2918)


segiddins:
Please see your test failures at https://travis-ci.org/CocoaPods/CocoaPods/jobs/48223521#L2177.

CalQL8ed-K-OS:
i was in the processing of installing rubocop to check, didn't expect someone to response that quick, props :)
addressed the issues just now.

CalQL8ed-K-OS:
¡hurray, passing tests!
locally i'm not getting passing tests though.
¿can anyone help me figure out what's wrong locally?
[Output of `bundle exec rake spec`](https://gist.github.com/CalQL8ed-K-OS/248732d2b4004de4d7d5)
it seems to just be a bunch of _undefined method_ errors on __ Pod::SourcesManager:Class__

kylef:
I don't think this should run only for `pod setup`, but instead for every invocation of `pod` (unless is slows down CocoaPods noticeably).

segiddins:
`xcrun clang  0.00s user 0.01s system 29% cpu 0.037 total`

CalQL8ed-K-OS:
looks like it's speed enough.
¿where should i move the method and method call to preempt any pod command?
¿anyone got any advice on my local test failures?

kylef:
@CalQL8ed-K-OS Hey, I think somewhere around https://github.com/CocoaPods/CocoaPods/blob/master/lib/cocoapods/command.rb#L43 would be best.

lib/cocoapods/command/setup.rb
```diff
@@ -44,6 +46,26 @@ def run
         UI.puts 'Setup completed'.green
       end
 
+      def check_xcode_license_approved
+        unless xcode_license_approved?
+          UI.puts xcode_license_message.red
```
segiddins: `raise Informative, xcode_license_message` ?

CalQL8ed-K-OS: - ¿do you propose adding that line, or replacing the `UI.puts` call?
- ¿what's the ramifications of that line? ¿who catches it?

CalQL8ed-K-OS:
@kylef , i've implemented the check for all commands now i believe.

i still have local failures when running `rake spec`.
at first glance they all seem to share the error message `NoMethodError: undefined method 'find_or_create_source_with_url' for Pod::SourcesManager:Class`.

i also still have no idea how to really test that this feature works since i don't have a machine without the license accepted.

lib/cocoapods/command.rb
```diff
@@ -121,5 +122,25 @@ def verify_lockfile_exists!
         raise Informative, "No `Podfile.lock' found in the project directory, run `pod install'."
       end
     end
+
+    def self.verify_xcode_license_approved!
+      unless xcode_license_approved?
+        UI.puts xcode_license_message.red
```
segiddins: `raise Informative, xcode_license_message`

CalQL8ed-K-OS: i'm not sure about this, it makes for uglier output in my opinion.

### neat
screen shot 2015-02-13 at 10 00 36 am


### ugly
screen shot 2015-02-13 at 10 05 38 am

orta: Yeah,  I prefer @CalQL8ed-K-OS it's not an assertion where someone needs the additional stack trace

segiddins: It only shows the backtrace in development mode...

orta: Ah, then probably better to have it in then :dancers:

CalQL8ed-K-OS: ahha @segiddins, that's what i was missing :)
will change now

CalQL8ed-K-OS:
just tried testing this on a fresh machine, but i can't.
you need to accept the license to do almost anything from the command line in the Rainforest setup process, including `git clone`

lib/cocoapods/command.rb
```diff
@@ -121,5 +122,24 @@ def verify_lockfile_exists!
...
+        raise Informative, xcode_license_message
+      end
+    end
+
+    def self.xcode_license_approved?
+      !(`/usr/bin/xcrun clang 2>&1` =~ /license/ && !$?.success?)
+    end
+
+    def self.xcode_license_message
+      <<-EOS.strip_heredoc
```
segiddins: why all the newlines in here?

CalQL8ed-K-OS: just making the message easier to read IMO

kylef: Messages are already word wrapped when displayed so this is unnecessary.

CalQL8ed-K-OS: okay, i don't agree, but i'll yield that over, no biggie.

segiddins:
Thanks so much @CalQL8ed-K-OS!

orta:
:star2:

PR #3188, opened 2015-02-24
Set SKIP_INSTALL=YES for all generated targets

### LINKED ISSUE #3007: 0.36.0.beta.1 breaks archiving project as Mac App Archive
#### Summary
After installation of 0.36.0.1.beta I noticed being unable to archive a Mac App project as a Mac application. Xcode just offers me "Generic Xcode Archive".

#### Steps to Reproduce
* create an empty "Cocoa Application" project
* do nothing but archiving of the product. As you can see, Xcode offers a valid *Mac App Archive*:

valid Archive Type

* close this project and create a Podfile within the project folder using this example (you can choose any pod):

```
platform :osx, "10.9"
xcodeproj "Test"

pod 'CCNStatusItemView'
```
* now fire `pod install`
* open the workspace
* archive again

#### Expected Results
The same valid *Mac App Archive* as shown above.

#### Actual Results
I can just archive as *Generic Xcode Archive*:

genericxcodearchive

mrackwitz:
Fixes #3007.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,11 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
+* Set `SKIP_INSTALL=YES` for all generated targets to avoid producing
+  *Generic Xcode Archives* on Archive.  
+  [Marius Rackwitz](https://github.com/mrackwitz)
+  [#3188](https://github.com/CocoaPods/CocoaPods/issues/3188)
+
 * Added support for .tpp C++ header files in specs (previously were getting
   filtered out and symlinks wouldn't get created in the Pods/Headers folder.)  
   [Honza Dvorsky](https://github.com/czechboy0)
--- a/lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
+++ b/lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
@@ -53,6 +53,7 @@ module Pod
             'OTHER_LIBTOOLFLAGS' => '$(OTHER_LDFLAGS)',
             'PODS_ROOT' => target.relative_pods_root,
             'GCC_PREPROCESSOR_DEFINITIONS' => '$(inherited) COCOAPODS=1',
+            'SKIP_INSTALL' => 'YES',
           }
 
           if target.requires_frameworks?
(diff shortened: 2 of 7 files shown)


segiddins:
Is there a chance this will have unintended side effects when using Pods in a framework, or something?

mrackwitz:
This could influence every build script, which rely on the build paths of generated build products.
If [`SKIP_INSTALL=YES`](https://developer.apple.com/library/ios/documentation/DeveloperTools/Reference/XcodeBuildSettingRef/1-Build_Setting_Reference/build_setting_ref.html#//apple_ref/doc/uid/TP40003931-CH3-SW36) and [`DEPLOYMENT_LOCATION=YES`](https://developer.apple.com/library/ios/documentation/DeveloperTools/Reference/XcodeBuildSettingRef/1-Build_Setting_Reference/build_setting_ref.html#//apple_ref/doc/uid/TP40003931-CH3-SW37), which is implied by [`ACTION=install`](https://developer.apple.com/library/ios/documentation/DeveloperTools/Reference/XcodeBuildSettingRef/1-Build_Setting_Reference/build_setting_ref.html#//apple_ref/doc/uid/TP40003931-CH3-SW45), which is the verb given on the CLI to xcodebuild, then the build product is placed into `$TARGET_TEMP_DIR/UninstalledProducts`.
Usually, this happens for a framework or library target only when the target is build as an implicit target dependency of an app target, which the user archives from Xcode.
This [does **not** apply](https://github.com/CocoaPods/cocoapods-packager/blob/master/lib/cocoapods-packager/builder.rb#L172) for [cocoapods-packager](/CocoaPods/cocoapods-packager).

segiddins:
@mrackwitz I'm not sure you've answered my question, which is, will this break things when you use CocoaPods with a framework target?

mrackwitz:
It could because the AggregateXCConfig is shared with the user target.
But it's only a matter if the user actually want to install the framework to the system's framework paths (`$(LOCAL_LIBRARY_DIR)/Frameworks`) or use `INSTALL_PATH` in a different way, which applies probably only in rare cases.
We could avoid that by inlining the build setting to Pods.xcodeproj.

segiddins:
So are you saying I shouldn't merge this PR now?

CHANGELOG.md
```diff
@@ -67,6 +67,11 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
+* Set `SKIP_INSTALL=YES` for all generated targets to avoid producing
+  *Generic Xcode Archives* on Archive.
```
segiddins: two spaces

segiddins:
@kylef I'll merge if this gets your :+1:

PR #3218, opened 2015-03-03
[IntegrationSpec] Rebuild for 0.36.0.rc.1

mrackwitz:


segiddins:
Why is this necessary?

mrackwitz:
Just to bump the version number, it is ignored in the diffs, but still present in files. If I rebuild it for a an intentional change, I want to commit only the intended changes. So by eagerly updating the version number, we ensure to have a cleaner history.

segiddins:
:/ Not really necessary imho, but go ahead and push this sort of non-functional change yourself. No need for a PR.

mrackwitz:
Pushed to master after Travis build succeed.

PR #3237, opened 2015-03-08
Adds concurrent operations

segiddins:
This is a total proof-of-concept at this point, but I'd really appreciate implementation help!


orta:
Oh wow, didn't know this existed already. 

So, roughly, do you have a sense of how many connections we should run parallel? Anyone know if there's a limit on github for simulations connections?

lib/cocoapods/command/repo/add.rb
```diff
@@ -38,12 +38,12 @@ def run
           prefix = @shallow ? 'Creating shallow clone of' : 'Cloning'
           UI.section("#{prefix} spec repo `#{@name}` from `#{@url}`#{" (branch `#{@branch}`)" if @branch}") do
             config.repos_dir.mkpath
-            Dir.chdir(config.repos_dir) do
+            Pod.chdir(config.repos_dir) do
```
kylef: I don't think this chdir is necessary, git checkout takes a directory to check out as it's last argument. Let's just make this the absolute path, something like:

```ruby
command = "clone '#{@url}' #{config.repos_dir + @name}"
```

lib/cocoapods/downloader.rb
```diff
@@ -57,6 +57,10 @@ def ui_sub_action(message)
         def ui_message(message)
           UI.puts message
         end
+
+        def chdir_monitor
```
kylef: This method doesn't appear to be used, can it be removed?

segiddins: It's used by the downloader

lib/cocoapods/sources_manager.rb
```diff
@@ -206,18 +206,16 @@ def update(source_name = nil, show_output = false)
...
-                output = git!('pull --ff-only')
-                UI.puts output if show_output && !config.verbose?
-              rescue Informative
-                UI.warn 'CocoaPods was not able to update the ' \
-                  "`#{source.name}` repo. If this is an unexpected issue " \
-                  'and persists you can inspect it running ' \
-                  '`pod repo update --verbose`'
-              end
+            begin
+              output = git!("--git-dir=#{(source.repo + '.git').to_s.shellescape} --work-tree=#{source.repo.to_s.shellescape} pull --ff-only")
```
kylef: Maybe it would make sense to unify out `git!` usage and make `git!` deal with `--git-dir`.

```ruby
git!('pull --ff-only', source.repo)
```

Then all the other git usages will not need to chdir either.

lib/cocoapods/downloader.rb
kylef: :white_check_mark:

lib/cocoapods/sources_manager.rb
segiddins: Sure, but then we'd need a custom way to create the `git` and `git!` methods that don't come directly from `Executable`

orta:
I'd like to see something like this:

```
$ pod install

Analyzing dependencies
Downloading dependencies
 ├ ⣻ Installing IGHTMLQuery 0.5.1 (was 0.5.1)
 └ ⡿ Installing PutioKit 0.1 (was 0.1)

Using AFNetworking (1.3.3)
Using ARAnalytics (2.8.0)
Using BPXLUUIDHandler (0.0.1)
Using MSVCLeakHunter (1.0.0)
Using Mixpanel (2.5.4)
Using NSData+Base64 (1.0.0)
Using OROpenSubtitleDownloader (1.0.0)
Using SEHumanizedTimeDiff (0.0.1)
Using TestFlightSDK (3.0.2)
Using xmlrpc (2.3.3)
Generating Pods project
Integrating client project
```

The spinners would cycle through `⣾ ⣽ ⣻ ⢿ ⡿ ⣟ ⣯ ⣷`

In _an ideal world_, when the download has finished the confirmation text would then get inserted back at the right alphabetical order. Making it very similar to how it is now:

```
Analyzing dependencies
Downloading dependencies
Using AFNetworking (1.3.3)
Using ARAnalytics (2.8.0)
Using BPXLUUIDHandler (0.0.1)
Installed IGHTMLQuery 0.5.1 (was 0.5.1)
Using MSVCLeakHunter (1.0.0)
Using Mixpanel (2.5.4)
Using NSData+Base64 (1.0.0)
Using OROpenSubtitleDownloader (1.0.0)
Installed PutioKit 0.1 (was 0.1)
Using SEHumanizedTimeDiff (0.0.1)
Using TestFlightSDK (3.0.2)
Using xmlrpc (2.3.3)
Generating Pods project
Integrating client project
```

Unsure if we can do that though.

segiddins:
I think we're going to punt on this for the time being, since download caching already provides an enormous speed boost.

orta:
:+1:

PR #3248, opened 2015-03-11
Fix #3247 by adding quote to $XCASSET_FILES

sodastsai:
Closes #3247.

PR #3253, opened 2015-03-11
[Generator] use absolute xcasset file path in copy resource script

crazytonyli:
Wrong `xcassets` path in generated `Pods-resources.sh` file. This bug can be reproduced by adding [1PasswordExtension spec](https://github.com/CocoaPods/Specs/blob/master/Specs/1PasswordExtension/1.1.2/1PasswordExtension.podspec.json) to demo app.

### DIFF
--- a/lib/cocoapods/generator/copy_resources_script.rb
+++ b/lib/cocoapods/generator/copy_resources_script.rb
@@ -128,7 +128,7 @@ install_resource()
       xcrun mapc "${PODS_ROOT}/$1" "${CONFIGURATION_BUILD_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}/`basename "$1" .xcmappingmodel`.cdm"
       ;;
     *.xcassets)
-      XCASSET_FILES="$XCASSET_FILES '$1'"
+      XCASSET_FILES="$XCASSET_FILES '${PODS_ROOT}/$1'"
       ;;
     /*)
       echo "$1"


orta:
Are #3248 and this attacking the same problem from different angles?

crazytonyli:
@orta No, I don't think so. That issue will cause shell script parsing error, this issue will cause Xcode unable to find the `xcasset` directory since the path is not correct.

```
/* com.apple.actool.notices */
/Users/crazytonyli/.../MyApp/1PasswordExtension/1Password.xcassets: warning: Failed to read file attributes for "/Users/crazytonyli/.../MyApp/1PasswordExtension/1Password.xcassets"
    Failure Reason: No such file or directory
/* com.apple.actool.compilation-results */
```

segiddins:
This looks good to me. @mrackwitz ?

mrackwitz:
Looks good! :+1:

segiddins:
Will this still allow `POD_ROOT` to be expanded, even though it's in single quotes?

mrackwitz:
>Will this still allow `POD_ROOT` to be expanded, even though it's in single quotes?
Yes, the expansion depends only on the outer-most quotes, and those are double quotes.

segiddins:
Thanks @crazytonyli!

postmechanical:
What's the ETA when the next cocoapods release containing this fix will be released? This issue is currently causing failures in our CI builds.

PR #3263, opened 2015-03-13
[xcassets/1546] WIP

AliSoftware:
This is a WIP for #1546 

> :warning: Don't merge this into `master` as initially requested in the original PR, but merge this (manually) into `stable-0.36.x`

Note: it will probably actually apply to all kind of pod resources, not just `xcassets`.

---

* [X] Add pods resources file references to the user project
* [X] Add pods resources files to appropriate user targets
* [x] Cleanup pods resources from user project if integrating with frameworks
    * WIP, having some `(null)` references (`Xcodeproj` bug?)
* [X] Remove pods resources from target if pod removed in `Podfile`
* [ ] ~~Remove pod resources from project if removed from all targets after deintegration (pod no more present in the `Podfile` — see `UserProjectIntegrator#integrate_user_targets`)~~ :arrow_right:  will be handled in [a separate PR dedicated to reintegration](https://github.com/CocoaPods/CocoaPods/issues/2471)
* [ ] ~~Remove pods resources if target itself not integrated (not listed in the `Podfile`) anymore (see `UserProjectIntegrator#integrate_user_targets`)~~ :arrow_right: will be handled in [a separate PR dedicated to reintegration](https://github.com/CocoaPods/CocoaPods/issues/2471)
* [x] Remove resources from the Pods project (see `FileReferencesInstaller#add_resource`)
* [x] Cleanup (or remove entierly?) the `Pods-resources.sh` script build phase
* [ ] Add integration specs for all the cases!
   * [ ] user project having assets/resources
   * [ ] pods with assets/resources but not linked with all user targets…
   * [ ] also test resource types other than `xcassets`
   * [ ] also test when integrating with dynamic frameworks (that should not require that step)
* [x] Cleanup code (debug output, etc)

---

Known bugs to investigate:

* [ ] [Resources duplicated in user project after 2nd pod install](https://github.com/CocoaPods/CocoaPods/issues/1546#issuecomment-90634423)
* [ ] [undefined method 'resources_build_phase' for PBXAggregateTarget](https://github.com/CocoaPods/CocoaPods/pull/3263#issuecomment-90592023)

PR #3303, opened 2015-03-21
[CopyResourcesScript] Do not discard `.xcassets` from the main project if there are any pods containing `.xcassets`.

lembacon:
If there are any pods being used containing `.xcassets` (for example, [1PasswordExtension](https://github.com/AgileBits/onepassword-app-extension)), while the main project also has a `.xcassets`, the one from the main project will be unexpectedly discarded.

### DIFF
--- a/lib/cocoapods/generator/copy_resources_script.rb
+++ b/lib/cocoapods/generator/copy_resources_script.rb
@@ -169,6 +169,7 @@ then
       TARGET_DEVICE_ARGS="--target-device mac"
       ;;
   esac
+  while read line; do XCASSET_FILES="$XCASSET_FILES '$line'"; done <<<$(find "$PWD" -name "*.xcassets" | egrep -v "^$PODS_ROOT")
   echo $XCASSET_FILES | xargs actool --output-format human-readable-text --notices --warnings --platform "${PLATFORM_NAME}" --minimum-deployment-target "${IPHONEOS_DEPLOYMENT_TARGET}" ${TARGET_DEVICE_ARGS} --compress-pngs --compile "${BUILT_PRODUCTS_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}"
 fi
 EOS


kylef:
Thanks @lembacon

segiddins:
Could someone please explain what exactly this change does?

kylef:
@segiddins This issues fixes the problem where the user's xcasset is overriden by any pod's xcasset. Note, it doesn't fix all the issues with assets and #3263 is still required.

segiddins:
No, I understand the issue that it's fixing, I'd like to understand _how_ it does so.

AliSoftware:
@segiddins it adds every ˋ*.xcassets` found in the project's folder — except the ones in `$PODS_ROOT` — to the `XCASSET_FILES` list of xcassets to compile.

segiddins:
Won't this overzealously include user xcassets, then?

AliSoftware:
Yes it will. That's why it's a temporary workaround until my #3263 PR gets merged and also solves this case

AliSoftware:
BTW we should be precise about this in the CHANGELOG (the thing being a workaround and not a permanent fix and all) so users won't get surprised.

sodastsai:
While the main project has 2 or more xcassets, this commit makes the ```$XCASSET_FILES``` become something like: 
```
'/<main_project_path>/Image1.xcassets /<main_project_path>/Image1.xcassets' '/<pod1_path>/pod1.xcassets' '/<pod2_path/pod2.xcassets'
```

but it should be
```
'/<main_project_path>/Image1.xcassets' '/<main_project_path>/Image1.xcassets' '/<pod1_path>/pod1.xcassets' '/<pod2_path/pod2.xcassets'
```
(the result of new added line in this commit is not correct when multiple xcassets in the main project)

sodastsai:
Or changing following line also works for me :)

```
while read line; do XCASSET_FILES="$XCASSET_FILES $line"; done <<<$(find "$PWD" -name "*.xcassets" | egrep -v "^$PODS_ROOT" | awk '{print "\x27" $0 "\x27"}')
```
(use `awk` to quote values)

PR #3315, opened 2015-03-24
Unique the resources by config

dtorres:
This helps generate a slimmer script when Pods define the same resources in their subspecs.

This is a diff of the generated scripts for the pod "FormatterKit":
https://www.diffchecker.com/z5ja4cve

### DIFF
--- a/lib/cocoapods/generator/copy_resources_script.rb
+++ b/lib/cocoapods/generator/copy_resources_script.rb
@@ -16,7 +16,10 @@ module Pod
       # @param  [Platform] platform @see platform
       #
       def initialize(resources_by_config, platform)
-        @resources_by_config = resources_by_config
+        @resources_by_config = {}
+        resources_by_config.each do |config, resources|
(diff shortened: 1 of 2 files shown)


segiddins:
Please add tests :)

segiddins:
(Also, we should probably just `uniq` where we construct the hash in the first place)

dtorres:
I also thought of doing the uniq in the AggregateTargetInstaller, but then the issue may resurface when/if CopyResourcesScript is used again somewhere else.

And besides, although that may be more efficient when running `pod update`, the gain is in the time of the generated script.

I've added tests and am running them before pushing them

segiddins:
I still think I'd prefer the `uniq` to be done when the hash is constructed, the generator should be 'dumb' and just use whatever is passed to it

dtorres:
I'll do another PR with those changes.

segiddins:
Closing in favor of https://github.com/CocoaPods/CocoaPods/issues/3327, thanks @dtorres !

PR #3327, opened 2015-03-29
[AggregateTargetInstaller] Unique resources passed to the script generator

dtorres:
As requested in #3315, an implementation on the `AggregateTargetInstaller` rather than `CopyResourcesScript`

I couldn't device a way to create a test this under AggregateTargetInstaller for this. Any help is appreciated.

### DIFF
--- a/lib/cocoapods/installer/target_installer/aggregate_target_installer.rb
+++ b/lib/cocoapods/installer/target_installer/aggregate_target_installer.rb
@@ -102,16 +102,11 @@ module Pod
         end
       end
 
-      # Creates a script that copies the resources to the bundle of the client
-      # target.
+      # Uniqued Resources grouped by config
       #
-      # @note   The bridge support file needs to be created before the prefix
-      #         header, otherwise it will not be added to the resources script.
+      # @return [Hash{ Symbol => Array<Pathname> }]
       #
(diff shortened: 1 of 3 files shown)


segiddins:
:+1: once we've added a test

segiddins:
Let's also add a changelog entry. Fwiw, it might be easiest to test this if you extract the creation of the `resources_by_config` hash into its own method, then you can just stub the return value of `resources` on the file accessors.

dtorres:
Great idea! I'll do that.

lib/cocoapods/installer/target_installer/aggregate_target_installer.rb
```diff
@@ -102,16 +102,11 @@ def create_bridge_support_file
...
+      # Uniqued Resources grouped by config
       #
-      # @note   The bridge support file needs to be created before the prefix
-      #         header, otherwise it will not be added to the resources script.
+      # @return [Hash{ Symbol => Array<Pathname> }]
       #
-      # @return [void]
-      #
-      def create_copy_resources_script
-        path = target.copy_resources_script_path
```
dtorres: This removed line makes every other test to fail.
PR #3337 adds it back.
Sorry again.

PR #3337, opened 2015-03-31
Provide an actual path to save the script to

dtorres:
In #3327 I inadvertently broke every other test (and cocoapods) except the one I added.

In this PR I added back the line that was removed, and all tests I broke now pass.

So sorry about that.


segiddins:
:+1: once travis runs

PR #3349, opened 2015-04-01
[Fixture] Add an asset catalog to the fixture podspec

mrackwitz:
In addition to that `FileAccessor#resource_bundles` will exclude asset catalogs. Even though that could be fixed on user-side by appropriate configuration, there is just no case, where it would make sense to have an asset catalog inside of an resource bundle.


segiddins:
@mrackwitz this needs a rebase?

kylef:
Looks like the PR was created from master and not the `ali-xcassets-1546`.

mrackwitz:
@segiddins @kylef: Yep, this was intentionally created from master as `ali-xcassets-1546` will need to be rebased anyway.

PR #3358, opened 2015-04-03
fix copy multiple .xcassets resource error

candyan:
if a project has multiple .xcassets folder, the copy_resources.sh will occur a error. 

like this:
`error: Failed to read file attributes for "/Users/Candyan/Joyshare/Mobile/Arrietty/Arrietty/Images.xcassets /Users/Candyan/Joyshare/Mobile/Arrietty/Arrietty/test.xcassets"
    Failure Reason: No such file or directory`

i found the problem in copy_resources.sh

'while read line' commond will read all result rather than one line of result.

### DIFF
--- a/lib/cocoapods/generator/copy_resources_script.rb
+++ b/lib/cocoapods/generator/copy_resources_script.rb
@@ -169,7 +169,7 @@ then
       TARGET_DEVICE_ARGS="--target-device mac"
       ;;
   esac
-  while read line; do XCASSET_FILES="$XCASSET_FILES '$line'"; done <<<$(find "$PWD" -name "*.xcassets" | egrep -v "^$PODS_ROOT")
+  while read -d $'\n' line; do XCASSET_FILES="$XCASSET_FILES '$line'"; done <<<$(find "$PWD" -name "*.xcassets" | egrep -v "^$PODS_ROOT")
   echo $XCASSET_FILES | xargs actool --output-format human-readable-text --notices --warnings --platform "${PLATFORM_NAME}" --minimum-deployment-target "${IPHONEOS_DEPLOYMENT_TARGET}" ${TARGET_DEVICE_ARGS} --compress-pngs --compile "${BUILT_PRODUCTS_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}"
 fi
 EOS


segiddins:
@AliSoftware please take a look at this?

AliSoftware:
This PR seems good.

The travis failure is only due to integration specs not being rebuilt with the fixed script, we should probably fix/rebuild them along the way but I won't have time today, so postponing the merge until then.

sdarlington:
I think both versions fail if there are spaces in the file name. My alternate version is this:

    XCASSET_FILES="$XCASSET_FILES $(find "$PWD" -path $PWD/Pods -prune -o -name "*.xcassets"  -print | sed "s/^\(.*\)$/'\1'/")"

It works for me but I've not done much testing beyond my one project that was failing.

(I can create a pull request if you like but it might be easier to include in this one?)

spenrose:
I had spaces in my file name because of the auto created watch kit extension. @sdarlington fixes the issue for me.

candyan:
@sdarlington @spenrose 

i can fix it. i also have the problem, but i replace the space with "-".

bonkey:
How about using `\0`?

```
find "$PWD" -name "*.xcassets" -not -path "$PODS_ROOT/*" -print0 | while read -d $'\0' xcasset ; do
  XCASSET_FILES="$XCASSET_FILES '$xcasset'"
done
```

sdarlington:
@bonkey Looks good (though I've not tried it). Didn't know about the -d option for `read`.

candyan:
@bonkey lgtm. i will use it.

bonkey:
It needs some adjustments as it looks like XCASSET_FILES in a loop is a local copy only

candyan:
@bonkey seems can use `\n`

bonkey:
Found at least sth what works in my case:
```
	XCASSET_FILES="$XCASSET_FILES $(find "$PWD" -name "*.xcassets" -not -path "$PODS_ROOT/*" -exec echo -n " '{}'" \;)"
```

sdarlington:
I still prefer @bonkey or my solution -- I think they're a lot clearer -- but if it works I'll be happy(!).

romanr:
I have same issue with large project containing multiple XCassets and App extension.
 
* I have **no spaces in the path or any of the file names** or xcassets or any of the image names.
* **The original fix in this pull request did not help** (which is changing ` while read line; do` to `while read -d $'\n' line; do`

My solution was to use @sdarlington fix:
`XCASSET_FILES="$XCASSET_FILES $(find "$PWD" -path $PWD/Pods -prune -o -name "*.xcassets"  -print | sed "s/^\\(.*\\)$/'\\1'/")"`

changing line 172 to 
`while read line; do XCASSET_FILES="$XCASSET_FILES $(find "$PWD" -path $PWD/Pods -prune -o -name "*.xcassets"  -print | sed "s/^\\(.*\\)$/'\\1'/")";`


### Here's step by step for those who has same problem:

1. Find out where your gems installed: `ruby -r rubygems -e "p Gem.path"`
2. modify the lines at the end of file: `sudo nano /Library/Ruby/Gems/2.0.0/gems/cocoapods-0.36.3/lib/cocoapods/generator/copy_resources_script.rb` (or whatever your path is) and change line 172 to the above.
3. `pod update` in your project

kylef:
Thanks for the pull request @candyan, this unfortunately doesn't completely fix the problem so I'm going to close this in favour of #3405.

PR #3383, opened 2015-04-10
Double-quote the output of the "find" heredoc that locates asset files.

williamsjj:
Fixes issue where newlines in output are converted to spaces which blows up actool.


kylef:
Thanks for the pull request @williamsjj, this unfortunately doesn't completely fix the problem so I'm going to close this in favour of #3405.

PR #3386, opened 2015-04-11
[pod repo list] Fix issue showing the URL for git

kylef:
I noticed a couple of bugs / nonsensical things with `pod repo list`. Here’s a pull request making this sane.

This pull request addresses does following:

- Use `source.url` instead of determining the git remote manually
- Show the branch instead of remote for git repositories, it makes no sense to show the remote since 1) it can only ever be origin according to `source.url` an 2) we’re printing the value of the remote anyway
- Always show the URI for local repositories (so people know what they can put in `source ‘’` in a Podfile

    Before
    
    ```
    b
    - Type: git ()
    - URL:
    - Path: /Users/kyle/.cocoapods/repos/b
    ```
    After
    ```
    b
    - Type: git (master)
    - URL:  file:///Users/kyle/.cocoapods/repos/b/.git
    - Path: /Users/kyle/.cocoapods/repos/b
    - Add tests around existing behaviour and the changes.
    ```

It fixes the following bugs:

- Fixes the output of git spec repositories which have a git remote, but does not have the upstream for the current branch configured to it.

    Before:
    
    ```
    test
    - Type: git ()
    - URL:
    - Path: /Users/kyle/.cocoapods/repos/test
    ```

    After:

    ```
    test
    - Type: git (master)
    - URL:  git@git.oschina.net:congni/IOSFrameLibrary.git
    - Path: /Users/kyle/.cocoapods/repos/test
    ```

- Suppresses git error output when failing to determine git branch (i.e, there is no commits yet)

    ```
    foo
    Could not get sha1 for HEAD. Skipping.
    - Type: git ()
    - URL:
    - Path: /Users/kyle/.cocoapods/repos/foo
    ```
    
    After:
    
    ```
    foo
    - Type: git (unknown)
    - URL:  file:///Users/kyle/.cocoapods/repos/foo/.git
    - Path: /Users/kyle/.cocoapods/repos/foo
    ```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,11 +13,18 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Marius Rackwitz](https://github.com/mrackwitz)
   [#3145](https://github.com/CocoaPods/CocoaPods/issues/3145)
 
+* Show the source URI for local Pod specification repositories in
+  `pod repo list`.  
+  [Kyle Fuller](https://github.com/kylef)
+
 ##### Bug Fixes
 
 * Do not pass code-sign arguments to xcodebuild when linting OS X targets.  
   [Boris Bügling](https://github.com/neonichu)
   [#3310](https://github.com/CocoaPods/CocoaPods/issues/3310)
+
+* Fixes an issue showing the URL to remote resources in `pod repo list`.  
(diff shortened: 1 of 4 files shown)


segiddins:
Looks good!

orta:
:shipit: 

Looks much better

Issue #1241, opened 2013-07-30
Allow to skip the deploy target version check


Whirlwind:
I have a question:
I use google map in my app and the google map only support 6.0 and later. :(
So I do something to support the ios 5.0: if the iOS is 5.0, I will use the apple map view and I will use the google map if it is 6.0 and later.

I know there is a keyword `platform` to set the deploy target version. But the version make it bad to my question:

I must set the platform to 6.0, otherwise pod install will not pass because the Google Map only support 6.0 and later! So I can NOT set it to 5.0!!
```
[!] The platform of the target `Pods` (iOS 5.0) is not compatible with `Google-Maps-iOS-SDK (1.4.0)` which has a minimum requirement of iOS 6.0.
```

And if I set the platform to 6.0, the all Pods target will be set to iOS 6.0!!

for Cocoapods v0.20.2, if my app's target version is iOS 5.0, the Pods target will be 5.0;
for Cocoapods v0.22.2, the Pods target will be 6.0, it ignore my app target setting.

How should I do?

kambala-decapitator:
This should really be integrated in CocoaPods in my opinion. It's completely developer's responsibility to write proper code (like make runtime checks for libraries that aren't supported on old OS versions), and not the dependency manager's.

Reflejo:
`:ignore_platform` sounds like the way to go. In the meantime, you can "workaround" the issue on your `Podfile` by  setting the platform to ios8 and then a post_install:

```ruby
# Hack to rollback the targets to iOS 7
post_install do |installer|
  installer.project.targets.each do |target|
    target.build_configurations.each do |config|
      config.build_settings["IPHONEOS_DEPLOYMENT_TARGET"] = "7.0"
    end
  end
end
```

Whirlwind:
Any update/best practice for this?

kylef:
I think @segiddins mentioned somewhere that this could become a warning and not an error.

@orta what do you think? Might make sense.

alloy:
Yeah, sound good :+1:

mrezk:
@segiddins I think there has been a regression here.

v0.39.0 throws the following message when trying to install a dependency with a higher deployment target:
```
[!] Unable to satisfy the following requirements:

- `MZFormSheetPresentationController (~> 2.2.0)` required by `Podfile`

Specs satisfying the `MZFormSheetPresentationController (~> 2.2.0)` dependency were found, but they required a higher minimum deployment target.
```

[...]

PR #3407, opened 2015-04-16
[Analyzer] Only warn when there is a minimum deployment target mismatch

segiddins:
Closes https://github.com/CocoaPods/CocoaPods/issues/1241.


orta:
Nice work

evilDave:
What happened to this? I am using 0.39.0 and it still reports a mismatch as an error.
(GoogleAppIndexing 2.0.2 is ios 9.0 whereas mine is ios 7.0)

PR #3408, opened 2015-04-16
[PrivatePodXCConfig] Public Headers are not needed to build the target itself

mrackwitz:
The build headers are always given as a superset of the public headers. To build the target itself all headers are needed typically, not only the public headers. The public headers are needed for other targets (pod targets + user targets) depending on a target.

/c @neonichu

### DIFF
--- a/lib/cocoapods/generator/xcconfig/private_pod_xcconfig.rb
+++ b/lib/cocoapods/generator/xcconfig/private_pod_xcconfig.rb
@@ -45,9 +45,7 @@ module Pod
         # @return [Xcodeproj::Config]
         #
         def generate
-          target_search_paths = target.build_headers.search_paths(target.platform)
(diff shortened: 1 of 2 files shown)


segiddins:
@mrackwitz to be clear, the issue here was that that header search path was being duplicated?

mrackwitz:
No, though that was the problem I was diagnosting, the HEADER_SEARCH_PATHS (mis-)configuration were not the issue, as the generated symlinks resolve to the same actual files. But I don't see how the public header sandbox dir would be needed.

segiddins:
Then what _was_ the issue/

mrackwitz:
It is unnecessary.

segiddins:
@mrackwitz I'm closing this, feel free to re-open / merge if you're sure it's correct.

PR #3412, opened 2015-04-16
[Linter] Allow to fail-fast

mrackwitz:
That's helpful for i.e. multi-platforms specs and/or specs with subspecs, especially in combination with `--no-clean`.

### DIFF
--- a/lib/cocoapods/command/lib.rb
+++ b/lib/cocoapods/command/lib.rb
@@ -113,6 +113,7 @@ module Pod
            ['--subspec=NAME', 'Lint validates only the given subspec'],
            ['--no-subspecs', 'Lint skips validation of subspecs'],
            ['--no-clean', 'Lint leaves the build directory intact for inspection'],
+           ['--fail-fast', 'Lint stops on the first failing platform or subspec'],
            ['--use-libraries', 'Lint uses static libraries to install the spec'],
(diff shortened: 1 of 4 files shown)


lib/cocoapods/command/lib.rb
```diff
@@ -123,6 +124,7 @@ def initialize(argv)
           @quick           = argv.flag?('quick')
           @allow_warnings  = argv.flag?('allow-warnings')
           @clean           = argv.flag?('clean', true)
+          @fail_fast       = argv.flag?('fail-fast', false)
```
kylef: Would it make sense for this to become the default, and `--no-fail-fast` to disable? /c @orta

mrackwitz: This should stay disabled at least unless `--allow-warnings` is given.

spec/unit/validator_spec.rb
```diff
@@ -487,6 +499,7 @@ def podspec_path
       end
 
       def setup_validator
+        @validator.instance_variable_set(:@results, [])
```
segiddins: why's this necessary?

segiddins:
:+1:

spec/unit/validator_spec.rb
mrackwitz: Because the variable is initialized in `validate`, which is not called directly in the tests, but instead private methods, still there are side-effects appearing on it.

PR #3414, opened 2015-04-16
[Spec|Doc|Fixture] Cleanup in preparation to a fixed xcasset handling

mrackwitz:
The fixture tarball `banana-lib.tar.gz` is a spec source inside the repo, containing the `BananaLib.podspec` and all associated files. Special about this is, that those live inside an own Git repo with multiple branches inside of the archive. Some of the files were historically not been committed there and just in the staging area, which feels noisy and vulnerable. This addresses those incosistencies in the fixtures. So this PR, specifically commit efa44c1, ships some hidden commits in the inner repo.
It brings some workflow improvements, that should ensure that those won't happen again so easily. It adds checks of the fixture tarballs for pending changes before rebuilding the tarball or wiping changes by overwriting them by unpacking the files. This also ensures indirectly over the dependency from the `spec` tasks, that there will be never pending changes in the inner repo.


segiddins:
This should be fine

segiddins:
Well, this was fun to rebase! Will merge once CI runs.

PR #3420, opened 2015-04-17
Fix module map installation

mrackwitz:
`PodSourceInstaller` didn't know about the new file accessor `module_map`, because of that it **deleted** the module map, before it was **copied** to the right directory, which means that it could exist now twice after the installation. This was not captured by the integration spec, because it uses a local podspec.
This is builded upon the updated fixture from #3414 to avoid merge conflicts.


segiddins:
Let's wait to merge this until the download cache is merged?

PR #3481, opened 2015-05-01
[pod] Write the UTF-8 warning to stderr, rather than stdout.

mjholgate:
🌈
This means operations like:

   POD_VERSION="`pod --version`"

Work correctly even if LANG isn't set (e.g. within Jenkins).

### DIFF
--- a/bin/pod
+++ b/bin/pod
@@ -1,7 +1,7 @@
 #!/usr/bin/env ruby
 
 if Encoding.default_external != Encoding::UTF_8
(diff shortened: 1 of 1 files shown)


mjholgate:
This problem bit me - hopefully this fix is useful to others too!

segiddins:
Seems good -- can you look into the CI failures?

mjholgate:
I think (but not sure), that the CI failures were happening anyway, even on a clean build?

PR #3498, opened 2015-05-04
[Cache] Fix the cache to handle re-downloads properly

mrackwitz:
If `destination` exists already, `source` is not copied directly `destination` but instead to `copy+source.basename`.

### DIFF
--- a/lib/cocoapods/downloader/cache.rb
+++ b/lib/cocoapods/downloader/cache.rb
@@ -184,6 +184,7 @@ module Pod
       def copy_and_clean(source, destination, spec)
         specs_by_platform = group_subspecs_by_platform(spec)
         destination.parent.mkpath
+        FileUtils.rm_rf(destination)
         FileUtils.cp_r(source, destination)
         Pod::Installer::PodSourcePreparer.new(spec, destination).prepare!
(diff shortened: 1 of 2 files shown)


kylef:
Can we get a test too?

mrackwitz:
Any ideas how to test that properly?
I was thinking about adding an integration test via `pod try`, which uses a custom `CP_HOME_DIR` to inject via custom `config.yaml` the following config settings:
```yaml
---
skip_download_cache: false
cache_root: .
```
Then the `before` and `after` dir could be used to describe the cache state.

segiddins:
:+1: 

CHANGELOG

segiddins:
@mrackwitz changelog this so we can release an update, tests can come later

mrackwitz:
Tests are nearly finished. Give me a moment.

lib/cocoapods/downloader/cache.rb
```diff
@@ -184,6 +184,7 @@ def in_tmpdir(&blk)
       def copy_and_clean(source, destination, spec)
         specs_by_platform = group_subspecs_by_platform(spec)
         destination.parent.mkpath
+        destination.rm_rf
```
segiddins: `FileUtils.rm_rf destination`

PR #3504, opened 2015-05-05
[TargetIntegrator] Diff lockfile and manifest as hashes

segiddins:
Probably closes https://github.com/CocoaPods/CocoaPods/issues/3502?

I don't know if this is too slow or too much of a ruby dependency.

### DIFF
--- a/lib/cocoapods/installer/user_project_integrator/target_integrator.rb
+++ b/lib/cocoapods/installer/user_project_integrator/target_integrator.rb
@@ -198,19 +198,22 @@ module Pod
         #
         def add_check_manifest_lock_script_phase
           phase_name = 'Check Pods Manifest.lock'
+          phase_shell_path = '/usr/bin/env ruby'
           native_targets_to_integrate.each do |native_target|
-            next if native_target.shell_script_build_phases.any? { |phase| phase.name == phase_name }
+            next if native_target.shell_script_build_phases.any? { |phase| phase.name == phase_name && phase.shell_path == phase_shell_path }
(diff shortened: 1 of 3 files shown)


kylef:
I'm worried about using Ruby in the build phases. I wonder if we can solve the problem a different way (and make Psych consistent across versions).

segiddins:
The linked issue links to the psych issue, where I ask if that's possible. 

-Samuel E. Giddins

On May 5, 2015, at 6:49 AM, Kyle Fuller <notifications@github.com> wrote:

I'm worried about using Ruby in the build phases. I wonder if we can solve the problem a different way (and make Psych consistent across versions).

lib/cocoapods/installer/user_project_integrator/target_integrator.rb
```diff
@@ -198,19 +198,23 @@ def add_copy_resources_script_phase
         #
         def add_check_manifest_lock_script_phase
           phase_name = 'Check Pods Manifest.lock'
+          phase_shell_path = '/usr/bin/env ruby'
           native_targets_to_integrate.each do |native_target|
-            next if native_target.shell_script_build_phases.any? { |phase| phase.name == phase_name }
+            next if native_target.shell_script_build_phases.any? { |phase| phase.name == phase_name && phase.shell_path == phase_shell_path }
```
segiddins: this doesn't seem to be properly updating existing script phases?

orta:
I'm wary of adding anything that can slow down the build phase unless absolutely essential, a few seconds saved here is many human hours.

PR #3509, opened 2015-05-05
[PodTargetInstaller] Don't generate an umbrella header when a custom mod...

segiddins:
...ule map is specified.

The integration spec failures are expected, and demonstrate the effect of this PR. I will update them post-merge.

\c @mrackwitz


mrackwitz:
This should be unit tested. Furthermore I think we should create an issue for what is fixed here, to refer it from the changelog.

segiddins:
There are currently no unit tests concerning custom module maps in the pod target installer spec, and I believe this is sufficiently covered by the integration spec.

PR #3548, opened 2015-05-10
[Resolver] Move side-effect of storing head pods from #specs_by_target

mrackwitz:
… to #resolve.
Methods, which work like getters, shouldn't have side-effects.

### DIFF
--- a/lib/cocoapods/resolver.rb
+++ b/lib/cocoapods/resolver.rb
@@ -53,7 +53,11 @@ module Pod
       dependencies = podfile.target_definition_list.map(&:dependencies).flatten
       @cached_sets = {}
       @activated = Molinillo::Resolver.new(self, self).resolve(dependencies, locked_dependencies)
-      specs_by_target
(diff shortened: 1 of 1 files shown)


segiddins:
Sure, this is fine, though I fail to see the motivation for changing this, since it's entirely non-functional.

mrackwitz:
I stumbled on that when working on something else and saw at as an origin of potential unwanted consequences and that I just spent a moment on double-checking that's not called twice was already too much.
… and less code is better code. :grimacing:

segiddins:
👍🏻🚢

In the future, feel free to just push refactors like this, as long as the specs pass

-Samuel E. Giddins

On May 10, 2015, at 12:50 PM, Marius Rackwitz <notifications@github.com> wrote:

I stumbled on that when working on something else and saw at as an origin of potential unwanted consequences and that I just spent a moment on double-checking that's not called twice was already too much.
… and less code is better code.

mrackwitz:
I will insist on going over CI and code review.

PR #3549, opened 2015-05-10
[TargetIntegrator] Ensure to reset the contents of the Manifest.lock Script Phase

mrackwitz:
I found that a while ago while reviewing the `TargetIntegrator`, but it staled locally.
Do we want to allow user changes to the script or ensure at this place that updates to the script are applied, if needed?
Meanwhile there is still a bigger issue, that existing targets are not re-integrated on changes to the integration process, which is to a certain extent blocked by CocoaPods/Xcodeproj#202.


segiddins:
I don't think this change is necessarily desirable -- I know people have changed that phase.

segiddins:
CHANGELOG and tests?

segiddins:
Actually, while we're at it, can we refactor the logic for 'create or update a script build phase' into it's own method?

segiddins:
@mrackwitz ping on this?

segiddins:
:+1:

Issue #3200, opened 2015-02-28
Compiler warning when adding Swift framework


steilerDev:
I have added two Swift frameworks ([Locksmith](https://github.com/matthewpalmer/Locksmith) and [SwiftyUserDefaults](https://github.com/radex/SwiftyUserDefaults)) to my project, integrated through CocoaPods (using version 0.36.0.rc.1 and explicitly stating `use_frameworks!` within my PodFile).

Both framework produce the same xCode compiler warning (using xCode 6.3 Beta):

```
[...]/Pods/<module-includes>:1:1: Umbrella header for module 'SwiftyUserDefaults' does not include header 'SwiftyUserDefaults-Swift.h'
```
and
```
[...]/Pods/<module-includes>:1:1: Umbrella header for module 'Locksmith' does not include header 'Locksmith-Swift.h'
```

After talking to the developer of the frameworks we conclude that the root of the problem might be CocoaPods.

segiddins:
What leads you to conclude the issue is caused by CocoaPods? That might help us fix it :)

steilerDev:
It might be a xCode or Swift bug, but since the problem occurred with pods of two independent devs and both of them don't know how to fix this problem I thought it might be a problem with cocoapods. 

On top of that the file causing the problem (the umbrella header) is as far as I can see a file created by cocoapods (isn't it?)

segiddins:
@mrackwitz any idea what could be causing this?

jshier:
I've been seeing this intermittently. It's usually fixable by clearing derived data. This is using the Xcode 6.3 beta too, as well as 0.36.beta.2, which I needed because I'm using pods with compiled libraries.

neonichu:
#3092 could be related. I'd be inclined to wait and see if this turns out to be an actual new warning or just beta noise.

steilerDev:
Apparently beta noise and/or resolved in a recent update of cocoapods. No more warnings in latest xCode beta. Thank you guys anyway.

andreamazz:
I'm having this issue using cocoapods `0.37` with Xcode `6.3.1`.  
The funny thing is that this happens only when I'm archiving the app for release on the AppStore, the debug build goes smoothly. 
The archive process results in a corrupted archive. 
Another detail that might help: if I strip down the watchkit and today extension's targets, no warning is raised and the archive is fine.  
I'm trying to release a new version of this project:
https://github.com/FancyPixel/gulps

and the output:
[...]

PR #3550, opened 2015-05-11
Deduplicate pod targets

mrackwitz:
This de-duplicates pod targets in the analyzer, if they share the same subspec set and target the same platform. If there is any dependency from at least one target definition, which doesn't fulfill this condition, then all dependencies with this root spec will be scoped.
This should fix diverse issues related to archiving with Xcode, especially when extensions are integrated. Beside that it will reduce build times, drastically.
A new configuration key was added (38ec3cd), which allows to opt-out of the feature, because these changes cause massive changes to the existing integration and are likely to break edge-case setups.

#### Open Todos:

The following things are still open to do:

* [x] A `PodTarget` can't be deduplicated if any of its transitive dependencies can't be deduplicated.
* [x] Validate that `PodTarget`s are only initialized legitimate: They can't be scoped when they have more then one target definition.
* [x] Add missing unit tests
* [x] _Decide whether to_ add integration tests with the old behavior
* [x] Tests & experience with real-life projects
* [x] Add a CHANGELOG entry

#### Dependencies:

This depends on the following PRs:

* [ ] CocoaPods/cocoapods-integration-specs#36

##### Deduplication / Descoping of PodTargets

This de-duplicates pod targets in the analyzer, if they share the same subspec set and target the same platform. If there is any dependency from at least one target definition, which doesn't fulfill this condition, then all dependencies with this root spec will be scoped.
A deduplicated pod target will be fully descoped, which means the prefix of all names is skipped and with frameworks, it builds in the `CONTENTS_FOLDER_PATH` as usual given by Xcode, without the introduced level of further scoping to avoid conflicts, which can't happen that way.
Concrete this means the following: If you used to integrate with the following Podfile:
```ruby
use_frameworks!
platform :ios, "8.1"
pod "Alamofire"
pod "Stargate"
pod "MRProgress/Circular"

target "ExampleCLI" do
  platform :osx, "10.10"
  pod "Alamofire"
end

target "ExampleWatchExtension" do
  pod "Stargate"
end

target "ExampleWidget" do
  pod "Alamofire"
  pod "MRProgress/ActivityIndicator"
  pod "MRProgress/Circular"
end
```

You would end up with an integration, which looks like:
* `Pods-Alamofire`
* `Pods-Stargate`
* `Pods-MRProgress`
* `Pods-ExampleCLI-Alamofire`
* `Pods-ExampleWatchExtension-Stargate`
* `Pods-ExampleWidget-Alamofire`
* `Pods-ExampleWidget-MRProgress`

Now you should end up with:
* `Pods-Alamofire`
* `Stargate` :tada:
* `Pods-MRProgress`
* `Pods-ExampleCLI-Alamofire`
* ~~`Pods-ExampleWatchExtension-Stargate`~~ :tada:
* `Pods-ExampleWidget-Alamofire`
* `Pods-ExampleWidget-MRProgress`

As you can see subspecs and different platforms prevent the deduplication, even when there is a subset of target definitions where the deduplication could been applied as seen for `Alamofire` in the example above.

##### Environment Header

All `PodTarget`s from the same `AggregateTarget` used to share an environment header. This won't be possible anymore for deduplicated pod targets. As the environment header defines precompiler macros to indicate the version for each integrated dependency in the same target definition, altering anything around this can cause unexpected failures when pods have implicit dependencies or offer additional behavior when used along other pods. This PR for now replaces the shared environment header with a dedicated environment header, which in the best case should only have information about all explicit dependencies of the given podspec. Another variant would be that it has additional access to the intersection of all dependencies of all target definitions it belongs to.

#### Highlighted Architecture Changes

* Pull down relation `target_definition` from `Target` into `AggregateTarget` and all getters, which depend on this and are only needed for the subclass.
* Pluralize prior singular inherited relation from `PodTarget` to `TargetDefinition`, now known as `PodTarget#target_definitions`
* Add a flag attribute `scoped` to `PodTarget` to indicate them explicitly as scoped, which is only possible if there is only one related `TargetDefinition`
* Extract `TargetInspector` and `TargetInspectionResult` out of `Analyzer`


segiddins:
:heart:

neonichu:
:sparkles: :sparkling_heart: :sparkles:

neonichu:
Tested with the example of <https://github.com/contentful-labs/Stargate>:

- Seemingly, warning settings are different between iPhone and WatchKit extension targets, so I get this a lot when installing:

```bash
[!] The pod `MMWormhole` is linked to different targets (["Pods-StargateExample", "Pods-StargateExample WatchKit Extension"]), which contain different settings to inhibit warnings. CocoaPods does not currently support different settings and will fall back to your preference set in the root target definition.
```

- The OS X target is no longer buildable, somehow `i386` ends up in the supported architectures, which isn't supported by Swift and also it becomes the default to build. Maybe related:

```bash
2015-05-11 08:43:05.125 xcodebuild[15650:1427685] stream error: stream error at offset 29: created by an unsupported XCDependencyGraph build
2015-05-11 08:43:05.148 xcodebuild[15650:1427686] stream error: stream error at offset 29: created by an unsupported XCDependencyGraph build
```

in the `xcodebuild` output.

lib/cocoapods/hooks/installer_representation.rb
```diff
@@ -69,10 +69,12 @@ def specs_by_lib
...
       def pods_by_lib
         result = {}
-        installer.aggregate_targets.map(&:pod_targets).flatten.each do |lib|
-          pod_names = [lib.root_spec.name]
-          pod_reps = pods.select { |rep| pod_names.include?(rep.name) }
-          result[lib.target_definition] = pod_reps
+        installer.aggregate_targets.map(&:pod_targets).flatten.uniq.each do |lib|
+          lib.target_definitions.each do |target_definition|
+            pod_names = [lib.root_spec.name]
+            pod_reps = pods.select { |rep| pod_names.include?(rep.name) }
```
segiddins: check with `==` ?

lib/cocoapods/target/aggregate_target.rb
```diff
@@ -28,6 +33,18 @@ def product_module_name
       c99ext_identifier(label)
     end
 
+    # @return [Platform] the platform for this target.
+    #
+    def platform
+      @platform ||= target_definition.platform
```
segiddins: I don't think the caching here is really necessary

spec/unit/generator/embed_frameworks_script_spec.rb
```diff
@@ -3,15 +3,14 @@
 module Pod
   describe Generator::EmbedFrameworksScript do
     it 'returns the embed frameworks script' do
-      target_definition = Podfile::TargetDefinition.new(:default, nil)
       frameworks = {
-        'Debug'   => %w(Loopback.framework Reveal.framework),
+        'Debug'   => %w(Pods/Loopback.framework Reveal.framework),
```
segiddins: why this change?

lib/cocoapods/target/aggregate_target.rb
mrackwitz: Full agree, it used to be there, so I kept it. Let's get rid of it.

spec/unit/generator/embed_frameworks_script_spec.rb
mrackwitz: Because the script should be able to handle paths and not only file base names correctly.

lib/cocoapods/hooks/installer_representation.rb
mrackwitz: :+1:

yonaskolb:
This is great! Been wondering about this

ItsTipTop:
Is it fixed now?
I just can't archive my app

mrackwitz:
@ItsTipTop: Are you using the latest released CocoaPods? That doesn't include those changes yet. Or are you using [a development installation](https://guides.cocoapods.org/contributing/dev-environment.html#using-latest-pod-command)? Then you could help if you would describe what exactly fails when you try to archive your app.

ItsTipTop:
@mrackwitz I'm using 0.37.1 version.
When I run my app on device/simulator all works fine. But when I try to archive my app I recieve warnings. Xcode successfully archived my application, but the Archives Organizer does not list my archive. My issue very similar to @andreamazz in #3200 



    Umbrella header for module 'UICKeyChainStore' does not include header 'Pods-TJShareExtension-UICKeyChainStore-umbrella.h'
    Umbrella header for module 'pop' does not include header 'Pods-TJPostKit-pop-umbrella.h'
    Umbrella header for module 'UIImage_Categories' does not include header 'Pods-TJPostKit-UIImage-Categories-umbrella.h'

Warnings appear only for frameworks that included in several targets. 

Here my podfile:

    platform :ios, '8.0'
    inhibit_all_warnings!
    use_frameworks!

       target 'AppTarget' do
           link_with 'AppTargetTests'
           pod 'pop', '~> 1.0'
           pod 'UIImage-Categories'
           pod 'UICKeyChainStore'
           pod 'AFNetworking'
        end
  
    target 'AppPostKit' do
           pod 'pop', '~> 1.0'
           pod 'UIImage-Categories'
    end

    target 'AppShareExtension' do
        pod 'UICKeyChainStore'
       pod 'AFNetworking'
   end

Now i can't archive my app and send update to App Store :(

Waiting for your reply. Thank you.

mrackwitz:
@neonichu: I could address the following issues:

>- Seemingly, warning settings are different between iPhone and WatchKit extension targets, so I get this a lot when installing:
>
>```bash
[!] The pod `MMWormhole` is linked to different targets (["Pods-StargateExample", "Pods-StargateExample WatchKit Extension"]), which contain different settings to inhibit warnings. CocoaPods does not currently support different settings and will fall back to your preference set in the root target definition.
```

> - In the shared Pods, the source files are seemingly added twice.

Left open is:

>- The OS X target is no longer buildable, somehow `i386` ends up in the supported architectures, which isn't supported by Swift and also it becomes the default to build.

**Comment by mrackwitz** [comment]
@ItsTipTop: Thanks for your detailed issue report, really sorry that this tooling issue prevents you from shipping  your app to the app store. But unless you have feedback to [a development installation with this branch](https://guides.cocoapods.org/contributing/dev-environment.html#using-latest-pod-command), this really shouldn't go in this PR, which tries to address the issues which are linked, which also apply for your specific issue. This makes it easier for us to work focused on an actual solution. Meanwhile, did you tried the [workaround here](https://github.com/CocoaPods/CocoaPods/issues/3493#issuecomment-99927503)? Hope that this will work for you.

**Comment by ItsTipTop** [comment]
@mrackwitz No. This workaround does not work for me.

**Comment by kexoth** [comment]
I'm having the same issue, I cannot archive my app & tried everything available on the Internet, but never suspected this is the issue until I found this on the Xcode Console:
     
    5/26/15 8:25:43.730 PM Xcode[88529]: Cannot make directory /Users/<my-username>/Library/Developer/Xcode/Archives/2015-05-26/App 5-26-15, 8.25 PM.xcarchive/dSYMs/Realm.framework.dSYM: File exists

Then I tried everything regarding the issue available here but with no success, here's my Podfile:

        platform :ios, '8.0'
        use_frameworks!
        pod 'RealmSwift', '~> 0.92'
        link_with 'AppToday', 'App WatchKit Extension'

        target 'App', :exclusive => true do
            pod 'RealmSwift', '~> 0.92'
            pod 'R.swift', '~> 0.7'
            pod 'NGAParallaxMotion', '~> 1.1.0'
        end

        post_install do |installer_representation|
            installer_representation.project.targets.each do |target|
            
                    config.build_settings['ONLY_ACTIVE_ARCH'] = 'NO'
            
                    if !config.to_s.include? 'Debug'
                        config.build_settings['CODE_SIGN_IDENTITY[sdk=iphoneos*]'] = 'iPhone Distribution'
                    end
                end
            end
        end

This way I get the duplicate thing & the app doesn't archive, although every target gets the dependancies it needs.

On the other hand with an approach for having a shared pod for every target & a dedicated pod for the iOS app in order to avoid the duplication, I changed the `Podfile`:

        platform :ios, '8.0'
        use_frameworks!
        pod 'RealmSwift', '~> 0.92'
        link_with 'App', 'AppToday', 'App WatchKit Extension'

        target 'App' do
            pod 'R.swift', '~> 0.7'
            pod 'NGAParallaxMotion', '~> 1.1.0'
        end

        post_install do |installer_representation|
            installer_representation.project.targets.each do |target|
            
                    config.build_settings['ONLY_ACTIVE_ARCH'] = 'NO'
            
                    if !config.to_s.include? 'Debug'
                        config.build_settings['CODE_SIGN_IDENTITY[sdk=iphoneos*]'] = 'iPhone Distribution'
                    end
                end
            end
        end

This way I get this error when installing the pods & I don't get the `Pods-App.xcconfig` with the App's specific dependancies:

    [!] CocoaPods did not set the base configuration of your project because your project already has a custom config set. In order for CocoaPods integration to work at all, please either set the base configurations of the target `App` to `Pods/Target Support Files/Pods-App/Pods-App.debug.xcconfig` or include the `Pods/Target Support Files/Pods-App/Pods-App.debug.xcconfig` in your build configuration.

We should wait for this patch to fix the duplication issue in a Cocoapods update?
What should be the right approach for this in the future between these 2?

I'm waiting for your reply, 
Thanks,
@kexoth

File: spec/unit/installer/analyzer_spec.rb [hunk 4]
```diff
@@ -111,6 +114,69 @@ module Pod
         target.platform.to_s.should == 'iOS 6.0'
       end
 
+      describe 'deduplication' do
```
mrackwitz: I added some further unit tests here. Feedback welcome!

lib/cocoapods/target/aggregate_target.rb
mrackwitz: I think this is desired, because it's a computed value and it's accessed quite often.

orta:
I've ran this on both [Eigen](https://github.com/artsy/eigen/pull/520) and [Eidolon](https://github.com/artsy/eidolon) and it's working :+1:

mrackwitz:
I've re-ran this on [Stargate](/contentful-labs/Stargate) and it's working there, too. :tada:

TosinAF:
Can we get this fix before 0.38 comes out, if not where is the link on how to run cocoapds from source? 
Thanks a lot for your work on this! @mrackwitz ! 🙌🏽

orta:
@TosinAF - https://guides.cocoapods.org/using/a-gemfile.html with the branch `mr-deduplicate-frameworks`

lib/cocoapods/installer/analyzer.rb
```diff
@@ -249,48 +250,119 @@ def generate_target(target_definition, specs)
...
-        pod_target = PodTarget.new(pod_specs, target.target_definition, sandbox)
+      def generate_pod_target(target_definitions, pod_specs, scoped: false)
+        pod_target = PodTarget.new(pod_specs, target_definitions, sandbox, scoped)
 
         if config.integrate_targets?
-          pod_target.user_build_configurations = target.user_build_configurations
-          pod_target.archs = @archs_by_target_def[target.target_definition]
+          target_inspections = result.target_inspections.select { |t, _| target_definitions.include?(t) }.values
+          pod_target.user_build_configurations = target_inspections.map(&:build_configurations).reduce({}, &:merge)
+          pod_target.archs = target_inspections.map(&:archs).uniq.sort
```
mrackwitz: This should be `flat_map(&:archs)`.

robruffler:
So this branch fixed the similar issue I was having w/ a watchkit extension project (unable to create archive), though I had to manually tweak the path being passed into install_framework in the Embed Pods Frameworks shell script in order for things to work correctly. Also, the Architectures for two of the shared targets (Bolts and Parse in this case) were blank by default.

Podfile contents for reference
```
platform :ios, '8.0'

use_frameworks!

link_with 'DBag', 'Dbag WatchKit Extension'

target 'DBag' do
pod 'Parse', '~> 1.7'
pod 'GoogleMaps', '~> 1.10'
end

target 'DBagTests' do

end

target 'DBag WatchKit Extension' do
pod 'Parse', '~> 1.7'
pod 'GoogleMaps', '~> 1.10'
end

target 'DBag WatchKit App' do

end
```

mrackwitz:
@robruffler:

>though I had to manually tweak the path being passed into install_framework in the Embed Pods Frameworks shell script in order for things to work correctly

Can you validate whether #3675 does also resolve this issue?

>Also, the Architectures for two of the shared targets (Bolts and Parse in this case) were blank by default.

I'm investigating this issue. Thanks for reporting!

robruffler:
@mrackwitz I'll check this out. Also of note, I'm only able to archive successfully once. After the first time I get this continually until I stop Xcode, delete Derived Data and restart Xcode. Not sure if this is related.

```
ValidateEmbeddedBinary Build/Intermediates/ArchiveIntermediates/DBag/InstallationBuildProductsLocation/Applications/dButton.app/PlugIns/dButton\ WatchKit\ Extension.appex
    cd /Users/rufflerr/Developer/iOS/dbutton
    export PATH="/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/usr/bin:/Applications/Xcode.app/Contents/Developer/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin"
    builtin-embeddedBinaryValidationUtility /Users/rufflerr/Developer/iOS/dbutton/Build/Intermediates/ArchiveIntermediates/DBag/InstallationBuildProductsLocation/Applications/dButton.app/PlugIns/dButton\ WatchKit\ Extension.appex -signing-cert 039628154C39273004CA5D0F25AD9FBDE504B01D -info-plist-path /Users/rufflerr/Developer/iOS/dbutton/Build/Intermediates/ArchiveIntermediates/DBag/InstallationBuildProductsLocation/Applications/dButton.app/Info.plist

error: warning: Is a directory
```

mrackwitz:
The issue with determining the architecture should be resolved now.

CHANGELOG.md
```diff
@@ -6,6 +6,25 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ## Master
 
+##### Highlighted Enhancement That Needs Testing
+
+* De-duplicate Pod Targets: CocoaPods now recognizes when a dependency is used
+  multiple times across different user targets, but need to be built only once.
```
segiddins: `needs`

CHANGELOG.md
```diff
@@ -6,6 +6,25 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
...
+  The targets in Pods.xcodeproj need to be duplicated when one of the following
```
segiddins: `Pods.xcodeproj` in backticks

CHANGELOG.md
```diff
@@ -6,6 +6,25 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
...
+  applies:
+  * They are used on different platforms.
+  * They are used with differents sets of subspecs.
+  * They have any dependency, which needs to be duplicated.
```
segiddins: ~~,~~

CHANGELOG.md
```diff
@@ -6,6 +6,25 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
...
+
+  You can opt-out of this behavior installation-wise, by setting the following
+  option in your ``~/.cocoapods/config.yaml`:
+  ```yaml
+  deduplicate_targets: false
+  ```
+
+  [Marius Rackwitz](https://github.com/mrackwitz)
+  [#3550](https://github.com/CocoaPods/CocoaPods/issues/3550)
```
segiddins: link to the issues fixed, rather than the PR.

lib/cocoapods/generator/embed_frameworks_script.rb
```diff
@@ -1,24 +1,15 @@
...
-
       # @return [Hash{String, Array{String}] Multiple lists of frameworks per
       #         configuration.
       #
       attr_reader :frameworks_by_config
 
-      # @param  [TargetDefinition] target_definition
-      #         @see #target_definition
-      #
       # @param  [Hash{String, Array{String}] frameworks_by_config
```
segiddins: this is missing a `}`

lib/cocoapods/generator/embed_frameworks_script.rb
```diff
@@ -72,9 +63,9 @@ def script
 
             # Embed linked Swift runtime libraries
             local basename
-            basename=$(echo $1 | sed -E s/\\\\..+// && exit ${PIPESTATUS[0]})
+            basename=$(echo $1 | sed -E "s/(.+\\/)?([^/]+)\\\\..+/\\2/" && exit ${PIPESTATUS[0]})
```
segiddins: ummm, I don't quite understand this change?

lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
```diff
@@ -49,26 +49,61 @@ def save_as(path)
...
+          XCConfigHelper.add_target_specific_settings(target, @xcconfig)
+
+          generate_vendored_build_settings
+          generate_other_ld_flags
+
+          # TODO: Need to decide how we are going to ensure settings like these
+          # are always excluded from the user's project.
+          #
+          # See https://github.com/CocoaPods/CocoaPods/issues/1216
+          @xcconfig.attributes.delete('USE_HEADERMAP')
```
segiddins: add a constant array of settings to delete?

lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
```diff
@@ -49,26 +49,61 @@ def save_as(path)
...
+
+        #---------------------------------------------------------------------#
+
+        private
+
+        # Add build settings, which ensure that the pod targets can be imported
+        # from the integrating target by all sort of imports, which are:
+        #  - `#import <…>`
+        #  - `#import "…"`
+        #  - `@import …` / `@import …;`
```
segiddins: #  - `import …` / `@import …;`

lib/cocoapods/generator/xcconfig/private_pod_xcconfig.rb
```diff
@@ -60,13 +60,13 @@ def generate
...
 
-          if target.requires_frameworks?
+          if target.requires_frameworks? && target.scoped?
             # Only quote the FRAMEWORK_SEARCH_PATHS entry, because it’s a setting that takes multiple values.
             # In addition, quoting CONFIGURATION_BUILD_DIR would make it be interpreted as a relative path.
             build_settings = {
               'PODS_FRAMEWORK_BUILD_PATH' => target.configuration_build_dir,
-              'CONFIGURATION_BUILD_DIR' => '$PODS_FRAMEWORK_BUILD_PATH',
               'FRAMEWORK_SEARCH_PATHS' => '"$PODS_FRAMEWORK_BUILD_PATH"',
+              'CONFIGURATION_BUILD_DIR' => '$PODS_FRAMEWORK_BUILD_PATH',
```
segiddins: `$PODS_FRAMEWORK_BUILD_PATH` should be in quotes?

lib/cocoapods/generator/xcconfig/private_pod_xcconfig.rb
```diff
@@ -60,13 +60,13 @@ def generate
             # 'USE_HEADERMAP' => 'NO'
           }
 
-          if target.requires_frameworks?
+          if target.requires_frameworks? && target.scoped?
```
segiddins: why does this only apply to scoped targets?

lib/cocoapods/installer.rb
```diff
@@ -350,7 +350,7 @@ def lock_pod_sources
     def determine_dependency_product_types
       aggregate_targets.each do |aggregate_target|
         aggregate_target.pod_targets.each do |pod_target|
-          pod_target.host_requires_frameworks = aggregate_target.requires_frameworks?
+          pod_target.host_requires_frameworks ||= aggregate_target.requires_frameworks?
```
segiddins: why this change?

lib/cocoapods/installer/analyzer.rb
```diff
@@ -249,48 +253,119 @@ def generate_target(target_definition, specs)
...
+          all_specs = specs_by_target.flat_map do |target_definition, dependent_specs|
+            dependent_specs.group_by(&:root).map do |root_spec, specs|
+              [root_spec, specs, target_definition]
+            end
+          end
+
+          distinct_targets = all_specs.each_with_object({}) do |dependency, hash|
+            root_spec, specs, target_definition = *dependency
+            hash[root_spec] ||= {}
+            (hash[root_spec][[specs, target_definition.platform]] ||= []) << target_definition
```
segiddins: this needs to not be on one line

lib/cocoapods/installer/analyzer.rb
```diff
@@ -249,48 +253,119 @@ def generate_target(target_definition, specs)
...
+          end
+
+          pod_targets = distinct_targets.flat_map do |_, targets_by_distinctors|
```
segiddins: `targets_by_distinctors`?

lib/cocoapods/installer/analyzer.rb
```diff
@@ -249,48 +253,119 @@ def generate_target(target_definition, specs)
...
+            if targets_by_distinctors.count > 1
+              # There are different sets of subspecs or the spec is used across different platforms
+              targets_by_distinctors.map do |distinctor, target_definitions|
+                specs, _ = *distinctor
```
segiddins: what is a `distinctor`?

lib/cocoapods/installer/analyzer.rb
```diff
@@ -249,48 +253,119 @@ def generate_target(target_definition, specs)
...
-      #         the aggregate target
+      # @param  [Hash{Podfile::TargetDefinition => Array<Specification>}] specs_by_target
+      #         the resolved specifications grouped by target.
       #
-      # @param  [Array<Specification>] specs
-      #         the specifications that need to be installed.
+      # @return [Array<PodTarget>]
+      #
+      def generate_pod_targets(specs_by_target)
+        if config.deduplicate_targets?
```
segiddins: this should be factored into multiple methods. as it is, it's _incredibly_ complex

lib/cocoapods/installer/analyzer.rb
```diff
@@ -249,48 +253,119 @@ def generate_target(target_definition, specs)
...
+          specs_by_target.flat_map do |target_definition, specs|
+            grouped_specs = specs.group_by.group_by(&:root).values.uniq
+            grouped_specs.flat_map do |pod_specs|
+              generate_pod_target([target_definition], pod_specs, :scoped => true)
+            end
+          end
+        end
+      end
+
+      # Finds the names of the Pods on which the given target _transitively_
```
segiddins: `upon which`

lib/cocoapods/installer/analyzer.rb
```diff
@@ -249,48 +253,119 @@ def generate_target(target_definition, specs)
...
+      #         All pod targets, which are integrated alongside.
       #
       # @return [Array<PodTarget>]
       #
-      def generate_pod_targets(target, specs)
-        grouped_specs = specs.group_by(&:root).values.uniq
-        grouped_specs.map do |pod_specs|
-          generate_pod_target(target, pod_specs)
+      def transitive_dependencies_for_pod_target(pod_target, targets)
+        if targets.count > 0
```
segiddins: `targets.any?`

lib/cocoapods/installer/analyzer.rb
```diff
@@ -542,305 +617,36 @@ def sources
...
-        return target_definition.platform if target_definition.platform
-        name = nil
-        deployment_target = nil
-
-        user_targets.each do |target|
-          name ||= target.platform_name
-          raise Informative, 'Targets with different platforms' unless name == target.platform_name
-          if !deployment_target || deployment_target > Version.new(target.deployment_target)
-            deployment_target = Version.new(target.deployment_target)
+      def check_platform_specifications
```
segiddins: `verify_platforms_specified!`

lib/cocoapods/installer/analyzer.rb
```diff
@@ -542,305 +617,36 @@ def sources
...
+        unless config.integrate_targets?
+          podfile.target_definition_list.map do |target_definition|
```
segiddins: `each`

lib/cocoapods/installer/analyzer.rb
```diff
@@ -542,305 +617,36 @@ def sources
...
-              targets = compute_user_project_targets(target_definition, user_project)
-              compute_platform_for_target_definition(target_definition, targets)
-              archs = compute_archs_for_target_definition(target_definition, targets)
-              @archs_by_target_def[target_definition] = archs
-            else
-              unless target_definition.platform
-                raise Informative, 'It is necessary to specify the platform in the Podfile if not integrating.'
-              end
-            end
+          podfile.target_definition_list.map do |target_definition|
```
segiddins: `each`, or use `reduce`

lib/cocoapods/installer/analyzer.rb
```diff
@@ -542,305 +617,36 @@ def sources
...
+            inspection_result[target_definition] = TargetInspector.new(target_definition).inspect!
           end
```
segiddins: this should be returning `inspection_result`?

segiddins: nvm, it is

lib/cocoapods/installer/analyzer/target_inspection_result.rb
```diff
@@ -0,0 +1,41 @@
...
+        attr_accessor :platform
+
+        # @return [Array<String>] the architectures used by user's targets
+        #
+        attr_accessor :archs
+
+        # @return [Bool] whether frameworks are recommended for the integration
+        #         due to the presence of Swift source in the user's targets
+        #
+        attr_accessor :recommends_frameworks
```
segiddins: should be requires?

lib/cocoapods/installer/analyzer/target_inspection_result.rb
```diff
@@ -0,0 +1,41 @@
...
+        #         inspected
+        #
+        attr_accessor :target_definition
+
+        # @return [Pathname] the path of the user project that the
+        #         #target_definition should integrate
+        #
+        attr_accessor :project_path
+
+        # @return [Array<String>] the uuid of the user'S targets
```
segiddins: downcase the `S`

lib/cocoapods/installer/analyzer/target_inspector.rb
```diff
@@ -0,0 +1,196 @@
...
+        #
+        def initialize(target_definition)
+          @target_definition = target_definition
+        end
+
+        # Inspect the #target_definition
+        #
+        # @return [TargetInspectionResult]
+        #
+        def inspect!
```
segiddins: `inspect`

lib/cocoapods/installer/analyzer/target_inspector.rb
```diff
@@ -0,0 +1,196 @@
...
+              raise Informative, 'Unable to find the Xcode project ' \
+              "`#{path}` for the target `#{target_definition.label}`."
+            end
+          else
+            xcodeprojs = config.installation_root.children.select { |e| e.fnmatch('*.xcodeproj') }
+            if xcodeprojs.size == 1
+              path = xcodeprojs.first
+            else
+              raise Informative, 'Could not automatically select an Xcode project. ' \
+              "Specify one in your Podfile like so:\n\n" \
```
segiddins: these two lines need to be indented

lib/cocoapods/installer/analyzer/target_inspector.rb
```diff
@@ -0,0 +1,196 @@
...
+        # @param  [Xcodeproj::Project] user_project
+        #         the user project
+        #
+        # @return [Array<PBXNativeTarget>]
+        #
+        def compute_targets(user_project)
+          native_targets = user_project.native_targets
+          if link_with = target_definition.link_with
+            targets = native_targets.select { |t| link_with.include?(t.name) }
+            raise Informative, "Unable to find the targets named `#{link_with.to_sentence}` to link with target definition `#{target_definition.name}`" if targets.empty?
```
segiddins: `link_with.to_sentence` shouldn't be inside backticks, since it would put the whole sentence inside the one pair

lib/cocoapods/installer/analyzer/target_inspector.rb
```diff
@@ -0,0 +1,196 @@
...
+        # @param  [Array<PBXNativeTarget] the user's targets of the project of
+        #         #target_definition which needs to be integrated
+        #
+        # @return [Hash{String=>Symbol}] A hash representing the user build
+        #         configurations where each key corresponds to the name of a
+        #         configuration and its value to its type (`:debug` or `:release`).
+        #
+        def compute_build_configurations(user_targets)
+          if user_targets
+            user_targets.map { |t| t.build_configurations.map(&:name) }.flatten.each_with_object({}) do |name, hash|
```
segiddins: `flat_map`?

lib/cocoapods/installer/analyzer/target_inspector.rb
```diff
@@ -0,0 +1,196 @@
+module Pod
+  class Installer
+    class Analyzer
+      class TargetInspector
+        include Config::Mixin
```
segiddins: since this is only used for `config.installation_root`, please pass that in as an arg to the initializer?

lib/cocoapods/installer/target_installer.rb
```diff
@@ -67,7 +67,7 @@ def add_target
       def custom_build_settings
         settings = {}
 
-        if target.archs
+        unless target.nil? || target.archs.empty?
```
segiddins: when will `target.nil?`

lib/cocoapods/target/pod_target.rb
```diff
@@ -3,31 +3,86 @@ module Pod
...
       @sandbox = sandbox
+      @scoped = scoped
       @build_headers  = Sandbox::HeadersStore.new(sandbox, 'Private')
       @file_accessors = []
       @resource_bundle_targets = []
     end
 
+    # @return [PodTarget] the same target, but scoped.
+    #
+    def scoped
```
segiddins: this will interfere with the `attire_reader` generated method

lib/cocoapods/target/pod_target.rb
```diff
@@ -3,31 +3,86 @@ module Pod
...
-      @target_definition = target_definition
+      @target_definitions = target_definitions
       @sandbox = sandbox
+      @scoped = scoped
       @build_headers  = Sandbox::HeadersStore.new(sandbox, 'Private')
       @file_accessors = []
       @resource_bundle_targets = []
     end
 
+    # @return [PodTarget] the same target, but scoped.
```
segiddins: returns an array

lib/cocoapods/target/pod_target.rb
```diff
@@ -125,14 +183,51 @@ def include_in_build_config?(configuration_name)
...
+        return false
+      elsif whitelists.count == 1
+        whitelists.first
+      else
+        UI.warn "The pod `#{pod_name}` is linked to different targets " \
+          "(#{target_definitions.map(&:label)}), which contain different " \
+          'settings to inhibit warnings. CocoaPods does not currently ' \
+          'support different settings and will fall back to your preference ' \
+          'set in the root target definition.'
+        return podfile.root_target_definitions.first.inhibits_warnings_for_pod?(root_spec.name)
```
segiddins: imho should default to `true` if any are true, but continue with the warning

spec/spec_helper/fixture.rb
```diff
@@ -9,6 +9,10 @@ def self.create_sample_app_copy_from_fixture(fixture_name)
     fixture_copy_path + "#{fixture_name}.xcodeproj"
   end
 
+  def self.test_repo_url
+    'https://github.com/CocoaPods/test_repo.git'
```
segiddins: 404s?

spec/unit/generator/xcconfig/aggregate_xcconfig_spec.rb
```diff
@@ -107,8 +111,20 @@ def spec
             @xcconfig.to_hash['OTHER_CFLAGS'].should == expected
           end
 
-          it 'links the pod targets with the aggregate target' do
-            @xcconfig.to_hash['OTHER_LDFLAGS'].should.include '-l"Pods-BananaLib"'
+          describe 'with a scoped pod target' do
+            def pod_target(spec)
```
segiddins: this doesn't seem to be used?

spec/unit/installer/analyzer/target_inspector_spec.rb
```diff
@@ -0,0 +1,262 @@
...
+        target_a.build_configuration_list.set_setting('ARCHS', %w(armv7 armv7s))
+        target_b = user_project.new_target(:application, 'Target', :ios)
+        target_b.build_configuration_list.set_setting('ARCHS', %w(armv7 i386))
+
+        target_definition = Podfile::TargetDefinition.new(:default, nil)
+        target_definition.set_platform(:ios, '4.0')
+        user_targets = [target_a, target_b]
+
+        archs = TargetInspector.new(target_definition).send(:compute_archs, user_targets)
+        %w(armv7 armv7s i386).each { |a| archs.should.include a }
```
segiddins: should be `==`? (can sort first)

spec/unit/installer/analyzer_spec.rb
```diff
@@ -87,16 +90,18 @@ module Pod
...
-          'Pods-AFNetworking',
-          'Pods-SVPullToRefresh',
+          'JSONKit',
+          'AFNetworking',
+          'SVPullToRefresh',
           'Pods-libextobjc',
         ].sort
         target.support_files_dir.should == config.sandbox.target_support_files_dir('Pods')
 
+        target.pod_targets.map(&:archs).uniq.should == [[]]
```
segiddins: why should this be empty?

spec/unit/target/pod_target_spec.rb
```diff
@@ -5,21 +5,38 @@ module Pod
...
       spec = fixture_spec('banana-lib/BananaLib.podspec')
       @target_definition = Podfile::TargetDefinition.new('Pods', nil)
-      @pod_target = PodTarget.new([spec], @target_definition, config.sandbox)
+      @pod_target = PodTarget.new([spec], [@target_definition], config.sandbox)
       @pod_target.stubs(:platform).returns(:ios)
     end
 
+    describe 'Meta' do
+      describe '#scoped' do
+        it 'returns a cloned target, which is scoped' do
```
segiddins: where is this checking the cloned target?

segiddins:
Amazing work, @mrackwitz !

basememara:
Thanks for all the hard work on this! I'm trying to use the `mr-deduplicate-frameworks` branch to archive but ran into the same issue as @robruffler. This is the error:
```
PhaseScriptExecution Embed\ Pods\ Frameworks /Users/.../Library/Developer/Xcode/DerivedData/.../ArchiveIntermediates/.../IntermediateBuildFilesPath/....build/Release-iphoneos/...-....sh
    cd "/Users/.../Solutions/Xcode/..."
    /bin/sh -c \"/Users/.../Library/Developer/Xcode/DerivedData/.../Build/Intermediates/ArchiveIntermediates/.../IntermediateBuildFilesPath/....build/Release-iphoneos/....build/Script-....sh\"

mkdir -p /Users/.../Library/Developer/Xcode/DerivedData/.../Build/Intermediates/ArchiveIntermediates/.../BuildProductsPath/Release-iphoneos/....app/Frameworks
rsync -av --filter - CVS/ --filter - .svn/ --filter - .git/ --filter - .hg/ --filter - Headers/ --filter - PrivateHeaders/ --filter - Modules/ /Users/.../Library/Developer/Xcode/DerivedData/.../Build/Intermediates/ArchiveIntermediates/.../BuildProductsPath/Release-iphoneos/Pods/MZTimerLabel.framework /Users/.../Library/Developer/Xcode/DerivedData/.../Build/Intermediates/ArchiveIntermediates/.../BuildProductsPath/Release-iphoneos/....app/Frameworks
building file list ... rsync: link_stat "/Users/.../Library/Developer/Xcode/DerivedData/.../Build/Intermediates/ArchiveIntermediates/.../BuildProductsPath/Release-iphoneos/Pods/MZTimerLabel.framework" failed: No such file or directory (2)
done

sent 29 bytes  received 20 bytes  98.00 bytes/sec
total size is 0  speedup is 0.00
rsync error: some files could not be transferred (code 23) at /SourceCache/rsync/rsync-45/rsync/main.c(992) [sender=2.6.9]
Command /bin/sh failed with exit code 23
```
So I'm trying to merge in #3675 as @mrackwitz suggested and modified the source code locally for `lib/cocoapods/generator/embed_frameworks_script.rb`. But can someone pls advise how to build the pod in the bin with this new change in the source so I can do another .../pod install? I'm almost there now to see if this fixes it but need help on that part.

lib/cocoapods/generator/embed_frameworks_script.rb
mrackwitz: This dropped before only the file-suffix, now it also drops the dir path.
```
a.framework => a
a/b.framework => b
a/b/c.framework => c
```

CHANGELOG.md
mrackwitz: There is no single proper issue.

lib/cocoapods/installer/analyzer.rb
mrackwitz: Maybe it's a good idea to rename that method as it is very near to stdlib's debugging helper `#inspect`.

lib/cocoapods/generator/embed_frameworks_script.rb
segiddins: Please add a comment to that effect

mrackwitz: A comment in generated source code? Nope :-1:
There is no appropriate place to add such comment.

lib/cocoapods/installer/target_installer.rb
mrackwitz: Probably tests only. Should be initialized as empty array. :+1:

lib/cocoapods/generator/embed_frameworks_script.rb
segiddins: ah true. please do find a place for such a comment through.

lib/cocoapods/target/pod_target.rb
mrackwitz: This is desired in this case. The reader is aliased before as `:scoped?`.

lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
mrackwitz: This wasn't introduced. This is just kept in the code as it was after some refactorings.

lib/cocoapods/generator/xcconfig/private_pod_xcconfig.rb
mrackwitz: Because setting the `CONFIGURATION_BUILD_DIR` here causes that scoped targets are actually scoped.

lib/cocoapods/installer.rb
mrackwitz: As the deduplication doesn't care about the build type, this will opt-in all `PodTarget` to frameworks, which are shared as a deduplicated dependency with a target definition, which is opted-in to use frameworks. Otherwise it would seem to be less deterministic, whether a `PodTarget` is build as framework or not.

lib/cocoapods/installer/analyzer.rb
mrackwitz: No skipping the first is something different. This would be perhaps more obvious:
```diff
- distinct_targets.flat_map do |_, targets_by_distinctors|
+ distinct_targets.values.flat_map do |targets_by_distinctors|
```

lib/cocoapods/installer/analyzer.rb
mrackwitz: An array of criterias, which must be equal for all root_specs of a dependency, so that it can be deduplicated. Any ideas how to express that shorter in a var name?

lib/cocoapods/installer/analyzer.rb
mrackwitz: It must be `each`, `UI.section` would swallow the return value otherwise.

lib/cocoapods/installer/analyzer/target_inspection_result.rb
mrackwitz: No, this is intentional. This is something, which wasn't propagated to UI yet. The analyzer can detect whether Swift source code is present in the user target and recommends to use `use_frameworks! `.

lib/cocoapods/installer/analyzer/target_inspector.rb
mrackwitz: :+1:

lib/cocoapods/installer/analyzer/target_inspector.rb
mrackwitz: see my comment above, other name ideas here?

spec/spec_helper/fixture.rb
mrackwitz: This will 404 until #3642 is implemented. But this shouldn't block this PR.

spec/unit/generator/xcconfig/aggregate_xcconfig_spec.rb
mrackwitz: But it is, it is defined in the root context and used in the before block there, but overwritten on this sub-context.

spec/unit/installer/analyzer_spec.rb
mrackwitz: Because the analyzer should recognize that the user targets do not overwrite this build setting and won't overwrite it in the pod targets. Before it was nil, a string or an array.

lib/cocoapods/installer.rb
mrackwitz: This is debatable, whether that should be a separate criteria or not. I would address this in a separate PR.

lib/cocoapods/installer/analyzer.rb
segiddins: I'd personally prefer `criteria` to `distinctor`, but that's entirely personal opinion

mrackwitz: I will take this into account in #3714.

CHANGELOG.md
kylef: It's fine to link to every issue that this closes. But you shouldn't link to the pull request that introduced the change.

The link you've used should be the pull request and not an issue:

```diff
- https://github.com/CocoaPods/CocoaPods/issues/3550
+ https://github.com/CocoaPods/CocoaPods/pull/3550
```

CHANGELOG.md
```diff
@@ -6,6 +6,25 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
...
+* De-duplicate Pod Targets: CocoaPods now recognizes when a dependency is used
+  multiple times across different user targets, but needs to be built only once.
+  The targets in `Pods.xcodeproj` need to be duplicated when one of the following
+  applies:
+  * They are used on different platforms.
+  * They are used with differents sets of subspecs.
+  * They have any dependency which needs to be duplicated.
+
+  You can opt-out of this behavior installation-wise, by setting the following
+  option in your ``~/.cocoapods/config.yaml`:
```
kylef: Wouldn't the use case of this flag be pre project not per installation?

spec/unit/generator/embed_frameworks_script_spec.rb
kylef: Perhaps that should be a new explicit test case than modifying an existing one to do this.

CHANGELOG.md
mrackwitz: I discussed that with @orta whether we want add something like `duplicate_targets!` instead to the Podfile, but as long as it's not broken, we don't really want offer both.

spec/unit/generator/embed_frameworks_script_spec.rb
mrackwitz: I don't see any advantage of a separate test for that. The history documents why this change was introduced. It's still a legit test scenario.

lib/cocoapods/installer/analyzer.rb
mrackwitz: See #3714.

neonichu:
:tada: :wine_glass: :beers:

k0nserv:
Great job
http://media3.giphy.com/media/EQIDJxDloq1zO/giphy.gif

sbhklr:
Great job! Already any plans for releasing it?

ItsTipTop:
Great! I installed 0.38.0.beta.1 and Xcode successfully archived my application. But when i'm trying to submit my app to App Store xcode fail:
"No matching provisioning profiles found for.." app and Extension too.

tanis2000:
What if you're using the same Pod like AFNetworking both in the application and the widget? You need to be able to set AF_APP_EXTENSIONS=1 in GCC_PREPROCESSOR_DEFINITIONS for the widget but you cannot do it as long as the targets get deduplicated. And since we're a team working on the same project it's not reasonable to force everyone to remember to add something to the config.yml 
Is there any other way around this issue?

neonichu:
See discussion in #3794

PR #3553, opened 2015-05-11
Reject adding a source file twice in the compile source build phase 

orta:
This happens to me when it has come in through two different subspecs. I think with static libs it would ignore this but with frameworks it's a different matter. 

Demoable by adding this to a podspec
```
pod 'ARAnalytics', :subspecs => ["Mixpanel", "HockeyApp", "DSL"]
```

Looking at tests etc now.

### DIFF
--- a/lib/cocoapods/installer/target_installer/pod_target_installer.rb
+++ b/lib/cocoapods/installer/target_installer/pod_target_installer.rb
@@ -63,7 +63,10 @@ module Pod
           }.each do |arc, files|
             files = files - headers - other_source_files
             flags = compiler_flags_for_consumer(consumer, arc)
-            regular_file_refs = files.map { |sf| project.reference_for_path(sf) }
+            regular_file_refs = files.map { |sf| project.reference_for_path(sf) }.reject do | file_ref|
+              native_target.source_build_phase.include? file_ref
+            end
+
             native_target.add_file_references(regular_file_refs, flags)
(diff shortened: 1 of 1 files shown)


orta:
Fails look unrelated?

lib/cocoapods/installer/target_installer/pod_target_installer.rb
```diff
@@ -63,7 +63,10 @@ def add_files_to_build_phases
           }.each do |arc, files|
             files = files - headers - other_source_files
             flags = compiler_flags_for_consumer(consumer, arc)
-            regular_file_refs = files.map { |sf| project.reference_for_path(sf) }
+            regular_file_refs = files.map { |sf| project.reference_for_path(sf) }.reject do | file_ref|
```
segiddins: extra space

segiddins:
This logic should probably also be duplicated for `header_file_refs` and `other_file_refs`

segiddins:
@orta do you want me to finish this one up for you?

orta:
Yeah!

orta:
I took a second look at it, but couldn't think of an elegant way to do it DRY-ly

segiddins:
Ah I can't because the branch is on your fork

orta:
I'll close this, feel free to re-implement yourself. I only added a line or two.

PR #3604, opened 2015-05-26
Symlink vendored framework headers into the public headers dir

### LINKED ISSUE #3161: Using a pod (which produce framework) in another pod
Hi,
recently Google has updated its pod `Google-Mobile-Ads-SDK`, which now produce a framework. I have a public podspec called `MUKAdMobViewController` which has a dependency to it, and I would like to support new major version.
I paste you podspec:
```
Pod::Spec.new do |s|
  s.name      = 'MUKAdMobViewController'
  s.platform  = :ios, '6.0'
  s.version   = '1.2.4'
  s.summary   = 'View controller which manages an AdMob banner, geolocation and interstitial presentation.'

  s.source = {
    :git => 'https://github.com/muccy/MUKAdMobViewController.git',
    :tag => s.version.to_s
  }
  
  s.requires_arc    = true
  s.source_files    = 'MUKAdMobViewController/*.{h,m}'
  s.compiler_flags  = '-Wdocumentation'
  s.frameworks      = 'CoreLocation'
  
  s.dependency 'Google-Mobile-Ads-SDK', '~> 6.6'
end
```

I bumped up dependency version (`s.dependency 'Google-Mobile-Ads-SDK', '~> 7.0'`) and problems started. I have an example app created to develop `MUKAdMobViewController`.
schermata 2015-02-19 alle 15 45 22

On top you can see two files of my pod: they are actually built by `Pods-MUKAdMobViewController` target, as you know. And that is why header files can not be found. There is no reference to `GoogleMobileAds.framework` there.
schermata 2015-02-19 alle 15 48 06

Before they adopted framework there was no problem. 
Before to post an issue I tried circumvent the problem by setting manually linker flags like `HEADER_SEARCH_PATHS` or `-framework` in `OTHER_LDFLAGS`, but with no success.

How can I do? Thanks.

vincentisambart:
Fixes #3161

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ To install or update CocoaPods see this [guide](http://docs.cocoapods.org/guides
 
 To install release candidates run `[sudo] gem install cocoapods --pre`
 
+## Master
+
+### Bug Fixes
+
+* Pods can now include headers from vendored frameworks from other pods.
+  [Vincent Isambart](https://github.com/vincentisambart)
+  [#3161](https://github.com/CocoaPods/CocoaPods/issues/3161)
+
 ## 0.37.2
 
 ##### Enhancements
--- a/lib/cocoapods/installer/file_references_installer.rb
+++ b/lib/cocoapods/installer/file_references_installer.rb
@@ -223,10 +223,6 @@ module Pod
       #         The consumer file accessor for which the headers need to be
       #         linked.
       #
-      # @return [Hash{Pathname => Array<Pathname>}] A hash containing the
-      #         headers folders as the keys and the absolute paths of the
-      #         header files as the values.
-      #
       def vendored_frameworks_header_mappings(headers_sandbox, file_accessor)
         mappings = {}
         file_accessor.vendored_frameworks.each do |framework|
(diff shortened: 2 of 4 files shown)


vincentisambart:
As proposed in #3161, it links the headers from vendored frameworks to the public headers directory.
Not perfect, no CHANGELOG entry yet, but first I wanted some feedback.
At first I thought about modifying FRAMEWORK_SEARCH_PATH of each target (adding the frameworks from it dependencies) but it seems that would require some heavy refactoring so I went with the easy solution.

neonichu:
This generally looks good to me, could you please fix the rubocop violations?

vincentisambart:
Thank you.
I fixed the rubocop violations, and updated a CHANGELOG entry.

neonichu:
Awesome, thanks a lot for your contribution, @vincentisambart :tada:

mrackwitz:
>At first I thought about modifying FRAMEWORK_SEARCH_PATH of each target (adding the frameworks from it dependencies) but it seems that would require some heavy refactoring so I went with the easy solution.

In the long run, that won't be a sufficient solution. Publicizing the headers does affect the ability to import headers from Objective-C / clang's tooling, but it won't be possible to use module imports, which makes it harder to use transitive vendored dependencies from Swift pods.

I think for now we are good to go with this solution, but once we solve #1993 and vendored dynamic frameworks are actually working, we will need to implement the more difficult solution.

neonichu:
@mrackwitz you're right, I added a point for this to the task list in #1993

PR #3610, opened 2015-05-27
[Travis] Only build master and stable branches

segiddins:
\c @kylef


kylef:
Won't this mean it won't test pull requests?

segiddins:
As you can see from this PR, it will still test pull requests into one of the tested branches

kylef:
What's the justification / reason for this change?

segiddins:
It means we won't build branches that are just PR-ed into master/stable twice.

segiddins:
Unless you seriously object, I'd like to merge this.

PR #3622, opened 2015-05-30
[Cache] Remove cache when CocoaPods version changes

segiddins:
Closes https://github.com/CocoaPods/CocoaPods/issues/3542


segiddins:
@kylef @mrackwitz this should be fine code-wise, as long as we're happy with the end effect

PR #3629, opened 2015-05-31
Added recursive support for header sub-directories

### LINKED ISSUE #3161: Using a pod (which produce framework) in another pod

vhariable:
Fixes #3161


vhariable:
Preparing CHANGELOG update now

vhariable:
This PR is ready for review. This fixes an issue with the previous fix for #3161 where sub-directories in the framework headers directory are not copied into Headers/Public/FrameworkName. This fix creates symlinks recursively, preserving the entire framework headers directory. Tested with GoogleMobileAds framework, which contains sub-directories.

segiddins:
:+1: can we just get a quick spec on this?

vhariable:
Sure, just let me know what you need. I am new to contributing to anything on GH. Do you mean a comment with more details or its own issue?

This is all related to #3161, but the previous fix didn't handle sub directories in a framework.

segiddins:
We just need a test added that shows this does what it says on the tin, and to make sure there are no future regressions

vhariable:
No problem, this is my first time to use Ruby and their term for a unit testing threw me. Will be later tomorrow evening before I get a chance to implement a test. In doing more testing I noticed an issue with the original fix for #3161 (not mine/this PR), and seeing if I can address that as part of this PR.

The original fix basically copies the headers from a vendor framework into Headers/Public/SpecName/FrameworkName/\*.h (ignored sub-directories completely). The naming convention is actually correct now as the framework name should be there so that existing code that needs to use \<Framework/Framework.h\> works, even when the spec.name of the podspec might be different than the vendor framework. This is an issue I noticed previous to finding the fix to #3161. So that is just background, as that now works with the previous fix for #3161.

The new issue I see as a side-effect, might not be that big of an issue, but if a podspec contains both vendor_frameworks and public_header_files entries, then headers are placed into Headers/Public/SpecName/\*.h and also into the Headers/Public/SpecName/FrameworkName/\*.h. I do not believe the public_header_files should even be imported if there is a matching vendor_framework already.

Here is an example of a pod that doesn't work at all in the stable build because of the different spec.name from the framework name as an example. https://github.com/CocoaPods/Specs/blob/master/Specs/Heyzap/8.2.1/Heyzap.podspec.json

In that situations all examples in their document show to use #import <HeyzapAds/HeyzapAds.h> and it will not be found, but requires changing to #import <Heyzap/HeyzapAds.h>. Again, this is before the original fix for #3161, just some background. Also, this is only an issue when creating a static framework that you use in another project (both using pods, the application refs a podspec for the custom static framework), as far as I know.

Unless anyone has any objections I will work up a spec, but might also try to tackle cleaning up the original fix for #3161 to not include public_header_files, if we are already going to import those same headers already via the vendor_frameworks.

vhariable:
@segiddins Please let me know if you need anything else. This includes spec updates, and also fixes the  issues I had described above with the duplicate header files in the Headers/Public when they are also included via the vendored_frameworks. This works for many different podspec configurations, seems like some of those shouldn't even be possible.

lib/cocoapods/installer/file_references_installer.rb
```diff
@@ -115,16 +115,17 @@ def link_headers
         UI.message '- Linking headers' do
           libraries.each do |library|
             library.file_accessors.each do |file_accessor|
+              framework_exp = Regexp.new "/(#{file_accessor.spec_consumer.send(:vendored_frameworks).join(')|(')})/"
```
segiddins: let's use a regex literal here i.e. just `/foo/`, no intermediary string necessary

lib/cocoapods/installer/file_references_installer.rb
```diff
@@ -115,16 +115,17 @@ def link_headers
...
               headers_sandbox = Pathname.new(file_accessor.spec.root.name)
               library.build_headers.add_search_path(headers_sandbox, library.platform)
               sandbox.public_headers.add_search_path(headers_sandbox, library.platform)
 
               header_mappings(headers_sandbox, file_accessor, file_accessor.headers).each do |namespaced_path, files|
-                library.build_headers.add_files(namespaced_path, files, library.platform)
+                library.build_headers.add_files(namespaced_path, files.select { |f| framework_exp.match(f.to_path).nil? }, library.platform)
```
segiddins: `f.to_path =~ framework_exp`

segiddins:
:+1:, just want @kylef or @neonichu to take a quick look over this!

neonichu:
LGTM :+1:

lib/cocoapods/sandbox/file_accessor.rb
```diff
@@ -168,7 +168,7 @@ def self.vendored_frameworks_headers_dir(framework)
       #
```
mrackwitz: See above in the title line: the file permissions changed from `0644` to `0775` for this file. Please revert this change.

neonichu:
Thanks for this, @vhariable :tada:

vhariable:
No problem!

PR #3640, opened 2015-06-03
Move source locking to installer, only lock source files (#3529, #3557)

### LINKED ISSUE #3529: Post-install hook: Permission denied
We have a post-install step which removes AFNetworkings UIImageView category:
```ruby
    pod.source_files.each do |pathname|
        if pathname.to_s.include? "/AFNetworking/UIImageView"
            File.open(pathname, 'w').write("/* REMOVED USE SDWEBIMAGE INSTEAD */")
        end
    end
```

As of 0.37 (0.37.0 and 0.37.1) this no longer works:
```
Permission denied - /PATH_TO_APP/Pods/AFNetworking/AFNetworking/UIImageView+AFNetworking.h
```

### LINKED ISSUE #3557: Read Only File Permissions
With the new Cocoapods update, all files are written under read-only permissions.
This is causing an issue for users who use .sqlite database files that were already created but only moved/copied to the device.
Because the permissions are read-only, when the database is open, it is open as read-only as well. This obviously breaks if we try to write any data to that database.

I understand that this was only changed recently. But can we have any options to all read-write files?
Meanwhile, I will downgrade to 0.36.1 as it was the last version that worked without read-only permissions

mglidden:
Fixes #3557 by only locking source files. I couldn't figure out how to get to the source files from the Pod::Installer::PodSourceInstaller, so I had to move it to Pod::Installer. 

I also unlocked the source files before hooks run, which fixes #3529.

### DIFF
--- a/lib/cocoapods/installer.rb
+++ b/lib/cocoapods/installer.rb
@@ -109,7 +109,6 @@ module Pod
       generate_pods_project
       integrate_user_project if config.integrate_targets?
       perform_post_install_actions
-      lock_source_files
     end
 
     def prepare
@@ -140,9 +139,9 @@ module Pod
       UI.section 'Downloading dependencies' do
         create_file_accessors
         install_pod_sources
-        unlock_source_files
         run_podfile_pre_install_hooks
         clean_pod_sources
+        lock_pod_sources
       end
     end
 
@@ -338,6 +337,19 @@ module Pod
       return unless config.clean?
       return unless @pod_installers
       @pod_installers.each(&:clean!)
+    end
+
+    # Locks the sources of the Pods if the config instructs to do so.
+    #
+    # @todo Why the @pod_installers might be empty?
+    #
+    def lock_pod_sources
+      return unless config.lock_pod_source?
+      return unless @pod_installers
+      @pod_installers.each do |installer|
+        pod_target =  pod_targets.detect {|target| target.name == installer.name}
+        installer.lock_files!(pod_target.file_accessors)
+      end
     end
 
     # Determines if the dependencies need to be built as dynamic frameworks or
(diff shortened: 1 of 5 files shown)


segiddins:
This belongs in the pod source installer -- use the file accessors to grab the source files.

mglidden:
How do I get to the file accessors from within the pod source installer?

segiddins:
I think it just has an array of them?

-Samuel E. Giddins

On Jun 2, 2015, at 11:44 PM, Mason Glidden <notifications@github.com> wrote:

How do I get to the file accessors from within the pod source installer?

lib/cocoapods/installer.rb
```diff
@@ -729,6 +720,44 @@ def run_podfile_post_install_hook
...
+    # @return [void]
+    #
+    def unlock_source_files
+      non_local_source_files do |file|
+        new_permissions = File.stat(file).mode | 0222
+        File.chmod(new_permissions, file)
+      end
+    end
+
+    def non_local_source_files
```
segiddins: please use an `&block` parameter

lib/cocoapods/installer.rb
```diff
@@ -729,6 +720,44 @@ def run_podfile_post_install_hook
...
+      installer = installer_rep
```
segiddins: also, let's not use the `installer_rep`?

mglidden:
I still can't figure out how to get from the Pod source installer to the file accessors. There's nothing directly in the class itself, and, as far as I can tell, neither it's sandbox or the specs_by_platform ivars have the file accessors either.

segiddins:
Yeah, you’re right. I’ll let @neonichu decide if it’s OK to leave this code in the installer

neonichu:
I'd say let's pass the `pod_targets` to `PodSourceInstaller`, similar to <https://github.com/CocoaPods/CocoaPods/blob/master/lib/cocoapods/installer.rb#L519> - then move the code back there.

segiddins:
@mglidden ping on @neonichu's comments?

kylef:
@mglidden Would you be able to update this? Would be great to get this in for 0.38.

mglidden:
I can update it this weekend – is that in time for 0.38?

segiddins:
@mglidden whenever you can update this would be great

mglidden:
Updated with master and incorporated the feedback today. I still have to pull in a fix for #3529 and make sure all the tests run.

segiddins:
@mglidden tests are failing

lib/cocoapods/installer/pod_source_installer.rb
```diff
@@ -66,8 +71,22 @@ def clean!
...
-        lock_installation unless local?
+      def lock_files!(file_accessors)
+        if local?
+          return
+        end
+
+        file_accessors.each do |file_accessor|
+          file_accessor.source_files.each do |source_file|
+            if File.exist?(source_file)
+              file = source_file.open
```
segiddins: this shouldn't be necessary?

lib/cocoapods/installer.rb
```diff
@@ -346,7 +346,10 @@ def clean_pod_sources
     def lock_pod_sources
       return unless config.lock_pod_source?
       return unless @pod_installers
-      @pod_installers.each(&:lock_files!)
+      @pod_installers.each do |installer|
+        pod_target =  pod_targets.detect {|target| target.name == installer.name}
```
segiddins: `select`

segiddins:
Manually merged via https://github.com/CocoaPods/CocoaPods/commit/e238edee3fe5c2bd5d94f768a20beeb163a52e8c.

PR #3657, opened 2015-06-06
[IntegrationSpecs] Use developer tools when available

segiddins:
\c @mrackwitz

### DIFF
--- a/spec/integration.rb
+++ b/spec/integration.rb
@@ -50,6 +50,11 @@ require 'clintegracon'
 require 'integration/xcodeproj_project_yaml'
 require 'tmpdir'
 
+if (developer_bin = `xcode-select -p 2>/dev/null`.strip) && $?.success?
+  developer_bin = Pathname(developer_bin) + 'usr/bin'
+  ENV['PATH'] = "#{developer_bin}#{File::PATH_SEPARATOR}#{ENV['PATH']}"
+end
+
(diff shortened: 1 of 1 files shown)


mrackwitz:
Do we really want that? We're lately very often on different versions of the toolchain and it would be better if we use similar versions of the tools, which is perhaps more likely if we use either the version of Xcode on the default path or the version shipped with the OS.

segiddins:
That's why this is using what apples ships with Xcode, since I have a never version of git that breaks the integration specs.

segiddins:
I'm merging this unless anyone objects

spec/integration.rb
```diff
@@ -94,6 +94,11 @@
   Process.wait(spawn('which hg', :err => :out, :out => '/dev/null'))
   has_mercurial = $?.success?
 
+  if (developer_bin = `xcode-select -p 2>/dev/null`.strip) && $?.success?
+    developer_bin = Pathname(developer_bin) + 'usr/bin'
+    ENV['PATH'] = "#{developer_bin}#{File::PATH_SEPARATOR}#{ENV['PATH']}"
```
mrackwitz: As this has global side-effects, I would also put it in the global scope at the beginning of the file.

segiddins: :+1:

PR #3753, opened 2015-06-29
[Installer] Add support for default plugins

segiddins:
\c @orta

### DIFF
--- a/lib/cocoapods/installer.rb
+++ b/lib/cocoapods/installer.rb
@@ -418,7 +418,7 @@ module Pod
     #
     def run_plugins_pre_install_hooks
       context = PreInstallHooksContext.generate(sandbox, podfile, lockfile)
-      HooksManager.run(:pre_install, context, podfile.plugins)
+      HooksManager.run(:pre_install, context, plugins)
     end
 
     # Performs any post-installation actions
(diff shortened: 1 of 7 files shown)


segiddins:
@orta that CI failure includes an exception in stats

orta:
Thanks, will take a look

orta:
alright, I have a sense of this, will see if I can take over this PR today/tomorrow

segiddins:
@orta, @kylef  this should be good to go, please review. I'll update the integration specs when I merge.

lib/cocoapods/installer.rb
```diff
@@ -453,6 +453,17 @@ def ensure_plugins_are_installed!
       end
     end
 
+    DEFAULT_PLUGINS = { 'cocoapods-stats' => {} }
```
kylef: Could we get an explicit test to check that the stats plugin is enabled by default? And also to ensure its gone when the disable stats end variable is set (unless it's handled elsewhere, but I feel this is the place).

segiddins: nope, it's disabled inside of stats itself

segiddins: (and that test is handled by the integration specs, no?)

segiddins:
Will merge when green.

Issue #3190, opened 2015-02-24
Proper Sources plugin API


segiddins:
We should make the SourcesManager more extensible for CocoaPods plugins.

clarkda:
! Thanks @kylef

What did you have in mind @segiddins?

segiddins:
No idea yet! As you're likely going to be the first consumer, it would be helpful to know what sort of API you'd need.

amorde:
Any news on this? I am currently writing a plugin to support declaring private repositories as described in #3061 because there is still no way to have a private repository that is not using git. Overriding the behavior of SourcesManager is clunky and could easily break

segiddins:
@amorde we're still waiting for someone to propose a concrete API and/or PR an implementation, if you'd like to help :)

amorde:
Would love to! I'm new to Ruby so I'm not the best person to do it, but I'll see what I can do.

clarkda:
Sorry guys! I've been covered up, but I'm working an issue that will likely reveal some pain points, let me fix this, get in a pr and I'll have more to say.

:cake: :cake:

amorde:
An option could be to allow plugins to register themselves as source providers, so that the sources manager will ask all source providers for their sources. 

For example:
    
    Pod::HooksManager::register('cocoapods-repo-svn',:source_providers) do |context|
         # logic to return list of Pod::Source objects
    end
    
    # in the Podfile
    plugin 'cocoapods-repo-svn', {
         :sources => [
              'http://svn.myrepo.com',
          ]
     }

    # somewhere in SourcesManager
    source_providers.each do |provider|
       provider_sources = run_source_provider_hooks
       provider_sources.each do |source|
           sources << source # <-- Pod::Source object
       end
    end

Super rough psuedo code but hopefully you get the idea. Does that sound like a good way to go about it? I'm not an experienced contributor to Cocoapods so I might me thinking of it in the wrong way

segiddins:
I suppose that approach could work -- it would basically allow the plugin to act as a factory for `Pod::Source`s

orta:
Yeah, this seems like a good idea to me :+1:

amorde:
The way I am doing it now is sending a subclass of Pod::Source, which also allows for more flexibility. I'm curious if @clarkda has other ideas

clarkda:
[...]

PR #3792, opened 2015-07-07
[Installer] Add support for source provider plugin hook

amorde:
This PR provides an initial/example implementation for the source provider plugin api described in #3190

I very much welcome suggestions for improvements - still new to both Ruby and the CocoaPods code base so it is very likely I am missing something.

The API would allow plugins to supply their own sources, whether they be un-versioned or using an unsupported VCS such as CVS, Hg, SVN, etc.

Example:

```ruby
# in a Podfile
plugin 'cocoapods-repo-svn', {
    :sources => [
        'https://svn.myrepository.com'
    ]
}

# Now in the plugin's cocoapods_plugin.rb file:
Pod::HooksManager.register('cocoapods-repo-svn', :source_provider) do |context, options|
    # options contains the hash defined in the Podfile
    source_names = options['sources']
    source_names.each do |name|
        source = MyCustomCreateSourceFunction(name)
        # Add any number of Pod::Source objects to the context
        context.add_source(source)
    end
end
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -136,10 +136,6 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   is now a default plugin.  
   [Samuel Giddins](https://github.com/segiddins)
 
-* Added a `:source_provider` hook to allow plugins to provide their own sources
-  [Eric Amorde](https://github.com/amorde)
-  [#3792](https://github.com/CocoaPods/CocoaPods/pull/3792)
-
 ##### Bug Fixes
 
 * Ensure that the `prepare_command` is run even when skipping the download
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -136,6 +136,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   is now a default plugin.  
   [Samuel Giddins](https://github.com/segiddins)
 
+* Added a `:source_provider` hook to allow plugins to provide their own sources
+  [Eric Amorde](https://github.com/amorde)
+  [#3792](https://github.com/CocoaPods/CocoaPods/pull/3792)
+
 ##### Bug Fixes
 
 * Ensure that the `prepare_command` is run even when skipping the download
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,11 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Samuel Giddins](https://github.com/segiddins)
   [#3926](https://github.com/CocoaPods/CocoaPods/issues/3926)
 
+* Add `:source_provider` hook to allow plugins to provide sources
+  [Eric Amorde](https://github.com/amorde)
+  [#3190](https://github.com/CocoaPods/CocoaPods/issues/3190)
+  [#3792](https://github.com/CocoaPods/CocoaPods/pull/3792)
+
 
 ## 0.38.2
 
--- a/lib/cocoapods/installer.rb
+++ b/lib/cocoapods/installer.rb
@@ -29,18 +29,19 @@ module Pod
   # source control.
   #
   class Installer
-    autoload :AggregateTargetInstaller, 'cocoapods/installer/target_installer/aggregate_target_installer'
-    autoload :Analyzer,                 'cocoapods/installer/analyzer'
-    autoload :FileReferencesInstaller,  'cocoapods/installer/file_references_installer'
-    autoload :PostInstallHooksContext,  'cocoapods/installer/post_install_hooks_context'
-    autoload :PreInstallHooksContext,   'cocoapods/installer/pre_install_hooks_context'
-    autoload :Migrator,                 'cocoapods/installer/migrator'
-    autoload :PodfileValidator,         'cocoapods/installer/podfile_validator'
-    autoload :PodSourceInstaller,       'cocoapods/installer/pod_source_installer'
-    autoload :PodSourcePreparer,        'cocoapods/installer/pod_source_preparer'
-    autoload :PodTargetInstaller,       'cocoapods/installer/target_installer/pod_target_installer'
-    autoload :TargetInstaller,          'cocoapods/installer/target_installer'
-    autoload :UserProjectIntegrator,    'cocoapods/installer/user_project_integrator'
+    autoload :AggregateTargetInstaller,   'cocoapods/installer/target_installer/aggregate_target_installer'
+    autoload :Analyzer,                   'cocoapods/installer/analyzer'
+    autoload :FileReferencesInstaller,    'cocoapods/installer/file_references_installer'
+    autoload :PostInstallHooksContext,    'cocoapods/installer/post_install_hooks_context'
+    autoload :PreInstallHooksContext,     'cocoapods/installer/pre_install_hooks_context'
+    autoload :SourceProviderHooksContext, 'cocoapods/installer/source_provider_hooks_context'
+    autoload :Migrator,                   'cocoapods/installer/migrator'
+    autoload :PodfileValidator,           'cocoapods/installer/podfile_validator'
+    autoload :PodSourceInstaller,         'cocoapods/installer/pod_source_installer'
+    autoload :PodSourcePreparer,          'cocoapods/installer/pod_source_preparer'
+    autoload :PodTargetInstaller,         'cocoapods/installer/target_installer/pod_target_installer'
+    autoload :TargetInstaller,            'cocoapods/installer/target_installer'
+    autoload :UserProjectIntegrator,      'cocoapods/installer/user_project_integrator'
 
     include Config::Mixin
 
@@ -124,6 +125,9 @@ module Pod
 
     def resolve_dependencies
       analyzer = create_analyzer
+
+      plugin_sources = run_source_provider_hooks
+      analyzer.sources.push(*plugin_sources)
 
       UI.section 'Updating local specs repositories' do
         analyzer.update_repositories
@@ -451,6 +455,16 @@ module Pod
     def run_plugins_post_install_hooks
       context = PostInstallHooksContext.generate(sandbox, aggregate_targets)
       HooksManager.run(:post_install, context, plugins)
+    end
+
+    # Runs the registered callbacks for the source provider plugin hooks.
+    #
+    # @return [void]
+    #
+    def run_source_provider_hooks
+      context = SourceProviderHooksContext.generate
+      HooksManager.run(:source_provider, context, plugins)
+      context.sources
     end
 
     # Ensures that all plugins specified in the {#podfile} are loaded.
--- a/lib/cocoapods/installer.rb
+++ b/lib/cocoapods/installer.rb
@@ -127,7 +127,7 @@ module Pod
       analyzer = create_analyzer
 
       plugin_sources = run_source_provider_hooks
-      analyzer.sources.push(*plugin_sources)
+      analyzer.sources.insert(0, *plugin_sources)
 
       UI.section 'Updating local specs repositories' do
         analyzer.update_repositories
--- a/lib/cocoapods/installer/source_provider_hooks_context.rb
+++ b/lib/cocoapods/installer/source_provider_hooks_context.rb
@@ -1,0 +1,32 @@
+module Pod
+  class Installer
+    # Context object designed to be used with the HooksManager which describes
+    # the context of the installer before spec sources have been created
+    #
+    class SourceProviderHooksContext
+      # @return [Array<Source>] The source objects to send to the installer
+      #
+      attr_reader :sources
+
+      # @return [SourceProviderHooksContext] Convenience class method to generate the
+      #         static context.
+      #
+      def self.generate
+        result = new
+        result
+      end
+
+      def initialize
+        @sources = []
+      end
+
+      # @param [Source] Source object to be added to the installer
+      #
+      def add_source(source)
+        unless source.nil?
+          @sources << source
+        end
+      end
+    end
+  end
+end
--- a/spec/unit/installer/source_provider_hooks_context_spec.rb
+++ b/spec/unit/installer/source_provider_hooks_context_spec.rb
@@ -1,0 +1,12 @@
+
+require File.expand_path('../../../spec_helper', __FILE__)
+
+module Pod
+  describe Installer::SourceProviderHooksContext do
+    it 'offers a convenience method to be generated' do
+      result = Installer::SourceProviderHooksContext.generate
+      result.class.should == Installer::SourceProviderHooksContext
+      result.sources.should == []
+    end
+  end
+end
--- a/spec/unit/installer_spec.rb
+++ b/spec/unit/installer_spec.rb
@@ -93,6 +93,51 @@ module Pod
         @installer.install!
       end
 
+      it 'runs source provider hooks before analyzing' do
+        config.skip_repo_update = true
+        @installer.unstub(:resolve_dependencies)
+        @installer.stubs(:validate_build_configurations)
+        @installer.stubs(:prepare_for_legacy_compatibility)
+        @installer.stubs(:clean_sandbox)
+        def @installer.run_source_provider_hooks
+          @hook_called = true
+        end
+        def @installer.analyze(*)
+          @hook_called.should.be.true
+        end
+        @installer.install!
+      end
+
+      it 'includes sources from source provider plugins' do
+        plugin_name = 'test-plugin'
+        Pod::HooksManager.register(plugin_name, :source_provider) do |context, options|
+          source_url = options['sources'].first
+          return unless source_url
+          source = Pod::Source.new(source_url)
+          context.add_source(source)
+        end
+
+        test_source_name = 'https://github.com/artsy/Specs.git'
+        plugins_hash = Installer::DEFAULT_PLUGINS.merge(plugin_name => { 'sources' => [test_source_name] })
+        @installer.podfile.stubs(:plugins).returns(plugins_hash)
+        @installer.unstub(:resolve_dependencies)
+        @installer.stubs(:validate_build_configurations)
+        @installer.stubs(:prepare_for_legacy_compatibility)
+        @installer.stubs(:clean_sandbox)
+        @installer.stubs(:ensure_plugins_are_installed!)
+        @installer.stubs(:analyze)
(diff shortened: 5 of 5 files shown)


amorde:
Looks like my own test failed, but I'm not even sure if I wrote the test correctly :0

lib/cocoapods/installer/source_provider_hooks_context.rb
```diff
@@ -0,0 +1,28 @@
+module Pod
+  class Installer
+    # Context object designed to be used with the HooksManager which describes
+    # the context of the installer before spec sources have been created
+    #
+    class SourceProviderHooksContext
+      # @return [Podfile] The Podfile for the project.
+      #
+      attr_accessor :podfile
```
segiddins: I don't see why the sources would need access to the pod file?

lib/cocoapods/installer.rb
```diff
@@ -445,6 +449,18 @@ def run_plugins_post_install_hooks
...
+    # Runs the registered callbacks for the source provider plugin hooks.
+    #
+    # @param [Analyzer] analyzer The analyzer that will be used to process sources
+    #
+    # @return [void]
+    #
+    def run_source_provider_hooks
+      context = SourceProviderHooksContext.generate(podfile)
+      HooksManager.run(:source_provider, context, plugins)
+      context.sources
```
segiddins: can we think of a way to do this without allowing other plugins to access each other's sources? That feels a bit weird to me -- maybe the context could have an `add_source` method?

segiddins:
I'd really love to see a test for this that actually uses a plugin source

amorde:
Hey @segiddins, thanks for the feedback! I have tested my own plugin which supports "local" sources (ex. "private" for ~/.cocoapods/repos/private) but I am unable to publish that code at the moment. That might change soon.

I'll work on the changes you suggested

lib/cocoapods/installer/source_provider_hooks_context.rb
amorde: You are totally correct, I only included it because before I was using the Podfile as a way to insert my sources manually. With this plugin, it isn't needed.

amorde:
I apologize for the numerous commits, I cannot get rubocop to run on my machine due to some weird issues with gemfiles so I don't know about violations until the travis build tells me. If anyone who has access would like to cancel those extra builds please do so!

amorde:
Ping @segiddins

segiddins:
`bundle exec rake spec` should run all of the specs, including rubocop

amorde:
Yes, I have tried that and I get the following:

```
Bundler could not find compatible versions for gem "cocoapods-core":
  In snapshot (Gemfile.lock):
    cocoapods-core (= 0.38.0.beta.1)

  In Gemfile:
    cocoapods (>= 0) ruby depends on
      cocoapods-core (= 0.38.0.beta.2) ruby

    cocoapods-core (>= 0) ruby

Running `bundle update` will rebuild your snapshot from scratch, using only
the gems in your Gemfile, which may resolve the conflict.
```

I have tried running `bundle update` but it never fixes the issue. I have even tried recreating the `Rainforest` repo from scratch and I get either this or some version conflict with `xcodeproj`...

Anyways, any comments on my changes? I was able to get a test in using the plugin API albeit a bit verbose as far as tests go

amorde:
The last Travis build seems to have failed due to some sort of ruby bundler error, but the second job succeeded.

amorde:
ping @orta

orta:
Hello! So, from my perspective this looks all good. It also seems a small enough change that I'd prefer it getting in with 0.38. I'm going to leave it to either @kylef / @neonichu / @segiddins to do the actual merging however, as they understand the codebase significantly better than I do.

amorde:
I'd be more than happy to make a new PR will all of those extra commits squashed into one if that makes it easier / cleaner, I did a terrible job of passing the rubocop tests heh..

orta:
Sure, do it, feel free to close this one - or you can squash the commits, and do a force push :+1:

amorde:
Chose the second option!

orta:
Great, consider yourself accomplished in your task. Someone will finish up the job.

lib/cocoapods/installer.rb
```diff
@@ -33,6 +33,7 @@ class Installer
     autoload :FileReferencesInstaller,  'cocoapods/installer/file_references_installer'
     autoload :PostInstallHooksContext,  'cocoapods/installer/post_install_hooks_context'
     autoload :PreInstallHooksContext,   'cocoapods/installer/pre_install_hooks_context'
+    autoload :SourceProviderHooksContext, 'cocoapods/installer/source_provider_hooks_context'
```
segiddins: please fix the alignment here

segiddins:
This seems good, but would like to also get @kylef to take a look at this.

:+1:

lib/cocoapods/installer.rb
amorde: Even without a space they will not be aligned.. Do you want me to rename it so it is shorter?

segiddins: Nope, probably the best thing to do is add a space to all the others so they're aligned again :)

amorde: Oops, I'm an idiot, that's a way better idea. Just pushed that in

kylef:
Code look fine, can you add a CHANGELOG entry @amorde?

The main concern I have with this implementation is that this API doesn't allow consumers to decide the order of sources with plugins. The [ordering of sources](https://guides.cocoapods.org/syntax/podfile.html#source) is important as it can determine which source a Pod may be used from.

```ruby
source 'https://github.com/CocoaPods/Specs'
source 'svn://foo'
source 'cvs://bar'
```

Is different to:

```ruby
source 'svn://foo'
source 'https://github.com/CocoaPods/Specs'
source 'cvs://bar'
```

amorde:
Sure, just pushed up the changelog now.

Regarding the order of sources, unless I am mistaken I believe changing the order of where you put the call to `plugin` would give you the desired behavior:
```ruby
# This SVN source will be evaluated before the master repo
plugin 'cocoapods-repo-svn', {
    :sources => [
        'https://svn.myserver.com'
    ]
}
source 'https://github.com/CocoaPods/Specs.git
```
This should work as expected, because the installer does a `push` onto the sources of the analyzer. The analyzer's `source` accessor will instantiate the array with `podfile.sources` first as seen [here](https://github.com/CocoaPods/CocoaPods/blob/master/lib/cocoapods/installer/analyzer.rb#L610)

segiddins:
I don't believe the order of `plugin` invocations will actually provide any ordering, though

amorde:
Yeah, I just had that realization as well. I'm not sure how to fix that to be honest

kylef:
@amorde I don't think that's the case, because the hooks won't be ran until `run_source_provider_hooks`, in your example the custom source (`https://svn.myserver.com`) will be after the `https://github.com/CocoaPods/Specs.git` source,

amorde:
@segiddins @kylef You are both correct. If there is a way we can support ordering I would be more than willing to add it but I'm not sure how to do that as of right now. I feel that this would still be a valuable feature and we could add the ordering later, but I understand if you do not want to include it until it supports ordering

amorde:
@kylef The plugin could support it theoretically, by declaring it in the Podfile:

```ruby
plugin 'cocoapods-repo-svn', {
    :sources => [
        'https://svn.myserver.com',
        'https://github.com/CocoaPods/Specs.git'
    ]
}
```

then in the plugin hook

``` ruby
Pod::HooksManager.register('cocoapods-repo-svn', :source_provider) do |context, options|
    source_names = options['sources']
    source_names.each do |name|
        if name == Pod::SourcesManager.master.url
            context.add_source(Pod::SourcesManager.master)
        else
            source = MyCustomCreateSourceFunction(name)
            context.add_source(source)
        end
    end
end
```

segiddins:
It might make sense to make the invocation of plugins ordered, but that still wouldn't solve the issue of ordering between native sources and plugin sources

amorde:
We could do something like this:

```ruby
# in `Installer.resolve_dependencies`
context = run_source_provider_hooks
analyzer.sources.insert(0, *context.sources_before_builtin)
analyzer.sources.push(*context.sources_after_builtin)
```

Then let the plugin decide the order, possibly allowing the user to provide a flag as to whether to put it before or after the builtin repos

amorde:
@kylef Any thoughts? I can implement some sort of source ordering, but I think this pull request has become a large enough discussion that it would be a good idea to discuss ordering elsewhere / in a separate issue

orta:
It might just be acceptable enough to say that if you care enough to be using a plugin for a source provider, they should be of the highest priority so that providers / users don't have to add a weird index in their pod files / APIs.

amorde:
Yeah, I agree. If using a private source that requires a plugin to support, you probably want your source to come first.

I will change the `push` to and `insert`, if that sounds reasonable

segiddins:
Sounds reasonable to me

amorde:
Anything else I need to do to get this merged?

clarkda:
looking forward to these changes! :+1: :cake:

amorde:
@segiddins ?

segiddins:
:+1:

alloy:
I loved reading this thread and how consensus was reached :+1:

amorde:
Just wondering if there's any info on this getting merged. We are looking to use this along with the `cocoapods-repo-svn` plugin at my workplace to manage our dependencies, but are waiting to switch our builds over until this is available

segiddins:
@amorde just waiting for @kylef to review :)

kylef:
What about executing the `source_provider` plugin hook right away when the plugin is used, along with supporting it's use as a source multiple times?

Example:

```ruby
plugin 'cocoapods-svn', :source => 'svn://foo'
source 'https://github.com/CocoaPods/Specs'
plugin 'cocoapods-cvs', :source => 'cvs://bar'
plugin 'cocoapods-svn', :source => 'svn://zzz'
```

This way the user would get full control of ordering.

amorde:
That sounds perfect, but I don't see how that would be implemented. Right now any call to `plugin` just [adds entries to a hash](https://github.com/CocoaPods/Core/blob/master/lib/cocoapods-core/podfile/dsl.rb#L577). Like you mentioned earlier, they won't be executed until later.

It seems that to implement that would require massive changes to the way the podfile is evaluated

amorde:
I think something that might be easier to implement would be to simply specify the master repo in the plugin's source, and let the plugin handle it (eg. just pass it off to the cocoapods implementation)

```ruby
plugin 'cocoapods-repo-svn', {
   :sources => [
       'svn://foo.com',
       'https://github.com/CocoaPods/Specs',
       'svn://bar.com'
    ]
}
```

Then the plugin could to do something like

```ruby
master = Pod::SourcesManager.master.first
if master.url == source_url
   context.add_source(master)
else
   # plugin source made here
end
```

amorde:
@kylef I've been thinking about this some more, and I had another idea I figured I'd toss out there. We could extend the `source` command to allow for a `:plugin` argument specifying that the source should be handled by a plugin

```ruby
source 'svn://foo', :plugin => 'cocoapods-repo-svn'
source 'https://github.com/CocoaPods/Specs'
source 'hg://bar', :plugin => 'cocoapods-repo-hg'
```

This would be a pretty breaking change, as `podfile.sources` can no longer be assumed to be an array of strings. I believe the only place in the cocoapods codebase that assumes that is in `analyzer.sources` but I am not sure.

I'm willing to do whatever is necessary to support ordering, however I *still* think that having plugin sources always come first is better than nothing - right now other VCS's are impossible to use as sources *at all*. Any team that is unable to use Git either cannot use specs repos or has to hack into the install process similar to what `cocoapods-keys` does.

amorde:
@kylef @segiddins I believe what I described above would allow the ordering of sources as long as plugins support the master repo, which is fairly easy. So if you're Podfile contained this:

```ruby
plugin 'cocoapods-repo-svn', :sources => [
       'svn://foo.com',
       'https://github.com/CocoaPods/Specs',
       'svn://bar.com'
    ]
```

It would behave as if you did this:

```ruby
source 'svn://foo.com',
source 'https://github.com/CocoaPods/Specs',
source 'svn://bar.com'
# There will be a duplicate master at the end, which should not matter
source 'https://github.com/CocoaPods/Specs',
```

The time I am able to dedicate to this is wearing thin, so if a consensus cannot be reached I will have to abandon this, unfortunately. If there is anything I can do to make this work for you guys please let me know

clarkda:
@kylef i agree with @amorde that supporting *true* ordering will likely require some pretty deep changes. i could maybe help with this if this is how we decide we need to do it.

but i'm kinda in the same camp as @orta and feel that if you are far enough off the beaten path to use a source plugin then we can probably just resolve the source plugins in order before our regular source. 

and thanks so much @amorde for the time and effort you're putting into this! this is great stuff!

eseilnacht:
+1

orta:
I feel like this is OK as is.

segiddins:
I'm waiting for @kylef to review before merging, since he stated he had objections before

Whirlwind:
It sounds good!

CHANGELOG.md
```diff
@@ -22,6 +22,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   is now a default plugin.  
   [Samuel Giddins](https://github.com/segiddins)
 
+* Added a `:source_provider` hook to allow plugins to provide their own sources
```
kylef: This should be moved to go under master, not beta 2.

kylef:
Other than that one comment about CHANGELOG. This is good to go.

Whirlwind:
```
source 'svn://foo'
source 'https://github.com/CocoaPods/Specs'
source 'cvs://bar'
```
I hope the `source` should be able to auto select the plugin. or we can pass the source type:
```
source 'svn://foo', :type=>"svn"
source 'https://github.com/CocoaPods/Specs', :type=>"git"
source 'cvs://bar', :type=>"cvs"
```
I think the `plugin` is not a good idea. Somebody can use another plugin to pull the same source.

amorde:
I'm having a difficult time including the upstream changes to the changelog.. could anyone help me out? It appears I've introduced a merge conflict

segiddins:
@amorde just reset `HEAD` to the previous commit, and rebase? If not, I can handle merging this manually tonight.

amorde:
Think I got it! Although now Travis is checking that the changelog didn't break the build :p

kylef:
Thanks for your contribution @amorde.

strawnut:
In my case, it is printed the error message.

```
NoMethodError - undefined method `check_version_information' for #<Pod::Source::Manager:0x007ffb4c04dc70>
/Library/Ruby/Gems/2.0.0/gems/cocoapods-repo-svn-2.0.1/lib/pod/command/repo_svn.rb:43:in `block in run'
/Library/Ruby/Gems/2.0.0/gems/cocoapods-1.0.0/lib/cocoapods/user_interface.rb:63:in `section'
/Library/Ruby/Gems/2.0.0/gems/cocoapods-repo-svn-2.0.1/lib/pod/command/repo_svn.rb:36:in `run'
/Library/Ruby/Gems/2.0.0/gems/cocoapods-repo-svn-2.0.1/lib/cocoapods_plugin.rb:32:in `update_or_add_source'
/Library/Ruby/Gems/2.0.0/gems/cocoapods-repo-svn-2.0.1/lib/cocoapods_plugin.rb:10:in `block (2 levels) in <top (required)>'
/Library/Ruby/Gems/2.0.0/gems/cocoapods-repo-svn-2.0.1/lib/cocoapods_plugin.rb:8:in `each'
/Library/Ruby/Gems/2.0.0/gems/cocoapods-repo-svn-2.0.1/lib/cocoapods_plugin.rb:8:in `block in <top (required)>'
/Library/Ruby/Gems/2.0.0/gems/cocoapods-1.0.0/lib/cocoapods/hooks_manager.rb:109:in `call'
/Library/Ruby/Gems/2.0.0/gems/cocoapods-1.0.0/lib/cocoapods/hooks_manager.rb:109:in `block (3 levels) in run'
/Library/Ruby/Gems/2.0.0/gems/cocoapods-1.0.0/lib/cocoapods/user_interface.rb:144:in `message'
/Library/Ruby/Gems/2.0.0/gems/cocoapods-1.0.0/lib/cocoapods/hooks_manager.rb:103:in `block (2 levels) in run'
/Library/Ruby/Gems/2.0.0/gems/cocoapods-1.0.0/lib/cocoapods/hooks_manager.rb:101:in `each'
/Library/Ruby/Gems/2.0.0/gems/cocoapods-1.0.0/lib/cocoapods/hooks_manager.rb:101:in `block in run'
/Library/Ruby/Gems/2.0.0/gems/cocoapods-1.0.0/lib/cocoapods/user_interface.rb:144:in `message'
/Library/Ruby/Gems/2.0.0/gems/cocoapods-1.0.0/lib/cocoapods/hooks_manager.rb:100:in `run'
/Library/Ruby/Gems/2.0.0/gems/cocoapods-1.0.0/lib/cocoapods/installer.rb:482:in `run_source_provider_hooks'
/Library/Ruby/Gems/2.0.0/gems/cocoapods-1.0.0/lib/cocoapods/installer.rb:143:in `resolve_dependencies'
/Library/Ruby/Gems/2.0.0/gems/cocoapods-1.0.0/lib/cocoapods/installer.rb:114:in `install!'
/Library/Ruby/Gems/2.0.0/gems/cocoapods-1.0.0/lib/cocoapods/command/install.rb:37:in `run'
/Library/Ruby/Gems/2.0.0/gems/claide-1.0.0/lib/claide/command.rb:334:in `run'
/Library/Ruby/Gems/2.0.0/gems/cocoapods-1.0.0/lib/cocoapods/command.rb:50:in `run'
/Library/Ruby/Gems/2.0.0/gems/cocoapods-1.0.0/bin/pod:55:in `<top (required)>'
/usr/local/bin/pod:23:in `load'
/usr/local/bin/pod:23:in `<main>'
```

If I insert below lines then, printed above error messages.
```
plugin 'cocoapods-repo-svn', :sources => [
	'svn://svn.code.sf.net/p/codeblocks/code/trunk'
]
```

Dont' mind about svn url. Anything cannot be worked.
I should use svn repo.
Actually, it has being worked using CocoaPods v0.33.1, but I guess CocoaPods does not support svn repo officially. I can't find out official article.

PR #3808, opened 2015-07-10
[Specs] Fixed plugin specs in the incorrect files

amorde:
The tests in `pre_install_hooks_context_spec.rb` and `post_install_hooks_context_spec.rb` are currently swapped, with the `pre_install` spec in the `post_install` file and vice versa.

Simple swap.


orta:
Hah!

PR #3826, opened 2015-07-16
Add language provisions to code of conduct

dcow:
It is the desire of the cocoapods team to facilitate calm, respectful discussion surrounding issues that may arise during the development of the CocoaPods projects. Make the community aware of these intentions by including prose in the formal code of conduct.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -2,11 +2,17 @@
 
 CocoaPods strongly values contributors from anywhere, regardless of gender, sexual orientation, disability, physical appearance, body size, race, or religion. As a result, the CocoaPods team has agreed to and enforces this code of conduct in order to provide a harassment-free experience for everyone who participates in the development of CocoaPods.
 
+CocoaPods also expects PG discussion on the issue tracker. Any use of explicit language is strictly prohibited.
+
 ### Summary
 
 Harassment in code and discussion or violation of physical boundaries is completely unacceptable anywhere in CocoaPods’ codebases, issue trackers, Slack, Campfire, mailing lists, meetups, and other events. Violators will be warned and then blocked or banned by the core team at or before the 3rd violation.
 
+For language violations, you will only receive one warning.
+
 ### In detail
+
+#### Harassment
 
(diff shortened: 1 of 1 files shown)


segiddins:
Thanks for this pull request. The code of conduct doesn't say anything about the exact language used. It is instead about the spirit that the language is used in. Combative or attacking language is not permitted, regardless of whether it is rated PG.

dcow:
@segiddins with all due respect, the code of conduct only explicitly calls out harassment. Nowhere is combative or "attacking" language mentioned. I think this is a necessary addition if you want to justify censoring explicit language that is neither attacking nor combating (or otherwise loosely harassing) any person. With respect to my comment on #13313, I was not attacking any person. I was simply expressing frustration at the idea that some*thing* that was in one state, is now in another state. Please believe me that I'm really not trying to be passive aggressive or anything here. If you misunderstood the intention of my comment and took it as harassment in the heat of the moment, then I'm truly sorry. But, I promise you that was not my intention and if you read the comment again I think you'll agree.

Inferis:
I'd tend to agree with the sentiment. If you go around waving "this language is not tolerated" here then you should at least mention it in the code of conduct to back your words up. Otherwise it's just an arbitrary statement from the repo owners. Which is fine by me, but is kind of counterproductive if you have a code of conduct in the first place.

alloy:
> It is the desire of the cocoapods team to facilitate calm, respectful discussion surrounding issues that may arise during the development of the CocoaPods projects.

This is correct.

> I think this is a necessary addition if you want to justify censoring explicit language

I disagree. It shouldn’t be that we are only ever justified to call something out and possibly remove it if we thought of it before hand.

Having said that, we decided on the following outcome: we frown upon explicit language, we don’t ever seeing it helping anybody, but we’ll not explicitly disallow it. In the future there won’t be official warnings about it, we’ll simply appeal to your better judgment for keeping a thread, that’s getting heated, calm, which will probably be enough.

Inferis:
Sounds very reasonable! 👍

dcow:
Thanks for discussing (=

alloy:
:heart:

PR #3830, opened 2015-07-16
[Installer] Improve error message when using Swift in static libs.

alloy:
Now it includes the offending Pods so that the user can take action
to remove the Pods, if they don’t want to move to frameworks yet.

### Before

```
[!] Pods written in Swift can only be integrated as frameworks; this feature is still in beta. Add `use_frameworks!` to your Podfile or target to opt into using it.
```

### After

```
[!] Pods written in Swift can only be integrated as frameworks; this feature is still in beta. Add `use_frameworks!` to your Podfile or target to opt into using it. The Swift Pod being used is: FBSnapshotTestCase
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,12 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   is now a default plugin.  
   [Samuel Giddins](https://github.com/segiddins)
 
+* Improve the message shown when trying to use Swift Pods without frameworks.
+  Now it includes the offending Pods so that the user can take action to remove
+  the Pods, if they don’t want to move to frameworks yet.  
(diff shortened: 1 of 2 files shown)


alloy:
Created a PR, added the CHANGELOG afterwards so that I had a ticket number to reference… :smile:

segiddins:
@alloy CHANGELOG shouldn't reference PRs, only issues

segiddins:
:+1:

alloy:
> CHANGELOG shouldn't reference PRs, only issues

Yeah I see that being better in this case to solve the chicken and egg, not sure if I necessarily think it’s actually required. The reason I want the link in the CHANGELOG is to provide more context, which my PR text does…

Oh well, I’ll let the computers win this battle and create a issue next time ¯\_(ツ)_/¯

Issue #3867, opened 2015-07-22
Headers for built dynamic frameworks are linked in the sandbox


perotinus:
When building a dynamic framework, all needed headers are put into the framework and should be referenced from their framework location. These headers, however, are still linked into ```$(PODS_ROOT)/Headers```. This doesn't cause breakages, but when those header paths are added to the ```HEADER_SEARCH_PATHS```, builds are slower than necessary.

It should never be necessary to link those headers into the sandbox when building as a dynamic framework, and removing them should speed up compilation slightly.

perotinus:
Fixed by the aforementioned PR.

PR #3868, opened 2015-07-22
When integrating Pods as frameworks, only link public headers in the sandbox for static Pods

perotinus:
When integrating Pods as frameworks, only link public headers into the sandbox if those headers are from a Pod that is not being built (that is, a static library Pod that's being statically linked into the user target).

Addresses #3867.

Depends on https://github.com/CocoaPods/cocoapods-integration-specs/pull/41


perotinus:
Is there any way to get the Travis build to run with this pull request and the one it depends on?

segiddins:
@perotinus don't worry about it, I usually update the integration specs post-merge :)

PR #3937, opened 2015-07-30
Add a command to check if dependencies are up to date

marcomorain:
:rainbow: :sunglasses: 

# Rational

At [CircleCi](https://circleci.com/mobile/ios) we automatically detect and run `pod install` when we detect a that contains a `Podfile` and an empty `Pods` directory. We cache the generated `Pods` folder from build to build, but we still need to run `pod install` to ensure that all dependencies are met. This call can take several minutes on every build. I want to add a command `pod check` that will exit with success when the Podfile.lock's manifest is satisfied, to allow us to run `pod check || pod install`, which should cut minutes off every build.

I've been writing an internal system to perform the same test, but I would like to push the changes back upstream so that others may benefit.

# Feature

Add a command `pod check` that returns success when the
pods specified in the lockfile are installed. This command
is designed to mirror the `bundle check` command provided
by bundler.

The expected use-case is to prevent having to run `pod install`
on every build in a CI scenario. The user should be able to
run `pod check || pod install` rather than pod install. This
can save minutes off CI builds.

http://bundler.io/v1.1/bundle_check.html

# TODO:

I wanted to open this PR for discussion and feedback before I spend any more time polishing. Here is a list of tasks that are needed to finish this:

- [ ] Write code docs
- [ ] Write tests
- [ ] Run linting tools

# Example

### Using `pod check` on an app with/without up-to-date pods 

Here is some example usage:
```
marc@skywarp ~/dev/KiwiTester $ rm Podfile.lock
marc@skywarp ~/dev/KiwiTester $ pod check
[!] No `Podfile.lock` found in the project directory, run `pod install`.
marc@skywarp ~/dev/KiwiTester $ pod install > /dev/null
marc@skywarp ~/dev/KiwiTester $ pod check

[!] The applications's dependencies are satisfied.
marc@skywarp ~/dev/KiwiTester $ rm -r Pods
marc@skywarp ~/dev/KiwiTester $ pod check
[!] Some dependencies are not met. Run `pod install` or update your CocoaPods installation.
marc@skywarp ~/dev/KiwiTester $ pod install > /dev/null
marc@skywarp ~/dev/KiwiTester $ pod check

[!] The applications's dependencies are satisfied.
marc@skywarp ~/dev/KiwiTester $ sed -i -e '3d' Pods/Manifest.lock # Delete line 3 of Manifest.lock
marc@skywarp ~/dev/KiwiTester $ pod check
[!] Some dependencies are not met. Run `pod install` or update your CocoaPods installation.
marc@skywarp ~/dev/KiwiTester $ pod install > /dev/null
marc@skywarp ~/dev/KiwiTester $ pod check

[!] The applications's dependencies are satisfied.
```

### Running `pod install` on an app with dependencies that all already specified.
```
marc@skywarp ~/dev/KiwiTester $ pod install > /dev/null  && /usr/bin/time pod install > /dev/null
        6.46 real         3.72 user         0.70 sys
marc@skywarp ~/dev/KiwiTester $ pod install > /dev/null  && /usr/bin/time ../CocoaPods/bin/pod check || pod install

[!] The applications's dependencies are satisfied.
        0.51 real         0.44 user         0.07 sys
```

### DIFF
--- a/lib/cocoapods/command.rb
+++ b/lib/cocoapods/command.rb
@@ -15,6 +15,7 @@ module Pod
 
   class Command < CLAide::Command
     require 'cocoapods/command/inter_process_communication'
+    require 'cocoapods/command/check'
     require 'cocoapods/command/lib'
     require 'cocoapods/command/list'
     require 'cocoapods/command/outdated'
--- a/lib/cocoapods/command/check.rb
+++ b/lib/cocoapods/command/check.rb
@@ -1,0 +1,30 @@
+module Pod
+  class Command
+    class Check < Command
+      self.summary = 'Quickly check if pods are installed'
+      self.description = <<-DESC
+Determine whether the requirements for your application are satisfied.
+The `check` command will quickly compare the installed Pods against the
+dependencies specified in Podfile.lock. The `check` command will exit with
+status 0 when the pods are up to date and status 1 otherwise.
+The `check` can be chained with the install command in the following way to
+determine if installation is necessary.
+
+    $ pod check || pod install
+DESC
+
+
+      def run
+        unless config.lockfile
+          raise Informative, 'No `Podfile.lock` found in the project directory, run `pod install`.'
+        end
+
+        unless config.lockfile == config.sandbox.manifest
+          raise Informative, 'Some dependencies are not met. Run `pod install` or update your CocoaPods installation.'
+        end
+        UI.notice('The applications\'s dependencies are satisfied.')
+      end
+
+    end
+  end
+end


marcomorain:
Here is some example usage:
```
marc@skywarp ~/dev/KiwiTester $ rm Podfile.lock
marc@skywarp ~/dev/KiwiTester $ pod check
[!] No `Podfile.lock` found in the project directory, run `pod install`.
marc@skywarp ~/dev/KiwiTester $ pod install > /dev/null

[!] The applications's dependencies are satisfied.
marc@skywarp ~/dev/KiwiTester $ rm -r Pods
marc@skywarp ~/dev/KiwiTester $ pod check
[!] Some dependencies are not met. Run `pod install` or update your CocoaPods installation.
marc@skywarp ~/dev/KiwiTester $ pod install > /dev/null
marc@skywarp ~/dev/KiwiTester $ pod check

[!] The applications's dependencies are satisfied.
marc@skywarp ~/dev/KiwiTester $ sed -i -e '3d' Pods/Manifest.lock # Delete line 3 of Manifest.lock
marc@skywarp ~/dev/KiwiTester $ pod check
[!] Some dependencies are not met. Run `pod install` or update your CocoaPods installation.
marc@skywarp ~/dev/KiwiTester $ pod install > /dev/null
marc@skywarp ~/dev/KiwiTester $ pod check

[!] The applications's dependencies are satisfied.
```

neonichu:
hm, isn't most of the additional time `pod install` spends updating the local copy of the spec repo?

marcomorain:
> hm, isn't most of the additional time pod install spends updating the local copy of the spec repo?

Yes, I think it is. I want to avoid having to update this repo on every build when the Pods are already present. The `~/.cocoapods` folder on a CI image is baked into the image, so that repo is quickly out dated, and has to be updated on every build. An alternative is to cache the `~/.cocoapods` folder from build to build, but that adds about half a GB that needs to saved/restored on every build.

The local spec repo on my machine is about half a gig –

```
marc@skywarp ~ $ du -d 0 -h ~/.cocoapods/
482M	/Users/marc/.cocoapods/
```

neonichu:
Would just adding `--no-repo-update` have the same effect? If `pod install` fails with that flag, you could retry without it.

segiddins:
I'm with @neonichu on this one -- I really don't see what this buys you over `bundle install --no-repo-update`, in addition to the fact that this implementation won't take into account a Podfile that has been modified without an install.

marcomorain:
> Would just adding --no-repo-update have the same effect? If pod install fails with that flag, you could retry without it.

`pod install --no-repo-update || pod install` might have the same effect. If so, so that would be awesome, and I can close this PR. I'm having difficulty testing how to run `pod install --no-repo-update` inside a project with Pods already installed, with `~/.cocoapods/repos/master` being outdated. I can checkout a previous version using something like `cd ~/.cocoapods/repos/master; git checkout 'master@{2014-01-01 18:30:00}'`. Do you think thats enough to simulate an 'old' `.cocoapods` cache?

Meanwhile I've been doing something like this locally (which lead me to open this PR):

```
diff --side-by-side --suppress-common-lines --ignore-matching-lines='^COCOAPODS: && cp Podfile.lock Pods/Manifest.lock || pod install
```

The idea is that we might try to build a repository where the `Podfile.lock` was created with a different version of CocoaPods than the one installed on the machin (hence the `--ignore-matching-lines`). I think copy the Podfile.lock over the manifest because the check to diff the lockfile and manifest that gets installed into the xcode project will refuse to build unless the two are identical.

marcomorain:
> I'm with @neonichu on this one -- I really don't see what this buys you over pod install --no-repo-update, in addition to the fact that this implementation won't take into account a Podfile that has been modified without an install.

It will be awesome if that will do that I need – if I can speed up `pod install` without having to change CocoaPods, that's a good thing.

segiddins:
@marcomorain don't do that, since the artifacts produced by different CocoaPods versions will be different.

marcomorain:
> @marcomorain don't do that, since the artifacts produced by different CocoaPods versions will be different.

Oh, interesting. What type of differences would you expect? Do you happen to know 2 versions of CocoaPods that would produce different artifacts?

segiddins:
0.38.1 and .2 produce different `Pods` directory contents

marcomorain:
Thanks, I'll take a look over the weekend to learn more about how cocoapods works.

segiddins:
@marcomorain I'm going to close this PR for now, but thanks for making it! Whatever route you end up going, i think it would be best as a plugin.

marcomorain:
Thanks for the feedback Samuel!

I'll let you know what direction I go for.

PR #3938, opened 2015-07-30
Update spec and lib commands to use all sources by default

dnkoutso:
re: https://github.com/CocoaPods/CocoaPods/pull/3932

It seems to me that all commands that deal with `--sources` should be unified and the behavior should be consistent.

Feel free to reject if I am missing some reason why these commands shouldn't default to all repos.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
+* Update `spec` and `lib` commands to use all sources by default.
+  [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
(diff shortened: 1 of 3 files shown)


segiddins:
@CocoaPods/core thoughts on this?

I'm kinda leaning against, since it won't reflect what `pod trunk push` does, but either way works for me.

AliSoftware:
I prefer `pod spec lint` and `pod lib lint` having the same behavior as `pod trunk push`. Forcing the pod author to use `--sources` with `pod spec lint` and `pod lib lint` ensures that the author understands that his/her pod will only work with private sources and can't be pushed to trunk but can only be `pod repo push` on a private specs repo.

dnkoutso:
SGTM! Closing.

PR #4012, opened 2015-08-12
[Validator] Print the name of the attribute that has the result

segiddins:
This means we'll actually print the part of the spec that's failing, which can be non-obvious in some cases, like when the invocation of xcodebuild fails.

\c @orta


orta:
More info on fails is definitely a good thing :+1:

PR #4028, opened 2015-08-16
[TargetIntegrator] Remove embed frameworks build phase where not needed

mrackwitz:
Fixes issues like #3905.
* [x] CHANGELOG


segiddins:
No point doing this when my PR will just add them back, given they'd be shipped in the same release?

mrackwitz:
No, your PR would just add them when integrating with static libraries for apps and tests. This ensures they are removed e.g. for extensions. This shouldn't be conflicting.

segiddins:
Gotcha.

segiddins:
This will conflict git-wise with my PR, but that's my fault for it being open so long. :+1: excellent work @mrackwitz !

PR #4097, opened 2015-08-28
[Resolver] Allow platform detection to work for nested dependencies

### LINKED ISSUE #4084: pod lint uses incompatible dependency version
My podspec specifies platform ```:ios, '6.0'``` and a dependency to AFNetworking ```~> 2.5```. When i reference this as a development pod the best compatible version of AFNetworking gets installed, which is 2.5.4 because 2.6 requires iOS 7. But when i lint this podspec i get the following notice

    [!] The platform of the target `Pods` (iOS 6.0) may not be compatible with `AFNetworking (2.6.0)` which has a minimum requirement of iOS 7.0 - OS X 10.9 - watchOS 2.0.

Obviously an incompatible version of AFNetworking is beeing used during lint.

segiddins:
Closes #4084.

- [x] Test
- [x] CHANGELOG

### DIFF
--- a/lib/cocoapods/installer/analyzer.rb
+++ b/lib/cocoapods/installer/analyzer.rb
@@ -537,7 +537,7 @@ module Pod
         end
 
         specs_by_target = nil
-        UI.section "Resolving dependencies of #{UI.path podfile.defined_in_file}" do
+        UI.section "Resolving dependencies of #{UI.path(podfile.defined_in_file) || 'Podfile'}" do
           resolver = Resolver.new(sandbox, podfile, locked_dependencies, sources)
           specs_by_target = resolver.resolve
           specs_by_target.values.flatten(1).each(&:validate_cocoapods_version)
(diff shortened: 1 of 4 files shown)


lib/cocoapods/installer/analyzer.rb
```diff
@@ -532,7 +532,7 @@ def resolve_dependencies
         end
 
         specs_by_target = nil
-        UI.section "Resolving dependencies of #{UI.path podfile.defined_in_file}" do
+        UI.section "Resolving dependencies of #{UI.path(podfile.defined_in_file) || 'Podfile'}" do
```
kylef: This change seems unrelated? Also, shouldn't `podfile.defined_in_file` always have a value?

segiddins: Not when linting

segiddins: (Which is what prompted this PR, so I made that change whilst debugging)

kylef: :white_check_mark:

segiddins:
Please review @kylef , this should be mergable.

kylef:
Good, but it needs to be :green_apple: green.

segiddins:
Will be green once https://github.com/CocoaPods/CocoaPods/pull/4104#issuecomment-136759535 is merged

PR #4104, opened 2015-08-29
[PodTargetInstaller] Mark all headers with availability levels

### LINKED ISSUE #4069: Documentation for public_header_files could use clarification
(I hope this is an acceptable issue 🌈)

The pod spec documentation around header files is potentially confusing for pod authors. The spec says:

```
A list of file patterns that should be used as public headers. 

These are the headers that will be exposed to the user’s project and from which documentation will be generated. If no public headers are specified then all the headers are considered public.
```

It is very clear that files matched by this pattern will become public headers, but what CP means by a public header is not exactly the same as what a stock Xcode project means by a public header, so it might be helpful to clarify the difference.

A pod author may expect for those headers to be marked as public in the Xcode target, which will only happen if the end user has specified that they will be building frameworks. As a pod author who has been mightily confused by this, I think it would be good to clarify exactly what the effect of this setting will be (they are copied to the **cocoapods** public header directory where they can be found by projects using the CP xcconfig files), which is not the same as the traditional Xcode public header file setting (they get copied to the built projects directory).

segiddins:
Closes #4069.

\c @mrackwitz

### DIFF
--- a/lib/cocoapods/installer/target_installer/pod_target_installer.rb
+++ b/lib/cocoapods/installer/target_installer/pod_target_installer.rb
@@ -89,15 +89,13 @@ module Pod
           header_file_refs = headers.map { |sf| project.reference_for_path(sf) }
           native_target.add_file_references(header_file_refs) do |build_file|
             # Set added headers as public if needed
-            if target.requires_frameworks?
-              build_file.settings ||= {}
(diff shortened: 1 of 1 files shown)


kylef:
:+1:

PR #4136, opened 2015-09-07
[EmbedFrameworksScript] Speedup by attempting to embed each stdlib just once

mrackwitz:
Before the Swift stdlibs were rsynced and codesigned multiple times but replaced everytime.
I updated the example to test that, but it's not a really suitable test case as it has just one dependency and doesn't replicate why the stdlib integration is needed at all.


segiddins:
👍🏻 great work!

lib/cocoapods/generator/embed_frameworks_script.rb
```diff
@@ -77,15 +79,15 @@ def script
...
+            code_sign_if_enabled "${FRAMEWORKS_DIR}/$(basename "$1")"
 
             # Embed linked Swift runtime libraries
             local swift_runtime_libs
             swift_runtime_libs=$(xcrun otool -LX "$binary" | grep --color=never @rpath/libswift | sed -E s/@rpath\\\\/\\(.+dylib\\).*/\\\\1/g | uniq -u  && exit ${PIPESTATUS[0]})
             for lib in $swift_runtime_libs; do
-              echo "rsync -auv \\"${SWIFT_STDLIB_PATH}/${lib}\\" \\"${destination}\\""
-              rsync -auv "${SWIFT_STDLIB_PATH}/${lib}" "${destination}"
-              code_sign_if_enabled "${destination}/${lib}"
+              if [[ $SWIFT_RUNTIME_LIBS_TO_INTEGRATE != *"$lib"* ]]; then
```
segiddins: could just use `uniq`, but this works too!

segiddins:
If we decide not to merge https://github.com/CocoaPods/CocoaPods/pull/4221, this is good with a CHANGELOG, but we should hold off on a merge since that PR, if it works, is preferable.

segiddins:
@mrackwitz is this worth rebasing, given this code no longer will be hit on Xcode 7?

PR #4146, opened 2015-09-09
Improve de-duplication for non-trivial cases

### LINKED ISSUE #4034: Target Deduplication removes necessary Pods project targets 🌈
I ran into this issue when upgrading from CocoaPods `0.37.2` to `0.38.0`, and it's not fixed as of `0.38.2`.

This issue is a bit long, but only because I wanted to make it very clear what is happening -- the [sample project](https://github.com/smileyborg/CPTargetDeduplication) should make this very easy to understand.

## Background
I ran into this issue with the QuickBooks app I work on, which has an unusually complex build configuration. Aside from CocoaPods, we have a workspace of multiple separate Xcode projects, and each Xcode project generally has 2 framework targets -- one for iOS and one for Mac. This is because we use these independent Xcode projects (and their frameworks) to share code between multiple iOS and OS X apps. We're also using CocoaPods to manage our third-party dependencies.

I've created a [small git repo](https://github.com/smileyborg/CPTargetDeduplication) that contains a sample workspace that represents a simple version of the structure of our real QuickBooks workspace. In this git repo:

* `ProjectABWorkspace`
  * `ProjectA` (has a dependency on `Pods`)
    * `ProjectA-iOS`
    * `ProjectA-Mac`
  * `ProjectB` (has a dependency on `ProjectA`, and on `Pods`)
    * `ProjectB-iOS`
    * `ProjectB-Mac`
  * `Pods`

As you can see, there are 2 separate Xcode projects and a Pods project (generated by CocoaPods). Each of the `ProjectA` and `ProjectB` projects has 2 framework targets it produces, one for iOS and one for OS X. Both `ProjectA` and `ProjectB` depend on a CocoaPod (INTUGroupedArray). And finally, `ProjectB` depends on `ProjectA`, by embedding & linking the `ProjectA.framework`.

The [Podfile in this repo](https://github.com/smileyborg/CPTargetDeduplication/blob/master/Podfile) has a separate Pod target for each framework target, making 4 total Pod targets.

## The Issue
With `0.37.2` all of the framework targets build fine. (You can checkout [this commit](https://github.com/smileyborg/CPTargetDeduplication/commit/dda3a6cdf5b6fe732c5fa390728e8dffac1694a5) to see that.)

However, with `0.38.0` or higher, both of the ProjectB framework targets fail to build (use the `ProjectB-iOS` and `ProjectB-Mac` schemes). The build failure itself is within the `ProjectBFile.m` import to the INTUGroupedArray CocoaPod.

The actual issue here is more clear when you look at the git log for the repo. If you look at the diff for [the commit that runs `pod install` using `0.38.2`](https://github.com/smileyborg/CPTargetDeduplication/commit/bbd8d1f16905aa653e7e2ee31326c5a561767d0f) (after already having run `pod install` using `0.37.2`), you can see two significant sets of deletions:
* All the files in `Pods/Target Support Files/Pods-ProjectB-Mac-INTUGroupedArray/` have been removed
* All the files in `Pods/Target Support Files/Pods-ProjectB-iOS-INTUGroupedArray/` have been removed

If you look at the Pods project itself at this point, it is also missing the `Pods-ProjectB-iOS-INTUGroupedArray` and `Pods-ProjectB-iOS-INTUGroupedArray` targets. So the root cause of the compilation error is that the INTUGroupedArray pod is no longer being linked with the ProjectB targets.

CocoaPods `0.38.0` introduced [target deduplication](http://blog.cocoapods.org/CocoaPods-0.38/#target-deduplication), which appears to be causing this issue. It looks to me like the *Target Deduplicator* decided these were not necessary (even though they were) and removed them.

This issue is blocking us from upgrading beyond CocoaPods `0.37.2`.

Let me know if you have any questions or want more information!

mrackwitz:
Attempts to fix #4034.
Attempts to fix https://github.com/CocoaPods/CocoaPods/issues/3838.

Given that there is an integration like:
```ruby
target 'FooApp' do
  platform :osx, '10.10'
  pod 'BananaKit'
end

target 'FooAppEx' do
  platform :osx, '10.10'
  pod 'BananaKit'
end

target 'BarApp' do
  platform :ios, '9.0'
  pod 'BananaKit'
end

target 'BarAppEx' do
  platform :ios, '9.0'
  pod 'BananaKit'
end
```

The analyzer will generate a model with the following pod targets:
* `Pods-FooApp-BananaKit` with aggregate targets `Pods-FooApp` and `Pods-FooAppEx`
* `Pods-BarApp-BananaKit` with aggregate targets `Pods-BarApp` and `Pods-BarAppEx`

So even though a pod is used on multiple platforms, the analyzer is able to recognize that it could be reused between app and its extension on the same platform in this example, but generally not limited to that.

The problem is that the pod target will be build in a scoped build dir, which is not available from all apps and pod targets. To stick with the example: the scoped build dir is added to the `FRAMEWORK_SEARCH_PATHS` of `FooApp` but not `FooAppEx`, because the deduplication/scoping logic makes at some places in the code the assumption if a pod target is scoped than just one target definition belongs to it. This may seem to be easily solved by just adding the same framework search path to `FooAppEx` as well, but this wouldn't work in any case, because there could be other dependencies from `FooApp`, which are build with e.g. other subspecs and so not thought for `FooAppEx`, but become visible which can cause indeterministic behavior and warnings because there are two (or more versions) of the same framework in the framework search paths.

So the solution is to change the way, they are scoped and ensure that pods targets are built to a directory, which doesn't cause clashes with other pod targets. But this will cause massive changes to the integration, likely breaking for anything relying on how scoped targets where named before.
Where a scoped target was named before `Pods-MyAppExtension-BananaKit`, the name/scope now indicates what's special about the build variant, which allow these cases:

1. `BananaKit-iOS`: if BananaKit was used on different platforms
* `BananaKit-iOS9.0`: if BananaKit was used on different platforms and versions
* `BananaKit-PalmTree-Leaf`: if BananaKit was used with different sets of subspecs, the name shows which subspecs are special to the specific target, leaving out common subspecs
* ~~`BananaKit-MyApp-MyAppExtension`: if both of the condition above apply, the name shows all target definitions where it is in use~~
* `BananaKit-iOS9.0-PalmTree-Leaf`: if both of the condition above apply, the name uses both schemes in combination

The order here is relatively arbitrary, but tries to prefer a shorter and simpler target name first, as especially 3. & 4. can become ugly and very long.

#### ToDo

* [x] Test for case 2, if relevant?
* [x] Test for case 3 with multiple subspec names
* [x] ~~Test for case 4 with multiple target names~~
* [x] Test for case 5 with multiple target names
* [x] Rebuild integration specs
* [x] Changelog

#### Note

In case of 1. the scoping of the build products could be dropped, because the default `$CONFIGURATION_BUILD_DIR` covers already a scoping by platform and architecture.

:warning: Transitive dependencies are still causing a duplication with the target based naming scheme.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 ##### Bug Fixes
 
 * Pods are built by default in another scoping level of the build products
-  directory identified by their name to exclude name clashes among
+  directory identified by their name to prevent name clashes among
   dependencies.  
   [Marius Rackwitz](https://github.com/mrackwitz)
   [#4146](https://github.com/CocoaPods/CocoaPods/pull/4146)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,12 +8,21 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
-* None.  
+* Allow deduplication to take effect even when the same pod is used with
+  different sets of subspecs across different platforms.
+  This changes the general naming scheme scoped pod targets. They are
+  suffixed now on base of what makes them different among others for the
+  same root spec instead of being prefixed by the dependant target.  
+  [Marius Rackwitz](https://github.com/mrackwitz)
+  [#4146](https://github.com/CocoaPods/CocoaPods/pull/4146)
 
 ##### Bug Fixes
 
-* None.  
-
+* Pods are built by default in another scoping level of the build products
+  directory identified by their name to exclude name clashes among
+  dependencies.  
+  [Marius Rackwitz](https://github.com/mrackwitz)
+  [#4146](https://github.com/CocoaPods/CocoaPods/pull/4146)
 
 ## 1.0.0.beta.3 (2016-02-03)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,11 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Marius Rackwitz](https://github.com/mrackwitz)
   [#4146](https://github.com/CocoaPods/CocoaPods/pull/4146)
 
+* Fix mixed integrations where static libraries are used along frameworks
+  from different target definitions in one Podfile.  
+  [Marius Rackwitz](https://github.com/mrackwitz)
+  [#4146](https://github.com/CocoaPods/CocoaPods/pull/4146)
+
 ## 1.0.0.beta.3 (2016-02-03)
 
 ##### Breaking
--- a/lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
+++ b/lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
@@ -97,7 +97,7 @@ module Pod
           if target.requires_frameworks?
             build_pod_targets = pod_targets.select(&:should_build?)
             framework_header_search_paths = build_pod_targets.map do |target|
-              "#{target.configuration_build_dir}/#{target.product_name}/Headers"
+              "#{target.build_product_path}/Headers"
             end
             build_settings = {
               # Make framework headers discoverable by `import "…"`
--- a/lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
+++ b/lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
@@ -59,6 +59,7 @@ module Pod
             'PODS_ROOT' => target.relative_pods_root,
             'GCC_PREPROCESSOR_DEFINITIONS' => '$(inherited) COCOAPODS=1',
             'FRAMEWORK_SEARCH_PATHS' => '$(inherited) ',
+            'LIBRARY_SEARCH_PATHS' => '$(inherited) ',
           }
           if pod_targets.any?(&:uses_swift?)
             config['EMBEDDED_CONTENT_CONTAINS_SWIFT'] = 'YES'
--- a/lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
+++ b/lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
@@ -94,15 +94,11 @@ module Pod
         #
         def settings_to_import_pod_targets
           if target.requires_frameworks?
-            framework_header_search_paths = pod_targets.select(&:should_build?).map do |target|
-              if target.scoped?
-                "$PODS_FRAMEWORK_BUILD_PATH/#{target.product_name}/Headers"
-              else
-                "$CONFIGURATION_BUILD_DIR/#{target.product_name}/Headers"
-              end
+            build_pod_targets = pod_targets.select(&:should_build?)
+            framework_header_search_paths = build_pod_targets.map do |target|
+              "#{target.relative_configuration_build_dir}/#{target.product_name}/Headers"
             end
             build_settings = {
-              'PODS_FRAMEWORK_BUILD_PATH' => XCConfigHelper.quote([target.scoped_configuration_build_dir]),
               # Make framework headers discoverable by `import "…"`
               'OTHER_CFLAGS' => '$(inherited) ' + XCConfigHelper.quote(framework_header_search_paths, '-iquote'),
             }
@@ -112,8 +108,10 @@ module Pod
               build_settings['HEADER_SEARCH_PATHS'] = '$(inherited) ' + XCConfigHelper.quote(library_header_search_paths)
               build_settings['OTHER_CFLAGS'] += ' ' + XCConfigHelper.quote(library_header_search_paths, '-isystem')
             end
-            if pod_targets.any? { |t| t.should_build? && t.scoped? }
-              build_settings['FRAMEWORK_SEARCH_PATHS'] = '$PODS_FRAMEWORK_BUILD_PATH'
+            scoped_pod_targets = build_pod_targets.select(&:scoped?)
+            unless scoped_pod_targets.empty?
+              framework_search_paths = scoped_pod_targets.map(&:relative_configuration_build_dir).uniq
+              build_settings['FRAMEWORK_SEARCH_PATHS'] = XCConfigHelper.quote(framework_search_paths)
             end
             build_settings
           else
--- a/lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
+++ b/lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
@@ -70,7 +70,6 @@ module Pod
           generate_settings_to_import_pod_targets
 
           XCConfigHelper.add_target_specific_settings(target, @xcconfig)
-          XCConfigHelper.add_settings_for_dependent_targets(target, pod_targets, @xcconfig)
 
           generate_vendored_build_settings
           generate_other_ld_flags
@@ -131,9 +130,11 @@ module Pod
         #  - `@import …;` / `import …`
         #
         def generate_settings_to_import_pod_targets
+          @xcconfig.merge! XCConfigHelper.settings_for_dependent_targets(target, pod_targets)
           @xcconfig.merge!(settings_to_import_pod_targets)
           target.search_paths_aggregate_targets.each do |search_paths_target|
             generator = AggregateXCConfig.new(search_paths_target, configuration_name)
+            @xcconfig.merge! XCConfigHelper.settings_for_dependent_targets(nil, search_paths_target.pod_targets)
             @xcconfig.merge!(generator.settings_to_import_pod_targets)
           end
         end
--- a/lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
+++ b/lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
@@ -70,6 +70,7 @@ module Pod
           generate_settings_to_import_pod_targets
 
           XCConfigHelper.add_target_specific_settings(target, @xcconfig)
+          XCConfigHelper.add_settings_for_dependent_targets(target, pod_targets, @xcconfig)
 
           generate_vendored_build_settings
           generate_other_ld_flags
@@ -107,11 +108,6 @@ module Pod
               library_header_search_paths = target.sandbox.public_headers.search_paths(target.platform)
               build_settings['HEADER_SEARCH_PATHS'] = '$(inherited) ' + XCConfigHelper.quote(library_header_search_paths)
               build_settings['OTHER_CFLAGS'] += ' ' + XCConfigHelper.quote(library_header_search_paths, '-isystem')
-            end
-            scoped_pod_targets = build_pod_targets.select(&:scoped?)
-            unless scoped_pod_targets.empty?
-              framework_search_paths = scoped_pod_targets.map(&:relative_configuration_build_dir).uniq
-              build_settings['FRAMEWORK_SEARCH_PATHS'] = XCConfigHelper.quote(framework_search_paths)
             end
             build_settings
           else
--- a/lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
+++ b/lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
@@ -97,7 +97,7 @@ module Pod
           if target.requires_frameworks?
             build_pod_targets = pod_targets.select(&:should_build?)
             framework_header_search_paths = build_pod_targets.map do |target|
-              "#{target.relative_configuration_build_dir}/#{target.product_name}/Headers"
+              "#{target.configuration_build_dir}/#{target.product_name}/Headers"
             end
             build_settings = {
               # Make framework headers discoverable by `import "…"`
--- a/lib/cocoapods/generator/xcconfig/pod_xcconfig.rb
+++ b/lib/cocoapods/generator/xcconfig/pod_xcconfig.rb
@@ -45,12 +45,14 @@ module Pod
           search_paths = target_search_paths.concat(sandbox_search_paths).uniq
 
           config = {
+            'FRAMEWORK_SEARCH_PATHS' => '$(inherited) ',
+            'GCC_PREPROCESSOR_DEFINITIONS' => '$(inherited) COCOAPODS=1',
+            'HEADER_SEARCH_PATHS' => XCConfigHelper.quote(search_paths),
+            'LIBRARY_SEARCH_PATHS' => '$(inherited) ',
             'OTHER_LDFLAGS' => XCConfigHelper.default_ld_flags(target),
             'PODS_ROOT' => '${SRCROOT}',
-            'HEADER_SEARCH_PATHS' => XCConfigHelper.quote(search_paths),
-            'GCC_PREPROCESSOR_DEFINITIONS' => '$(inherited) COCOAPODS=1',
-            'SKIP_INSTALL' => 'YES',
             'PRODUCT_BUNDLE_IDENTIFIER' => 'org.cocoapods.${PRODUCT_NAME:rfc1034identifier}',
+            'SKIP_INSTALL' => 'YES',
             # 'USE_HEADERMAP' => 'NO'
           }
 
@@ -61,7 +63,7 @@ module Pod
             @xcconfig.merge!(file_accessor.spec_consumer.pod_target_xcconfig)
           end
           XCConfigHelper.add_target_specific_settings(target, @xcconfig)
-          XCConfigHelper.add_settings_for_dependent_targets(target, target.dependent_targets, @xcconfig)
+          @xcconfig.merge! XCConfigHelper.settings_for_dependent_targets(target, target.dependent_targets)
           @xcconfig
         end
 
--- a/lib/cocoapods/generator/xcconfig/pod_xcconfig.rb
+++ b/lib/cocoapods/generator/xcconfig/pod_xcconfig.rb
@@ -8,6 +8,13 @@ module Pod
       # required by CocoaPods.
       #
       class PodXCConfig
+        # @return [String] Defined to hold the default Xcode build path, so
+        #         that when this is overridden per {PodTarget}, it is still
+        #         possible to reference other build products relative to the
+        #         original path.
+        #
+        SHARED_BUILD_DIR_VARIABLE = 'PODS_SHARED_BUILD_DIR'.freeze
+
         # @return [Target] the target represented by this xcconfig.
         #
         attr_reader :target
@@ -57,13 +64,16 @@ module Pod
 
           @xcconfig = Xcodeproj::Config.new(config)
 
-          if target.requires_frameworks? && target.scoped?
+          if target.requires_frameworks?
             build_settings = {
+              SHARED_BUILD_DIR_VARIABLE => '$(BUILD_DIR)/$(CONFIGURATION)$(EFFECTIVE_PLATFORM_NAME)',
               'CONFIGURATION_BUILD_DIR' => target.configuration_build_dir,
             }
             scoped_dependent_targets = target.dependent_targets.select { |t| t.should_build? && t.scoped? }
             unless scoped_dependent_targets.empty?
-              framework_search_paths = scoped_dependent_targets.map(&:relative_configuration_build_dir).uniq
+              framework_search_paths = scoped_dependent_targets.map do |target|
+                target.relative_configuration_build_dir("$#{SHARED_BUILD_DIR_VARIABLE}")
+              end
               build_settings['FRAMEWORK_SEARCH_PATHS'] = XCConfigHelper.quote(framework_search_paths)
             end
             @xcconfig.merge!(build_settings)
--- a/lib/cocoapods/generator/xcconfig/pod_xcconfig.rb
+++ b/lib/cocoapods/generator/xcconfig/pod_xcconfig.rb
@@ -8,13 +8,6 @@ module Pod
       # required by CocoaPods.
       #
       class PodXCConfig
-        # @return [String] Defined to hold the default Xcode build path, so
-        #         that when this is overridden per {PodTarget}, it is still
-        #         possible to reference other build products relative to the
-        #         original path.
-        #
-        SHARED_BUILD_DIR_VARIABLE = 'PODS_SHARED_BUILD_DIR'.freeze
-
         # @return [Target] the target represented by this xcconfig.
         #
         attr_reader :target
@@ -57,33 +50,18 @@ module Pod
             'HEADER_SEARCH_PATHS' => XCConfigHelper.quote(search_paths),
             'GCC_PREPROCESSOR_DEFINITIONS' => '$(inherited) COCOAPODS=1',
             'SKIP_INSTALL' => 'YES',
-            'FRAMEWORK_SEARCH_PATHS' => '$(inherited) ',
             'PRODUCT_BUNDLE_IDENTIFIER' => 'org.cocoapods.${PRODUCT_NAME:rfc1034identifier}',
             # 'USE_HEADERMAP' => 'NO'
           }
 
           @xcconfig = Xcodeproj::Config.new(config)
 
-          if target.requires_frameworks?
-            build_settings = {
-              SHARED_BUILD_DIR_VARIABLE => '$(BUILD_DIR)/$(CONFIGURATION)$(EFFECTIVE_PLATFORM_NAME)',
-              'CONFIGURATION_BUILD_DIR' => target.configuration_build_dir,
-            }
-            scoped_dependent_targets = target.dependent_targets.select { |t| t.should_build? && t.scoped? }
-            unless scoped_dependent_targets.empty?
-              framework_search_paths = scoped_dependent_targets.map do |target|
-                target.relative_configuration_build_dir("$#{SHARED_BUILD_DIR_VARIABLE}")
-              end
-              build_settings['FRAMEWORK_SEARCH_PATHS'] = XCConfigHelper.quote(framework_search_paths)
-            end
-            @xcconfig.merge!(build_settings)
-          end
-
           XCConfigHelper.add_settings_for_file_accessors_of_target(target, @xcconfig)
           target.file_accessors.each do |file_accessor|
             @xcconfig.merge!(file_accessor.spec_consumer.pod_target_xcconfig)
           end
           XCConfigHelper.add_target_specific_settings(target, @xcconfig)
+          XCConfigHelper.add_settings_for_dependent_targets(target, target.dependent_targets, @xcconfig)
           @xcconfig
         end
 
--- a/lib/cocoapods/generator/xcconfig/pod_xcconfig.rb
+++ b/lib/cocoapods/generator/xcconfig/pod_xcconfig.rb
@@ -59,10 +59,13 @@ module Pod
 
           if target.requires_frameworks? && target.scoped?
             build_settings = {
-              'PODS_FRAMEWORK_BUILD_PATH' => XCConfigHelper.quote([target.configuration_build_dir]),
-              'FRAMEWORK_SEARCH_PATHS' => '$PODS_FRAMEWORK_BUILD_PATH',
-              'CONFIGURATION_BUILD_DIR' => '$PODS_FRAMEWORK_BUILD_PATH',
+              'CONFIGURATION_BUILD_DIR' => target.configuration_build_dir,
             }
+            scoped_dependent_targets = target.dependent_targets.select { |t| t.should_build? && t.scoped? }
+            unless scoped_dependent_targets.empty?
+              framework_search_paths = scoped_dependent_targets.map(&:relative_configuration_build_dir).uniq
+              build_settings['FRAMEWORK_SEARCH_PATHS'] = XCConfigHelper.quote(framework_search_paths)
+            end
             @xcconfig.merge!(build_settings)
           end
 
--- a/lib/cocoapods/generator/xcconfig/xcconfig_helper.rb
+++ b/lib/cocoapods/generator/xcconfig/xcconfig_helper.rb
@@ -219,23 +219,23 @@ module Pod
           dependent_targets = dependent_targets.select(&:should_build?)
           has_configuration_build_dir = target.respond_to?(:configuration_build_dir)
           if has_configuration_build_dir
+            build_dir_var = "$#{SHARED_BUILD_DIR_VARIABLE}"
             build_settings = {
-              'CONFIGURATION_BUILD_DIR' => target.configuration_build_dir,
+              'CONFIGURATION_BUILD_DIR' => target.configuration_build_dir(build_dir_var),
               SHARED_BUILD_DIR_VARIABLE => '$(BUILD_DIR)/$(CONFIGURATION)$(EFFECTIVE_PLATFORM_NAME)',
             }
-            build_dir_var = "$#{SHARED_BUILD_DIR_VARIABLE}"
           else
-            build_settings = {}
             build_dir_var = '$CONFIGURATION_BUILD_DIR'
+            build_settings = {}
           end
           unless dependent_targets.empty?
             framework_search_paths = []
             library_search_paths = []
             dependent_targets.each do |dependent_target|
               if dependent_target.requires_frameworks?
-                framework_search_paths << dependent_target.relative_configuration_build_dir(build_dir_var)
+                framework_search_paths << dependent_target.configuration_build_dir(build_dir_var)
               else
-                library_search_paths << dependent_target.relative_configuration_build_dir(build_dir_var)
+                library_search_paths << dependent_target.configuration_build_dir(build_dir_var)
               end
             end
             build_settings['FRAMEWORK_SEARCH_PATHS'] = '$(inherited) ' + XCConfigHelper.quote(framework_search_paths.uniq) unless framework_search_paths.empty?
--- a/lib/cocoapods/generator/xcconfig/xcconfig_helper.rb
+++ b/lib/cocoapods/generator/xcconfig/xcconfig_helper.rb
@@ -56,6 +56,8 @@ module Pod
         # @param  [Xcodeproj::Config] xcconfig
         #         The xcconfig to edit.
         #
+        # @return [void]
+        #
         def self.add_settings_for_file_accessors_of_target(target, xcconfig)
           target.file_accessors.each do |file_accessor|
             XCConfigHelper.add_spec_build_settings_to_xcconfig(file_accessor.spec_consumer, xcconfig)
@@ -80,6 +82,8 @@ module Pod
         # @param [Spec::FileAccessor] file_accessor
         #        The file accessor, which holds the list of static frameworks.
         #
+        # @return [void]
+        #
         def self.add_static_dependency_build_settings(target, xcconfig, file_accessor)
           file_accessor.vendored_static_frameworks.each do |vendored_static_framework|
             XCConfigHelper.add_framework_build_settings(vendored_static_framework, xcconfig, target.sandbox.root)
@@ -96,6 +100,8 @@ module Pod
         #
         # @param [Xcodeproj::Config] xcconfig
         #        The xcconfig to edit.
+        #
+        # @return [void]
         #
         def self.add_dynamic_dependency_build_settings(target, xcconfig)
           target.file_accessors.each do |file_accessor|
@@ -117,6 +123,8 @@ module Pod
         # @param  [Xcodeproj::Config] xcconfig
         #         The xcconfig to edit.
         #
+        # @return [void]
+        #
         def self.add_spec_build_settings_to_xcconfig(consumer, xcconfig)
           xcconfig.libraries.merge(consumer.libraries)
           xcconfig.frameworks.merge(consumer.frameworks)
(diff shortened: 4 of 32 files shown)


neonichu:
:tada: 

I wonder though if we should simply throw an error in the application extension case, since we already know that duplicated dependencies will lead to problems when archiving apps. Maybe at least show a warning that this probably won't work?

lib/cocoapods/installer/analyzer.rb
```diff
@@ -326,6 +327,45 @@ def generate_pod_targets(specs_by_target)
...
+            result = targets_by_distinctors.map do |distinctor, target_definitions|
+              names = target_definitions.map do |target_definition|
+                target_definition.root? ? "Pods" : target_definition.name
+              end
+              [distinctor, names.join('-')]
+            end
+          end
+        end
+
+        Hash[result.map { |d,name| [d,name.gsub('/', '_')] }]
```
segiddins: spacing. Use `tr` instead of `gsub`

mrackwitz:
>I wonder though if we should simply throw an error in the application extension case, since we already know that duplicated dependencies will lead to problems when archiving apps. Maybe at least show a warning that this probably won't work?

Throw an error or print a warning when app and app extension don't share the same dependency? We would need first be able to recognize whether both belong to each other.

neonichu:
@mrackwitz true, but shouldn't be too hard to obtain that information. Will look into that.

lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
```diff
@@ -94,15 +94,15 @@ def generate
         #
         def generate_settings_to_import_pod_targets
           if target.requires_frameworks?
-            framework_header_search_paths = pod_targets.select(&:should_build?).map do |target|
+            build_pod_targets = pod_targets.select(&:should_build?)
+            framework_header_search_paths = build_pod_targets.map do |target|
               if target.scoped?
```
segiddins: looks like it no longer needs to be a conditional

lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
```diff
@@ -112,8 +112,12 @@ def generate_settings_to_import_pod_targets
...
               build_settings['OTHER_CFLAGS'] += ' ' + XCConfigHelper.quote(library_header_search_paths, '-isystem')
             end
-            if pod_targets.any? { |t| t.should_build? && t.scoped? }
-              build_settings['FRAMEWORK_SEARCH_PATHS'] = '"$PODS_FRAMEWORK_BUILD_PATH"'
+            scoped_pod_targets = build_pod_targets.select(&:scoped?)
+            unless scoped_pod_targets.empty?
+              framework_search_paths = scoped_pod_targets.map do |target|
+                target.relative_configuration_build_dir
+              end.uniq
+              build_settings['FRAMEWORK_SEARCH_PATHS'] = XCConfigHelper.quote(framework_search_paths)
```
segiddins: needs `$(inherited)` also

lib/cocoapods/installer/analyzer.rb
```diff
@@ -326,6 +327,45 @@ def generate_pod_targets(specs_by_target)
...
+            common_specs = all_spec_variants.reduce(all_spec_variants, &:&)
+            result = targets_by_distinctors.map do |distinctor, _|
+              specs, _ = *distinctor
+              specs -= common_specs
+              [distinctor, specs.map { |spec| spec.name.split('/')[1..-1] }.join('-')]
+            end
+          else
+            result = targets_by_distinctors.map do |distinctor, target_definitions|
+              names = target_definitions.map do |target_definition|
+                target_definition.root? ? "Pods" : target_definition.name
```
segiddins: single quotes

lib/cocoapods/target/pod_target.rb
```diff
@@ -61,7 +68,7 @@ def initialize(specs, target_definitions, sandbox, scoped = false)
     #
     def scoped
       target_definitions.map do |target_definition|
-        PodTarget.new(specs, [target_definition], sandbox, true).tap do |target|
+        PodTarget.new(specs, [target_definition], sandbox, target_definition.label).tap do |target|
```
segiddins: is that always the correct scope suffix?

lib/cocoapods/target/pod_target.rb
```diff
@@ -75,7 +82,7 @@ def scoped
     #
     def label
       if scoped?
-        "#{target_definitions.first.label}-#{root_spec.name}"
+        "#{root_spec.name}-#{scope_suffix}"
```
segiddins: Might want to make it `Pods-`, purely for aesthetics

segiddins:
This looks very good! Biggest thing for me is seeing a bevy of unit tests around the method that generates scope suffixes.
We also need to investigate if there's a max length we can make a pod target's name

lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
mrackwitz: Indeed.

lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
mrackwitz: This will cause that the aggregate target inherits `FRAMEWORK_SEARCH_PATHS` from the user project. I see how that would make sense for the user project, but it could cause unexpected issues with the aggregate target through misconfiguration. I don't think that this should belong into this PR.

lib/cocoapods/target/pod_target.rb
mrackwitz: Increase the length and reduces the overview because it breaks the order with fully de-duplicated pods.

lib/cocoapods/target/pod_target.rb
mrackwitz: You can't reproduce with that, what the analyzer archives. But the label of the target definition is always a satisfying denominator. This is used in the fallback case and as helper for our tests.

lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
segiddins: aggregate target never actually links any frameworks, so it shouldn't be an issue?

segiddins: Since this is set as the user target's configuration, it really does need to have inherited

lib/cocoapods/target/pod_target.rb
segiddins: :+1:

lib/cocoapods/target/pod_target.rb
segiddins: ok, so the target definition label is always _a_ correct scope suffix, but it is the most broad and therefore undesirable? cool :+1:

segiddins:
@mrackwitz I can pick this up if you want, just let me know what needs to be done

segiddins:
@mrackwitz I tried to rebase this yesterday but ran into some issues. Could you please handle rebasing it?

benasher44:
Noted in #3838 that I think I'm having a similar issue, and I'm using 1.0.0.beta.2. I'm happy to give this branch a try and see if it solves my issue described there and report back, once this gets PR rebased.

jafara:
Would love to see this land!

lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
mrackwitz: Turns out that this is already set in `AggregateXCConfig#generate`. The merging keeps it intact.

mrackwitz:
I have reawaked this PR and extended it with the outstanding tests. In progress of that, I fixed the naming scheme as seen in the updated comment above as well, so that the names stay consistent. But this needs still further testing with real-world projects, especially with those, which where concerned from this issue.

CHANGELOG.md
```diff
@@ -6,6 +6,16 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ## Master
 
+##### Breaking
```
segiddins: I think this is a bug fix, not a breaking change?

lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
```diff
@@ -112,8 +108,12 @@ def settings_to_import_pod_targets
...
               build_settings['OTHER_CFLAGS'] += ' ' + XCConfigHelper.quote(library_header_search_paths, '-isystem')
             end
-            if pod_targets.any? { |t| t.should_build? && t.scoped? }
-              build_settings['FRAMEWORK_SEARCH_PATHS'] = '$PODS_FRAMEWORK_BUILD_PATH'
+            scoped_pod_targets = build_pod_targets.select(&:scoped?)
+            unless scoped_pod_targets.empty?
+              framework_search_paths = scoped_pod_targets.map do |target|
+                target.relative_configuration_build_dir
+              end.uniq
+              build_settings['FRAMEWORK_SEARCH_PATHS'] = XCConfigHelper.quote(framework_search_paths)
```
segiddins: Inherited

lib/cocoapods/generator/xcconfig/pod_xcconfig.rb
```diff
@@ -59,10 +59,15 @@ def generate
...
           if target.requires_frameworks? && target.scoped?
             build_settings = {
-              'PODS_FRAMEWORK_BUILD_PATH' => XCConfigHelper.quote([target.configuration_build_dir]),
-              'FRAMEWORK_SEARCH_PATHS' => '$PODS_FRAMEWORK_BUILD_PATH',
-              'CONFIGURATION_BUILD_DIR' => '$PODS_FRAMEWORK_BUILD_PATH',
+              'CONFIGURATION_BUILD_DIR' => XCConfigHelper.quote([target.configuration_build_dir]),
             }
+            scoped_dependent_targets = target.dependent_targets.select { |t| t.should_build? && t.scoped? }
+            unless scoped_dependent_targets.empty?
+              framework_search_paths = scoped_dependent_targets.map do |target|
```
segiddins: `map(&:relative_...)`

lib/cocoapods/target/pod_target.rb
```diff
@@ -59,15 +66,17 @@ def initialize(specs, target_definitions, sandbox, scoped = false)
 
     # @param [Hash{Array => PodTarget}] cache
     #        the cached PodTarget for a previously scoped (specs, target_definition)
+    # @param  [String] scope_suffix
+    #         @see PodTarget#scope_suffix
     # @return [Array<PodTarget>] a scoped copy for each target definition.
     #
-    def scoped(cache = {})
+    def scoped(cache = {}, scope_suffix=nil)
```
segiddins: Missing spaces around `=`

benasher44:
Okay, I just tried this out, and it still won't compile, but the error is slightly different. To start out, there's one thing that stands out as a something that looks weird/incorrect. The file in question that won't compile has a framework search path that looks like this:

-F/Users/basher/Code/iphone/Pods/\"/Users/basher/Library/Developer/Xcode/DerivedData/Yelp-fyngslzoeilfvcblkbskorhepqvd/Build/Products/Debug-iphonesimulator/Pods-YelpShared-Yelp\"

It looks like it's trying to make it so that it has the framework search paths of /Users/basher/Code/iphone/Pods/ and /Users/basher/Library/Developer/Xcode/DerivedData/Yelp-fyngslzoeilfvcblkbskorhepqvd/Build/Products/Debug-iphonesimulator/Pods-YelpShared-Yelp. The syntax here for specifying multiple framework search paths seems funny/incorrect, and the correct include paths should be
- /Users/basher/Library/Developer/Xcode/DerivedData/Yelp-fyngslzoeilfvcblkbskorhepqvd/Build/Products/Debug-iphonesimulator (this is where the pod framework is whose subspec is used by one target and whose pod framework in its entirety is used by a pod in the main target)
- /Users/basher/Library/Developer/Xcode/DerivedData/Yelp-fyngslzoeilfvcblkbskorhepqvd/Build/Products/Debug-iphonesimulator/Pods-YelpShared-Yelp

Let me know if you need more info to help debug this.

lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
segiddins: :+1:

CHANGELOG.md
mrackwitz: I classified it as breaking because it will cause huge changes to the integration. Actually it shouldn't break anything beside podfile installation hooks, which could be affected by any change. But then I would rather say it's an enhancement. This PR is still labeled as defect, because it predates #4147. Since that was merged, everything what is left here to do was more a known feature limitation, I guess.

segiddins:
Is there any chance this will also fix https://github.com/CocoaPods/CocoaPods/issues/4751?

mrackwitz:
@benasher44: Thanks for the quick feedback. This was a wrongly quoted absolute path in a build setting which only takes single values and Xcode so interprets as relative. I could track it down and fix it.

mrackwitz:
>Is there any chance this will also fix #4751?

For reference: No, this was done in #4818.

benasher44:
@mrackwitz no problem! Was the build setting you were referring to something I should look into that might be incorrect? I'm happy to dig around if that's the case.

I've pulled in your latest changes, and now I'm back to the original error. Here's a high level of what the Podfile looks like:

```
abstract_target 'YelpShared' do
  pod 'A/subspec'

  target 'Yelp' do
    pod 'A'
    pod 'B'
    pod 'C' (in its podspec, depends on A and B)
  end

  target 'Extension'
end
```

When it goes to compile C, it has a compiler error looking for pod B, but it might as well be A too because the framework search path is set such that the compiler wouldn't be able to find either. It just happens to fail looking for a header in B first. The issue is that C depends on pods, A and B, which have been built and installed in $BUILT_PRODUCTS_DIR, but the framework search path is set to $BUILT_PRODUCTS_DIR/Pods-YelpShared-Yelp.

Does this help?

segiddins:
RuboCop failures

mrackwitz:
@benasher44: Thanks for the feedback, but I'm still missing some info here. Can you post a full list of targets and their `CONFIGURATION_BUILD_DIR`?
What puzzles me is that you have the target definitions label in the framework search path. That may happen because C has dependencies, which are scoped. But if C really depends on plain A and B and not different subspecs to the shared subspec in your abstract target, then the analyzer should only generate a scoped target for `A/subspec`.

Can you also show me the resolved build setting `FRAMEWORK_SEARCH_PATHS` for your target C? I thought it would contain the default build dir via `$(inherited)`.

benasher44:
@mrackwitz since this is getting increasingly difficult to discuss while trying to discuss pods so abstractly (can't giveaway our whole podfile unfortunately), I've put together a an example that reproduces the issue and I've [posted it here](https://github.com/benasher44/CocoaPods4146).

The project fails to build with the podfile as is in the project:
```

platform :ios, '8.0'
use_frameworks!

abstract_target 'SharedDeps' do
  pod 'AFNetworking/NSURLConnection'

  target 'PodsNonTrivalDepsTest' do
    pod 'AFNetworking'
    pod 'AFOAuth2Manager'
  end

end
```

If you remove the first pod in abstract target to make it look like this:
```

platform :ios, '8.0'
use_frameworks!

abstract_target 'SharedDeps' do
  target 'PodsNonTrivalDepsTest' do
    pod 'AFNetworking'
    pod 'AFOAuth2Manager'
  end

end
```
then the project will build

benasher44:
@mrackwitz so this most recent update fixed that issue, but now I'm having an issue related to module maps. I've updated [the test project](https://github.com/benasher44/CocoaPods4146) to reproduced the issue.

benasher44:
@mrackwitz I see you've made some updates. I'll try to have some feedback for you tomorrow. I'm traveling today, and I'm trying to avoid doing work that uses xcode to savor what battery life I have throughout my travels today :)

mrackwitz:
@benasher44: No worries, I cloned your project meanwhile and was already running the latest changes on it, but I think there is still an issue left to resolve. I'll let you know, when I could fix it until then you can safe some battery cycles. :wink:

benasher44:
@mrackwitz ah okay cool. Thanks! I'll wait for more updates then :)

mrackwitz:
The names for the targets generated for @benasher44's updated Podfile are not perfectly intuitive.

```ruby
abstract_target 'SharedDeps' do
  pod 'AFNetworking/NSURLConnection'

  target 'App' do
    pod 'AFNetworking'
  end

  target 'Extension'
end
```

Because AFNetworking has quite a bunch of default subspecs, this leads to the following names:

* `AFNetworking` for the extension
* `AFNetworking-NSURLSession-UIKit-root` for the app

That's because all subspecs which are common among all variants are ignored and only those subspecs are added, which differentiate the pod from all other variants, where the root spec itself is marked as `root` for distinction.

More intuitive would be to take the default subspecs into account:

* `AFNetworking-NSURLConnection` for the extension
* `AFNetworking` for the app

It is not sufficient though to ignore the names of all default subspecs.

Given e.g. the following change to the above Podfile:

```diff
abstract_target 'SharedDeps' do
  pod 'AFNetworking/NSURLConnection'

  target 'App' do
    pod 'AFNetworking'
  end

  target 'Extension'
end

+target 'Test' do
+  pod 'AFNetworking'
+end
```

This would led to the situation that there is a variant which requires all default subspecs and the optional subspec `NSURLConnection` while another variant requires just the subspec `NSURLConnection`. Their target labels would clash.

One idea could be to differentiate those variants by an additional prefix:

* `AFNetworking` for the default variant
* `AFNetworking-default-NSURLConnection` for an extended default variant
* `AFNetworking-NSURLConnection` for an exclusive variant

One problem here is that `default` could be also the name of a subspec. (Yes, that's [in use](https://github.com/CocoaPods/Specs/blob/master/Specs/GDRealtime/0.3.0/GDRealtime.podspec.json#L28) in [the wild](https://github.com/CocoaPods/Specs/blob/master/Specs/NetworkEye/0.9.0/NetworkEye.podspec.json#L29).) Other ideas for a better naming scheme, which is still conflict-free, would be welcome.

Why all this effort for just a this tiny implementation detail? Because the current model causes huge changes to the Pods project whenever a pod e.g. adds or remove a new subspec or when I as a user add another subspec to a target. Furthermore it gets harder to reason about integrations for us and pod maintainers.

/c @orta

neonichu:
@mrackwitz I think the best way to ensure the default is conflict free is to break one of the rules for spec names. My vote would be for `.default` as this is not a valid spec name (see <https://github.com/CocoaPods/Core/blob/master/lib/cocoapods-core/specification/linter.rb#L199>), but still valid in the name of an Xcode target, AFAIK. It impairs readability slightly, but seems like a good compromise?

benasher44:
+1 Xcode build logs were never known for their readability anyway.

orta:
> My vote would be for .default as this is not a valid spec name

Seems reasonable to me

benasher44:
@mrackwitz your most recent update seems to have fixed one framework search paths issue, but I've found another one and have updated [the sample project](https://github.com/benasher44/CocoaPods4146). Before my sample project update, AFOAuth2Manager was the pod in this project that used one of the pods depended on by both the main target and the extension target. Given that AFOAuth2Manager only depended on AFNetworking/NSURLConnection, your most recent changes fixed the build.

I've now updated the project for a different case, and I've dropped AFOAuth2Manager in favor of AFNetworking-RACExtensions. This pod depends on all of AFNetworking, and not just AFNetworking/NSURLConnection, and the project now fails building AFHTTPSessionManager+RACSupport.m because it can't find "AFHTTPSessionManager.h" (I've tried manually replacing the import with the framework-style <AFNetworking/AFHTTPSessionManager.h>, and this didn't fix it). If you look at the framework search path flags when trying to compile this file, they look like this:

```
-F/Users/basher/Library/Developer/Xcode/DerivedData/PodsNonTrivalDepsTest-fincreuzyxmnwdgnbjhsjaxpthad/Build/Products/Debug-iphonesimulator/AFOAuth2Manager
-F/Users/basher/Library/Developer/Xcode/DerivedData/PodsNonTrivalDepsTest-fincreuzyxmnwdgnbjhsjaxpthad/Build/Products/Debug-iphonesimulator/AFNetworking-NSURLConnection-Reachability-Security-Serialization
-F/Users/basher/Library/Developer/Xcode/DerivedData/PodsNonTrivalDepsTest-fincreuzyxmnwdgnbjhsjaxpthad/Build/Products/Debug-iphonesimulator/AFNetworking.default-NSURLConnection-NSURLSession-Reachability-Security-Serialization-UIKit
```
This doesn't seem correct, since it doesn't need both AFNetworking search paths to compile this file. It should probably look like this:

```
-F/Users/basher/Library/Developer/Xcode/DerivedData/PodsNonTrivalDepsTest-fincreuzyxmnwdgnbjhsjaxpthad/Build/Products/Debug-iphonesimulator/AFOAuth2Manager
-F/Users/basher/Library/Developer/Xcode/DerivedData/PodsNonTrivalDepsTest-fincreuzyxmnwdgnbjhsjaxpthad/Build/Products/Debug-iphonesimulator/AFNetworking.default-NSURLConnection-NSURLSession-Reachability-Security-Serialization-UIKit
```

mrackwitz:
Wow, great! That was fast. I'll have another look at it and might include your sample in our integration tests.

orta:
I've got [a branch of Artsy/Eigen](https://github.com/artsy/eigen/tree/orta-de-dupe) with de-dupes, I went from 0.39 -> this. First up, running `pod install` feels slow. Secondly, this seems to break an implicit dependency we have within two of our pods. This pod doesn't have a dependency on `Artsy+UIColors` but would be able to find it if it needed colours.

screen shot 2016-02-08 at 21 12 09

segiddins:
@orta breaking that is fine -- you need to specify something as a dependency if you want to guarantee its search paths are setup correctly

jvanmetre:
I also am having the problem where 1.0.3 beta will have framework linking issues, that this PR is supposed to help with. My Podfile is a multi OS with use_frameworks! similar to:
```
def shared_pods
  pod 'AFNetworking', '~> 3'
  pod 'LocalPod', :path => './LocalPod/'
end
target 'iOSTarget' do
  platform :ios, '8.0'
  shared_pods
end
target 'iOSTodayWidget' do
  platform :ios, '8.0'
  shared_pods
end
target 'AppleTVTarget' do
  platform :tvos, '9.0'
  shared_pods
end
```
I have tried this branch, and it is working great except that the generated Pods-AppleTVTarget-frameworks.sh file still has the old framework names under the install_framework portion.
It looks like:
```
  install_framework "Pods-AppleTVTarget/AFNetworking.framework"
fi
```
when it should look like:
```
  install_framework "AFNetworking-tvOS/AFNetworking.framework"
fi
```

mrackwitz:
@benasher44: I found the root cause for the issue, you've report and I'm working on a fix.

@jvanmetre: Excellent! Thanks for your report, didn't caught that yet.

benasher44:
@mrackwitz awesome! 🎉🎉🎉

jvanmetre:
This appears to fix the issue I found
```
diff --git a/lib/cocoapods/installer/target_installer/aggregate_target_installer.rb b/lib/cocoapods/installer/target_installer/aggregate_target_installer.rb
index 1b7cbad..5d8bd4c 100644
--- a/lib/cocoapods/installer/target_installer/aggregate_target_installer.rb
+++ b/lib/cocoapods/installer/target_installer/aggregate_target_installer.rb
@@ -151,7 +151,7 @@ module Pod
           end
           frameworks_by_config[config] = relevant_pod_targets.flat_map do |pod_target|
             frameworks = pod_target.file_accessors.flat_map(&:vendored_dynamic_artifacts).map { |fw| "${PODS_ROOT}/#{fw.relative_path_from(sandbox.root)}" }
-            frameworks << "#{target_definition.label}/#{pod_target.product_name}" if pod_target.should_build? && pod_target.requires_frameworks?
+            frameworks << "#{pod_target.configuration_build_dir}/#{pod_target.product_name}" if pod_target.should_build? && pod_target.requires_frameworks?
             frameworks
           end
         end
```

segiddins:
@mrackwitz I see this is passing, is it ready for review?

benasher44:
I'll give this a try tomorrow, and then I'll update

mrackwitz:
Yeah, I would appreciate your feedback.

benasher44:
@mrackwitz overall, this is a huge huge improvement. The framework search paths for each framework seems correct. I did manage to find one issue though related to frameworks containing hyphens, and I've updated [the sample project](https://github.com/benasher44/CocoaPods4146). In the sample project, you could previously import Mailcheck-ObjectiveC with the following: `#import <Mailcheck-ObjectiveC/Mailcheck.h>`. With your branch, this is a build error, and you have to do this instead: `#import <Mailcheck_ObjectiveC/Mailcheck.h>`. It appears that frameworks with hyphens in their names are having their hyphens converted to underscored. This would be okay, if the change was just within the sample project. Unfortunately, Mailcheck-ObjectiveC depends on NSString-Email, and so the import for that framework within the Mailcheck-ObjectiveC pod would also have to be changed. Can we go back to allowing hyphens?

lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
```diff
@@ -94,15 +95,11 @@ def generate
...
           if target.requires_frameworks?
-            framework_header_search_paths = pod_targets.select(&:should_build?).map do |target|
-              if target.scoped?
-                "$PODS_FRAMEWORK_BUILD_PATH/#{target.product_name}/Headers"
-              else
-                "$CONFIGURATION_BUILD_DIR/#{target.product_name}/Headers"
-              end
+            build_pod_targets = pod_targets.select(&:should_build?)
+            framework_header_search_paths = build_pod_targets.map do |target|
+              "#{target.configuration_build_dir}/#{target.product_name}/Headers"
```
segiddins: can this be moved to be a method on `PodTarget` ?

lib/cocoapods/generator/xcconfig/xcconfig_helper.rb
```diff
@@ -196,6 +203,47 @@ def self.add_target_specific_settings(target, xcconfig)
...
+            framework_search_paths = []
+            library_search_paths = []
+            dependent_targets.each do |dependent_target|
+              if dependent_target.requires_frameworks?
+                framework_search_paths << dependent_target.configuration_build_dir(build_dir_var)
+              else
+                library_search_paths << dependent_target.configuration_build_dir(build_dir_var)
+              end
+            end
+            build_settings['FRAMEWORK_SEARCH_PATHS'] = '$(inherited) ' + XCConfigHelper.quote(framework_search_paths.uniq) unless framework_search_paths.empty?
```
segiddins: i don't think the `unless empty?`s are necessary

lib/cocoapods/installer/analyzer/pod_variant.rb
```diff
@@ -0,0 +1,59 @@
...
+        end
+        alias_method :eql?, :==
+
+        # Hashes the instance by all its attributes.
+        #
+        # This adds support to make instances usable as Hash keys.
+        #
+        # @!visibility private
+        def hash
+          [specs, platform, requires_frameworks].hash
```
segiddins: would be more performant to multiply each hash by a prime number and `xor` them

lib/cocoapods/installer/target_installer/aggregate_target_installer.rb
```diff
@@ -151,7 +151,7 @@ def create_embed_frameworks_script
           end
           frameworks_by_config[config] = relevant_pod_targets.flat_map do |pod_target|
             frameworks = pod_target.file_accessors.flat_map(&:vendored_dynamic_artifacts).map { |fw| "${PODS_ROOT}/#{fw.relative_path_from(sandbox.root)}" }
-            frameworks << "#{target_definition.label}/#{pod_target.product_name}" if pod_target.should_build? && pod_target.requires_frameworks?
+            frameworks << "#{pod_target.label}/#{pod_target.product_name}" if pod_target.should_build? && pod_target.requires_frameworks?
```
segiddins: attr on `PodTarget`  ?

lib/cocoapods/target/pod_target.rb
```diff
@@ -82,7 +90,11 @@ def scoped(cache = {})
     #
     def label
       if scoped?
-        "#{target_definitions.first.label}-#{root_spec.name}"
+        if scope_suffix[0] == '.'
```
segiddins: `.first`

lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
mrackwitz: Why? It's just used at this one place. I can't recognize any advantage of having a quick accessor for it.

lib/cocoapods/generator/xcconfig/xcconfig_helper.rb
mrackwitz: Just an attempt into minimizing the footprint of the xcconfig. I'm happy to waive that.

lib/cocoapods/installer/analyzer/pod_variant.rb
mrackwitz: Ok, fair. As the performance might have dropped already, improvements like that would be welcome.

lib/cocoapods/installer/target_installer/aggregate_target_installer.rb
mrackwitz: Okay, I see that's actually the same. :+1:

lib/cocoapods/target/pod_target.rb
mrackwitz: :cop::-1: That's a `String`. :wink:

lib/cocoapods/installer/analyzer/pod_variant.rb
mrackwitz: So I was curious and try it out. A quick benchmark shows that a solution like you suggests is in average 20% slower than what I came up with originally. It seems like being able to use the static C implementation of the stdlib outweighs the additional array allocation.

lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
segiddins: because then the logic won't be duplicated if it's used elsewhere

lib/cocoapods/target/pod_target.rb
segiddins: ah good point :P

lib/cocoapods/generator/xcconfig/xcconfig_helper.rb
```diff
@@ -196,6 +203,47 @@ def self.add_target_specific_settings(target, xcconfig)
...
+        # @param  [Target] target
+        #         The target.
+        #
+        # @param  [Array<PodTarget>] dependent_targets
+        #         The pod targets the given target depends on.
+        #
+        # @param  [Xcodeproj::Config] xcconfig
+        #         The xcconfig to edit.
+        #
+        def self.add_settings_for_dependent_targets(target, dependent_targets, xcconfig)
```
segiddins: document returning void

spec/unit/target/pod_target_spec.rb
```diff
@@ -162,8 +169,13 @@ module Pod
       end
 
       it 'returns the path for the CONFIGURATION_BUILD_DIR build setting' do
-        @pod_target.configuration_build_dir.should == '$(BUILD_DIR)/$(CONFIGURATION)$(EFFECTIVE_PLATFORM_NAME)'
-        @pod_target.scoped.first.configuration_build_dir.should == '$(BUILD_DIR)/$(CONFIGURATION)$(EFFECTIVE_PLATFORM_NAME)/Pods'
+        @pod_target.configuration_build_dir.should == '$CONFIGURATION_BUILD_DIR/BananaLib'
+        @pod_target.scoped.first.configuration_build_dir.should == '$CONFIGURATION_BUILD_DIR/BananaLib-Pods'
```
segiddins: please also test the version of the method that accepts a parameter

lib/cocoapods/generator/xcconfig/xcconfig_helper.rb
```diff
@@ -196,6 +203,47 @@ def self.add_target_specific_settings(target, xcconfig)
...
+          dependent_targets = dependent_targets.select(&:should_build?)
+          has_configuration_build_dir = target.respond_to?(:configuration_build_dir)
+          if has_configuration_build_dir
+            build_dir_var = "$#{SHARED_BUILD_DIR_VARIABLE}"
+            build_settings = {
+              'CONFIGURATION_BUILD_DIR' => target.configuration_build_dir(build_dir_var),
+              SHARED_BUILD_DIR_VARIABLE => '$(BUILD_DIR)/$(CONFIGURATION)$(EFFECTIVE_PLATFORM_NAME)',
```
segiddins: does this need to be quoted?

segiddins:
This mostly looks amazing, @mrackwitz ! Great work, thank you so much for doing this. I'll give you a few days to address comments, and hopefully either @neonichu or @AliSoftware will have a few minutes to look it over as well. :rocket:

lib/cocoapods/generator/xcconfig/xcconfig_helper.rb
mrackwitz: No, just builtin variables allowing multiple values need to be quoted. If that would be quoted, the quotations would be taken over, where this variable is used and Xcode would interpret it as relative path beginning with a quotation mark.

spec/unit/target/pod_target_spec.rb
mrackwitz: :white_check_mark: Done!

mrackwitz:
@benasher44:
>In the sample project, you could previously import Mailcheck-ObjectiveC with the following: `#import <Mailcheck-ObjectiveC/Mailcheck.h>`. With your branch, this is a build error, and you have to do this instead: `#import <Mailcheck_ObjectiveC/Mailcheck.h>`. It appears that frameworks with hyphens in their names are having their hyphens converted to underscores.

Are you sure that this was really the case before with frameworks? Because a framework name must equal the module name, which is limited to C99ext identifiers. These don't allow hyphens. So this would only work with libraries.

benasher44:
@mrackwitz ah woops! You're right. Okay then, well I'll just test your most recent changes and update again. Thanks!!

benasher44:
@mrackwitz looks good from my end! Nice work!!

lib/cocoapods/installer/analyzer/pod_variant.rb
segiddins: :+1:

lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
```diff
@@ -112,9 +109,6 @@ def settings_to_import_pod_targets
               build_settings['HEADER_SEARCH_PATHS'] = '$(inherited) ' + XCConfigHelper.quote(library_header_search_paths)
               build_settings['OTHER_CFLAGS'] += ' ' + XCConfigHelper.quote(library_header_search_paths, '-isystem')
             end
-            if pod_targets.any? { |t| t.should_build? && t.scoped? }
-              build_settings['FRAMEWORK_SEARCH_PATHS'] = '$PODS_FRAMEWORK_BUILD_PATH'
```
mrackwitz: Removing this here would likely break search path inheritance.
@segiddins: Can you confirm that?

segiddins: yes, it breaks search path inheritance -- we'll need to fix it so the proper `FRAMEWORK_SEARCH_PATH` is set, as well as not setting the `OTHER_CFLAGS` to the `CONFIGURATION_BUILD_DIR` for inherited pods

CHANGELOG.md
```diff
@@ -8,12 +8,26 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
...
+  suffixed now on base of what makes them different among others for the
+  same root spec instead of being prefixed by the dependant target.  
+  [Marius Rackwitz](https://github.com/mrackwitz)
+  [#4146](https://github.com/CocoaPods/CocoaPods/pull/4146)
 
 ##### Bug Fixes
 
-* None.  
+* Pods are built by default in another scoping level of the build products
+  directory identified by their name to exclude name clashes among
```
neonichu: exclude => prevent?

spec/fixtures/matryoshka/matryoshka.podspec
```diff
@@ -0,0 +1,30 @@
+Pod::Spec.new do |s|
+  s.name             = "matryoshka"
+  s.version          = "1.0.0"
+  s.author           = { "Matryona Malyutin" => "matryona@malyutin.local" }
```
neonichu: :clap:

neonichu:
Great work, @mrackwitz :clap: Looks good from my side, except for the small Changelog fix :)

benasher44:
Commit 4333212 still working for me.

segiddins:
@mrackwitz if this is no longer a WIP, it gets a :+1: from me. Great job! :rocket:

mrackwitz:
Okay, let's :ship: it!

orta:
oh, wow

MRW I overhear somebody in my office make a really clever pun

PR #4147, opened 2015-09-09
[Analyzer] Build scoped pods once for each target definition

### LINKED ISSUE #4034: Target Deduplication removes necessary Pods project targets 🌈

mrackwitz:
Fixes #4034. Alternative approach to #4146.

#### ToDo:

* [x] Changelog


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
+* Either generate just one pod target or generate it once for each target definition.
```
segiddins: wrong section

segiddins:
:+1: if we can verify that this indeed fixes things

CHANGELOG.md
mrackwitz: but just wrong version meanwhile?

segiddins: ah yeah

segiddins:
I've now rebased this, it's good to merge once CI passes (I manually verified that it fixed the linked issue).

segiddins:
:fireworks: thanks marius!

mrackwitz:
Thanks for taking over the last mile, Sam!

Issue #4197, opened 2015-09-17
`pod install` prints wrong error message


manuyavuz:
Hi,

During adding a private pod into a project, if a specific version is not specified in the Podfile and there are only pre-release versions available for the pod, `pod install` command fails with the following output:

```
[!] Unable to satisfy the following requirements:

- `PrivatePodName` required by `Podfile`

Specs satisfying the `PrivatePodName` dependency were found, but they required a higher minimum deployment target.
```

If I explicitly specify pre-release version in the Podfile, `pod install` succeeds without any problem.

I dug into the issue a little, and discovered that the conflict that resulted from this case was not handled correctly in `handle_resolver_error` method of `resolver.rb` file. There should be another check that detects the case and print a meaningful output to user.

I've actually implemented this and will provide a PR.

segiddins:
Closed by https://github.com/CocoaPods/CocoaPods/issues/4198.

PR #4198, opened 2015-09-17
[Resolver] Give a meaningful message for the case where there is no stable version for a pod, and user do not explicitly specify the required version.

manuyavuz:
As stated in #4197, `pod install` and `pod update` commands print wrong error information for the following case:

- There are only pre-release versions available for a pod specified in the Podfile (no stable versions).
- User does not explicitly set a version for the pod. She just uses `pod 'PodName'`

This PR fixes the case by  providing a meaningful error message for the occurring dependency conflict. 

While handling resolver error in `handle_resolver_error` method of `resolver.rb` file, the code will check if the specification that is not satisfying given requirement was a pre-release version, and if the requirement was including any explicit version or source. If spec is a pre-release version, and there is no explicit version or source for the requirement, the code will print a message directing the user to specify pre-release version explicitly like this:

```
[!] There is no corresponding stable version for `PodName`. 
You should explicitly specify the version in order to install a pre-release version of `PodName`
```

However, I'm not fully sure that the proposed check handles all scenarios. A double check from anyone would be appreciated.

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,13 +8,12 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
-* Use watchsimulator when validating pods with the watchOS platform.  
 * Give a meaningful message for the case where there is no available stable version for a pod,
   and user do not explicitly specify the required version.
   [Muhammed Yavuz Nuzumlalı](https://github.com/manuyavuz)
   [#4197](https://github.com/CocoaPods/CocoaPods/issues/4197)
 
-* Use watchsimulator when validatng PodSpecs with watchOS compatibility.  
+* Use watchsimulator when validating pods with the watchOS platform.
   [Thomas Kollbach](https://github.com/toto)
   [#4130](https://github.com/CocoaPods/CocoaPods/issues/4130)
   
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 ##### Bug Fixes
 
 * Use watchsimulator when validating pods with the watchOS platform.  
+* Give a meaningful message for the case where there is no available stable version for a pod,
+  and user do not explicitly specify the required version.
+  [Muhammed Yavuz Nuzumlalı](https://github.com/manuyavuz)
+  [#4197](https://github.com/CocoaPods/CocoaPods/issues/4197)
+
+* Use watchsimulator when validatng PodSpecs with watchOS compatibility.  
   [Thomas Kollbach](https://github.com/toto)
   [#4130](https://github.com/CocoaPods/CocoaPods/issues/4130)
   
--- a/lib/cocoapods/resolver.rb
+++ b/lib/cocoapods/resolver.rb
@@ -383,7 +383,12 @@ module Pod
               'version requirement to your Podfile ' \
               "(e.g. `pod '#{name}', '#{lockfile_reqs.map(&:requirement).join("', '")}'`) " \
               "or revert to a stable version by running `pod update #{name}`."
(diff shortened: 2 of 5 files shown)


segiddins:
Looks good! Just needs specs and a CHANGELOG entry

orta:
Yeah, nice message @manuyavuz  :+1:

manuyavuz:
Thanks for the feedback! I'll update CHANGELOG now. 

I have a question, travis failed with the following offense:

lib/cocoapods/resolver.rb:372:5: C: Perceived complexity for handle_resolver_error is too high. [15/10]
    def handle_resolver_error(error)
    ^^^

Do you have any suggestions on splitting this method in order to reduce complexity?

segiddins:
Nah, we should be ignoring that in the rubocop config anyways, we prefer to judge complexity manually

manuyavuz:
At last! My fingers were resisting to use 2 space indentations :)

I've also added to specs for `resolver.rb`. Only thing fails is 'perceived complexity' thing.

I think it can be merged now?

segiddins:
I'll take a look when I'm next at a computer, but thanks so much @manuyavuz !

manuyavuz:
Hi again, 

My previous implementation was not complete. It was not giving a good message for the following sample case:

`pod 'AFNetworking', '< 1.0', '> 0.10.1'`

For the above scenario, there are three possibilities (1.0RC3, 1.0RC2, 1.0RC1) all of which are pre-release versions. Previous message would be misleading. Here is the new message for the above case:

```
[!] There are only pre-release versions available satisfying the following requirements:

	'AFNetworking', '< 1.0, > 0.10.1'

You should explicitly specify the version in order to install a pre-release version
```

orta:
Clever :+1:

segiddins:
Brilliant! I'll merge this if you can rebase it! :+1: 
Thanks @manuyavuz!

manuyavuz:
Rebase completed! 👋🏽

CHANGELOG.md
```diff
@@ -9,6 +9,12 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 ##### Bug Fixes
 
 * Use watchsimulator when validating pods with the watchOS platform.
```
segiddins: this looks extraneous, perhaps a rebase error?

manuyavuz: Yes, seems like auto-merge have failed. Will fix now.

segiddins:
:ship: thanks so much!

CHANGELOG.md
```diff
@@ -8,7 +8,12 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
-* Use watchsimulator when validating pods with the watchOS platform.  
+* Give a meaningful message for the case where there is no available stable version for a pod,
+  and user do not explicitly specify the required version.
```
mrackwitz: Add two-spaces at the end of the line for a line break.

CHANGELOG.md
```diff
@@ -8,7 +8,12 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
-* Use watchsimulator when validating pods with the watchOS platform.
```
mrackwitz: You / presumably your editor accidentally removed the two spaces here.

segiddins:
@mrackwitz I just clean that up when doing the release

manuyavuz:
@mrackwitz I removed these because git diff --check was giving 'trailing spaces' warning. I did not think that two spaces needed for the line break. Is it a need for GitHub Markdown format, or required by default markdown format?

mrackwitz:
@manuyavuz: It's a general markdown thing, not specific to GitHub.
Nice work, btw. :+1:

manuyavuz:
I'll keep that in mind, thanks for the remind! :+1:

PR #4215, opened 2015-09-19
add CoreData .datamodeld into source files of pod target, so that we may simply add data model as source file

pinxue:
Currently, we may add .datamodeld as source file, it will be downloaded during pod installation. But it will not be added to Pod's compiling list. This simple change just add it back.

BTW, I feel there may be a better way to integrate the logic, I just spend about one half day to learn ruby, read the source code and test the patch, far away from understanding file_accessor totally. Let me know if the feeling is right.

### DIFF
--- a/lib/cocoapods/installer/file_references_installer.rb
+++ b/lib/cocoapods/installer/file_references_installer.rb
@@ -187,7 +187,7 @@ module Pod
           end
         end
       end
-  
+
       def add_datamodels_to_pods_group
         file_accessors.each do |file_accessor|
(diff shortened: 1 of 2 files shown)


segiddins:
I'd rather we figure out why the current compilation as a resource isn't working for you than going down this route, if at all possible

pinxue:
You are right. It is quite weird, I have tried several times with resource method, but the data model just didn't appear in project after building. But I just cleanup everything and tried again, it just works.

PR #4249, opened 2015-09-23
Changes search index format to fasten `pod search --full` command.

manuyavuz:
Detailed discussion can be found here: CocoaPods/cocoapods-search#8 
- Previous implementation was having the drawback of traversing and performing `gsub` on all index strings for each pod specification.
- New implementation stores words as keys and list of pods containing corresponding word inside their specification as the values for corresponding hash keys. Therefore, while searching for a query, we only need to check if query is matched with any key in index hash, if so, we will add corresponding list of spec names to a Set object. Resulted Set object gives us the search result. Using this policy, sources manager can perform a faster search operation.

### DIFF
--- a/lib/cocoapods/sources_manager.rb
+++ b/lib/cocoapods/sources_manager.rb
@@ -155,18 +155,16 @@ module Pod
       # @return [Hash{String => Hash{String => Array<String>}}] The up to date search data.
       #
       def updated_search_index
-        index = stored_search_index
-        unless index
-          # Create index from scratch
-          UI.puts 'Creating search index..'
-          index = {}
-          all.each do |source|
-            source_name = source.name
+        index = stored_search_index || {}
+        all.each do |source|
+          source_name = source.name
+          unless index[source_name]
+            UI.print "Creating search index for spec repo '#{source_name}'.."
             index[source_name] = aggregate.generate_search_index_for_source(source)
+            UI.puts " Done!"
           end
-          save_search_index(index)
-          UI.puts 'Search index creation done!'
         end
+        save_search_index(index)
         index
       end
 
@@ -219,28 +217,28 @@ module Pod
       # @param  [Hash{Source => Array<String>}] changed_spec_paths
       #                  A hash containing changed specification paths for each source.
       def update_search_index_if_needed(changed_spec_paths)
-        search_index = nil
+        search_index = stored_search_index
+        return unless search_index
         changed_spec_paths.each_pair do |source, spec_paths|
-          source_name = source.name
-          next unless spec_paths.length > 0
-          search_index = stored_search_index
+          index_for_source = search_index[source.name]
+          next unless index_for_source && spec_paths.length > 0
           updated_pods = source.pods_for_specification_paths(spec_paths)
+
           new_index = aggregate.generate_search_index_for_changes_in_source(source, spec_paths)
-          next unless search_index && search_index[source_name]
           # First traverse search_index and update existing words
           # Removed traversed words from new_index after adding to search_index,
           # so that only non existing words will remain in new_index after enumeration completes.
-          search_index[source_name].each_pair do |word, _|
+          index_for_source.each_pair do |word, _|
             if new_index[word]
-              search_index[source_name][word] |= new_index[word]
+              index_for_source[word] |= new_index[word]
             else
-              search_index[source_name][word] -= updated_pods
+              index_for_source[word] -= updated_pods
             end
           end
           # Now add non existing words remained in new_index to search_index
-          search_index[source_name].merge!(new_index)
+          index_for_source.merge!(new_index)
         end
-        save_search_index(search_index) if search_index
+        save_search_index(search_index)
       end
 
       # Updates search index for changed pods in background
@@ -271,7 +269,8 @@ module Pod
         changed_spec_paths = {}
         sources.each do |source|
           UI.section "Updating spec repo `#{source.name}`" do
-            changed_spec_paths[source] = source.update(show_output && !config.verbose?)
+            changed_source_paths = source.update(show_output && !config.verbose?)
+            changed_spec_paths[source] = changed_source_paths if changed_source_paths.count > 0
             check_version_information(source.repo)
           end
         end
--- a/lib/cocoapods/sources_manager.rb
+++ b/lib/cocoapods/sources_manager.rb
@@ -110,25 +110,18 @@ module Pod
       #
       # @raise  If no source including the set can be found.
       #
-      # @note   Full text search requires to load the specification for each
-      #         pod, hence is considerably slower.
-      #
       # @return [Array<Set>]  The sets that contain the search term.
       #
       def search_by_name(query, full_text_search = false)
         if full_text_search
-          set_names = []
+          set_names = Set.new
           query_regexp = /#{query}/i
-          updated_search_index.each do |name, set_data|
-            texts = [name]
-            if full_text_search
-              texts << set_data['authors'].to_s if set_data['authors']
-              texts << set_data['summary']      if set_data['summary']
-              texts << set_data['description']  if set_data['description']
+          updated_search_index.each_value do |word_spec_hash|
+            word_spec_hash.each_pair do |word, spec_names|
+              set_names.merge(spec_names) unless word !~ query_regexp
             end
-            set_names << name unless texts.grep(query_regexp).empty?
           end
-          sets = set_names.sort.map do |name|
+          sets = set_names.map do |name|
             aggregate.representative_set(name)
           end
         else
@@ -142,39 +135,62 @@ module Pod
         sets
       end
 
-      # Creates or updates the search data and returns it. The search data
-      # groups by name the following information for each set:
+      # Returns the search data. If a saved search data exists, retrieves it from file and returns it.
+      # Else, creates the search data from scratch, saves it to file system, and returns it.
+      # Search data is grouped by source repos. For each source, it contains a hash where keys are words
+      # and values are the pod names containing corresponding word.
       #
+      # For each source, list of unique words are generated from the following spec information.
       #   - version
       #   - summary
       #   - description
       #   - authors
       #
-      # @note   This operation is fairly expensive, because of the YAML
-      #         conversion.
-      #
-      # @return [Hash{String => String}] The up to date search data.
+      # @return [Hash{String => Hash{String => Array<String>}}] The up to date search data.
       #
       def updated_search_index
+        index = stored_search_index
+        unless index
+          # Create index from scratch
+          UI.puts 'Creating search index..'
+          index = {}
+          all.each do |source|
+            source_name = source.name
+            index[source_name] = aggregate.generate_search_index_for_source(source)
+          end
+          save_search_index(index)
+          UI.puts 'Search index creation done!'
+        end
+        index
+      end
+
+      # Returns the search data stored in the file system.
+      # If existing data in the file system is not valid, returns nil.
+      #
+      def stored_search_index
         unless @updated_search_index
           if search_index_path.exist?
-            require 'yaml'
-            stored_index = YAML.load(search_index_path.read)
-            if stored_index && stored_index.is_a?(Hash)
-              search_index = aggregate.update_search_index(stored_index)
-            else
-              search_index = aggregate.generate_search_index
+            require 'json'
+            index = JSON.parse(search_index_path.read)
+            if index && index.is_a?(Hash) # TODO: should we also check if hash has correct hierarchy?
+              return @updated_search_index = index
             end
-          else
-            search_index = aggregate.generate_search_index
           end
-
-          File.open(search_index_path, 'w') do |file|
-            file.write(search_index.to_yaml)
-          end
-          @updated_search_index = search_index
+          @updated_search_index = nil
         end
         @updated_search_index
+      end
+
+      # Stores given search data in the file system.
+      # @param [Hash] index
+      #        Index to be saved in file system
+      #
+      def save_search_index(index)
+        require 'json'
+        @updated_search_index = index
+        File.open(search_index_path, 'w') do |file|
+          file.write(@updated_search_index.to_json)
+        end
       end
 
       # Allows to clear the search index.
@@ -192,6 +208,35 @@ module Pod
       extend Executable
       executable :git
 
+      # Updates the stored search index if there are changes in spec repos while updating them.
+      # Update is performed incrementally. Only the changed pods' search data is re-generated and updated.
+      # @param  [Hash{Source => Array<String>}] changed_spec_paths
+      #                  A hash containing changed specification paths for each source.
+      def update_search_index_if_needed(changed_spec_paths)
+        search_index = nil
+        changed_spec_paths.each_pair do |source, spec_paths|
+          source_name = source.name
+          next unless spec_paths.length > 0
+          search_index = stored_search_index
+          updated_pods = source.pods_for_specification_paths(spec_paths)
+          new_index = aggregate.generate_search_index_for_changes_in_source(source, spec_paths)
+          next unless search_index && search_index[source_name]
+          # First traverse search_index and update existing words
+          # Removed traversed words from new_index after adding to search_index,
+          # so that only non existing words will remain in new_index after enumeration completes.
+          search_index[source_name].each_pair do |word, _|
+            if new_index[word]
+              search_index[source_name][word] |= new_index[word]
+            else
+              search_index[source_name][word] -= updated_pods
+            end
+          end
+          # Now add non existing words remained in new_index to search_index
+          search_index[source_name].merge!(new_index)
+        end
+        save_search_index(search_index) if search_index
+      end
+
       # Updates the local clone of the spec-repo with the given name or of all
       # the git repos if the name is omitted.
       #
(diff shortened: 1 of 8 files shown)


floere:
This is great! :rocket: 

Were you able to run the specs? There are two failures, reproduced here:

```
Pod::Informative: [!] Unable to find a pod with name, author, summary, or descriptionmatching `Chunky`
	/Users/travis/build/CocoaPods/CocoaPods/lib/cocoapods/sources_manager.rb:130:in `search_by_name': In general - can perform a full text search of the sets
	spec/unit/sources_manager_spec.rb:75:in `block (3 levels) in <module:Pod>'
	/Users/travis/build/CocoaPods/CocoaPods/spec/spec_helper/pre_flight.rb:33:in `call'
	/Users/travis/build/CocoaPods/CocoaPods/spec/spec_helper/pre_flight.rb:33:in `block in <class:Context>'
	spec/unit/sources_manager_spec.rb:73:in `block (2 levels) in <module:Pod>'
	spec/unit/sources_manager_spec.rb:37:in `block in <module:Pod>'
	spec/unit/sources_manager_spec.rb:30:in `<module:Pod>'
	spec/unit/sources_manager_spec.rb:29:in `<top (required)>'
Pod::Informative: [!] Unable to find a pod with name, author, summary, or descriptionmatching `Ch[aeiou]nky`
	/Users/travis/build/CocoaPods/CocoaPods/lib/cocoapods/sources_manager.rb:130:in `search_by_name': In general - can perform a full text regexp search of the sets
	spec/unit/sources_manager_spec.rb:82:in `block (3 levels) in <module:Pod>'
	/Users/travis/build/CocoaPods/CocoaPods/spec/spec_helper/pre_flight.rb:33:in `call'
	/Users/travis/build/CocoaPods/CocoaPods/spec/spec_helper/pre_flight.rb:33:in `block in <class:Context>'
	spec/unit/sources_manager_spec.rb:80:in `block (2 levels) in <module:Pod>'
	spec/unit/sources_manager_spec.rb:37:in `block in <module:Pod>'
	spec/unit/sources_manager_spec.rb:30:in `<module:Pod>'
	spec/unit/sources_manager_spec.rb:29:in `<top (required)>'
```

manuyavuz:
Yeah, I observed them also, but couldn't have time to fix it. Hoping to fix these tomorrow.

floere:
That's great! :rocket:

manuyavuz:
Hey @floere,

I checked again and discovered that failures were caused by Core repo here not containing my changes in CocoaPods/Core#265.

How should I manage this? This PR is tightly coupled to changes in Core repo, so I guess it will not pass unless the PR in Core repo is merged.

Need suggestions.

floere:
Are the Core changes standalone? (Thery work independently?)

manuyavuz:
Core include changes on `aggregate.rb` and can run standalone. However, `sources_manager.rb` here is dependent on the changes in aggregate code. Therefore, CocoaPods/Core#265 can pass individually, but this won't pass.

segiddins:
You can point the gem file here to your branch until the Core PR is merged

manuyavuz:
@segiddins I tried your suggestion but probably missing something.

Added the following line to Gemfile of CocoaPods project:

```
  gem 'cocoapods-core', :git => 'https://github.com/manuyavuz/Core.git', :branch => 'improvement/searchPerformance'
```

When I run rake spec I get the following error:

```
[!] There was an error parsing `Gemfile`: You cannot specify the same gem twice coming from different sources.
You specified that cocoapods-core (>= 0) should come from https://github.com/CocoaPods/Core.git (at master) and https://github.com/manuyavuz/Core.git (at improvement/searchPerformance)
. Bundler cannot continue.

 #  from /Users/manuyavuz/Development/OpenSource/Cocoapods/Rainforest/CocoaPods/Gemfile:29
 #  -------------------------------------------
 #    gem 'cocoapods-dependencies'
 >    gem 'cocoapods-core', :git => 'https://github.com/manuyavuz/Core.git', :branch => 'improvement/searchPerformance'
 #    gem 'bacon'
 #  -------------------------------------------
```

Is this default Core dependency come from `gem 'cocoapods-dependencies'`? If so, how can I solve this? Setting local dependency won't work because travis do not run using local dependencies I guess?

manuyavuz:
Ah, got it. Removed cp_gem line for Core dependency.

manuyavuz:
@floere, previous errors have gone now. However, there are two fails for functional specs. I was not observing them in the previous commit. 

I think they may be related to changes included from that time until now because I rebased the branch with master before sending the last commit. 

Also the failure seems not related to any source file that I've changed.

manuyavuz:
Hi, 

I've completed update index policy. All tests pass but got a strange error during running Inch.

```
--------------------------------------------------------------------------------
Running Inch
--------------------------------------------------------------------------------
Parse docs …
.............................................................................................
Evaluating …
................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
rake aborted!
ArgumentError: wrong number of arguments (2 for 1)
/Users/travis/build/CocoaPods/CocoaPods/vendor/bundle/ruby/2.0.0/gems/colored-1.2/lib/colored.rb:83:in `color'
/Users/travis/build/CocoaPods/CocoaPods/vendor/bundle/ruby/2.0.0/gems/term-ansicolor-1.3.0/lib/term/ansicolor.rb:226:in `on_color'
/Users/travis/build/CocoaPods/CocoaPods/vendor/bundle/ruby/2.0.0/gems/inch-0.5.10/lib/inch/utils/ui.rb:65:in `__header'
/Users/travis/build/CocoaPods/CocoaPods/vendor/bundle/ruby/2.0.0/gems/inch-0.5.10/lib/inch/utils/ui.rb:47:in `header'
/Users/travis/build/CocoaPods/CocoaPods/vendor/bundle/ruby/2.0.0/gems/inch-0.5.10/lib/inch/cli/command/output/list.rb:27:in `block in display_list'
/Users/travis/build/CocoaPods/CocoaPods/vendor/bundle/ruby/2.0.0/gems/inch-0.5.10/lib/inch/cli/command/output/list.rb:22:in `each'
/Users/travis/build/CocoaPods/CocoaPods/vendor/bundle/ruby/2.0.0/gems/inch-0.5.10/lib/inch/cli/command/output/list.rb:22:in `display_list'
/Users/travis/build/CocoaPods/CocoaPods/vendor/bundle/ruby/2.0.0/gems/inch-0.5.10/lib/inch/cli/command/output/list.rb:16:in `initialize'
/Users/travis/build/CocoaPods/CocoaPods/Rakefile:344:in `new'
/Users/travis/build/CocoaPods/CocoaPods/Rakefile:344:in `block (2 levels) in <top (required)>'
/Users/travis/build/CocoaPods/CocoaPods/Rakefile:146:in `block (2 levels) in <top (required)>'
Tasks: TOP => inch:spec
(See full trace by running task with --trace)
```

What could be the reason?

segiddins:
That happens when inch fails, so something isn't properly documented.

manuyavuz:
@floere with the last commit all tests pass. You may check out them now.

segiddins:
W00t! Excellent work, @manuyavuz , really look forward to reviewing this! If you want, feel free to give me your email so we can get you into our slack org?

manuyavuz:
Thanks @segiddins! I would be very happy to join the slack team because I'm hoping to contribute more. You can use the email in my github profile.

Waiting to hear for more!

lib/cocoapods/sources_manager.rb
```diff
@@ -208,11 +253,14 @@ def update(source_name = nil, show_output = false)
           sources =  git_sources
         end
 
+        changed_spec_paths = {}
         sources.each do |source|
           UI.section "Updating spec repo `#{source.name}`" do
             Dir.chdir(source.repo) do
               begin
+                prev_commit_hash = (git! %w(rev-parse HEAD)).strip
```
segiddins: can we move this updating logic into `Source` ? i.e. so it has an `update` method that returns the changes spec paths?

lib/cocoapods/sources_manager.rb
```diff
@@ -224,6 +272,11 @@ def update(source_name = nil, show_output = false)
             check_version_information(source.repo)
           end
         end
+        # Perform search index update operation as a subprocess.
+        fork do
```
segiddins: should the forked process do `Process.daemon` so that it doesn't keep the parent process alive?

spec/unit/sources_manager_spec.rb
```diff
@@ -241,11 +231,12 @@ module Pod
         UI.output.should.match /is up to date/
       end
 
-      it 'uses the only fast forward git option' do
-        set_up_test_repo_for_update
-        SourcesManager.expects(:git!).with { |options| options.should.include? '--ff-only' }
-        SourcesManager.update(test_repo_path.basename.to_s, true)
-      end
+      # TODO: how should we change this? Update code runs `git rev-parse HEAD` which fails for this spec
```
segiddins: stub out the git things to return what you need them to return

lib/cocoapods/sources_manager.rb
manuyavuz: You mean moving also the following, right?

```
output = git! %w(pull --ff-only)
```

segiddins: yup!

manuyavuz: @segiddins should we also move UI code there into source?

I don't think it's suitable because we would need config object inside source to handle UI things.

I'm thinking of making a method returning a tuple variable including git pull output and changed paths. It would be sth like this:

```
[output, changed_spec_paths]  = source.update
```
Does this fit?

segiddins:
Yup

lib/cocoapods/sources_manager.rb
```diff
@@ -142,41 +137,64 @@ def search_by_name(query, full_text_search = false)
...
       end
 
+      # Stores given search data in the file system.
+      # @param [Hash] index
+      #        Index to be saved in file system
+      #
+      def save_search_index(index)
+        require 'json'
+        @updated_search_index = index
+        File.open(search_index_path, 'w') do |file|
```
segiddins: `search_index_path` should be a `Pathname`

lib/cocoapods/sources_manager.rb
```diff
@@ -208,22 +255,19 @@ def update(source_name = nil, show_output = false)
...
-                  'and persists you can inspect it running ' \
-                  '`pod repo update --verbose`'
-              end
-            end
+            changed_spec_paths[source] = source.update(show_output && !config.verbose?)
             check_version_information(source.repo)
           end
         end
+        # Perform search index update operation as a subprocess.
+        Process.fork do
```
segiddins: can we extract this bit into a new method?
`update_search_index_if_needed_in_background` or something?

spec/unit/sources_manager_spec.rb
```diff
@@ -241,11 +231,12 @@ module Pod
         UI.output.should.match /is up to date/
       end
 
-      it 'uses the only fast forward git option' do
-        set_up_test_repo_for_update
-        SourcesManager.expects(:git!).with { |options| options.should.include? '--ff-only' }
-        SourcesManager.update(test_repo_path.basename.to_s, true)
-      end
+      # TODO: given that update moved to source.rb, should this still exist here?
```
segiddins: this is good to delete

lib/cocoapods/sources_manager.rb
manuyavuz: I think this way works too. Current implementation also performs open in the same way as the following:

```
      File.open(search_index_path, 'w') do |file|
        file.write(search_index.to_yaml)
      end
```

What kind of problem may we face?

segiddins: No problem, we just try to stay away from using strings as paths and use pathnames instead

manuyavuz: I think it indeed gives a pathname. It's a method of sources_manager as the following (line 202):

```
      # @return [Pathname] The path where the search index should be stored.
      #
      def search_index_path
        Config.instance.search_index_file
      end
```

segiddins: in that case, lets just use `search_index_path.open`

manuyavuz:
I've completed implementation and fixed multi word search issue. Solving that required a change in cocoapods-search plugin also (CocoaPods/cocoapods-search#10)

I think this is ready for final review too.

lib/cocoapods/sources_manager.rb
```diff
@@ -192,6 +214,44 @@ def search_index_path
...
+      # Updates the stored search index if there are changes in spec repos while updating them.
+      # Update is performed incrementally. Only the changed pods' search data is re-generated and updated.
+      # @param  [Hash{Source => Array<String>}] changed_spec_paths
+      #                  A hash containing changed specification paths for each source.
+      def update_search_index_if_needed(changed_spec_paths)
+        search_index = nil
+        changed_spec_paths.each_pair do |source, spec_paths|
+          source_name = source.name
+          next unless spec_paths.length > 0
+          search_index = stored_search_index
```
floere: Could this be `||=`?

lib/cocoapods/sources_manager.rb
```diff
@@ -192,6 +214,44 @@ def search_index_path
...
+          updated_pods = source.pods_for_specification_paths(spec_paths)
+          new_index = aggregate.generate_search_index_for_changes_in_source(source, spec_paths)
+          next unless search_index && search_index[source_name]
+          # First traverse search_index and update existing words
+          # Removed traversed words from new_index after adding to search_index,
+          # so that only non existing words will remain in new_index after enumeration completes.
+          search_index[source_name].each_pair do |word, _|
```
floere: You could store the result of the `search_index[source_name]` access in `source_search_index` (example name) to clarify that it's always about the same piece of the search index.

floere:
Only have two small comments. Thanks @manuyavuz!

lib/cocoapods/sources_manager.rb
manuyavuz: We do not want to create a search index from scratch while updating repos if there is no search index created before. So i think it should stay as it is, and exit from block at line 229.

floere: What do you mean when you say "exit from block"?

manuyavuz: Ah man I got the possible bug now. You're right. Fixing it.

manuyavuz:
Consider this case:

- User performs a repo update.
- Update method detects changes in one of the repositories.
- There is an existing search_index, but it does not contain inverted index for updated repo (possibly because user added repo after she performed a pod search).

Should this code generate an inverted index for only the changed files? Or should it ignore them?

manuyavuz:
This decision will also affect search policy.

I think the most appropriate flow is the following:

- During repo update, search index for a source repo is updated only if there exist a previous search index generated for the source repo. If there is no index for the source, it will not be generated, even if there are changes in the repo.

- During a `pod search --full` search, code will look for each source, and will generate search index from scratch if there is no index generated for that source before.

If that's OK, i'll finish development accordingly.

Thoughts, @floere, @segiddins, @alloy?

floere:
@manuyavuz That makes a lot of sense to me.

manuyavuz:
Guys, I have a problem. 

As you can see, travis fails on ` updates a repository` spec of `update_spec.rb` file.

Because the new update method starts a daemon process in background, it keeps search_index.json file open. Specs try to clear `tmp` folder in `preflight.rb` file before any specs, but because the file is in use by daemon process, remove fails with `Errno::ENOTEMPTY` error.

I've solved this in unit test via stubbing `update_search_index_if_needed_in_background` method which starts the daemon process. However, I'm not sure if this is the correct way to handle it in a functional spec. 

What should I do to resolve this?

segiddins:
@manuyavuz make it so `update_search_index_if_needed_in_background` returns a PID you can wait on?

manuyavuz:
@segiddins Wouldn't original update method (therefore the `pod repo update` command) be blocked if I wait for that pid to finish?

segiddins:
Good point. I'd just expect that the background method get's called then, manually call the synchronous method, and then verify

manuyavuz:
I actually did as you suggest in unit tests of SourcesManager class. However, not sure if this is an appropriate thing to do in a functional spec. Failed test is the following:

``` ruby
    it 'updates a repository' do
      upstream = SpecHelper.temporary_directory + 'upstream'
      FileUtils.cp_r(test_repo_path, upstream)
      Dir.chdir(test_repo_path) do
        `git remote add origin #{upstream}`
        `git remote -v`
        `git fetch -q`
        `git branch --set-upstream-to=origin/master master`
      end
      lambda { command('repo', 'update').run }.should.not.raise
    end
```

Is it alright to change it like the following?:

``` ruby
    it 'updates a repository' do
      upstream = SpecHelper.temporary_directory + 'upstream'
      FileUtils.cp_r(test_repo_path, upstream)
      Dir.chdir(test_repo_path) do
        `git remote add origin #{upstream}`
        `git remote -v`
        `git fetch -q`
        `git branch --set-upstream-to=origin/master master`
      end
      # Add expect for bg method
      SourcesManager.expects(:update_search_index_if_needed_in_background).with({}).returns(nil)
      lambda { command('repo', 'update').run }.should.not.raise
    end
```

If this is OK, I'll make a last rebase and push final version.

segiddins:
IMHO that's fine

manuyavuz:
Offenses made me sick 😤

I think this is ready to 🚀 also. You can now make a last glance.

spec/unit/sources_manager_spec.rb
```diff
@@ -235,28 +225,47 @@ module Pod
...
 
-      it 'update source backed by a git repository' do
+      it 'updates source backed by a git repository' do
         set_up_test_repo_for_update
+        SourcesManager.expects(:update_search_index_if_needed_in_background).with({}).returns(nil)
         SourcesManager.update(test_repo_path.basename.to_s, true)
         UI.output.should.match /is up to date/
       end
 
-      it 'uses the only fast forward git option' do
```
segiddins: This test can come back

segiddins:
😄 super well done!

floere:
:+1:

manuyavuz:
It appeared again 😉

segiddins:
:rocket: :ship:
Great work @manuyavuz!

manuyavuz:
Great!

PR #4261, opened 2015-09-25
[CopyResourcesScript][EmbedFrameworksScript] Only create directories …

segiddins:
…when they will have contents

This might help with some of the issues we've been getting about extensions?

* [ ] Figure out if this _actually_ fixes things
* [ ] Changelog 

\c @kylef @mrackwitz

### DIFF
--- a/lib/cocoapods/generator/copy_resources_script.rb
+++ b/lib/cocoapods/generator/copy_resources_script.rb
@@ -94,11 +94,9 @@ module Pod
         script
       end
 
-      INSTALL_RESOURCES_FUNCTION = <<EOS
+      INSTALL_RESOURCES_FUNCTION = <<-SH
 #!/bin/sh
 set -e
-
-mkdir -p "${CONFIGURATION_BUILD_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}"
 
 RESOURCES_TO_COPY=${PODS_ROOT}/resources-to-copy-${TARGETNAME}.txt
 > "$RESOURCES_TO_COPY"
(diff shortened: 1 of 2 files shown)


lib/cocoapods/generator/copy_resources_script.rb
```diff
@@ -154,20 +154,20 @@ def script
...
-
-mkdir -p "${CONFIGURATION_BUILD_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}"
-rsync -avr --copy-links --no-relative --exclude '*/.svn/*' --files-from="$RESOURCES_TO_COPY" / "${CONFIGURATION_BUILD_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}"
-if [[ "${ACTION}" == "install" ]] && [[ "${SKIP_INSTALL}" == "NO" ]]; then
-  mkdir -p "${INSTALL_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}"
-  rsync -avr --copy-links --no-relative --exclude '*/.svn/*' --files-from="$RESOURCES_TO_COPY" / "${INSTALL_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}"
+SH
+
+      RSYNC_CALL = <<-SH
+if [ "$(cat "$RESOURCES_TO_COPY")" ]; then
```
mrackwitz: Does this actually work? Can't we just do `[ -s "$RESOURCES_TO_COPY" ]`?

segiddins: `-s` is bash-specific, I believe

segiddins:
(Also, we need to test if this _actually_ fixes things or not)

lib/cocoapods/generator/copy_resources_script.rb
kylef: `sh` is bash on OS X. We could use `#!/usr/bin/env bash` instead of `#!/bin/sh` and switch to use this.

segiddins: but `sh` isn't guaranteed to be bash, merely a POSIX-compliant shell. Since right now we use `/bin/sh`, I didn't feel it was worth changing it

PR #4267, opened 2015-09-26
[PodTargetInstaller] Preserve header_mappings_dir when copying header…

segiddins:
…s to a framework

I got bored. Needs a new integration spec added, and manual verification the built framework, you know, works. Beyond that, we should also validate that, when a `header_mappings_dir` exists, _all_ of the pod's headers are in that directory (otherwise invalid? frameworks might be created).

- [x] CHANGELOG
- [x] Integration spec

\c @samdmarshall


segiddins:
Can someone please confirm this is something we want? If so, I'll scope it for 1.0.

skydivedan:
It's something that I want, if it helps fix the referenced RestKit problem.

PR #4281, opened 2015-09-29
[Analyzer] Don't require platform for podless targets when not integrating

sguillope:
When running `pod install --no-integrate`, it's not required to validate that the `platform` attribute is specified for targets that don't declare any dependencies.

### DIFF
--- a/lib/cocoapods/installer/analyzer.rb
+++ b/lib/cocoapods/installer/analyzer.rb
@@ -636,7 +636,7 @@ module Pod
       def verify_platforms_specified!
         unless config.integrate_targets?
           podfile.target_definition_list.each do |target_definition|
-            if target_definition.dependencies.size > 0 && target_definition.platform.nil?
+            if !target_definition.empty? && target_definition.platform.nil?
               raise Informative, 'It is necessary to specify the platform in the Podfile if not integrating.'
             end
           end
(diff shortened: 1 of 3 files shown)


lib/cocoapods/installer/analyzer.rb
```diff
@@ -636,7 +636,7 @@ def sources
       def verify_platforms_specified!
         unless config.integrate_targets?
           podfile.target_definition_list.each do |target_definition|
-            unless target_definition.platform
+            if target_definition.dependencies.size > 0 && target_definition.platform.nil?
```
segiddins: `target_definition.empty?`

segiddins:
Awesome!

lib/cocoapods/installer/analyzer.rb
sguillope: :+1: 
Do you mean this though?
```
if !target_definition.empty? && target_definition.platform.nil?
```

segiddins:
@sguillope yup!

lib/cocoapods/installer/analyzer.rb
sguillope: @segiddins is it correct to exclude inherited dependencies in the check?
If have a `Podfile` like:
```ruby
platform :ios
pod 'InheritedPod'
target 'osx' do
  # empty
end
```
should the `osx` target fail the platform validation?

segiddins:
@sguillope I believe that's ok, since the platform will also be inherited?

segiddins:
Excellent. Thanks so much, @sguillope ! :shipit:

PR #4318, opened 2015-10-05
[HooksManager] Send user options as a HashWithIndifferentAccess

### LINKED ISSUE #3354: [Plugins] Provide plugin options as a HashWithIndifferentAccess.
Currently it’s too easy to run into issues where a user provides options as symbols and the plugin expects them to be strings or vice-versa. Instead of having the plugin implementers deal with this every time, it’s probably safer to use ActiveSupport’s [HashWithIndifferentAccess](http://api.rubyonrails.org/classes/ActiveSupport/HashWithIndifferentAccess.html).

segiddins:
Closes #3354.

### DIFF
--- a/lib/cocoapods/hooks_manager.rb
+++ b/lib/cocoapods/hooks_manager.rb
@@ -1,4 +1,4 @@
-require 'rubygems'
+require 'active_support/core_ext/hash/indifferent_access'
 
 module Pod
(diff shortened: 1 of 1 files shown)


orta:
wooooooo

segiddins:
@orta you shouldn't have merged this without tests and a changelog entry, please add those?

orta:
sorry, too excited :dancers:

PR #4320, opened 2015-10-05
[PodTargetInstaller] Set deployment target to maximum spec deployment…

### LINKED ISSUE #3454: respect the deployment_target of a dependency when building
XCode 6.3 has started to use the deployment target of a library when performing static analysis.

Currently CocoaPods overrides the deployment target of the podspec with the platform directive in each target. For example:

```
target 'MyApp' do
    platform :ios, '6.1'
    pod 'CorePlot' # <-- has a deployment_target of 5.0
    pod 'AFNetworking'
end
```

The library `libPods-MyApp-CorePlot.a` will be built with a deployment target of 6.1 instead of 5.0. This causes a lot of warnings in static analysis because CorePlot has shims in the code that make the library work on ios 5.0. However, XCode realizes those shims are no longer necessary if deploying to 6.1+ and emits warnings.

A possible work around for this was to create multiple targets with their own deployment targets:

```
target 'MyApp-50' do
    platform :ios, '5.0'
    link_with 'MyApp'
    pod 'CorePlot'
end

target 'MyApp-61' do
    platform :ios, '6.1'
    link_with 'MyApp'
    pod 'AFNetworking'
end
```

Unfortunately each of these targets has its own xcconfig and CocoaPods doesn't merge these xcconfigs into one, causing issues. So CocoaPods doesn't seem to support two pods linking against the same entry point (`MyApp`).

The correct solution (it seems to me) is that the deployment target should be used from the podspec when building the library. For example, CorePlot has `spec.ios.deployment_target '5.0'`, so when building `libPods-MyApp-CorePlot.a` the deployment target should be set to 5.0. This seems reasonable to me since we're talking about configuring the dependencies, not `MyApp` (which has its own xcodeproj that I configure). Everything works fine if I manually click to set the deployment target on there in the generated xcodeproj, but obviously this gets wiped out when I run `pod update` again.

segiddins:
… target

Closes #3454.

- [x] CHANGELOG
- [x] Specs

### DIFF
--- a/lib/cocoapods/installer/target_installer/pod_target_installer.rb
+++ b/lib/cocoapods/installer/target_installer/pod_target_installer.rb
@@ -310,6 +310,15 @@ module Pod
         end
       end
 
+      # The deployment target for the pod target, which is the maximum of all
+      # the deployment targets for the current platform of the target.
+      #
+      # @return [String] The deployment target.
+      #
+      def deployment_target
+        target.specs.map { |spec| spec.deployment_target(target.platform) }.max
+      end
+
       #-----------------------------------------------------------------------#
     end
   end
--- a/lib/cocoapods/installer/target_installer/pod_target_installer.rb
+++ b/lib/cocoapods/installer/target_installer/pod_target_installer.rb
@@ -317,7 +317,9 @@ module Pod
       #
       def deployment_target
         default = Podfile::TargetDefinition::PLATFORM_DEFAULTS[target.platform.name]
-        target.specs.map { |spec| spec.deployment_target(target.platform) || default }.max
+        target.specs.map do |spec|
+          Pod::Version.new(spec.deployment_target(target.platform) || default)
+        end.max.to_s
       end
 
       #-----------------------------------------------------------------------#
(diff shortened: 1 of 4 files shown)


lib/cocoapods/installer/target_installer/pod_target_installer.rb
```diff
@@ -285,6 +285,15 @@ def custom_module_map
         @custom_module_map ||= target.file_accessors.first.module_map
       end
 
+      # The deployment target for the pod target, which is the maximum of all
+      # the deployment targets for the current platform of the target.
+      #
+      # @return [String] The deployment target.
+      #
+      def deployment_target
+        target.specs.map { |spec| spec.deployment_target(target.platform) }.max
```
segiddins: this needs to `||` with `PLATFORM_DEFAULTS[target.platform]`

lib/cocoapods/installer/target_installer/pod_target_installer.rb
```diff
@@ -285,6 +285,16 @@ def custom_module_map
...
       end
 
+      # The deployment target for the pod target, which is the maximum of all
+      # the deployment targets for the current platform of the target.
+      #
+      # @return [String] The deployment target.
+      #
+      def deployment_target
+        default = Podfile::TargetDefinition::PLATFORM_DEFAULTS[target.platform.name]
+        target.specs.map { |spec| spec.deployment_target(target.platform) || default }.max
```
floere: Won't this fail in the standard problem case `ruby -e 'p ["10.10", "10.9"].max'`?

segiddins: Ah you're right. Need to wrap into a Pod::Version

segiddins:
Will merge once CI runs

PR #4372, opened 2015-10-12
[AbstractExternalSource] Quickly validate specs

segiddins:
- [x] CHANGELOG
- [x] Tests

### DIFF
--- a/lib/cocoapods/external_sources/abstract_external_source.rb
+++ b/lib/cocoapods/external_sources/abstract_external_source.rb
@@ -141,17 +141,29 @@ module Pod
       # @return [void]
       #
       def store_podspec(sandbox, spec, json = false)
-        if spec.is_a? Pathname
-          spec = Specification.from_file(spec).to_pretty_json
-          json = true
-        elsif spec.is_a?(String) && !json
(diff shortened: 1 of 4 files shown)


lib/cocoapods/external_sources/abstract_external_source.rb
```diff
@@ -141,17 +141,29 @@ def download_request
...
+        case spec
+        when Pathname
+          spec = Specification.from_file(spec)
+        when String
+          path = "#{name}.podspec"
+          path << '.json' if json
+          spec = Specification.from_string(spec, path)
+        end
+        validate_podspec(spec)
+        sandbox.store_podspec(name, spec.to_pretty_json, true, true)
```
floere: Why can the fourth parameter here only be `true`?

segiddins: because it's always saved as json

floere: I think I was asking because of the `json` parameter and the `path << '.json' if json` code. Is there still any point to that code?

segiddins:
Will merge once CI passes

segiddins:
:ship:. I've been waiting for this for years!

Issue #4374, opened 2015-10-13
Pod install spends a lot of time on Xcodeproj::Project.open


danielribeiro:
🌈

Hi,

We have found that pod install spends a lot of time on Xcodeproj::Project.open. Since it is called multiple times with the same path, with a simple cache, we have managed to speed it up by 100%.

A sample of the speedup and how we achieved it is in this public fork of Braintree's sdk: https://github.com/danielribeiro/braintree_ios/commit/627f24202e2065266a76a05b5dc0f6b7b5588dfc#diff-4a25b996826623c4a3a4910f47f10c30

The profile output showing the time spend can be seen [here](http://danielribeiro.github.io/braintree_ios/ProfileForCallStackPrinter.html).

On one of our largest iOS repos, this 2x speedup amounts to a few minutes saved. I'd love to see this performance improvement be made more broadly available. Any thoughts?

segiddins:
Thanks for investigating! I don't think the cache in Xcodeproj is the right answer, but we probably could do a better job of holding onto the projects in the installer instead of re-opening them?

danielribeiro:
No problem :smile: 

I agree caching in Xcodeproj feels a bit out of place. Maybe caching on the call sites inside CococaPods would be better?

I've summed up the call sites [here](https://github.com/danielribeiro/braintree_ios/blob/danielrb/logging-who-calls-XcodeprojProject.open/callsiteanalysis.txt) (the raw data was generated by [this code](https://github.com/danielribeiro/braintree_ios/pull/3/files)) . The summary is:

```
This callsite was called 6 times: 
$GEMS/cocoapods-0.39.0/lib/cocoapods/installer/analyzer/target_inspector.rb:32:in `compute_results'
$GEMS/cocoapods-0.39.0/lib/cocoapods/installer/analyzer.rb:659:in `block (2 levels) in inspect_targets_to_integrate'
$GEMS/cocoapods-0.39.0/lib/cocoapods/installer/analyzer.rb:657:in `each'

This callsite was called 6 times: 
$GEMS/cocoapods-0.39.0/lib/cocoapods/installer.rb:552:in `block (2 levels) in prepare_pods_project'
$GEMS/cocoapods-0.39.0/lib/cocoapods/installer.rb:551:in `map'
$GEMS/cocoapods-0.39.0/lib/cocoapods/installer.rb:551:in `block in prepare_pods_project'

This callsite was called 6 times: 
[...]

PR #4399, opened 2015-10-16
improving the performance of Pod::Installer::Analyzer#generate_pod_targets

danielribeiro:
…, by avoiding computing previously computed dependency trees.

On our cases, this makes what would be a 20min+ `pod install`, back into a 2min (measured with the performance gains described on the https://github.com/CocoaPods/CocoaPods/issues/4374 issue).

This gain is huge for projects with a lot of targets (20+) and pods (100+), when the de-duplication process happens.

### DIFF
--- a/lib/cocoapods/target/pod_target.rb
+++ b/lib/cocoapods/target/pod_target.rb
@@ -57,13 +57,15 @@ module Pod
       @dependent_targets = []
     end
 
+    # @param [Hash{Array => PodTarget}] cache
+    #        the cached PodTarget for a previously scoped (specs, target_definition)
     # @return [Array<PodTarget>] a scoped copy for each target definition.
     #
     def scoped(cache = {})
       target_definitions.map do |target_definition|
-        cash_key = [specs, target_definition]
-        if cache[cash_key]
-          cache[cash_key]
+        cache_key = [specs, target_definition]
+        if cache[cache_key]
+          cache[cache_key]
         else
           target = PodTarget.new(specs, [target_definition], sandbox, true)
           target.file_accessors = file_accessors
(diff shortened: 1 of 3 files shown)


segiddins:
@mrackwitz please review?

mrackwitz:
Isn't the only legit case, when the cache may be hit, when the `PodTarget` is already scoped? I'd just return the same `PodTarget` by it's method `#scoped` in that case. That may help with the performance already through saving some unnecessary allocations.

segiddins:
@mrackwitz that's the point -- it caches the scoped dependencies for cases when they're transitive.

lib/cocoapods/target/pod_target.rb
```diff
@@ -59,14 +59,19 @@ def initialize(specs, target_definitions, sandbox, scoped = false)
 
     # @return [Array<PodTarget>] a scoped copy for each target definition.
     #
-    def scoped
+    def scoped(cache = {})
       target_definitions.map do |target_definition|
-        PodTarget.new(specs, [target_definition], sandbox, true).tap do |target|
+        cash_key = [specs, target_definition]
```
segiddins: `cache`

lib/cocoapods/target/pod_target.rb
```diff
@@ -59,14 +59,19 @@ def initialize(specs, target_definitions, sandbox, scoped = false)
 
     # @return [Array<PodTarget>] a scoped copy for each target definition.
     #
-    def scoped
+    def scoped(cache = {})
```
segiddins: docs need to be updated for the new argument

danielribeiro: Added. And rebased off master.

segiddins:
:+1: looks good to me!

mrackwitz:
Okay, makes perfectly sense and I can see how this will have huge performance gains. :+1:
It will also guarantee again that structural equivalent pod targets have always the same object identity. :tada:

mrackwitz:
@danielribeiro: Please add a changelog entry, then this is good to merge.

danielribeiro:
@mrackwitz Just updated it. Thanks :smile:

mrackwitz:
Ace! :+1:

PR #4408, opened 2015-10-19
Fix handling of localized files in Pods installed as frameworks

### LINKED ISSUE #2597: Storyboard in Resource Bundle
Hi guys,

I've searched all through here and SO today, and I'm struggling to find a solution to my issue. I'm trying to include a library I've made into a project. The library includes localized strings and localized storyboards, so I've added a `resource_bundle` into it's podspec, which has a `**/*.{lproj,png}` specification. The strings are coming through fine, as expected (it pulls in the whole lproj folder). However, the storyboards are *not* working, and I believe it's because they aren't compiled. Instead of being .storyboardc files, the lproj's just contain .storyboards. Am I doing this the right way? Is there something I'm missing? Is this supported?

Thanks. Here's your :rainbow: btw.

timbodeit:
Fixes #2597 

When a resource pattern like `**/*.{storyboard,strings}` is specified in a Pod that gets installed as a framework, files in `*.lproj` folders will now be grouped together by file name and base path in a `PBXVariantGroup`.

##### Terminology 

path = `<basepath>/<language>.lproj/<filename>.<extension>`

In the following

- `basepath` means the path of the parent directory of the .lproj directory that a localized file is in.
- `filename` means the name of the file without the extension

#### Reason for change

This is needed, so that Xcode recognizes them as localized versions of the the same file. It had already been possible to use a `**/*.lproj` pattern to copy localized `.strings` files. However with storyboards this does not work as Xcode (when building a framework) doesn't compile them when the `.lproj` folder is treated as a file reference rather than a folder.

##### Variant group naming

`Base.lproj/Foo.storyboard` and `Base/Foo.strings` will be combined into a variant group named `Foo` (without extension). 
This is a little different from what you would get if you dragged the file into the project yourself and clicked the language checkboxes on the storyboard file manually.  
The usual project would have a variant group called `Foo.storyboard`, that contains `Base.lproj/Foo.storyboard` and the various `*.lproj/Foo.strings`.

It seems to me, that the variant group name is only shown to the Xcode user and plays no role in the actual build process.
As I didn't want to define "extension precendence", dropping the file extension seemed like the best option to me.

I came to this conclusion after:

- Trying a localized `Image.jpg` would still be retrieved by `UIImage(named: "Image.jpg")` and not `UIImage(named: "Image")`
- Inspecting the compiled framework in the DerivedData folder and finding this folder structure
```
Base.lproj
┗ Foo.storyboardc
┗ Image.jpg
de.lproj
┗ Foo.strings
┗ Image.jpg
...
```

##### Multiple localized files with the same name

It is possible that a a lib developer wants to include multiple files localized files with the same name (not counting extension). An example of this might be a `FooLib` that includes a `Foo.storyboard` and a localized `Foo.png`.

To not make the grouping logic overly complicated I used the following grouping scheme:
- All files with the same basepath and filename are considered localized versions of the same file.
  Meaning: they will be put into a single variant group.  
- File extension is not part of the compared file name to allow for localized storyboards, xibs etc.

If a lib developer wants to include two unrelated, localized files with the same name, I highly recommend putting them in two different folders (using two different base paths). If not, the structure of the xcode project will look a little weird to the user, though it doesn't seem to change the resulting framework.
I did a quick test on how Xcodes behaves when a variant group contains multiple files for a single language and couldn't notice any difference to having two separate variant groups.

##### Type signature for `project#variant_groups_by_path_and_name`

The type signature includes `[Pathname, String]`. What I meant by this is an array with two elements where the first one is a Pathname and the second is a String.

Ruby not being one of my strongest languages skills, I'm not sure that this is actually the correct meaning of the `[Foo,Bar]` notation (see `project#initialize`s path argument).
Did I use this correctly?

### DIFF
--- a/lib/cocoapods/project.rb
+++ b/lib/cocoapods/project.rb
@@ -20,6 +20,7 @@ module Pod
       super(path, skip_initialization, object_version)
       @support_files_group = new_group('Targets Support Files')
       @refs_by_absolute_path = {}
+      @variant_groups_by_path_and_name = {}
       @pods = new_group('Pods')
       @development_pods = new_group('Development Pods')
       self.symroot = LEGACY_BUILD_ROOT
@@ -265,6 +266,11 @@ module Pod
     #
     attr_reader :refs_by_absolute_path
 
+    # @return [Hash{[Pathname, String] => PBXVariantGroup}] The variant groups
+    #         grouped by absolute path of parent dir and name.
+    #
+    attr_reader :variant_groups_by_path_and_name
+
     # Returns the group for an absolute file path in another group.
     # Creates subgroups to reflect the file system structure if
     # reflect_file_system_structure is set to true.
@@ -308,7 +314,9 @@ module Pod
       if relative_dir.basename.to_s =~ lproj_regex
         filename = absolute_pathname.basename.sub_ext('').to_s
         lproj_parent_dir = absolute_pathname.dirname.dirname
-        group = group[filename] || group.new_variant_group(filename, lproj_parent_dir)
+        group = @variant_groups_by_path_and_name[[lproj_parent_dir, filename]] ||
+          group.new_variant_group(filename, lproj_parent_dir)
+        @variant_groups_by_path_and_name[[lproj_parent_dir, filename]] = group
       end
 
       group
--- a/lib/cocoapods/project.rb
+++ b/lib/cocoapods/project.rb
@@ -183,14 +183,7 @@ module Pod
         raise ArgumentError, "Paths must be absolute #{absolute_path}"
       end
 
-      if reflect_file_system_structure
-        relative_path = file_path_name.relative_path_from(group.real_path)
-        relative_dir = relative_path.dirname
-        relative_dir.each_filename do|name|
-          next if name == '.'
-          group = group[name] || group.new_group(name, name)
-        end
-      end
+      group = group_for_path_in_group(absolute_path, group, reflect_file_system_structure)
 
       if ref = reference_for_path(absolute_path)
         ref
@@ -271,6 +264,55 @@ module Pod
     #         by absolute path.
     #
     attr_reader :refs_by_absolute_path
+
+    # Returns the group for an absolute file path in another group.
+    # Creates subgroups to reflect the file system structure if
+    # reflect_file_system_structure is set to true.
+    # Makes a variant group if the path points to a localized file inside a
+    # *.lproj folder. The same variant group is returned for files with the
+    # same name, even if their file extensions differ.
+    #
+    # @param  [Pathname] absolute_pathname
+    #         The pathname of the file to get the group for.
+    #
+    # @param  [PBXGroup] group
+    #         The parent group used as the base of the relative path.
+    #
+    # @param  [Bool] reflect_file_system_structure
+    #         Whether group structure should reflect the file system structure.
+    #         If yes, where needed, intermediate groups are created, similar to
+    #         how mkdir -p operates.
+    #
+    # @return [PBXGroup] The appropriate group for the filepath.
+    #         Can be PBXVariantGroup, if the file is localized.
+    #
+    def group_for_path_in_group(absolute_pathname, group, reflect_file_system_structure)
+      unless absolute_pathname.absolute?
+        raise ArgumentError, "Paths must be absolute  #{absolute_path}"
+      end
+
+      relative_pathname = absolute_pathname.relative_path_from(group.real_path)
+      relative_dir = relative_pathname.dirname
+      lproj_regex = /\.lproj/
+
+      # Add subgroups for folders, but treat .lproj as a file
+      if reflect_file_system_structure
+        relative_dir.each_filename do|name|
+          break if name.to_s =~ lproj_regex
+          next if name == '.'
+          group = group[name] || group.new_group(name, name)
+        end
+      end
+
+      # Turn .lproj into a variant group
+      if relative_dir.basename.to_s =~ lproj_regex
+        filename = absolute_pathname.basename.sub_ext('').to_s
+        lproj_parent_dir = absolute_pathname.dirname.dirname
+        group = group[filename] || group.new_variant_group(filename, lproj_parent_dir)
+      end
+
+      group
+    end
 
     #-------------------------------------------------------------------------#
   end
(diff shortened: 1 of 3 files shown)


timbodeit:
# Depends on [Xcodeproj#318](https://github.com/CocoaPods/Xcodeproj/pull/318)

The new code in project.rb relies on the `PBXGroup#new_variant_group`. This pull request in which this method was introduced has been merged into Xcodeproj:master, but since then no new releases where made.

segiddins:
@timbodeit ❤️ this is basically a perfect PR!

lib/cocoapods/project.rb
```diff
@@ -183,14 +184,7 @@ def add_file_reference(absolute_path, group, reflect_file_system_structure = fal
         raise ArgumentError, "Paths must be absolute #{absolute_path}"
```
segiddins: can now remove this, since the check is made later?

lib/cocoapods/project.rb
```diff
@@ -272,6 +266,62 @@ def add_build_configuration(name, type)
...
+    #         Can be PBXVariantGroup, if the file is localized.
+    #
+    def group_for_path_in_group(absolute_pathname, group, reflect_file_system_structure)
+      unless absolute_pathname.absolute?
+        raise ArgumentError, "Paths must be absolute  #{absolute_path}"
+      end
+
+      relative_pathname = absolute_pathname.relative_path_from(group.real_path)
+      relative_dir = relative_pathname.dirname
+      lproj_regex = /\.lproj/
```
segiddins: `/xi`

lib/cocoapods/project.rb
```diff
@@ -272,6 +266,62 @@ def add_build_configuration(name, type)
...
+          next if name == '.'
+          group = group[name] || group.new_group(name, name)
+        end
+      end
+
+      # Turn .lproj into a variant group
+      if relative_dir.basename.to_s =~ lproj_regex
+        filename = absolute_pathname.basename.sub_ext('').to_s
+        lproj_parent_dir = absolute_pathname.dirname.dirname
+        group = @variant_groups_by_path_and_name[[lproj_parent_dir, filename]] ||
```
segiddins: use `||=` here ?

lib/cocoapods/project.rb
timbodeit: @segiddins :+1:  on the `i`. But what's the purpose of the `x` if the expression contains no whitespace or comments?
Or is it something that you just apply by default?

lib/cocoapods/project.rb
timbodeit: True ... Except the stack trace being a little more concise when the error is thrown, there is no reason to keep it there.

segiddins:
This is failing on ci (https://travis-ci.org/CocoaPods/CocoaPods/jobs/86253825#L1618-L1732), otherwise I'm good to merge this, but someone else on the team should first verify it

timbodeit:
That's what happens when you think "those changes are so small ... no need to run `rake spec` locally".

I overlooked that the previous value of group is actually used in the line where you suggested using `||=`. I reversed the changes I made to that line. Using `||=` would be possible, but only by assigning to a different variable name and assigning this variable back to group. I'm happy to do that, but I feel like a multi line `||` is the better way to go. 
Noticed and fixed a couple other issues (see commit message for details).

segiddins:
No, my suggestion was to do `dictionary[key] ||= value`

timbodeit:
Oh ok .. I thought the comment was concerning the line above instead of the one below

timbodeit:
How about the dependency on the Xcodeproj changes? Would you do a release before merging or can that wait until the preparation of the next Cocoapods release?

segiddins:
Nah no need to wait for a release of Xcodeproj to merge

timbodeit:
Rebased onto CocoaPods:master and removed b1dbfe6 `[Gemspec] Update xcodeproj to current master hash` as a similar update has been made in efc2ffde37b4928452a3771b11427be8dce1cc3a

ankamsandeep:
@timbodeit +1, Thanks for the fix.

I have a problem, I am using static libraries and trying to use your fix. I see this following error 

NoMethodError - undefined method `new_variant_group' for #<Xcodeproj::Project::Object::PBXGroup:0x007fc28cec2970>

more logs https://gist.github.com/ankamsandeep/e253b3512091109c8875

timbodeit:
@ankamsandeep What commit of Xcodeproj are you using?

ankamsandeep:
@timbodeit I am using the master commit of Cocoapods, which include your fix. I guess master commit has the latest code from Xcodeproj.

timbodeit:
@ankamsandeep Did you run `bundle install`? I don't know the ruby build environment of CocoaPods that well, but this should also be done by `rake bootstrap` if I'm not mistaken.

I've checked the Gemfile.lock on the current master and that part seems to be ok. 

It seems to me like you're using an older version of Xcodeproj. If you're using the Rainforest with `rake local_dependencies_set`, also make sure that the `Xcodeproj` submodule is up to date.

ankamsandeep:
@timbodeit: I will try updating Xcodeproj submodule and see if that fixes. 

As per your previous comments "When a resource pattern like **/*.{storyboard,strings} is specified in a Pod that gets installed as a framework", will your solution work for Static Lib's?

ankamsandeep:
@timbodeit : The error thats posted on https://gist.github.com/ankamsandeep/e253b3512091109c8875 is thrown when using the following resources pattern.
s.resources_bundle = {'Resources' => [Pod/ResourcesBundle/**/*.{storyboard,xib}]}

timbodeit:
@ankamsandeep About support for static libraries: On first thought I would have said no. But with resource bundles it should actually work. If you could give it a try and tell us what you have found out, that would be awesome.

---
**Search Path**

`Pod/ResourceBundle` is the directory that contains your `*.lproj` folders?
Then you're looking for something more like: 
```
s.resource_bundles = {
    'Resources' => ['Pod/ResourcesBundle/**/*.{storyboard,xib}']
}
```
- `resource_bundles`, not `resources_bundle`
- File Pattern in single quotes (though I'm not sure that is necessary)
- `**` to match any directory structure
- `*` to match any file name before the named extensions

---

**Your error**

The error you're experiencing is unrelated to this though. I can only say, that I think it related to the version of Xcodeproj. This is everything that I can think of, that might solve this issue:
- Check out a clean copy of CocoaPods/Rainforest and run `rake bootstrap`. Use the cocoapods executable at `Rainforest/CocoaPods/bin/pod` for testing
- If you are trying to use it via `rake install`,  also run `rake install` for Xcodeproj. Otherwise you might be using the newest master of CocoaPods with an older version of Xcodeproj.  
To be safe, before installing, remove all versions of Xcodeproj you already have on your system by running `gem uninstall xcodeproj`

Issue #4388, opened 2015-10-14
Pods copy resources script does not set the minimum deployment target when compiling storyboards


aleh:
We have a pod with storyboards where "Builds for" is set to "Project Deployment Target". Our project is targeting iOS 7.0, but when the pods resource script compiles these storyboards it calls ibtool without specifying the minimum deployment version (unlike XCode, which would use --minimum-deployment-target 7.0  switch). It looks like with XCode 6.4+ ibtool assumes iOS 8.0+ when the target deployment version is not specified, which leads to crashes when storyboards are loaded on iOS 7.0.

As a workaround we can specify a concrete version in "Builds for" field, but this might be not desirable for  reusable pods.

I am using Cocoapods 0.38.2, btw.

segiddins:
Can you make a pull request adding this to the `ibtool` invocation in the copy resources script?

manuyavuz:
Closing the issue regarding problem solved with #4410 and there is no validation feedback. Feel free to reopen again if solution does not work for you.

PR #4410, opened 2015-10-19
Add deployment target option for ibtool command in copy resources script

manuyavuz:
This solves #4388

However, I'm not sure if this also works for os x targets because I'm adding deployment target using `IPHONEOS_DEPLOYMENT_TARGET` environment variable.

PS: Travis will fail for integration tests. I will send a PR updating integration specs repo also.

### DIFF
--- a/lib/cocoapods/generator/copy_resources_script.rb
+++ b/lib/cocoapods/generator/copy_resources_script.rb
@@ -116,12 +116,12 @@ install_resource()
 {
   case $1 in
     *\.storyboard)
-      echo "ibtool --reference-external-strings-file --errors --warnings --notices --minimum-deployment-target ${IPHONEOS_DEPLOYMENT_TARGET} --output-format human-readable-text --compile ${CONFIGURATION_BUILD_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}/`basename \\"$1\\" .storyboard`.storyboardc ${PODS_ROOT}/$1 --sdk ${SDKROOT}"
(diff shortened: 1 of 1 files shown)


segiddins:
Hm so I think we'll need to calculate the deployment target based upon the platform were building for

segiddins:
Maybe `${${SDKROOT}_DEPLOYMENT_TARGET}`?

manuyavuz:
It won't work because SDKROOT is the path of the sdk.

I see the following env variable:

```
DEPLOYMENT_TARGET_SETTING_NAME=IPHONEOS_DEPLOYMENT_TARGET
```
I think it could work via `${${DEPLOYMENT_TARGET_SETTING_NAME}}`

Thoughts?

segiddins:
ah yeah, that should do it

manuyavuz:
Even though I don't fully get what `!` character provides, true way is the following:

```
${!DEPLOYMENT_TARGET_SETTING_NAME}
```

I will update accordingly.

manuyavuz:
Made the change. I think we need to update cocoapods-deintegrate plugin along with cocoapods-integration-specs submodule. Can send a PR for that also.

segiddins:
Don't worry about the integration specs, I update them once I merge the CocoaPods PR. Why would deintegrate also require a change?

manuyavuz:
I see that it also contains test project structures inside spec fixtures which contain copy resource script in the old way. 

Great for integration specs, I'll close the PR on that repo then.

segiddins:
:+1: will merge this this afternoon with everything else, well done @manuyavuz

Issue #4142, opened 2015-09-08
Don't crash if the argument provided to `pod repo lint` is not a directory


yellokrow:
After uninstalling cocoapods-* using gem and deleting the .cocoapods folder, I reinstalled it. I issued `pod setup` after the install. When I issue the command `pod repo lint PodSpecName.podspec` for my previously working podspecs I get the error 'ENOTDIR - Not a directory'. I even tried to copy and paste the Artsy podspec from the tutorials on cocoapods.com and I get the same error when running the command. Help would be appreciated.

Linting spec repo `artsy.podspec`

[in /Users/MyUserName]

――― MARKDOWN TEMPLATE ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――

### Command

```
/usr/bin/pod repo lint artsy.podspec
```

### Report

* What did you do?

* What did you expect to happen?

* What happened instead?

### Stack

```
   CocoaPods : 0.38.2
        Ruby : ruby 2.0.0p481 (2014-05-08 revision 45883) [universal.x86_64-darwin14]
    RubyGems : 2.4.6
        Host : Mac OS X 10.10.5 (14F27)
       Xcode : 6.4 (6E35b)
         Git : git version 2.3.2 (Apple Git-55)
Ruby lib dir : /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib
Repositories : master - https://github.com/CocoaPods/Specs.git @ cb3d21caae2d3bbc729584b2818628a75155d259
```

### Plugins

```
cocoapods-plugins : 0.4.2
cocoapods-stats   : 0.5.3
cocoapods-trunk   : 0.6.4
cocoapods-try     : 0.4.5
```

### Podfile

```ruby

```

### Error

```
Errno::ENOTDIR - Not a directory - artsy.podspec
/Library/Ruby/Gems/2.0.0/gems/cocoapods-core-0.38.2/lib/cocoapods-core/source.rb:90:in `open'
/Library/Ruby/Gems/2.0.0/gems/cocoapods-core-0.38.2/lib/cocoapods-core/source.rb:90:in `entries'
/Library/Ruby/Gems/2.0.0/gems/cocoapods-core-0.38.2/lib/cocoapods-core/source.rb:90:in `pods'
/Library/Ruby/Gems/2.0.0/gems/cocoapods-core-0.38.2/lib/cocoapods-core/source/health_reporter.rb:53:in `analyze'
[...]

PR #4442, opened 2015-10-24
Improve repo lint error messaging

manuyavuz:
Solves #4142.

When the example command `pod repo lint artsy.podspec` runs, it will print the following message:

```
% poddev repo lint artsy.podspec

Linting spec repo `artsy.podspec`

[!] Unable to find a source named: `artsy.podspec`
```

I did not add another raise check, instead rely on the checks performed in `Source` class. Also used `SourceManager`'s methods to remove duplicate code.


segiddins:
Can we get a spec in for the bug that's fixed, and a changelog entry? Otherwise, 👍🏻

segiddins:
:+1:

PR #4443, opened 2015-10-24
Improve message when there is no spec in repos for a given dependency

manuyavuz:
This solves #4430.

Currently, when there is no existing specification inside any of the spec repos with given requirements, user would get the following error:

```
[!] Unable to satisfy the following requirements:
- `Lib (= 4.0.0)` required by `Podfile`
```

Now, user will get sth like this:

```
[!] Unable to find a specification for the following requirement:

	'Lib', '= 4.0.0'
```

I think this is enough for user to understand that their repo either not up-to-date or there is really no such specification with given dependency.

I actually wanted to use `Molinillo::NoSuchDependencyError` rather manually adding another message, but it's message does not contain name of the dependency, just it's version. 

I'm open to better management suggestions.

### DIFF
--- a/lib/cocoapods/resolver.rb
+++ b/lib/cocoapods/resolver.rb
@@ -400,7 +400,13 @@ module Pod
             message << "\nYou should explicitly specify the version in order to install a pre-release version"
           elsif !conflict.existing
             conflict.requirements.values.flatten.each do |r|
-              unless search_for(r).empty?
+              if search_for(r).empty?
+                # There is no existing specification inside any of the spec repos with given requirements.
+                message << "\n\nNone of the spec sources contain a spec satisfying the `#{r}` dependency." \
+                  "\nYou have either; mistyped the name or version," \
+                  ' not added the source repo that hosts the Podspec to your Podfile,' \
+                  ' or not got the latest versions of your source repos.'
+              else
                 message << "\n\nSpecs satisfying the `#{r}` dependency were found, " \
                   'but they required a higher minimum deployment target.'
               end
--- a/spec/unit/resolver_spec.rb
+++ b/spec/unit/resolver_spec.rb
@@ -405,6 +405,10 @@ module Pod
         e = lambda { resolver.resolve }.should.raise Informative
         e.message.should.match(/Unable to satisfy the following requirements/)
         e.message.should.match(/`AFNetworking \(= 3.0.1\)` required by `Podfile`/)
+        e.message.should.match(/None of the spec sources contain a spec satisfying the `AFNetworking \(= 3.0.1\)` dependency./)
+        e.message.should.match(/You have either; mistyped the name or version,/)
+        e.message.should.match(/ not added the source repo that hosts the Podspec to your Podfile,/)
+        e.message.should.match(/ or not got the latest versions of your source repos./)
       end
 
       it 'takes into account locked dependencies' do
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
+* Improve message when there is no spec in repos for dependency set in Podfile.  
+  [Muhammed Yavuz Nuzumlalı](https://github.com/manuyavuz)
+  [#4430](https://github.com/CocoaPods/CocoaPods/issues/4430)
+
 * Reduce the number of times the user's Xcode project is opened, speeding up
   installation.  
   [Samuel Giddins](https://github.com/segiddins)


lib/cocoapods/resolver.rb
```diff
@@ -403,6 +403,10 @@ def handle_resolver_error(error)
               unless search_for(r).empty?
```
segiddins: this needs to be an `if` now, since there's an `else`

segiddins:
hmmm, so there already should be a no such dependency error being thrown in this case? Do you think you can cook up a spec so I can have a look at what's going on?

manuyavuz:
Actually flow is the following when requested spec does not exist in any repo:

- `Resolver#resolve` method calls `Molinillo::Resolver#resolve` method.
- `resolve` calls `process_topmost_state`
- For this specific case, possibilities parameter is empty. 
- Because there is no `possibility`, code creates conflict using `create_conflict`

On `handle_resolve_error`, it falls into last elsif case, with `search_for` method returning empty array because there is no such spec. This is where I added the corresponding message.

I think it's better to throw a NoSuchDependencyError and improve messaging of that error, because it feels like the ideal match for that case. 
Maybe we can check if possibilities array is empty as a first step in Molinillo::Resolver#resolve method, and then throw a NoSuchDependencyError.

I modified the spec testing `raises if no such version of a dependency exists` case, and verified id new error message is actually called.

What type of spec do you think is needed?

segiddins:
Just a spec I can use to reproduce the issue, it sounds like a bug in molinillo to me

manuyavuz:
Umm ok I'll look at it

manuyavuz:
I added a failing test scenario but it's not exactly the same with the original scenario.

I think original scenario include this flow:

- New Podfile and Podfile.lock files which include a version update on a dependency are pulled from project repository.
- User performs `pod install --no-repo-update`
- Because repo is not up-to-date, CP can't find specific version, but fail with an ambiguous message. Message also includes `required by 'Podfile.lock'`because Podfile.lock also include the new version.

In current scenario I'm giving an empty locked dependencies array. To generate the exact scenario, we should find a way to give a sample Dependency array.

orta:
```
[!] Unable to find a specification for the following requirement:

    'Lib', '= 4.0.0'
```

Is awesome, great work @manuyavuz

spec/unit/resolver_spec.rb
```diff
@@ -396,15 +396,25 @@ module Pod
...
+
+      it 'test which should raise with no such dependency error' do
         podfile = Podfile.new do
           platform :ios
           pod 'AFNetworking', '3.0.1'
         end
         resolver = Resolver.new(config.sandbox, podfile, empty_graph, SourcesManager.all)
         e = lambda { resolver.resolve }.should.raise Informative
-        e.message.should.match(/Unable to satisfy the following requirements/)
-        e.message.should.match(/`AFNetworking \(= 3.0.1\)` required by `Podfile`/)
```
segiddins: it still should say where the failing requirement came from

manuyavuz: Current implementation actually satisfies the above two lines. What it should do is to satisfy the line below. I didn't add a check about the failing pod because this spec was to show the erroneous case. Correct spec will be the commented out lines.

segiddins:
Hmmm, why not just take the existing exception message and `gsub` it? That way we keep the metadata intact?

manuyavuz:
Sorry I think I don't get it. Do you mean changing the default message of VersionConflict class (`Unable to satisfy the following requirements bla bla`)?

segiddins:
Nope, just checking if the conflict is due to there being no possibilities, and if so raising an informative with `version_conflict.message.gsub(/unable blah blah/, 'other blah blah')`

manuyavuz:
Should we do this check inside Resolution#unwind_for_conflict where VersionConflict is raised, or inside Resolver#handle_resolver_error method which is my first proposal (changing message 
`if search_for(r).empty?`)?

What do you think?

segiddins:
No no, I think it should be handled in CocoaPods's resolver, not Molinillo

manuyavuz:
I've decided to give an additional information about the possible causes rather than writing the message from scratch. Error message will be like this:
```
[!] Unable to satisfy the following requirements:

- `Lib (= 4.0.0)` required by `Podfile`

None of the spec sources contain a spec satisfying the `Lib (= 4.0.0)` dependency.
Either your source repositories are not up-to-date, or there is really no such specification at all.
```

This is similar to error message structure we gave about minimum deployment target, so i think this way would be better. Any suggestions on wording?

orta:
Perhaps:

```
[!] Unable to satisfy the following requirements:

- `Lib (= 4.0.0)` required by `Podfile`

None of the spec sources contain a spec satisfying the `Lib (= 4.0.0)` dependency.
You have either; mistyped the name, not added the source repo that hosts the Podspec or do not have the latest versions of your source repos.
```

manuyavuz:
`You have either; mistyped the name or version, not added the source repo that hosts the Podspec to your Podfile, or not got the latest versions of your source repos.`

Error could be cause by mistyped version also. It's better to specify where user should add source (Podfile). Not sure for the change in the last part; my change felt more grammatically correct to me, but as an alien I may miss something.

segiddins:
👍🏻

manuyavuz:
I think it's ready to ⛵️ now!

orta:
awesssome

manuyavuz:
Thanks for help guys!

Issue #4420, opened 2015-10-21
XCode 7.1 - Include of non-modular header inside framework


solomon23:
I have a project that was building fine that suddenly stopped building in xcode 7.1

The project has a development pod and also includes the GPUImage pod.  The development pod used to be able to reference GPUImage via:

````
#import <GPUImage/GPUImage.h>
````

But that now gives a compile error of "Include of non-modular header inside framework module"

I've tried setting the flag: 

````
CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES = 'YES'
````

but that makes no difference.

I tried to import GPUImage via:

````
@import GPUImage;
````

but that won't compile.  Is this a know issue ?

segiddins:
Could you please provide a project that demonstrates this issue? Thanks!

neonichu:
There was some mention of this in the beta release notes, see [this comment](https://github.com/CocoaPods/CocoaPods/issues/4056#issuecomment-143187681) — not sure if still applicable.

solomon23:
Here's a simple project that repros this:

https://www.dropbox.com/s/hlidogntavn8cvq/BadPod.zip?dl=0

It looks like it doesn't error out until you mix swift and objc in the same development module.

solomon23:
@neonichu looks like "Always Search User Paths"  is defaulting to 'no' now.  That sample project has it set to 'no' for every target and still has the issue.

neonichu:
Oh, the issue with the auto-generated `*-Swift.h` headers *is* indeed a known issue, see #3767

solomon23:
Is that the same issue ? This has been working fine for us in multiple projects until 7.1 was installed this morning :/

solomon23:
@neonichu doesn't seem like the same issue after reading through the thread.  Sounds like they're trying to reference swift code from ObjC inside a pod.  That's not what I'm trying to do.  I'm trying to import the header from one pod into another pod all in ObjC.  I loaded up XCode 6.4 and confirmed it builds the given sample project fine.  Something happened with how the projects are constructed and XCode 7.1

neonichu:
Yep, seems to be a different issue. Not sure what exactly changed with Xcode 7.1, but using a proper module import `@import GPUImage;` makes it work fine.

solomon23:
[...]

PR #4453, opened 2015-10-27
[PodTarget] Infer umbrella header

segiddins:
Might help with https://github.com/CocoaPods/CocoaPods/issues/4420?
Closes https://github.com/CocoaPods/CocoaPods/issues/3767

### DIFF
--- a/lib/cocoapods/target.rb
+++ b/lib/cocoapods/target.rb
@@ -140,14 +140,14 @@ module Pod
     #         module map.
     #
     def umbrella_header_path
-      support_files_dir + "#{label}-umbrella.h"
+      support_files_dir + "#{product_module_name}.h"
     end
 
     # @return [Pathname] the absolute path of the LLVM module map file that
     #         defines the module structure for the compiler.
     #
     def module_map_path
-      support_files_dir + "#{label}.modulemap"
+      support_files_dir + "#{product_module_name}.modulemap"
     end
 
     # @return [Pathname] the absolute path of the prefix header file.
--- a/lib/cocoapods/installer/target_installer/pod_target_installer.rb
+++ b/lib/cocoapods/installer/target_installer/pod_target_installer.rb
@@ -270,7 +270,8 @@ module Pod
       end
 
       def create_umbrella_header
-        return super unless custom_module_map
+        inferred_umbrella_header = target.build_headers.root + "#{target.product_module_name}/#{target.product_module_name}.h"
+        return super unless custom_module_map || inferred_umbrella_header.file?
       end
 
       def custom_module_map


segiddins:
Note that, with this change, spec authors will have to make their umbrella headers conditional if they have subspecs (i.e. checking `#if __has_include`)

segiddins:
(The other alternative might be to append imports of all public headers to the existing header file?)

segiddins:
@CocoaPods/core what are the thoughts on this one?

AliSoftware:
This breaks `OHHTTPStubs`: `pod lib lint` returns errors, while it passes with the current 0.39).

```
 -> OHHTTPStubs (4.5.0)
    - WARN  | [OHHTTPStubs/Default,OHHTTPStubs/JSON] xcodebuild:  <module-includes>:1:1: warning: umbrella header for module 'OHHTTPStubs' does not include header 'OHHTTPStubsResponse+JSON.h' [-Wincomplete-umbrella]
    - WARN  | [OHHTTPStubs/Default] xcodebuild:  <module-includes>:1:1: warning: umbrella header for module 'OHHTTPStubs' does not include header 'OHPathHelpers.h' [-Wincomplete-umbrella]
    - ERROR | [OHHTTPStubs/Default, OHHTTPStubs/Core, OHHTTPStubs/NSURLSession, and more...] xcodebuild: Returned an unsuccessful exit code.
    - WARN  | [OHHTTPStubs/HTTPMessage] xcodebuild:  <module-includes>:1:1: warning: umbrella header for module 'OHHTTPStubs' does not include header 'OHHTTPStubsResponse+HTTPMessage.h' [-Wincomplete-umbrella]
    - WARN  | [OHHTTPStubs/Mocktail] xcodebuild:  <module-includes>:1:1: warning: umbrella header for module 'OHHTTPStubs' does not include header 'OHHTTPStubs+Mocktail.h' [-Wincomplete-umbrella]

[!] OHHTTPStubs did not pass validation, due to 1 error and 4 warnings.
```
[Full log of `pod lib lint --verbose` here](https://github.com/CocoaPods/CocoaPods/files/24832/lint-OHHTTPStubs-infer-umbrella-header.txt)

---

Will try to investigate later, but have to go and will be busy until 17pm Paris time. Maybe I'm just missing some `__has_include` stuff at some point (I had them but not sure I kept them when migrating to a custom umbrella header), will have to double-check that.

Note that `OHHTTPStubs`:
* Has subspecs
* Has a custom umbrella header and module map
* Is iOS/watchOS/tvOS/OSX and has been successfully tested to work on Xcode 7.1 before.
* Umbrella header issue was recently fixed via AliSoftware/OHHTTPStubs#127

lib/cocoapods/target.rb
```diff
@@ -140,14 +140,14 @@ def xcconfig_path(variant = nil)
     #         module map.
     #
     def umbrella_header_path
-      support_files_dir + "#{label}-umbrella.h"
+      support_files_dir + "#{product_module_name}.h"
```
AliSoftware: Won't this change only work if the umbrella header is the inferred one, but break if the module map has a custom umbrella header instead (like in `OHHTTPStubs` where `OHHTTPStubs.h` exists but **isn't** the umbrella header)?

AliSoftware:
I think this change will do more harm than solve stuff, because:

* People will have to add `__has_include` everywhere in their umbrella headers, whereas they didn't have to care before as CocoaPods did it right
* This inferred umbrella header can guess wrong (like in the case of `OHHTTPStubs` if the header with the same name already exists but isn't the umbrella header (which is probably the most common reason why people would do a custom module map in the first place)

Wouldn't a `custom_module_map` DSL attribute help fix that? People wanting CocoaPods to use their own modulemap & umbrella header instead of letting CocoaPods generate its own could explicitly declare it in their `.podspec` that way, instead of guessing and risking to break things?

segiddins:
@AliSoftware if you have a custom module map, then CocoaPods already doesn't generate an umbrella header.

lib/cocoapods/target.rb
segiddins: this code path already isn't used when a custom module map is specified

AliSoftware:
Mmmh how come does it still work if a user only fetch some subspecs and not all, then?

Note that I didn't mention anything my `podspec` about that custom umbrella header, because:
* if people are using my lib thru anything other than CocoaPods, like Carthage or manual copy of the framework or whatnot, they'll have all the subspecs in that generated framework and my custom umbrella header will import them all
* if people are using my lib thru CocoaPods, I let it generate the umbrella header for me so that it only import the headers that are actually fetched by the subspecs selected in the Podfile

You're saying that there's a DLS attribute to specify a custom module map already, and that I should use it instead so that CocoaPods _never_ generate its own module map in my case? (And thus I'll need to add `__has_include` tests in that custom umbrella headers so it can work with subspecs then)?

segiddins:
I'm saying if you're using a custom module map, CocoaPods will not generate an umbrella header.

PR #4509, opened 2015-11-06
Use correct path for import in umbrella header when preserving header_mappings_dir

timbodeit:
#4267 preserves the directory structure inside the Pod's header_mappings_dir, if one is specified in the podspec.

This led to problems with the pod target's umbrella header, since the (now nested) path of the included header files needs to be taken into account when generating the import statements.

I would classify this rather as a bugfix than an enhancement. Should I create a Changelog entry for this, even though the Bug has not been released yet?


segiddins:
Looks good to me, but I'd like someone else on the team to confirm this is the correct fix. 
No need for a CHANGELOG entry

segiddins:
@mrackwitz can you take a quick look at this please?

mrackwitz:
Looks good to me! :+1:

PR #4512, opened 2015-11-06
Remove newline causing color problem when printing to a pager

manuyavuz:
Related to CocoaPods/cocoapods-search#17. Removes newline included while printing pod information in order to make coloring work in a pager.


segiddins:
Shouldn't we compensate for this by printing out a plain new line elsewhere?

manuyavuz:
Actually currently system puts a newline before and after printing each pod. This results in double lines between pods. I felt one line is enough, so just removed it. I can add if two lines would be better.

PR #4519, opened 2015-11-08
Add IO type setting to UI module

manuyavuz:
Related to CocoaPods/cocoapods-search#17

### DIFF
--- a/lib/cocoapods/user_interface.rb
+++ b/lib/cocoapods/user_interface.rb
@@ -20,6 +20,10 @@ module Pod
       attr_accessor :title_level
       attr_accessor :warnings
 
+      # @return [IO] IO object to which UI output will be directed.
+      #
+      attr_accessor :output_io
+
       # @return [Bool] Whether the wrapping of the strings to the width of the
       #         terminal should be disabled.
       #
@@ -322,7 +326,12 @@ module Pod
       #        The message to print.
       #
       def puts(message = '')
-        STDOUT.puts(message) unless config.silent?
+        return if config.silent?
+        begin
+          (output_io || STDOUT).puts(message)
+        rescue Errno::EPIPE
+          exit 0
+        end
       end
 
       # prints a message followed by a new line unless config is silent.
@@ -331,7 +340,12 @@ module Pod
       #        The message to print.
       #
       def print(message)
-        STDOUT.print(message) unless config.silent?
+        return if config.silent?
+        begin
+          (output_io || STDOUT).print(message)
+        rescue Errno::EPIPE
+          exit 0
+        end
       end
 
       # gets input from $stdin


segiddins:
Definitely needs tests, and I'm not sure that's the correct way to handle exceptions

manuyavuz:
Do you think this type of availability would be needed in future usages? If not, I think we can keep this in search code to keep general logic clean. 

If that would be valuable for future, I'll definitely add tests and improve exception cases, I needed an early PR to discuss things

segiddins:
Let's see what @Karla-Isabel-Sandoval thinks, since she was in the process of refactoring all the UI code

manuyavuz:
Curiously awaiting your contribution @Karla-Isabel-Sandoval, thanks!

manuyavuz:
I had been totally messed in previous commit :( It's now much more consistent, it changes implementations of `puts` and `print` methods included inside UI module.

We should still consider how to handle exceptions and add specs, however, adding such functionality here would be more consistent across the project. I see nearly every plugin somewhat edits implementation of UI, which makes it hard to follow code flow. Adding these things as options feel better to me.

manuyavuz:
How should I add tests for this?

Currently `UI` module's `puts`, `print` methods are overridden inside `spec_helper/user_interface.rb` file, so my code will not be run. Not sure of a convenient way of doing this.

lib/cocoapods/user_interface.rb
```diff
@@ -322,7 +323,16 @@ def choose_from_array(array, message)
...
       def puts(message = '')
-        STDOUT.puts(message) unless config.silent?
+        return if config.silent?
+        begin
+          if io_type
+            io_type.puts(message)
+          else
+            STDOUT.puts(message) unless config.silent?
+          end
+        rescue Errno::EPIPE
```
segiddins: Why doesn't this exit 0?

lib/cocoapods/user_interface.rb
```diff
@@ -322,7 +323,16 @@ def choose_from_array(array, message)
       #        The message to print.
       #
       def puts(message = '')
-        STDOUT.puts(message) unless config.silent?
+        return if config.silent?
+        begin
+          if io_type
+            io_type.puts(message)
+          else
+            STDOUT.puts(message) unless config.silent?
```
segiddins: Redundant.

lib/cocoapods/user_interface.rb
manuyavuz: It should definitely be 0.

manuyavuz:
Code is cleaner now. But still need suggestions on how to test ;)

lib/cocoapods/user_interface.rb
```diff
@@ -331,13 +337,18 @@ def puts(message = '')
...
+        rescue Errno::EPIPE
+          exit 0
+        end
       end
 
       # gets input from $stdin
       #
       def gets
-        $stdin.gets
+        (io || $stdin).gets
```
segiddins: hmm, since most pipes used here won't be bidirectional, let's keep this as `stdin` for now?

manuyavuz: Makes sense

Gemfile.lock
```diff
@@ -7,7 +7,7 @@ GIT
 
 GIT
   remote: https://github.com/CocoaPods/Core.git
-  revision: ce26e1eb797a6ad1f1d94b0304a50c491bc41237
+  revision: 6f85c4fcff81960c0983616082bc345763a15dc6
```
segiddins: are all these changes necessary?

lib/cocoapods/user_interface.rb
```diff
@@ -19,6 +19,7 @@ class << self
       attr_accessor :indentation_level
       attr_accessor :title_level
       attr_accessor :warnings
+      attr_accessor :io
```
segiddins: rename this to `output` and I think we're good to go, unless @Karla-Isabel-Sandoval wants to give any feedback

manuyavuz: Regarding we only use it for output, seems reasonable, but I want to make sure it expects an IO object. I guess I can add type as documentation.

manuyavuz:
@segiddins I think we can't rename it to `output` because `@output` variable is used in almost every `spec_helper.rb` files in which `UI#puts`, `UI#print`, `UI#warn` methods are reimplemented. What about `output_io`?

Gemfile.lock
manuyavuz: No, reverted it to master version.

segiddins:
`output_io` works for me

manuyavuz:
@segiddins I'll merge this after CI passes then.

PR #4527, opened 2015-11-10
Prevent installer to be run from inside sandbox directory

manuyavuz:
This prevents crashes which can happen when `pod install` command is run from inside subdirectories of Pods directory.

### DIFF
--- a/lib/cocoapods/installer.rb
+++ b/lib/cocoapods/installer.rb
@@ -115,9 +115,9 @@ module Pod
 
     def prepare
       # Raise if pwd is inside Pods
-      if (Pathname.pwd <=> sandbox.root) >= 0
+      if Dir.pwd.start_with?(sandbox.root.to_path)
         message = 'Command should be run from a directory outside Pods directory.'
-        message << "\n\n\tCurrent directory is '#{Pathname.pwd}'\n"
+        message << "\n\n\tCurrent directory is #{UI.path(Pathname.pwd)}\n"
         raise Informative, message
       end
       UI.message 'Preparing' do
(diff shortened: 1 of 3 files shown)


lib/cocoapods/installer.rb
```diff
@@ -114,6 +114,12 @@ def install!
     end
 
     def prepare
+      # Raise if pwd is inside Pods
+      if (Pathname.pwd <=> sandbox.root) >= 0
+        message = 'Command should be run from a directory outside Pods directory.'
+        message << "\n\n\tCurrent directory is '#{Pathname.pwd}'\n"
```
segiddins: Use `UI.path` here

lib/cocoapods/installer.rb
```diff
@@ -114,6 +114,12 @@ def install!
     end
 
     def prepare
+      # Raise if pwd is inside Pods
+      if (Pathname.pwd <=> sandbox.root) >= 0
```
segiddins: Is this guaranteed to work? I'm not sure pathname's comparison operator is that clever

manuyavuz: You're right, I got it totally wrong. What about this:

```ruby
relative_path = Pathname.pwd.relative_path_from(sandbox.root)
unless relative_path.each_filename.first == '..'
  #raise
end
```

segiddins:
I'd recommend doing `.expand_path.to_path` on each and checking if one is the prefix of the other

manuyavuz:
Used prefix comparison, it feels safer now.

segiddins:
👍🏻

PR #4544, opened 2015-11-14
Ensure all user projects are properly integrated

segiddins:
Depends upon https://github.com/CocoaPods/Xcodeproj/pull/340.

- [x] CHANGELOG
- [x] Specs
  - [x] Rebuild integration specs
  - [x] XCConfig Integrator handling old CocoaPods base configuration paths
  - [x] Not saving user project if unnecessary
  - [x] Proper deintegration taking place
  - [x] Major version change
- [x] Documentation

\c @manuyavuz


lib/cocoapods/installer/user_project_integrator/target_integrator.rb
```diff
@@ -259,9 +225,9 @@ def add_check_manifest_lock_script_phase
           end
         end
 
-        private
+        public
 
-        # @!group Private helpers
+        # @!group Helpers
```
segiddins: can avoid making these public

manuyavuz:
My first test trials failed for a use case scenario, but I'm not sure where it came from, or even if I made a wrong setup. Will update after a comprehensive local test.

lib/cocoapods/installer/user_project_integrator.rb
```diff
@@ -112,8 +115,38 @@ def create_workspace
...
-          TargetIntegrator.new(target).integrate!
+        target_integrators = targets_to_integrate.sort_by(&:name).map do |target|
+          TargetIntegrator.new(target)
+        end
+
+        Config.instance.with_changes(:silent => true) do
+          deintegrator = Deintegrator.new
+          all_project_targets = target_integrators.map(&:user_project).flat_map(&:native_targets).uniq
+          all_native_targets = target_integrators.flat_map(&:native_targets).uniq
+          targets_to_deintegrate = all_project_targets - all_native_targets
```
manuyavuz: @segiddins I think this case does not handle the following podfile transition from :

```
pod 'MBProgressHUD'
```
to :

```
target 'test' do
	pod 'MBProgressHUD'
end
```

Because both two cases integrates with the user project target _`test`_, both `all_project_targets` and `all_native_targets` include `test` target. However, we should still deintegrate this target because in the first podfile `libPods.a` library had been linked to it, but in the second case `libPods-test.a` library will be linked, and we have to get rid of `libPods.a`.

segiddins: not too worried about that since the link with first target thing is going away in the new DSL

manuyavuz: OK with that because it will be fixed with new DSL. 
However, we should handle the following transition from:

```
target 'test' do
    pod 'MBProgressHUD'
end
```
to:
```
target 'test' do

end
```
In this case, `all_native_targets` still contains `test` target because there is definition in podfile, but no entries generated inside `Target Support Files` for the target. Therefore, build fails.

Therefore we should also deintegrate targets in podfile with no pods.

manuyavuz: Wrong alarm, this is working for the above case also :)

manuyavuz:
Looks great :+1:  Only thing remains is removal of not existing xcconfig file references from user project.

manuyavuz:
@segiddins I think I handled xcconfig deintegration with CocoaPods/cocoapods-deintegrate#15. We can focus on specs now.

manuyavuz:
@segiddins, seems like there is a problem with printing of `pod deintegrate` command. It waits jobs to be done and prints all output at once. But using released deintegrate plugin prints output incrementally. Do you experience the same?

orta:
ace

segiddins:
@manuyavuz I can't reproduce that? It seems to stream output when I run `dpod deintegrate`

segiddins:
Now, deintegrate will be run when the major version changes -- including deleting the Pods directory -- whenever the major version changes. This means we can both delete the migration code once we bump the version to 1.x.x and be sure that everything will be fully up-to-date. This will mean a 1-time redownload of everything and re-inetgration of the user project, but I'm personally OK with that.

orta:
> This will mean a 1-time redownload of everything and re-inetgration of the user project, but I'm personally OK with that.

Me too,

manuyavuz:
This is a very acceptable tradeoff :+1:

segiddins:
I think all I still need to do on this is add specs?

segiddins:
Time to :ship: this, I think

orta:
Woo!

CHANGELOG.md
```diff
@@ -89,6 +89,18 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Samuel Giddins](https://github.com/segiddins)
   [#3347](https://github.com/CocoaPods/CocoaPods/issues/3347)
 
+* Targets that are no longer integrated will CocoaPods will be properly
```
manuyavuz: integrated **with** CocoaPods

spec/unit/installer/user_project_integrator/target_integrator/xcconfig_integrator_spec.rb
```diff
@@ -105,5 +105,20 @@ module Pod
 
       UI.warnings.should.not.match /not set.*base configuration/
     end
+
+    it 'handles when xcconfig is set to another sandox xcconfig' do
```
manuyavuz: another **sandox** xcconfig

segiddins: already fixed :)

spec/unit/installer_spec.rb
```diff
@@ -177,6 +177,38 @@ def @installer.analyze(*)
...
+          should_not_deintegrate = %w(1.0.0 1.0.1 1.1.0 1.2.2)
+          should_not_deintegrate.each do |version|
+            lockfile = generate_lockfile(:lockfile_version => version)
+            installer = Pod::Installer.new(config.sandbox, generate_podfile, lockfile)
+            Pathname.expects(:glob).never
+            installer.send(:deintegrate_if_different_major_version)
+          end
+        end
+
+        it 'does deintegrate when the major version is the same' do
```
manuyavuz: major version is **different**

spec/unit/installer/user_project_integrator/target_integrator/xcconfig_integrator_spec.rb
manuyavuz: You pushed faster than I read 😁

manuyavuz:
Sweet! :shipit:

PR #4560, opened 2015-11-19
[User Interface] add method to pipe outputs to a pager

manuyavuz:
Do we need a CHANGELOG for it?

### DIFF
--- a/lib/cocoapods/user_interface.rb
+++ b/lib/cocoapods/user_interface.rb
@@ -368,6 +368,16 @@ module Pod
         warnings << { :message => message, :actions => actions, :verbose_only => verbose_only }
       end
 
+      # Pipes all output inside given block to a pager.
+      #
+      def with_pager
+        IO.popen((ENV['PAGER'] || 'less -R'), 'w') do |io|
+          Signal.trap('INT','IGNORE')
+          UI.output_io = io
+          yield
+        end
+      end
+
       private
 
       # @!group Helpers
--- a/lib/cocoapods/user_interface.rb
+++ b/lib/cocoapods/user_interface.rb
@@ -370,12 +370,15 @@ module Pod
 
       # Pipes all output inside given block to a pager.
       #
+      # @yield Code block in which inputs to {#puts} and {#print} methods will be printed to the piper.
+      #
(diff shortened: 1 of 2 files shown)


segiddins:
No changelog, yes a test 👍🏻

manuyavuz:
Will look for a test :+1: Do you know a way of showing need to the block in documentation? I think it should be more descriptive, but couldn't find a convenient way.

segiddins:
`@yields` I think

manuyavuz:
Spec added. 🚀?

lib/cocoapods/user_interface.rb
```diff
@@ -368,6 +368,19 @@ def warn(message, actions = [], verbose_only = false)
         warnings << { :message => message, :actions => actions, :verbose_only => verbose_only }
       end
 
+      # Pipes all output inside given block to a pager.
+      #
+      # @yield Code block in which inputs to {#puts} and {#print} methods will be printed to the piper.
+      #
+      def with_pager
+        IO.popen((ENV['PAGER'] || 'less -R'), 'w') do |io|
+          Signal.trap('INT', 'IGNORE')
```
segiddins: Do we have to do anything to reset this?

lib/cocoapods/user_interface.rb
```diff
@@ -368,6 +368,19 @@ def warn(message, actions = [], verbose_only = false)
...
+          UI.output_io = io
+          yield
+        end
+        UI.output_io = nil
```
segiddins: Should be in `ensure`

lib/cocoapods/user_interface.rb
manuyavuz: Yeah, good catch :+1:

lib/cocoapods/user_interface.rb
manuyavuz: Should we raise an informative in case of rescue?

segiddins: Nah, just putting this line inside `ensure` is sufficient

segiddins:
:+1:

PR #4564, opened 2015-11-20
`pod spec create` to use `s.version.to_s` as the git tag when possible

Ashton-W:
Changes to make `pod spec create` produce *podspecs* with `s.version.to_s` as the git tag when possible

```ruby
s.source = { :git => 'https://github.com/Ashton-W/CRToast.git', :tag => s.version.to_s }
```

If there are no tags or the tags contain extra characters (eg: `v1.4`) a string with the tag is used (current behaviour).

### DIFF
--- a/lib/cocoapods/command/spec/create.rb
+++ b/lib/cocoapods/command/spec/create.rb
@@ -60,7 +60,7 @@ module Pod
           data[:author_email]  = `git config --get user.email`.strip
           data[:source_url]    = "http://EXAMPLE/#{name}.git"
           data[:ref_type]      = ':tag'
-          data[:ref]           = 's.version.to_s'
+          data[:ref]           = '#{s.version}'
           data
         end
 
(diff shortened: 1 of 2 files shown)


segiddins:
Why is this preferable?

Ashton-W:
Common (best?) practise is to avoid duplicating the version string in podspecs, and instead refer to that string in the tag. When manually updating a podspec, it's one less thing to update.

segiddins:
Ah. I've always used `"v#{s.version}"`

Ashton-W:
Right, that works if you are tags are prefixed with `v`.
We could change it to use `"#{s.version}"`, that might be easier to understand and to modify.

I didn't handle the case of `s.version == "v#{s.version}"`, but that could be added.

segiddins:
I prefer the string interpolation route personally. Would also be awesome to add a test for this

Ashton-W:
Rubocop didn't like me writing `%q(#{s.version})` instead of `'#{s.version}'`, even though I think the former is clearer in it's intent to not interoplate variables in the string. That's why the latter is used.

lib/cocoapods/command/spec/create.rb
```diff
@@ -60,7 +60,7 @@ def default_data_for_template(name)
           data[:author_email]  = `git config --get user.email`.strip
           data[:source_url]    = "http://EXAMPLE/#{name}.git"
           data[:ref_type]      = ':tag'
-          data[:ref]           = '0.0.1'
+          data[:ref]           = '#{s.version}'
```
alloy: This does not work as expected. Single quoted strings do not interpolate.

Ashton-W: It is intentional (see last comment in PR). The non interpolated string will be used in the generated podspec.

segiddins:
:+1:

orta:
Cool

PR #4722, opened 2016-01-04
Don't warn about setting base config when existing is identical to new one

jedlewison:
Issue here is the scenario where a project gets added to a new workspace temporarily and has its pods built in a new directory. In that case, the location of the newly generated config file may differ from the existing one, but the config is actually identical. So when the config files are the same, it seems reasonable to silence the ominous "CocoaPods may not work at all" message since it definitely works.

If there's a different or better way to handle this edge case, that'd be great too — I'm not experienced with Ruby so I easily could have missed something else.

### DIFF
--- a/lib/cocoapods/installer/user_project_integrator/target_integrator/xcconfig_integrator.rb
+++ b/lib/cocoapods/installer/user_project_integrator/target_integrator/xcconfig_integrator.rb
@@ -88,12 +88,14 @@ module Pod
               if existing.real_path.to_path.start_with?(pod_bundle.sandbox.root.to_path << '/')
                 set_base_configuration_reference.call
               elsif !xcconfig_includes_target_xcconfig?(config.base_configuration_reference, path)
-                UI.warn 'CocoaPods did not set the base configuration of your ' \
-                'project because your project already has a custom ' \
-                'config set. In order for CocoaPods integration to work at ' \
-                'all, please either set the base configurations of the target ' \
-                "`#{target.name}` to `#{path}` or include the `#{path}` in your " \
-                "build configuration (#{UI.path(existing.real_path)})."
+                unless existing_config_is_identical_to_pod_config?(existing.real_path, pod_bundle.xcconfig_path(config.name))
+                  UI.warn 'CocoaPods did not set the base configuration of your ' \
+                  'project because your project already has a custom ' \
+                  'config set. In order for CocoaPods integration to work at ' \
+                  'all, please either set the base configurations of the target ' \
+                  "`#{target.name}` to `#{path}` or include the `#{path}` in your " \
+                  "build configuration (#{UI.path(existing.real_path)})."
+                end
               end
             elsif config.base_configuration_reference.nil? || file_ref.nil?
               set_base_configuration_reference.call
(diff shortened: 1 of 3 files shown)


orta:
This seems like a good idea to me :+1:

segiddins:
Can you add a spec and a CHANGELOG entry for this?

jedlewison:
@segiddins Yep. Dumb question tho -- what do I need to do to make a CHANGELOG entry? Added a spec and refactored to make things a little clearer -- just running specs now and getting back up to date with master before pushing to my PR.

orta:
Not a dumb question!

So edit this file: https://github.com/CocoaPods/CocoaPods/blob/master/CHANGELOG.md

You need to add a `h1` of `Master` as this is the first PR post 1.0 e.g.:

```
## Master

##### Enhancement

* Don't show the thingy when the thing.
  [me](https://github.com/ok)

```

lib/cocoapods/installer/user_project_integrator/target_integrator/xcconfig_integrator.rb
```diff
@@ -88,12 +88,14 @@ def self.set_target_xcconfig(pod_bundle, target, config)
               if existing.real_path.to_path.start_with?(pod_bundle.sandbox.root.to_path << '/')
                 set_base_configuration_reference.call
               elsif !xcconfig_includes_target_xcconfig?(config.base_configuration_reference, path)
-                UI.warn 'CocoaPods did not set the base configuration of your ' \
-                'project because your project already has a custom ' \
-                'config set. In order for CocoaPods integration to work at ' \
-                'all, please either set the base configurations of the target ' \
-                "`#{target.name}` to `#{path}` or include the `#{path}` in your " \
-                "build configuration (#{UI.path(existing.real_path)})."
+                unless File.exist?(path) && File.exist?(existing.real_path) && !FileUtils.compare_file(path, existing.real_path)
```
segiddins: these are  'Pathname` objects, so do `path.file?`, `existing.real_path.file?`

jedlewison: Thanks, will do.

jedlewison:
Specs question -- Unit tests are passing, but there are 44 failures integration tests. Is that likely because I'm not cleaning up something properly after a test? Or is it possible that the code change could break integration but not unit tests? Or perhaps something else entirely? Example error message is "Bacon::Error: File comparison error `Pods/Pods.xcodeproj` for update_selected"; ends with "775 specifications (1502 requirements), 44 failures, 0 errors"

segiddins:
@jedlewison lets see what happens on CI -- it's possible you didn't `git submodule update` ? Also, you'll need to rebase, since we don't want the merge master commit to get back into master.

jedlewison:
@segiddins Yeah, I may open a new PR unless I can change my source -- haven't done much with github PRs. I just merged, which triggered a new CI, but that has no new changes from me.

segiddins:
`git fetch && git rebase origin/master` should do it for you!

orta:
and, alas, again :P

segiddins:
I can merge this manually, don't worry!

jedlewison:
Haha cool -- no worries either way. Thanks guys!

segiddins:
Thanks for the contribution, @jedlewison !

jedlewison:
You're quiet welcome. And thank you both for the help with this and also for CocoaPods in general!

PR #4734, opened 2016-01-06
Fix for resource bundles TARGETED_DEVICE_FAMILY issue

arifken:
We found that the TARGETED_DEVICE_FAMILY is not set correctly for resource bundles configured in a podspec. The application using the resource bundle from the pod did not compile assets for iPad because this value was defaulted to "1" (iPhone only). Now, it correctly configures the pod resource bundle target to include both iPhone and iPad


segiddins:
Manually merged via https://github.com/CocoaPods/CocoaPods/commit/ddbffba9e7c0b1e630e4a444ab5326059a3e08ef, thanks!

arifken:
thanks!

Issue #4762, opened 2016-01-13
resource_bundle is broken in 1.0.0.beta.2


djbe:
Any pod in my Podfile that uses the resource_bundles property is broken: the copy bundle resources step for each corresponding bundle target contains no items. If I rollback to 0.39.0, everything works as it is supposed to.

Example pod with (now) broken resource_bundles step:
https://github.com/TimOliver/TOWebViewController/

Also happens in a private podspec with the following bundles:

	s.resource_bundles = {
		'Appwise-iOS' => ['Appwise-iOS/Resources/*.lproj']
	}

efirestone:
Looks like this is related to my change in https://github.com/CocoaPods/CocoaPods/pull/4635

Looking into it now.

djbe:
Beta 3 fixes this for me.

Thx @efirestone!

danielgalasko:
Can anyone confirm if this is working in beta 6? One of my dependencies `pod NYTPhotoViewer` is now failing for the same error

    error: /Path/To/Build/Products/Debug-iphonesimulator/NYTPhotoViewer/NYTPhotoViewer.bundle: No such file or directory

danielgalasko:
In fact I just checked with beta 3 now and its still broken

PR #4765, opened 2016-01-14
Fix bug where lproj directories aren't added to project

efirestone:
https://github.com/CocoaPods/CocoaPods/issues/4762

Root of the issue was that the resources were being added to the xcodeproj in all-lowercase, and when the target installer went to look for items using properly-cased paths they didn't match, so it didn't add them to the targets.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
+* Properly add resource files to resources build phase.
+  [Eric Firestone](https://github.com/efirestone)
+  [#4762](https://github.com/CocoaPods/CocoaPods/issues/4762)
+
 * Fix suggestion of sudo when it actually isn't needed.  
   [Marcel Jackwerth](https://github.com/sirlantis)
 
--- a/lib/cocoapods/installer/file_references_installer.rb
+++ b/lib/cocoapods/installer/file_references_installer.rb
@@ -199,29 +199,29 @@ module Pod
         lproj_paths = Set.new
         lproj_paths_with_files = Set.new
         allowable_paths = paths.select do |path|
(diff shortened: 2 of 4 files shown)


efirestone:
@segiddins FYI. Sorry about introducing this in the first place.

CHANGELOG.md
```diff
@@ -37,6 +37,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
+* Properly add resource files to resources build phase.
```
segiddins: needs to go in a `Master` section

segiddins:
Could you add a test for different cases?

efirestone:
I added checks to (case-sensitively) look at the path as well as the fact that the file ref exists, which checks that this won't regress. I also added a check that the build resources target includes `en.lproj`. Both of these would have caught this issue.

CHANGELOG.md
efirestone: Good catch, fixed.

segiddins:
Great, thanks!

nkibler:
When can we expect this to be released? This is blocking us from moving off 0.39.0, which is causing us issues that #4635 solves.

cnoon:
Ping here as well. This issue is blocking us from being able to switch over to test with the beta releases.

segiddins:
https://guides.cocoapods.org/using/unreleased-features.html

cnoon:
Thanks for your quick reply @segiddins...much appreciated. Unfortunately for my current situation, it's not realistic for me to ask 20+ engineers to switch over to the unreleased version. I merely wanted to call attention to the fact that this particular bug is a blocker for us. I figured the more info coming your way about the betas, the better.

We have no expectations here, I realize how hectic scheduling big releases can be. We greatly appreciate all the hard work your fantastic team does every day! 🍻

PR #4809, opened 2016-01-23
Handle CoreData mapping model resources

efirestone:
When using recursive resource globs, CoreData mapping models should not have their child items added to the project directly.


efirestone:
This was the last resource that's specially handled by the copy-resources build phase script, so I believe it's the only one that was missing.

segiddins:
:+1: with a rebase

segiddins:
Thanks so much @efirestone !

efirestone:
Happy to help!

PR #4820, opened 2016-01-26
[PodTargetInstaller] Fix #4819

### LINKED ISSUE #4819: Project headers are copied into build directory when header_mapping_dir is set
If a target is integrated with frameworks and the podspec specifies a value for `header_mapping_dir`, project headers are copied into the build directory instead of an expected framework header path. This is because the build phase created by the `PodTargetInstaller` relies on `PROJECT_HEADERS_FOLDER_PATH`, which doesn't exist by default.

mrackwitz:
Neither PROJECT_HEADERS_FOLDER_PATH exists nor it makes sense to bundle project headers into the built framework, so skip them here.

Fixes #4819.


segiddins:
Can you confirm projects with a header mappings dir and project headers can build again after this change?

mrackwitz:
>Can you confirm projects with a header mappings dir and project headers can build again after this change?

No, at least for the Realm podspec it doesn't change anything, but it wasn't supposed to fix that. It doesn't spam the build folder anymore with headers, which wouldn't be picked up anyway.

PR #4829, opened 2016-01-27
[XCConfig] Don't recurse for transitive dependencies

mrackwitz:
The list of dependencies on `PodTarget` contains already all transitive dependencies, that means recursively calling the method for any of its elements won't catch additional targets and just wastes some cycles.


segiddins:
👍🏻

PR #4844, opened 2016-01-29
[Command Project] xcodeproj was renamed to project

mrackwitz:
I skipped references to `xcodeproj` in strings, so I missed these occurrences. I double checked all remaining occurrences and it seems like these were the only ones which had to be changed.


orta:
:+1:

segiddins:
👍🏻

PR #4868, opened 2016-02-05
[Gemfile] Add option to cp_gem to make using gem from path easier

segiddins:


segiddins:
Going to merge this unless there are any objections?

orta:
Makes sense :+1:

PR #4875, opened 2016-02-08
[SandboxPod] Fix escaping paths with quotes in them

### LINKED ISSUE #4874: Double quotes in project's name causes unbound variable exception upon sandbox-pod install
Simple repro:

``` BASH
$ mkdir a\"b
$ cd a\"b
$ echo pod 'IQKeyboardManager' > Podfile
$ sandbox-pod install
```

Fails with:
`sandbox-exec: line 73: unbound variable: b"`

Whereas executing `$ pod install` passes that phase.

The problem lies [here](https://github.com/CocoaPods/CocoaPods/blob/master/bin/sandbox-pod#L97) because `Pod::Config.instance.project_root` needs to be escaped in order for the regex to be properly constructed.

segiddins:
Closes #4874.

\c @alloy 

- [ ] CHANGELOG

### DIFF
--- a/bin/sandbox-pod
+++ b/bin/sandbox-pod
@@ -31,7 +31,7 @@ require 'cocoapods/config'
 require 'rbconfig'
 require 'erb'
 
-PROFILE_ERB_TEMPLATE = <<-EOS
+PROFILE_ERB_TEMPLATE = <<-'ERB'
 (version 1)
 (debug allow)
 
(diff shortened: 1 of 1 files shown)


mrackwitz:
:+1:

Mitko-Kerezov:
@segiddins have you tried the steps to reproduce I've mentioned in the issue?
As I said, I'm not great at ruby at all, but even with your fix I'm able to reproduce the issue and get an error:
`sandbox-exec: line 55: unbound variable: b"`

PR #4909, opened 2016-02-18
[XCConfig] Fix search path inheritance support

segiddins:
\c @mrackwitz

### DIFF
--- a/lib/cocoapods/generator/xcconfig/pod_xcconfig.rb
+++ b/lib/cocoapods/generator/xcconfig/pod_xcconfig.rb
@@ -45,12 +45,14 @@ module Pod
           search_paths = target_search_paths.concat(sandbox_search_paths).uniq
 
           config = {
+            'FRAMEWORK_SEARCH_PATHS' => '$(inherited) ',
+            'GCC_PREPROCESSOR_DEFINITIONS' => '$(inherited) COCOAPODS=1',
+            'HEADER_SEARCH_PATHS' => XCConfigHelper.quote(search_paths),
+            'LIBRARY_SEARCH_PATHS' => '$(inherited) ',
             'OTHER_LDFLAGS' => XCConfigHelper.default_ld_flags(target),
             'PODS_ROOT' => '${SRCROOT}',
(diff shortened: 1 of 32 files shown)


mrackwitz:
There should be tests for `settings_for_dependent_targets` and `settings_to_import_pod_targets`.

segiddins:
Agreed, but I figured you might want to refactor this. It just gets everything working again.

mrackwitz:
How should I do non-trivial refactorings without tests?

segiddins:
Integration specs cover this.

mrackwitz:
Do they really? Can you point me where? When I was able to break them in the first place, are they really sufficient? I don't think so. We need unit tests to specify behavior in such cases.

segiddins:
You broke them by regenerating them... The integration specs breakage here demonstrates the fix. I was just trying to help you get your PR ready to merge, I'm not looking for an argument.

mrackwitz:
Neither I'm. Just trying to figure out whether that's sufficient or not. I don't think so and can take of that. Thank you for this first step.

lib/cocoapods/generator/xcconfig/pod_xcconfig.rb
```diff
@@ -45,12 +45,14 @@ def generate
           search_paths = target_search_paths.concat(sandbox_search_paths).uniq
 
           config = {
+            'FRAMEWORK_SEARCH_PATHS' => '$(inherited) ',
```
mrackwitz: Why do you have these `FRAMEWORK_SEARCH_PATHS` and `LIBRARY_SEARCH_PATHS` here overwritten with `$(inherited) `? On my branch these values are kept anyway when these are overwritten and as long as these aren't overwritten, they don't need to be explicitly set. I ask because somehow this causes double occurrences of `$(inherited)`.

segiddins:
@mrackwitz I actually think that was a remnant of me trying to refactor something, so those two lines can likely be deleted.

mrackwitz:
I cherry-picked your changes on my branch for easier rebasing.

PR #4915, opened 2016-02-18
Add "TARGET_DEVICE_ARGS" to ibtool steps for storyboards

### LINKED ISSUE #4913: The ibtool processing of storyboards needs the "target-device" flags when building for less than iOS 8.0
We were getting hit with the build process creating a zero bytes compiled storyboard, and poking around things, we found the culprit to be the `ibtool` step,

    ibtool --reference-external-strings-file --errors --warnings --notices --minimum-deployment-target 6.0 --output-format human-readable-text --compile BuildDir/SwrveConversation.storyboardc SourceDir/SwrveConversation.storyboard --sdk /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS9.2.sdk

And then reading up about this on stackoverflow, I found,
    http://stackoverflow.com/questions/29563618/xcode-no-valid-compiled-storyboard-at-path/29563997

Which pointed to the `Use Size Classes` which I think essentially turns a Universal storyboard into a per-device one.

Then looking at the stop that XCode do themselves, it seems this has been resolved in their build process, but CocoaPods has fallen behind.

In their step, they do,

    ibtool --errors --warnings --notices --module TestyBedy --output-partial-info-plist /Developer/workspaces/xcode/TestyBedy/Build/Intermediates/TestyBedy.build/Debug-iphonesimulator/TestyBedy.build/SwrveConversation-SBPartialInfo.plist --auto-activate-custom-fonts --target-device iphone --target-device ipad --minimum-deployment-target 6.0 --output-format human-readable-text --compilation-directory /Developer/workspaces/xcode/TestyBedy/Build/Intermediates/TestyBedy.build/Debug-iphonesimulator/TestyBedy.build /Developer/workspaces/xcode/TestyBedy/TestyBedy/SwrveConversation.storyboard

So then, for us, altering the top command to just,

    ibtool --reference-external-strings-file --errors --warnings --notices --minimum-deployment-target 6.0 --output-format human-readable-text --compile BuildDir/SwrveConversation.storyboardc SourceDir/SwrveConversation.storyboard --sdk /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS9.2.sdk  --target-device iphone --target-device ipad

Then produced a correctly compiled storyboard, and fixed all our issues.

seaders:
Closes #4913.

Move the case statement to the top of the file, where we set "TARGET_DEVICE_ARGS" so it's available to all scripts, and add it to the `ibtool` storyboard processing step.

### DIFF
--- a/lib/cocoapods/generator/copy_resources_script.rb
+++ b/lib/cocoapods/generator/copy_resources_script.rb
@@ -106,6 +106,21 @@ RESOURCES_TO_COPY=${PODS_ROOT}/resources-to-copy-${TARGETNAME}.txt
 
 XCASSET_FILES=()
 
+case "${TARGETED_DEVICE_FAMILY}" in
+  1,2)
+    TARGET_DEVICE_ARGS="--target-device ipad --target-device iphone"
+    ;;
+  1)
+    TARGET_DEVICE_ARGS="--target-device iphone"
+    ;;
+  2)
+    TARGET_DEVICE_ARGS="--target-device ipad"
+    ;;
+  *)
+    TARGET_DEVICE_ARGS="--target-device mac"
+    ;;
+esac
+
 realpath() {
   DIRECTORY="$(cd "${1%/*}" && pwd)"
   FILENAME="${1##*/}"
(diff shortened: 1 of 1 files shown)


segiddins:
:+1: 
@efirestone this look good to you?

efirestone:
Looks good to me.

Tangentially related (and not a blocker for this PR), it might make sense to remove all of the manual compiling in `copy_resources_script.sh` now that we have the semantic resource understanding that I added. Because the elements are added properly in the Xcode project, Xcode will compile them properly, with no post-processing from Pods being necessary. For example, previously, Xcode just knew to copy `en.lproj/*` and then Pods manually compiled things like `en.lproj/Main.xib`, but now Xcode knows about `en.lproj/Main.xib` directly in the project and will compile it.

This would eliminate any inconsistencies between how Pods and Xcode compile things, and would speed up the build. The difficult part is that we'd likely break lots of existing pre-1.0.0 podspecs since they're using `Resources/*` rather than `Resources/**/*`, so even with 1.0.0 things like nested XIBs wouldn't be added to the project and wouldn't get compiled. We could only manually compile if we detect something like that (but it won't be perfect), or could add a compatibility flag somewhere, or just make everyone update to do the new thing.

Happy to discuss in a better forum if there is one.

seaders:
I definitely agree with that perspective, @efirestone, getting those manual steps as basic as possible (as in *maybe* just copying them into the correct directory), and then letting XCode deal with them as it knows how to, would I think clear a lot of the inconsistencies some people encounter with CocoaPods, especially now with more build steps changing with XCode.

But as you said, if it breaks a load of pre- things, you gotta give people a bit of a heads up & notice about it first.

jafara:
To be fair, CocoaPods 1.0.0 introduces a lot of breaking changes so if there is a time to do it, the time is now.

efirestone:
I actually think we should be able to remove the `copy resources` script completely when using the new functionality. It would be worth commenting it out on some bigger projects (using `Resources/**/*`) to give it a test.

segiddins:
@efirestone even when not using frameworks?

efirestone:
I believe so. The compilation parts definitely shouldn't be necessary, and copying things into the resources bundle should also happen via Xcode. The two parts I'm not sure about:
- Copying the resource bundle itself into the main bundle (app or framework). Does the `copy resources` script do this? It didn't look like it in a quick glance, but I could be missing it. Even if it does, we could add the resources bundle to the appropriate `copy resources` build phase and not need the script.
- Not entirely sure about the behavior when using `spec.resources` (vs `resource_bundles`). I don't think there are issues here, but I haven't used `resources` enough to know all of the intricacies.

seaders:
Hey guys, just wondering if there's anything more needed from me?

If there's edits needed for the integration side, I can make those too.  My read of it is that two files need editing,

https://github.com/CocoaPods/cocoapods-integration-specs/blob/05c3fc9beea098b5e1ce3aaa882d454511e6348c/install_resources/after/Pods/Target%20Support%20Files/Pods-SampleApp/Pods-SampleApp-resources.sh

and

https://github.com/CocoaPods/cocoapods-integration-specs/blob/05c3fc9beea098b5e1ce3aaa882d454511e6348c/install_resources_no_source_files/after/Pods/Target%20Support%20Files/Pods-SampleApp/Pods-SampleApp-resources.sh

I could create a PR for those as well, if need be?

segiddins:
@seaders nah, I'll merge and bump the integration specs as soon as I have a chance.

seaders:
Ok cool @segiddins.  Thanks again for all the quickness on this, there's a few Swrve customers affected by this, and having an official answer for the issue would be great.

segiddins:
Manually merged via https://github.com/CocoaPods/CocoaPods/commit/b5813f3bbfa4ccb979a65917c6b179de5bd8fec0.

PR #4933, opened 2016-02-23
Propagate the CocoaPods version to the prepare_command

mrackwitz:
This gives non-trivial podspecs one way to adopt the breaking changes of 1.0 while staying backwards-compatibility with 0.39. This isn't suitable when two versions of CocoaPods are used alongside as the `prepare_command` is run once on the cached sources. We could mitigate this by making the cache version-specific.

### DIFF
--- a/lib/cocoapods/installer/pod_source_preparer.rb
+++ b/lib/cocoapods/installer/pod_source_preparer.rb
@@ -59,9 +59,11 @@ module Pod
         UI.section(' > Running prepare command', '', 1) do
           Dir.chdir(path) do
             ENV.delete('CDPATH')
+            ENV['COCOAPODS_VERSION'] = Pod::VERSION
             prepare_command = spec.prepare_command.strip_heredoc.chomp
             full_command = "\nset -e\n" + prepare_command
             bash!('-c', full_command)
+            ENV.delete('COCOAPODS_VERSION')
           end
         end
       end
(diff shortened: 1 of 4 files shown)


segiddins:
@mrackwitz cache is already cleaned when a different version of CocoaPods runs

lib/cocoapods/installer/pod_source_preparer.rb
```diff
@@ -59,9 +59,11 @@ def run_prepare_command
         UI.section(' > Running prepare command', '', 1) do
           Dir.chdir(path) do
             ENV.delete('CDPATH')
+            ENV['COCOAPODS_VERSION'] = Pod::VERSION
             prepare_command = spec.prepare_command.strip_heredoc.chomp
             full_command = "\nset -e\n" + prepare_command
             bash!('-c', full_command)
+            ENV.delete('COCOAPODS_VERSION')
```
segiddins: this needs to be inside an `ENSURE`

segiddins:
:+1: with CHANGELOG

lib/cocoapods/installer/pod_source_preparer.rb
mrackwitz: :+1:

mrackwitz:
:tada:

spec/unit/installer/pod_source_installer_spec.rb
```diff
@@ -58,6 +58,18 @@ module Pod
           @spec.prepare_command = 'cd Classes;ls Banana.h'
           lambda { @installer.install! }.should.not.raise
         end
+
+        it 'sets the $COCOAPODS_VERSION environment variable' do
+          @spec.prepare_command = "[ \"$COCOAPODS_VERSION\" == \"#{Pod::VERSION}\" ] || exit 1"
+          lambda { @installer.install! }.should.not.raise
```
orta: this is a nice test :+1:

PR #4934, opened 2016-02-23
[PodVariantSet] Naive algorithm for #scope_by_specs

mrackwitz:
The scope suffix name algorithm implemented by #4146 works, but brings up some lengthy suffixes when a pod is used in different subspec variant, e.g. in a case like AFNetworking, where the subspecs define internal dependencies. So depending on just a single subspec of AFNetworking might bring up some more, which might be virtual default subspecs, because they keep popping up everywhere as dependencies. It would be possible to recognize that and leave them out of the name. But as the dependencies can be itself platform-dependent that would open another can of worms.
So instead of that the pod targets could be just enumerated. That has a better performance obviously, but it is then not so easy to grasp anymore what's going on. That makes debugging and supporting harder. Only by taking a look into each aggregate targets' dependencies within Xcode, it's possible to see, which pod target belongs to which target in the user project.

### DIFF
--- a/lib/cocoapods/installer/analyzer/pod_variant_set.rb
+++ b/lib/cocoapods/installer/analyzer/pod_variant_set.rb
@@ -1,5 +1,3 @@
-require 'set'
-
 module Pod
   class Installer
     class Analyzer
@@ -23,7 +21,7 @@ module Pod
         #
         def scope_suffixes
           return { variants.first => nil } if variants.count == 1
-          scope_by_specs
+          scope_by_build_type
         end
 
         # Groups the collection by result of the block.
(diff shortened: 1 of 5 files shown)


segiddins:
Given we already have this logic written and tested, and the simple `-\d` suffix provides no context, I'm not really convinced there's any benefit to this change?

mrackwitz:
Just had it already laying around and wanted to make sure that we come together to the best decision, not just what was the first thing which came to my mind when I implemented it once. I'm perfectly fine with closing that unmerged.

mrackwitz:
@orta: When we discussed that briefly internally, you seem to lean more towards this option. Do you have a strong opinion about that?

segiddins:
Thanks for opening this for discussion, but I'm favor of keeping the existing implementation, if that's OK with you?

mrackwitz:
Yes, absolutely, just wanted to keep it around so that we might get other input on that, but I'm also in favor of closing this unmerged.

PR #4942, opened 2016-02-24
[ISSUE_TEMPLATE] Bigger headings

mrackwitz:
I was a bit shy with my initial proposal. These bigger headings bring an underline with GitHub's stylesheets which gives the different sections better visual distinction.


orta:
This feels good:
screen shot 2016-02-24 at 12 35 36 pm

:ship:

segiddins:
Shouldn't we just get people to share the output of `pod env` if they're on 1.0 instead of just the podfile?

mrackwitz:
I wrote a [bash script](https://gist.github.com/mrackwitz/9c567523d4576537f120), which could be piped to shell for previous versions. (:scream:) It tries to mimic the outputs of `pod env`, but has to make some compromises. What do you think, shall we go so far, put that in our repo and recommend that in the meantime?

Issue #4872, opened 2016-02-08
Resource bundle has weird path?


a2:
There's something with quotes that it doesn't like? I'm not sure. This is on 1.0.0.beta.3.

<img width="891" alt="screen shot 2016-02-07 at 6 52 47 pm" src="https://cloud.githubusercontent.com/assets/241156/12882023/c3784412-ce4e-11e5-8730-05eee818d043.png">
![screen shot 2016-02-07 at 6 53 03 pm](https://cloud.githubusercontent.com/assets/241156/12882025/c398cbb0-ce4e-11e5-8d96-a5b8c46e5723.png)

segiddins:
@efirestone could this be caused by your recent changes?

efirestone:
Quite possible. I don't believe I'm generating any resource paths myself, but it's possible this is a combination of my changes and some less-tested code path somewhere. The issue definitely appears to be the generated path (note the `/Users/a2/Developer/Hydr8/Pods` at the beginning, followed by another full path). Perhaps this is an issue with the resource path being relative to the group, but also being defined as an absolute path.

@a2, some questions:

- Is it possible to provide your podspec, or at least the `.resources =` or `.resource_bundles =` part?
- Is this consistently reproducible?

In general, repro steps would help me look into it more. Thanks!

a2:
@efirestone Yeah, it was reproducible, so I went back to 0.39.0.

The resource_bundles part of the podspec was:
```ruby
s.resource_bundles = { "Hydr8KitResources" => "Hydr8Kit/Resources/**/*" }
```

The path structure was:
Hydr8 > Hydr8Kit > Resources > Base.lproj > UnitTemplates.stringsdict

efirestone:
I'm having trouble reproducing this locally.

Another thought: @segiddins could this be related to target de-duping? The issue is around the bundle itself and not the resources in it, so it seems like it might be at the target level.

@a2 Do you have multiple subspecs in the podspec that might be consuming this resources bundle (possibly indirectly)? Is it possible to get either a small sample project (ideal), or at least more of the podspec? Sorry for the trouble, and thanks for the help.

a2:
@efirestone There aren't any subspecs. I'll try to get a small repro project together later today.

basememara:
[...]

PR #4954, opened 2016-02-26
Fix resource bundle paths for improved deduplication

mrackwitz:
While #4146 fixed one problem described in #4872, it brought up another one. The paths in the `CopyResourcesScript` passed to the `install_resource` function for resource bundles are wrong.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
-* None.  
+* Fixes paths passed for resources bundles in the copy resources script.  
+  [Marius Rackwitz](https://github.com/mrackwitz)
+  [#4954](https://github.com/CocoaPods/CocoaPods/pull/4954)
 
 
 ## 1.0.0.beta.4 (2016-02-24)
--- a/lib/cocoapods/installer/target_installer/aggregate_target_installer.rb
+++ b/lib/cocoapods/installer/target_installer/aggregate_target_installer.rb
@@ -107,15 +107,18 @@ module Pod
         library_targets = target.pod_targets.reject do |pod_target|
           pod_target.should_build? && pod_target.requires_frameworks?
         end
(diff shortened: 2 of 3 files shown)


segiddins:
\c @efirestone

mrackwitz:
This changes the used base path variable from `BUILT_PRODUCTS_DIR` to `CONFIGURATION_BUILD_DIR`, while the [documentation](https://developer.apple.com/library/mac/documentation/DeveloperTools/Reference/XcodeBuildSettingRef/1-Build_Setting_Reference/build_setting_ref.html#//apple_ref/doc/uid/TP40003931-CH3-SW42) states:
>Run Script build phases can use the value of this build setting as a convenient way to refer to the product files built by one or more targets even when these files are scattered throughout a directory hierarchy (for example, when `DEPLOYMENT_LOCATION` is set to `YES`.

But according to the documented value of the setting `BUILT_PRODUCTS_DIR`, it takes only a different value when `RETAIN_RAW_BINARIES` is also set to `YES`, what doesn't seem to have much practical relevance.

segiddins:
RuboCop failures: https://travis-ci.org/CocoaPods/CocoaPods/jobs/111974486#L2693-L2703

efirestone:
LGTM

arifken:
Thanks for fixing this! i've been running into this issue as well. looking forward to it in the next beta. 👍

gcox:
Not sure this is totally fixed.  It did fix the path issue I was having with the ```1PasswordExtension``` but it caused a new issue for the ```google-cast-sdk``` pod.

The path passed to ```install_resource``` for this pod is ```"$CONFIGURATION_BUILD_DIR/google-cast-sdk/CastFrameworkAssets.bundle"``` but the actual path is ```"$CONFIGURATION_BUILD_DIR/CastFrameworkAssets.bundle"```. Not sure if this is a bug in the generation of the path or placement of ```CastFrameworkAssets.bundle```.

segiddins:
@gcox please open a new issue with a project that reproduces the issue

Issue #4989, opened 2016-03-07
Issues Cloning Spec repo - GitHub taking a very long time to download changes to the Specs Repo


jlubeck:
Note from @orta - 

If you are here because your Specs repo isn't updating, run: `cd ~/.cocoapods/repos/master && git fetch --depth=2147483647` - this will convert your local repository of Podspecs to be a full clone, as opposed to a shallow copy.

---

## What did you do?

Run `pod setup`

## What did you expected to happen?

Clone Spec repo master

## What happened instead?

It only downloads a few bytes and then throws error:

> fatal: unable to access 'https://github.com/CocoaPods/Specs.git/': transfer closed with outstanding read data remaining

## Podfile

No Podfile yet

I also tried cloning the repo manually or with the githhub desktop app with no avail. 
I´m having no issues cloning any other repo in github. Only with this one. Is it possible there is something wrong with it???

Thanks

jlubeck:
Tried again, new error:

>   error: RPC failed; result=18, HTTP code = 200
>   fatal: The remote end hung up unexpectedly
> [!] /usr/bin/git clone https://github.com/CocoaPods/Specs.git master --depth=1
> 
> Cloning into 'master'...
> error: RPC failed; result=18, HTTP code = 200
> fatal: The remote end hung up unexpectedly

Very weird...

ecch531:
I got same issue, too.

mangofever:
Same.
Cannot clone spec repo.

art-divin:
+1, same issue 
`git clone https://github.com/CocoaPods/Specs.git` takes forever

stringsanbu:
+1, been messing around with this for awhile. I doubled the buffer, didn't work. It actually seems to be getting "something" but fails. Using verbose didn't say much, just said it had issues accessing it. 

I tried accessing my other repos and it seemed to be OK, but it was definitely slower than normal.

kshuin:
+1 

I got  same issue, too.
my pod version was "0.39.0"

I tried cloning master repos directly (by `git clone git@github.com:CocoaPods/Specs.git master --depth=1 --verbose`), but also failed.

oronbz:
+1

aceontech:
+1

pedrocid:
+1

MarkMolina:
+1

stringsanbu:
Temporary workaround which might work: https://github.com/CocoaPods/Specs/master.zip haven't tested though

[...]

PR #5001, opened 2016-03-08
[WIP] Use the GitHub API to determine if we need to fetch a repository

DanielTomlinson:
- [x] Basic implementation
- [ ] Update tests

As recommended https://github.com/CocoaPods/CocoaPods/issues/4989#issuecomment-193801376

My current issue is mostly around how best to update the specs, especially because of networking being a dependency, and us using the real specs repo.
The current addition of
```
`git checkout HEAD~1`
`git clean -fd`
```
is just me trying to check the existing behaviour, and will be removed before removing [WIP]

### DIFF
--- a/lib/cocoapods/master_source.rb
+++ b/lib/cocoapods/master_source.rb
@@ -13,10 +13,10 @@ module Pod
     #
     def update(show_output)
       if requires_update
-        return super.update(show_output)
+        super
+      else
+        []
       end
-
-      return []
     end
 
     # Returns whether a source requires updating. 
@@ -30,19 +30,21 @@ module Pod
     # @return [Bool] Whether the given source should be updated.
     #
     def requires_update
+      current_commit_hash = ''
       Dir.chdir(repo) do
         current_commit_hash = (`git rev-parse HEAD`).strip
       end
-      url = URI.parse('https://api.github.com/repos/CocoaPods/Specs/commits/master')
-      req = Net::HTTP::Get.new(url.path)
-      req.add_field('Accept', 'application/vnd.github.chitauri-preview+sha')
-      req.add_field('If-None-Match', current_commit_hash)
+      uri = URI.parse('https://api.github.com/repos/CocoaPods/Specs/commits/master')
 
-      res = Net::HTTP.new(url.host, url.port).start do |http|
-        http.request(req)
-      end
+      http = Net::HTTP.new(uri.host, uri.port)
+      http.use_ssl = true
+      headers = {
+        'Accept' => 'application/vnd.github.chitauri-preview+sha',
+        'If-None-Match' => current_commit_hash
+      }
+      code = http.head(uri.path, headers).code.to_i
 
-      return res.code != 304 
+      return code != 304
     end
   end
 end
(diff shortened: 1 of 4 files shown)


lib/cocoapods/sources_manager.rb
```diff
@@ -41,7 +41,15 @@ def aggregate_for_dependency(dependency)
       #
       def sources(names)
         dirs = names.map { |name| source_dir(name) }
-        dirs.map { |repo| Source.new(repo) }
+        dirs.map do |repo|
+          if Pathname(repo).basename == 'master'
```
mrackwitz: There should be a method like `source_from_path` which takes care of that logic.

mrackwitz: Given that `Source::Aggregate` also instantiates sources, I would be in favor of moving `MasterSource` into CocoaPods/Core and adding a static method e.g. `Source[path]`, which does this check here. Everywhere where we are currently using `Source.new`, we could use this new method then instead.

segiddins: or even do it in `Source.new`

DanielTomlinson:
Ok, unit tests are passing locally now. Will wait for travis to run the rest.

DanielTomlinson:
Ok cool, looks like I just need to add network stubs.

lib/cocoapods/master_source.rb
```diff
@@ -0,0 +1,50 @@
...
+      end
+      uri = URI.parse('https://api.github.com/repos/CocoaPods/Specs/commits/master')
+
+      http = Net::HTTP.new(uri.host, uri.port)
+      http.use_ssl = true
+      headers = {
+        'Accept' => 'application/vnd.github.chitauri-preview+sha',
+        'If-None-Match' => current_commit_hash
+      }
+      code = http.head(uri.path, headers).code.to_i
```
alloy: Use https://github.com/Fingertips/Nap instead, which is the one that should be used everywhere in the codebase already.

lib/cocoapods/master_source.rb
```diff
@@ -0,0 +1,50 @@
...
+
+      return code != 304
```
k0nserv: What happens in the event of 5xx codes? It might be okay to treat them the same as 2xx, but it's worth considering.

DanielTomlinson: @k0nserv My current thinking is to just try the fetch in those cases for now (especially as the API is in preview and likely to change). We could also raise an error.

k0nserv: Okay that's :) Just wanted to see if you'd considered it.

dtorres:
Suggestion: Wouldn't it be better to do a GithubSource instead, so that any sources hosted on github get the same advantages?

segiddins:
@dtorres no, since wwe may want to special-case the master repo in more ways in the future.

dtorres:
@segiddins that makes sense. But, a GithubSource may still be beneficial since so many private sources are hosted in github as well.
Maybe some sort of composition where MasterSource adopts internally GithubSource and then does these future special cases as well?

segiddins:
@dtorres maybe in the future -- our immediate goal is to improve specifically the master repo. Once that's done, someone else is free to make a PR extending appropriate improvements to other GH repos.

DanielTomlinson:
Superseded by https://github.com/CocoaPods/Core/pull/313

PR #5021, opened 2016-03-10
[Validator] Validate that header files are in source_files

### LINKED ISSUE #4936: PodDirCleaner doesn't preserve public_header_files or private_header_files
I was writing a podspec for a vendored library, so I included `public_header_files` but no `source_files`. Turns out that `public_header_files` are stripped out by `PodDirCleaner`, so my pod worked great during development and `pod lib lint`, but the necessary files were being stripped when I would `pod spec lint`.

Perhaps it's expected that only `source_files` will be preserved. In that case maybe we should update [the documentation](https://guides.cocoapods.org/syntax/podspec.html#public_header_files) for `public`/`private_header_files` to indicate that the files should only be a subset of the files specified in `source_files`?

segiddins:
Closes #4936.

Hard to test because the current validator specs all use JSONKit, which doesn't have enough source files to reproduce the issues.

### DIFF
--- a/lib/cocoapods/validator.rb
+++ b/lib/cocoapods/validator.rb
@@ -560,11 +560,17 @@ module Pod
     # Ensures that a list of header files only contains header files.
     #
     def _validate_header_files(attr_name)
-      non_header_files = file_accessor.send(attr_name).
+      header_files = file_accessor.send(attr_name)
+      non_header_files = header_files.
         select { |f| !Sandbox::FileAccessor::HEADER_EXTENSIONS.include?(f.extname) }.
         map { |f| f.relative_path_from(file_accessor.root) }
       unless non_header_files.empty?
         error(attr_name, "The pattern matches non-header files (#{non_header_files.join(', ')}).")
+      end
+      non_source_files = header_files - file_accessor.source_files
+      unless non_source_files.empty?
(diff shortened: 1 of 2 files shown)


lib/cocoapods/validator.rb
```diff
@@ -560,12 +560,18 @@ def _validate_module_map
...
     def _validate_header_files(attr_name)
-      non_header_files = file_accessor.send(attr_name).
+      header_files = file_accessor.send(attr_name)
+      non_header_files = header_files.
         select { |f| !Sandbox::FileAccessor::HEADER_EXTENSIONS.include?(f.extname) }.
         map { |f| f.relative_path_from(file_accessor.root) }
       unless non_header_files.empty?
         error(attr_name, "The pattern matches non-header files (#{non_header_files.join(', ')}).")
       end
+      non_source_files = header_files - file_accessor.source_files
```
dshirley: I'm not completely sure, but I think that this screws up linting / uploading my Podspec.
Our source files are packaged in a .zip.  Our headers are in the .zip file (the zip can also be downloaded from our website).

Here's our spec:  https://github.com/CocoaPods/Specs/blob/master/Specs/CrittercismSDK/5.5.5/CrittercismSDK.podspec.json

Issue #5048, opened 2016-03-17
Add an example with resource bundles to the test harness


mrackwitz:
CocoaPods supports different possibilities for podspecs to declare resources. One of these are resource bundles. At installation time, CocoaPods generates an additional target for each resources bundles. The resources are added to that. At runtime, Xcode applies the built-in build rules to these files, just like it is done for resources in app or framework targets. The built product is a bundle directory, which holds the processed resources. This resources bundle has to be copied into the app bundle, which is done by the generated copy resources script.
The test suite defines via unit and integration specs how a working installation should look like. But this working installation is mostly based on human assumptions and reasoning. There are no specs currently which actually execute a copy resources script, which copies a resource bundle from a build directory into an app bundle. If any path around that is broken that might not be caught just by code review (as happened with #4954, #5028).
Our [examples](https://github.com/CocoaPods/CocoaPods/tree/master/examples) would be suitable for adding an example with a podspec, which uses resource bundles.
## Notes
- The examples are validated via the exit code of `xcodebuild`. This can be triggered via a Rake task which can be executed by [`rake examples:build`](https://github.com/CocoaPods/CocoaPods/blob/master/Rakefile#L244-L282).
- Only iOS and OS X examples are actually build and contribute in such a way to test coverage.
- Local podspecs / development pods as seen in the existing examples are preferable. Large downloads increase the execution time of the test suite, which makes them undesirable and should be avoided at all cost.
- Ideas for additional inspections / validations of the build products are welcome!

benasher44:
@mrackwitz I was having a look at what would be involved to do this, and I'd want to have example podspecs that illustrate all of the different ways resources can be added. Is there some CocoaPods.org search foo that can help me find pods that have these different kinds of specs? Or, could I just make local pods with a parent project Podfile that use :path to point to where the pods/specs are? It seemed like this was what you were implying with "local podspecs / development pods," but just by looking at some of the existing Podfiles in the example projects, I don't see development/local pods.

mrackwitz:
@benasher44: Coverage for all different ways of integrating resources would be even better. What you describe was exactly what I meant. But the Alamofire example for example is using a local podspec. (Which I added back then before there was an official podspec. Given it's popularity / the resulting likeliness it stays around and is updated, it might make sense to switch the approach there.)

benasher44:
Ah yes I see it is. Alright, I'll have a look at this

benasher44:
[...]

PR #5049, opened 2016-03-17
[WIP][PodTargetInstaller] Fix build dir for resource bundles

### LINKED ISSUE #5034: Specs using resource_bundle broken on CocoaPods 1.0.0.beta.6

# Report

The [documentation](http://guides.cocoapods.org/syntax/podspec.html#resource_bundles) for the usage of resource bundles suggests that Spec files use a format like so:

```ruby
spec.resource_bundles = {
    'MapBox' => ['MapView/Map/Resources/*.png'],
    'OtherResources' => ['MapView/Map/OtherResources/*.png']
  }
```

When this declaration is done cocoapods generates a new target for the copy resources phase of the bundle. It looks like the path for this bundle is incorrect as it now ends up in `Pod/Pod/Pod.bundle` where the expected path from the copy script phase is `Pod/Pod.bundle`.

Changing the spec to use `spec.resource = 'Path/To/Bundle.bundle'` resolves the issue and removes the extra target.

## What did you do?

Created a simple podfile under CocoaPods 1.0.0.beta.2 -> 1.0.0.beta.6 using a Spec that uses `resource_bundle`. Ran pod install and then attempt to build the resulting project. 

The project to build and a happy simulator to open up with the app running.

## What happened instead?

The copy bundle resources phase fails to identify the bundle and the bundle is also not present in the project structure.

It seems as though I should find a `.bundle` under the specific pods directory but instead I see a folder structure of `POD_NAME/Assets/ios/{Resources that should be in a bundle but are now clearly visible}`

Interestingly enough Specs that use the `resources` tag and specify bundles work normally. As an example the [GoogleAppIndexing](https://github.com/CocoaPods/Specs/blob/master/Specs/GoogleAppIndexing/2.0.3/GoogleAppIndexing.podspec.json) pod works fine and one will see the `.bundle` in the correct location.

My suspicion is that this is a regression of #4762 ?

## Podfile

```ruby
use_frameworks!

target 'NYTPhotoViewerFun' do
	pod 'NYTPhotoViewer'
end
```

## Working Spec

I forked 'NYTPhotoViewer' [https://github.com/danielgalasko/NYTPhotoViewer/blob/danielgalasko-cocoapods-1.0.0/NYTPhotoViewer.podspec#L19] and this works

## Failing Spec

The original Spec that was working on CocoaPods 0.39.0 can be found [here](https://github.com/NYTimes/NYTPhotoViewer/blob/develop/NYTPhotoViewer.podspec#L19)

mrackwitz:
It was accidentally recursively defined before, which led to the target build product being scoped one level to deep.
For consistency sake I've based the parent dir on base of what we define by the XCConfigHelper as `PODS_SHARED_BUILD_DIR`. We could do that here as well. This value is the default value in Xcode for `CONFIGURATION_BUILD_DIR`, which is different from what the documentation states. (:tada:) We might also use the documented value instead or `TARGET_BUILD_DIR`. But the latter is also based on `CONFIGURATION_BUILD_DIR`, so that could lead to the same recursive definition problem we had before, depending on variables are resolved.

* [x] Depends on #5048
* [x] Re-evaluate on what build setting this is defined
* [x] Changelog entry

Fixes #5034.

/c @samdmarshall


jafara:
Is this still happening? It feels like development of CocoaPods has slowed down a bit. Its a pity cause its very close to 1.0.0 release....

segiddins:
@jafara a bunch of us have been on vacation recently, but this will still happen as soon as all of the action items above have been addressed, either by Marius (who is on vacation now) or someone else.

PR #5056, opened 2016-03-18
[PodTargetInstaller] Fix CONFIGURATION_BUILD_DIR var for resource bun…

chipp:
…dles

Now it is generating with help of XCConfigHelper.


orta:
@mrackwitz might be best to look at this

mrackwitz:
In principle a nice idea, but that doesn't work here as the constant defines a variable name, which is only available to the xcconfigs of pod targets.

PR #5063, opened 2016-03-19
Example with resource bundles for the test harness

### LINKED ISSUE #5048: Add an example with resource bundles to the test harness

benasher44:
This fixes #5048 by adding some samples for testing different ways of adding resources to pods


benasher44:
@mrackwitz so `rake examples:build` doesn't report any failures, but if I actually go to build my example project then, the build definitely fails as intended (because of #5034). Is there some other way these example projects are intended to be used, or should I fix up the `rake examples:build` to fail if `xcodebuild` fails?

benasher44:
I should also mention that after doing all of this, this failure is just as easily discovered by doing `pod lib lint ResourcesBundleExample.podspec`. Knowing this, maybe there's a lighter-weight way to catch this in the test suite, given that `pod lib lint` basically does what I'm doing here (integrate a pod into a test project and try building)?

examples/Resource Bundle Example/Example Pods/Images.xcassets/Contents.json
```diff
@@ -0,0 +1,6 @@
+{
+  "info" : {
+    "version" : 1,
+    "author" : "xcode"
+  }
+}
```
segiddins: trailing newlines on these files, please?

segiddins:
This is awesome!

examples/Resource Bundle Example/Example Pods/Images.xcassets/Contents.json
benasher44: Sure I can do that. Xcode (7.2) generated these though, so if someone goes and modifies this later, you'll run into the same issue. Would you still like this fixed?

segiddins: Not strictly necessary, but trailing newlines can help with diffing

benasher44: Gotcha. Will fix!

benasher44:
Done. Would still like to wait to see #5069 merged, so then I can update this PR and ensure the build fails with this change, since #5034 isn't fixed.

benasher44:
I'm not sure what's going on with  `execution_output.txt`, but this has flaked on me a few times now.

segiddins:
@benasher44 don't worry about it

benasher44:
@mrackwitz travis is now failing in `xcodebuild`'s `CpResource` as expected.

segiddins:
@mrackwitz if you could rebase your fix on top of this branch, I think we'll be good to merge?

mrackwitz:
:white_check_mark: Done. That means, we can close this PR as the changes will be merged via #5048.

PR #5114, opened 2016-04-03
[Installer] List all found configurations when an unknown one is encountered

### LINKED ISSUE #5113: Error message when using :configuration in a Podfile could show more information
So, I've added a configuration to one of my Pods in my Podfile:

``` ruby 
target 'Artsy' do
  pod 'ALPValidator', :configuration => 'Release'
end
```

But the name doesn't match up with my configurations:

screen shot 2016-04-03 at 21 13 31

And the message I receive is:

``` sh
[!] Unknown configuration whitelisted: release.
```

This could be expanded to include the configurations CP found, and the Pod:

```
[!] Unknown configuration for the Pod ALPValidator, "release". 
    CocoaPods found "Debug", "Store" and "Demo" - did you mean one of these?
```

segiddins:
Closes #5113.

\c @orta

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
-* None.  
+* When an unknown build configuration is mentioned in the Podfile, CocoaPods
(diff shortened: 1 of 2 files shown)


orta:
Yep :+1:

PR #5127, opened 2016-04-07
[Validator] Add analyze build action to validator xcodebuild command

Ruenzuo:
This adds a new option to both `pod lib lint` and `pod spec lint` to optionally run xcodebuild analyze build action. Tests are missing, I wanted to check what do you guys think about this first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   will suggest the build configurations found in the user project.  
   [Samuel Giddins](https://github.com/segiddins)
   [CocoaPods#5113](https://github.com/CocoaPods/CocoaPods/issues/5113)
+* Add analyze build action to CocoaPods linter xcodebuild command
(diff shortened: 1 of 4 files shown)


segiddins:
@orta thoughts on adding this feature?

orta:
Sorry for the delay!

Thanks @Ruenzuo  - I'm not too sold on how much this adds to the process in terms of advantages vs the amount of people opting to use `--allow-warnings` - we've tried to keep the Xcode verification more a "this builds" than "this is perfect".

Ruenzuo:
For us, we would like to use the CocoaPods linter as a way to find defects on our internal libraries, because we're already using it as a part of our release process.

But you have a fair point, didn't realise that the linter validation was more oriented to "this builds" rather than to detect defects. Probably running the `analyze` build action separately, or using a tool like [Infer](http://fbinfer.com), is a better choice.

 Thanks for looking into it ❤️

orta:
No worries - thanks to you @Ruenzuo  - it's always worth bringing these things up, and this is the first I've seen of infer! :+1:

Issue #5133, opened 2016-04-08
 Unit Test Targets do not copy resources (specifically storyboards) to the correct location


seaders:
* [:rainbow:] I've read and understood the [*CONTRIBUTING guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

Added a Unit Test Testing target to my project with reference to storyboards references in a CocoaPod (SwrveSDK)

## What did you expected to happen?

The test to be able to properly reference the storyboard at runtime

## What happened instead?

```2016-04-08 16:24:18.389 UnitTestsCocoaPods[42397:839885] Unable to load Conversation Item View Controller. Could not find a storyboard named 'SwrveConversation' in bundle NSBundle </Developer/repos/git/swrve/swrve-sdk/ios/UnitTestsCocoaPods/Build/Debug-iphonesimulator/UnitTestsCocoaPods.app/PlugIns/UnitTestsCocoaPods.xctest> (loaded)```

Very similar to https://github.com/CocoaPods/CocoaPods/issues/5022, this is (likely) due to a different TARGET_BUILD_DIR causing @rpath to be further down the tree from the main CONFIGURATION_BUILD_DIR where the resources are.

## Podfile

```ruby
platform :ios, '6.0'

abstract_target 'SwrveCocoaPods' do
    pod 'SwrveSDKCommon', :path => '../public'
    pod 'SwrveConversationSDK', :path => '../public'
    pod 'SwrveLocationSDK', :path => '../SwrveLocationSDK'
    pod 'SwrveSDK', :path => '../public'
    target 'UnitTestsCocoaPods'
    target 'UnitTestsCocoaPodsUnitTestsCocoaPods'
end
```

segiddins:
This has been merged!

Issue #5022, opened 2016-03-11
UI Test Targets do not generate valid test bundles when using frameworks


mtitolo:
* [:rainbow:] I've read and understood the [*CONTRIBUTING guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

Added a KIF UI Testing target to my project

## What did you expected to happen?

I expect to be able to run my UI Test target using CP's generated frameworks

## What happened instead?

The test bundle cannot be loaded:
```
2016-03-10 17:20:03.344 FrameworksIntegration[4274:1256151] The bundle “FrameworksIntegrationKIFTests” couldn’t be loaded because it is damaged or missing necessary resources. Try reinstalling the bundle.
2016-03-10 17:20:03.345 FrameworksIntegration[4274:1256151] (dlopen_preflight(/Users/michele/Library/Developer/Xcode/DerivedData/FrameworksIntegration-cneyifjmzbeardhlhwpbojjoaxrp/Build/Products/Debug-iphonesimulator/FrameworksIntegrationKIFTests-Runner.app/PlugIns/FrameworksIntegrationKIFTests.xctest/FrameworksIntegrationKIFTests): Library not loaded: @rpath/KIF.framework/KIF
  Referenced from: /Users/michele/Library/Developer/Xcode/DerivedData/FrameworksIntegration-cneyifjmzbeardhlhwpbojjoaxrp/Build/Products/Debug-iphonesimulator/FrameworksIntegrationKIFTests-Runner.app/PlugIns/FrameworksIntegrationKIFTests.xctest/FrameworksIntegrationKIFTests
  Reason: image not found)
```

This is (likely) due to a different `TARGET_BUILD_DIR` causing `@rpath` to be further down the tree from the main `CONFIGURATION_BUILD_DIR` where the frameworks are:
```
TARGET_BUILD_DIR = /Users/michele/Library/Developer/Xcode/DerivedData/FrameworksIntegration-gqwvjrovjkymuvgxnryvnnwzcchu/Build/Products/Debug-iphoneos/FrameworksIntegrationKIFTests-Runner.app//PlugIns
```
Aka Xcode is complaining that `KIF.framework` is not in `PlugIns`.

[Sample project is here](https://github.com/mtitolo/FrameworksIntegration/releases/tag/v0.0.1).

## Podfile

```
platform :ios, '9.0'

use_frameworks!

[...]

PR #5134, opened 2016-04-08
 [CopyResourcesScript] Rely on TARGET_BUILD_DIR to locate destination

seaders:
Fixes https://github.com/CocoaPods/CocoaPods/issues/5133.  It seems the new destination for Unit Tests in XCode 7.3 has changed, and resources like storyboards need to come from `DESTINATION/Plugins/UnitTests.xctest`.  Without the change to TARGET_BUILD_DIR, resources get built to just `DESTINATION` and the tests can't find the resources at runtime.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,11 +19,12 @@ before_install:
   # There is a bug in travis. When using system ruby, bundler is not
   # installed and causes the default install action to fail.
   - sudo gem install bundler
-  - bundle install
-  - bundle exec pod repo update
 
 after_install:
   - ./bin/pod repo update --silent
+
+before_script:
+  - bundle exec pod repo update --silent
 
 script: bundle exec rake
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,8 @@ before_install:
   # There is a bug in travis. When using system ruby, bundler is not
   # installed and causes the default install action to fail.
   - sudo gem install bundler
+  - bundle install
+  - bundle exec pod repo update
 
 after_install:
   - ./bin/pod repo update --silent
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,15 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 * None.  
 
 
+## 1.0.0.beta.7
+
+* Rely on `TARGET_BUILD_DIR` instead of `CONFIGURATION_BUILD_DIR` in the
+  generated embed resources build phase's script, so that UI test targets can
+  be run.  
+  [seaders](https://github.com/seaders)
+  [#5133](https://github.com/CocoaPods/CocoaPods/issues/5133)
+
+
 ## 1.0.0.beta.6 (2016-03-15)
 
 ##### Breaking
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,11 +15,6 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
-* None.  
-
-
-## 1.0.0.beta.7
-
 * Rely on `TARGET_BUILD_DIR` instead of `CONFIGURATION_BUILD_DIR` in the
   generated embed resources build phase's script, so that UI test targets can
   be run.  
--- a/examples/TestInclusions/Podfile
+++ b/examples/TestInclusions/Podfile
@@ -1,0 +1,12 @@
+workspace 'Examples.xcworkspace'
+project 'TestInclusions.xcodeproj'
+platform :ios, '6.0'
+
+abstract_target 'TestInclusionsPods' do
+    pod 'SwrveConversationSDK', '~> 4.3.0'
+
+    target 'TestInclusions'
+    target 'TestInclusionsTests' do
+        pod 'OCHamcrest', '~> 5.0'
+    end
+end
--- a/examples/TestInclusions/Podfile
+++ b/examples/TestInclusions/Podfile
@@ -6,7 +6,5 @@ abstract_target 'TestInclusionsPods' do
     pod 'SwrveConversationSDK', '~> 4.3.0'
 
     target 'TestInclusions'
-    target 'TestInclusionsTests' do
-        pod 'OCHamcrest', '~> 5.0'
-    end
+    target 'TestInclusionsTests'
 end
--- a/Rakefile
+++ b/Rakefile
@@ -273,8 +273,8 @@ begin
             when :ios
               # Specifically build against the simulator SDK so we don't have to deal with code signing.
               # Need to use the iPhone 6s, since this is the default simulator paired with a watch in Xcode 7.3
-              testCmd = (scheme_name.start_with? 'Test') ? 'test' : ''
-              execute_command "xcodebuild -workspace '#{workspace_path}' -scheme '#{scheme_name}' clean build #{testCmd} ONLY_ACTIVE_ARCH=NO -destination 'platform=iOS Simulator,name=iPhone 6s'"
+              test_flag = (scheme_name.start_with? 'Test') ? 'test' : ''
+              execute_command "xcodebuild -workspace '#{workspace_path}' -scheme '#{scheme_name}' clean build #{test_flag} ONLY_ACTIVE_ARCH=NO -destination 'platform=iOS Simulator,name=iPhone 6s'"
             else
               raise "Unknown platform #{platform}"
             end
--- a/Rakefile
+++ b/Rakefile
@@ -273,7 +273,8 @@ begin
             when :ios
               # Specifically build against the simulator SDK so we don't have to deal with code signing.
               # Need to use the iPhone 6s, since this is the default simulator paired with a watch in Xcode 7.3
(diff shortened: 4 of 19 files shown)


segiddins:
@mrackwitz @efirestone how does this change look to you?

efirestone:
LGTM.

Can confirm that with Xcode 7.2 the `xctest` bundle was installed alongside the host app, but with 7.3 it's installed inside of the host app (something like `Host.app/Plugins/testBundle.xctest`). To correspond with this change, they've made `TARGET_BUILD_DIR` point inside the host app where before `TARGET_BUILD_DIR` and `CONFIGURATION_BUILD_DIR` were the same. Can also confirm that these two variables are still the same for the production (non-test) target, so it shouldn't break anything there.

segiddins:
Awesome -- don't worry about the integration spec changes, I'll handle those before the next update goes out

seaders:
In this, can I ask about the tests you guys run? The reason I want to ask is that for this, and the last thing we did, our unit tests, for the Swrve SDK, showed up multiple of these issues.

Is there any way we could contribute some of those stuff to you guys, to make CocoaPods maybe that bit more robust.

segiddins:
Yes! Integration tests that build into `./build` would be acceptable, I think.

seaders:
Sorry @segiddins I don't fully get you?  Where would the integration tests live?  In this repo, or a different one?  Is there an XCode sample project, with a PodFile that I should make a PR for, or a similar process to that?  I've looked through the contribution docs, but haven't found somewhere detailing that side.

segiddins:
In the rakefile, there's code to build examples that you could probably build upon. Or the integration specs in `spec/integration.rb`

seaders:
Ok, that last commit has a simple test project, and I've altered the Rakefile to run `xcodebuild clean build test` if one of the Examples starts with `Test`.  If there's anything from what I've checked in against how you do things, just let me know and I'll fix it up.

seaders:
@segiddins I've put in a PR for the spec update, and will add another commit to this with an updated subproject reference if / when that gets merged in as well.

examples/TestInclusions/Podfile
```diff
@@ -0,0 +1,12 @@
+workspace 'Examples.xcworkspace'
+project 'TestInclusions.xcodeproj'
+platform :ios, '6.0'
+
+abstract_target 'TestInclusionsPods' do
+    pod 'SwrveConversationSDK', '~> 4.3.0'
+
+    target 'TestInclusions'
+    target 'TestInclusionsTests' do
+        pod 'OCHamcrest', '~> 5.0'
```
mrackwitz: For the sake of simplicity, I'd avoid unnecessary dependencies here and would rather use XCTest's built-in matchers.

mrackwitz:
I can confirm the problem and the solution from what I've seen with #5022. This looks to good to me. :+1: Thanks for going the extra mile and contributing to the test harness!

examples/TestInclusions/Podfile
seaders: Removed.  Sorry about that, I'm just so used to using OCHamcrest that I didn't even think about it!

(And here was me thinking I was showing off proper abstract_target usage too!)

seaders:
Not at all, I just hate seeing the ci checks failing on the PR.

seaders:
@mrackwitz sorry, I thought you were mentioning the changes to the specs repo.

Yeah for the test harness, I can definitely say there's selfish reasons behind that.  That side of our SDK has caused us loads of trouble with how XCode has changed what it does with resources, and the mess that is storyboards, in the last year or two, so having that as part of the tests for CocoaPods will only make our SDK more stable.

On the test, we only use storyboards, so I've only put a test in for that, but in the `*-resources.sh` file, you do things with all of

- storyboard
- xib
- framework
- xcdatamodel
- xcdatamodeld
- xcmappingmodel
- xcassets

And a generic `*` at the end, so I do think a few more tests could be done as well, in the inclusion test project.

efirestone:
Without trying to derail the conversation, it's worth again mentioning that we may be able to remove the copy-resources script almost completely now given my changes for handling resources in Xcode. The gotcha is that people will need to switch to specifying resources as `**/*` rather than `*/*`. The latter used to be used as a workaround because Pods didn't properly handle the internals of things like xcdatamodels or localization directories. It now understands these things, but they have to be included in the glob for it to work. We can't really automatically recurse with the old syntax since you may want to intentionally omit some resources within a bundle.

seaders:
Yeah, I remember this discussion from my last PR!  If/when you do get around to removing that copying, it'll still make the system more stable to have that bit more testing around and about to make sure it's working correctly, specifically on the latest and greatest XCode, where most of these issues normally occur.

efirestone:
Oh yeah, I'm absolutely all for more integration tests. Getting rid of the copy script just might have avoided this issue in the first place.

I'll leave it up to the core maintainers on whether or not they're willing to break lots of existing podspecs in the interest of letting Xcode do more of the heavy lifting.

orta:
> Without trying to derail the conversation, it's worth again mentioning that we may be able to remove the copy-resources script almost completely now given my changes for handling resources in Xcode.

The problem here is when you support per-config Pods, I believe

mrackwitz:
I can confirm what @orta states here. That wouldn't work with Xcode and is one reason, why we need to use this script instead.

seaders:
Any idea when the pair for this, https://github.com/CocoaPods/cocoapods-integration-specs/pull/64 (and this one) might get the go ahead?  Seems like ye are ok with these updates.

segiddins:
Merged the other PR

seaders:
@segiddins not sure what's going on with that build, it seems like it can't find our CocoaPod?

```Resolving dependencies of `Podfile`

[!] Unable to find a specification for `SwrveConversationSDK (~> 4.3.0)```

But running a `pod install` locally gets everything fine?

**Comment by segiddins** [comment]
@seaders might need to run a `pod repo update` on Travis?

**Comment by seaders** [comment]
Would that be better done in the rakefile, or can it be done manually there @segiddins?

**Comment by segiddins** [comment]
@seaders I'd put it in the `.travis.yml` for now

**Comment by seaders** [comment]
Gah! All sorts of odd fails going on.  Any chance you can take a look @segiddins?

**Comment by segiddins** [comment]
@seaders it needs to be `bundle exec pod repo update`

**Comment by seaders** [comment]
@segiddins Still nope :(

```
/Users/travis/build.sh: line 45: bundle: command not found

before_install.1
sudo gem install bundler

bundle exec pod repo update

bundler: command not found: pod

Install missing gem executables with 'bundle install'
```

**Comment by orta** [comment]
hey, nice, a :green_heart:

**Comment by seaders** [comment]
Ha, yeah!  Took more than a few tweaks and pushes and pokes, but got there in the end :)

There's always issues like this when your local rake is different to what's on the CI servers (which was one of the main reasons for this PR initially).

**Comment by mrackwitz** [comment]
Nice! :green_heart: Could you please squash some of these commits together? (It doesn't have to be just a single commit, but those which are reverted by succeeding commits don't really contribute to meaningful history. 🙂)

**Comment by seaders** [comment]
@mrackwitz done.  That mess squashed down to 2 commits now.

File: .travis.yml [hunk 2]
```diff
@@ -19,6 +19,8 @@ before_install:
   # There is a bug in travis. When using system ruby, bundler is not
   # installed and causes the default install action to fail.
   - sudo gem install bundler
+  - bundle install
+  - bundle exec pod repo update
 
 after_install:
   - ./bin/pod repo update --silent
```
segiddins: Is this redundant now?

CHANGELOG.md
```diff
@@ -18,6 +18,15 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 * None.  
 
 
+## 1.0.0.beta.7
```
segiddins: This should go under the master section

.travis.yml
seaders: I'm not really sure.  I don't know what the story is with Travis-CI at all, tbh.  With the rest of the tweaks I needed to get done (and how long it took), after I got everything working, I just left it in.

seaders:
@segiddins: no idea what's failing the build now :/

```
Bacon::Error: File comparison error `execution_output.txt` for install_vendored_dynamic_framework:

--- DIFF -----------------------------------------------------------------------

     git checkout -b new_branch_name

     Checking out files: 100% (630/630), done.

  > Running prepare command

    $ BASH_BIN -c  set -e sh build.sh cocoapods-setup without-core

--- END ------------------------------------------------------------------------
```

Never seen that before, and have no idea what it's related to, nor what might have caused it.  Doesn't seem related to my changes at all.

segiddins:
@seaders don't worry about that failure

seaders:
Ok grand.  Last thing is the travis edits then, I'll remove them if they are redundant, but is there a way we can check?  I'd be under the slight assumption that a pod repo update should always happen in the CI build?  I was fairly surprised that it wasn't happening, in that failure that was hitting me.

.travis.yml
segiddins: ok, so after reading the travis docs, I think we can back out the changes to `before_install`, and change this section to

```yaml
before_script:
  - bundle exec pod repo update --silent
```

seaders:
@segiddins so we're all good?  I see there's conflicts now, but it'd be great if this could get closed off.  Thanks again.

segiddins:
Yup!

mrackwitz:
I rebased this and manually merged it to master after CocoaPods/cocoapods-integration-specs#64 was already merged. :rocket:

Issue #5135, opened 2016-04-08
Improve helpfulness of spec-repo out-of-date possibility.


alloy:
```
[!] Unable to satisfy the following requirements:
- `Nimble-Snapshots` required by `Podfile`
- `Nimble-Snapshots (= 4.0.0)` required by `Podfile.lock`

None of the spec sources contain a spec satisfying the `Nimble-Snapshots` dependency.
You have either; mistyped the name or version, not added the source repo that hosts the Podspec to your Podfile, or not got the latest versions of your source repos.

None of the spec sources contain a spec satisfying the `Nimble-Snapshots (= 4.0.0)` dependency.
You have either; mistyped the name or version, not added the source repo that hosts the Podspec to your Podfile, or not got the latest versions of your source repos.
```

There are a few things that can be improved here:
* Don’t specify basically the same error message twice.
* With the change to no longer auto-update repos on `pod install`, it’s much more likely that the repo is out-of-date than the user mistyping anything, so that should move to the front otherwise you risk the user mentally signing-off after the first part of the message. This can be improved even more by separating the various suggestions into a list.

  > None of the spec sources contain a spec satisfying the `Nimble-Snapshots (= 4.0.0)` dependency. You have either:
  > * out-of-date source repos which you can update with `pod repo update`
  > * not added the source repo that hosts the Podspec
  > * or mistyped the name or version.

orta:
Probably worth mentioning the change:

> None of the spec sources contain a spec satisfying the `Nimble-Snapshots (= 4.0.0)` dependency. You have either:
> * out-of-date source repos which you can update with `pod repo update`
> * not added the source repo that hosts the Podspec
> * or mistyped the name or version.
>
>Note: as of CocoaPods 1.0, `pod repo update` is not called before `pod [install or update]` 

( where we change the update/install depending on the command given )

alloy:
`pod repo update` is still being called before `pod update`, no?

orta:
> pod repo update is still being called before pod update, no?

yes

PR #5136, opened 2016-04-08
Improve 1.0 error messages around not-found dependencies

orta:
The majority of the discussion is in #5135 - for now this will handle the implementation aspects

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,11 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Samuel Giddins](https://github.com/segiddins)
   [CocoaPods#5113](https://github.com/CocoaPods/CocoaPods/issues/5113)
 
+* Improved error messages, around the 1.0 changes when we cannot find a dependency 
+  in local sources, noting that CP > 1 doesnt run `pod repo update`
+  [Orta Therox](https://github.com/orta)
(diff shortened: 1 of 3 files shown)


orta:
Is this worth adding a changelog entry?

segiddins:
@orta, yes an entry under enhancements is called for since its closing an issue.

orta:
Done'd :+1:

segiddins:
You also need to update the tests :P

orta:
Hah - so I can, and have fixed one of the tests, but the other is the example where it appears twice, any thoughts on what the logic should look like to only show it once if it's `Podfile` & `Podfile.lock`?

```
- `Nimble-Snapshots` required by `Podfile`
- `Nimble-Snapshots (= 4.0.0)` required by `Podfile.lock`
```

```
None of the spec sources contain a spec satisfying the `Nimble-Snapshots` dependency.
You have either; mistyped the name or version, not added the source repo that hosts the Podspec to your Podfile, or not got the latest versions of your source repos.

None of the spec sources contain a spec satisfying the `Nimble-Snapshots (= 4.0.0)` dependency.
You have either; mistyped the name or version, not added the source repo that hosts the Podspec to your Podfile, or not got the latest versions of your source repos.
```

orta:
Or not, was just some dodgy string interpolation 🎉 - though that problem still remains

alloy:
👍

PR #5140, opened 2016-04-11
Sort Dir.glob result in PathList case insensitively

soutaro:
`Dir.glob` result on case sensitive file system may be different from the result on case insensitive file system. This would cause unintentional diffs on `-umbrella.h` files depending on the config of file system in which `pod install` is done, when `Pods` directory is in the repository.

1. `pod install` on case insensitive file system → files are sorted case insensitively
2. `git commit && git push`
3. `pod install` on case sensitive file system → files are sorted case sensitively
4. Diffs may appear on `-umbrella.h` files

This pull request is to fix this issue by sorting `Dir.glob` result explicitly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
-* None.  
+* Sort files from Dir.glob explicitly to produce same result on case sensitive
+  file system as result on case insensitive file system.  
+  [Soutaro Matsumoto](https://github.com/soutaro)
 
 
 ## 1.0.0.beta.6 (2016-03-15)
--- a/lib/cocoapods/sandbox/path_list.rb
+++ b/lib/cocoapods/sandbox/path_list.rb
@@ -49,7 +49,7 @@ module Pod
         end
         root_length  = root.to_s.length + 1
         escaped_root = escape_path_for_glob(root)
-        paths = Dir.glob(escaped_root + '**/*', File::FNM_DOTMATCH)
+        paths = Dir.glob(escaped_root + '**/*', File::FNM_DOTMATCH).sort_by(&:upcase)
         absolute_dirs  = paths.select { |path| File.directory?(path) }
         relative_dirs  = absolute_dirs.map  { |p| p[root_length..-1] }
         absolute_paths = paths.reject { |p| p == "#{root}/." || p == "#{root}/.." }
--- a/spec/unit/sandbox/path_list_spec.rb
+++ b/spec/unit/sandbox/path_list_spec.rb
@@ -166,6 +166,20 @@ module Pod
       end
     end
 
+    describe 'Reading file system' do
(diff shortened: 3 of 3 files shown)


segiddins:
Awesome, thanks! Think you could add a test + CHANGELOG entry so we can get this in the next release?

soutaro:
Added CHANGELOG entry.

CHANGELOG.md
```diff
@@ -20,7 +20,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
-* None.  
+* Sort files from Dir.glob explicitly to produce same result on case sensitive
+  file system as result on case insensitive file system.
+  [Soutaro Matsumoto](https://github.com/soutaro)
+  [#5140](https://github.com/CocoaPods/CocoaPods/pull/5140)
```
segiddins: don't link to the PR, since it isn't an issue

CHANGELOG.md
```diff
@@ -20,7 +20,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
-* None.  
+* Sort files from Dir.glob explicitly to produce same result on case sensitive
+  file system as result on case insensitive file system.
```
segiddins: please add two trailing spaces at the end of this line

soutaro:
Updated CHANGELOG, and I did `--amend`.

* Removed link to PR
* Added two spaces

segiddins:
Awesome! This is :+1: from me with the addition of a test case

mrackwitz:
@soutaro: I think it would be sufficient to test that by adding two or more files to the `banana-lib` fixture, which would be ordered differently and validating in an unit test in [path_list_spec.rb](https://github.com/CocoaPods/CocoaPods/blob/master/spec/unit/sandbox/path_list_spec.rb) that your change always keeps them in order, which would be expected for a case-sensitive file system.

soutaro:
Thanks!

I think I can add test to confirm my change does not break existing behavior. I will add files to `banana-lib`.

However I don't think it easy to have test for case sensitive file system case. Adding files to `banana-lib` does not make sense when the test is run on case insensitive file system. The best way would be to have another CI job to run tests on case sensitive file system. (or another way is to stub `Dir.glob` to simulate case sensitive case.)

segiddins:
Stubbing `Dir.glob` would be sufficient

soutaro:
Added a test stubbing `Dir.glob`.

segiddins:
https://travis-ci.org/CocoaPods/CocoaPods/jobs/122836236#L2936-L2953

soutaro:
Oh... Fixed.

segiddins:
👍

soutaro:
May I squash 8eb3519 and bbb5c74 ? (or others?)

segiddins:
Yup, you can squash if you'd like!

soutaro:
I did squash test commits and `git push -f`.

PR #5146, opened 2016-04-13
Fix non-modular header errors for vendored frameworks.

### LINKED ISSUE #4805: Still getting non-modular header inside framework errors with 1.0.0-beta2
Hi,
we are creating a vendored framework that is distributed as a binary in cocoapods.

When including our framework with swift code we were getting errors just like case:
XCode 7.1 - Include of non-modular header inside framework #4420

We noticed that these errors went away if we removed the Public headers directory created by cocoapods. This directory appears to just be symlinking to all of our public headers from our framework.

Taking a look at the commit for 4420 it looks like you removed the generation of headers for the Private headers that are generated from frameworks but not for the Public headers. I would suggest that this should also be done for the Public headers.

Please let us know if this is something that is going to be addressed. Of if we can modify our podspec to prevent the public headers being generated from our framework.

Thanks.

markspanbroek:
:rainbow:

Currently when including a vendored_framework in your podspec, its headers will be linked into the Pods/Headers/Public directory. This directory is added to the HEADER_SEARCH_PATH of any pod target that depends on it. Which will lead to a "Include of non-modular header inside framework module" error, because Xcode doesn't like it when header files are both available through header search paths and framework search paths (https://forums.developer.apple.com/thread/23554#81066).

This pull request fixes this issue. It is basically doing for vendored frameworks what already has been done for the build headers in PR #4476.

It is also likely fixes #4805.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,11 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
-* None.  
-
+* Headers from vendored frameworks no longer end up in the HEADER_SEARCH_PATH when 
+  using frameworks. They are assumed to be already present as modular headers in the
+  framework itself.  
+  [Mark Spanbroek](https://github.com/markspanbroek)
+  [#5146](https://github.com/CocoaPods/CocoaPods/pull/5146)
 
 ## 1.0.0.beta.7 (2016-04-15)
 
--- a/examples/Vendored Framework Example/Example Pods/VendoredFrameworkExample.podspec
+++ b/examples/Vendored Framework Example/Example Pods/VendoredFrameworkExample.podspec
@@ -1,0 +1,14 @@
+Pod::Spec.new do |s|
+  s.name                    = "VendoredFrameworkExample"
+  s.version                 = "0.0.1"
+  s.summary                 = "Vendored Framework in a spec test pod."
+  s.description             = "This spec specifies a vendored framework."
+
+  s.ios.deployment_target   = '8.0'
+  s.homepage                = "https://cocoapods.org"
+  s.license                 = { :type => "MIT", :file => "../../../../LICENSE" }
+  s.author                  = "Mark Spanbroek"
+  s.source                  = { :http => "https://github.com/AFNetworking/AFNetworking/releases/download/3.1.0/AFNetworking.framework.zip" }
+  s.ios.vendored_frameworks = "**/iOS/AFNetworking.framework"
+end
+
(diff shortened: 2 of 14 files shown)


segiddins:
@mrackwitz could you take a look at this one?

mrackwitz:
The change itself looks good to me with a unit spec and a changelog entry. :+1:
But history has shown that changes to the header search paths are quite volatile, so we should do some further testing beyond that. It would be also good, if we would have a build example with a vendored framework to ensure that such integrations keep building.

markspanbroek:
Apparently there is already a vendored framework integration spec: https://github.com/CocoaPods/cocoapods-integration-specs/tree/master/install_vendored_dynamic_framework
Does that suffice, or are you looking for something more specific?

Is there anything else that I can do to help test the change?

segiddins:
@markspanbroek since that test case is specifically for _dynamic_ vendored frameworks, its not going to show these changes anyways

markspanbroek:
Thanks @segiddins, I will see if I can come up with a new integration spec for this change.

markspanbroek:
I just added a build example that checks whether an app with a vendored framework builds correctly.

segiddins:
And that example fails without this fix applied?

markspanbroek:
Yes indeed, it fails with an "include of non-modular header" without the fix.

jafara:
Nice! Merge dat shit!

CHANGELOG.md
```diff
@@ -12,8 +12,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
-* None.  
-
+* Headers from vendored frameworks no longer end up in the HEADER_SEARCH_PATH when 
+  using frameworks. They are assumed to be already present as modular headers in the
+  framework itself.  
+  [Mark Spanbroek](https://github.com/markspanbroek)
```
mrackwitz: Please add a reference to either #4805 or this PR, which ever of both you think describe better what was the issue and what had to be done. While we tend to prefer adding the issue, if it can be considered "canoncial" / very descriptive for the problem and there are not a bunch of duplicates around it, which don't really describe the root of the problem.

mrackwitz:
Beside my outstanding comment, this looks indeed good to go. Terrific work! Thank you very much @markspanbroek for tackling this. :+1:

markspanbroek:
Thanks @mrackwitz, I decided upon adding a reference to this PR in the changelog because I think it describes the problem a little better than issue #4805.

segiddins:
Excellent!

PR #5162, opened 2016-04-15
[SourcesManager] Provide a deprecation path for access to Pod::SourcesManager

segiddins:
Who said a constant should be constant?

This is evil, but I think its necessary to ease the pain for third-party transition to CocoaPods 1.0. We will remove this hack very very soon, I promise.

\c @alloy @mrackwitz


alloy:
Edgy, I love it 👌

PR #5199, opened 2016-04-25
Make the utf-8 warning message respect --no-ansi

### LINKED ISSUE #5013: Hardcoded ansi color codes displayed when --no-ansi is passed on the command line
# Report

## What did you do?

Ran `pod install --no-ansi`

No text with ansi color codes will be displayed.

## What happened instead?

I get a warning message that uses ansi color codes to inform me to update my `~/.profile` with the text encoding of UTF-8.

The lines in question appear to be hardcoded to use color codes: https://github.com/CocoaPods/CocoaPods/blob/master/bin/pod#L3-L11

Kapin:
Closes #5013 

Just prints a different docstring now if we find --no-ansi in ARGV. I tried just printing the colour codes conditionally but it didn't seem to work in my local terminal. I'm sure there is a more "ruby" way to do this 😛

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
-* None.  
+* The warning for not having utf-8 set as the default encoding for a
+  terminal now properly respects the `--no-ansi` argument.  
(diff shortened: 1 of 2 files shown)


segiddins:
👍 with a CHANGELOG entry

segiddins:
👍

PR #5222, opened 2016-04-28
[XCConfig] Support customized build dirs in user project

### LINKED ISSUE #5217: Allow to use customized build dirs
# Report

## What did you do?

1. Setup a project and customized the `CONFIGURATION_BUILD_DIR`.
2. Run `pod install`.
3. Try to build the project.

In step 2, I would expect a warning telling me either what I need to do when I'm using a custom value for `CONFIGURATION_BUILD_DIR` or recommending me to remove the build settings from the target.

Alternatively, it would be nice if step 3 would succeed.

## What happened instead?

In step 2, I wasn't seeing a warning for conflicting overridden build settings.
In step 3 instead of successfully building my target, the build failed with a linker error, because the linker cannot find the pod targets' build products.

## Why does that happen?

The pods build products cannot be found because the user target XCConfig injects custom `LIBRARY_SEARCH_PATHS` / `FRAMEWORK_SEARCH_PATHS` which rely on a build setting, which has a different value in my project than in the `Pods.xcodeproj`. We shouldn't avoid relying on a user-customized value for the build directory. But we could allow customizing the `CONFIGURATION_BUILD_DIR` by just relying on `BUILD_DIR`. Even better would be to defining aliases in the user target XCConfig, which can then be overriden in the user target, if needed. Such overwrites should then obviously not cause warnings on installation-time.

```bash
PODS_BUILD_DIR=$(BUILD_DIR)
PODS_CONFIGURATION_BUILD_DIR="$(PODS_BUILD_DIR)/$(CONFIGURATION)$(EFFECTIVE_PLATFORM_NAME)"
```

Without a change around that, users would be required to redefine all these custom scoped search paths themselves, beginning from 1.0 with the implementation of the different scoping strategy.

mrackwitz:
Fixes #5217.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,11 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Orta Therox](https://github.com/orta)
   [#4714](https://github.com/CocoaPods/CocoaPods/issue/4714)
 
+* Support customized build directories by letting user xcconfig definitions
+  rely on the new overridable alias build variable `PODS_BUILD_DIR`.  
+  [Marius Rackwitz(https://github.com/mrackwitz]
+  [#5217](https://github.com/CocoaPods/CocoaPods/issue/5217)
+
 ##### Bug Fixes
 
 * Fix for `pod repo push --help` throwing an error.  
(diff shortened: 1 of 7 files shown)


segiddins:
👍

PR #5296, opened 2016-05-12
Remove 'Until we reach 1.0' from upgrading message since this happened

kylef:
It's no longer necessary to inform users that versions of CocoaPods less than 1 will frequently change when they are already running 1.0.0.

> Until we reach version 1.0 the features of CocoaPods can and will change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
-* None.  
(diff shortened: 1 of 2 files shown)


kylef:
🎊 Btw, congratualtions on shipping 1.0. You all did a great job!. 🎊

orta:
Thanks @kylef ! 

This is a dupe of https://github.com/CocoaPods/CocoaPods/pull/5280 ( which also include tests ) - so appreciated, but closing - I also spotted it just after 1.0

PR #5326, opened 2016-05-14
[PathList] Simplify read_file_system implementation, deal with unicode paths

marcboquet:
When working on #5324 I started with this file, but ended up seeing that the issue could be solved more broadly by normalizing `config.installation_root` at the very start.

The unicode issue (#5294) still remains when calling `PathList` by its own though, even if CocoaPods seems to carry the path from config so I don't think it would happen in normal use. But I thought maybe a fix here would be good too?

Also, while investigating the issue I changed `PathList#read_file_system` quite a bit hoping it would help. It ended up not helping but I liked the idea of leveraging `Pathname` for most of the work, I think it makes the method more readable.


lib/cocoapods/sandbox/path_list.rb
```diff
@@ -47,15 +50,20 @@ def read_file_system
...
+        dirs = []
         escaped_root = escape_path_for_glob(root)
-        paths = Dir.glob(escaped_root + '**/*', File::FNM_DOTMATCH).sort_by(&:upcase)
-        absolute_dirs  = paths.select { |path| File.directory?(path) }
-        relative_dirs  = absolute_dirs.map  { |p| p[root_length..-1] }
-        absolute_paths = paths.reject { |p| p == "#{root}/." || p == "#{root}/.." }
-        relative_paths = absolute_paths.map { |p| p[root_length..-1] }
-        @files = relative_paths - relative_dirs
-        @dirs  = relative_dirs.map { |d| d.gsub(/\/\.\.?$/, '') }.reject { |d| d == '.' || d == '..' } .uniq
+        paths = Pathname.glob(escaped_root + '**/*', File::FNM_DOTMATCH) do |path|
```
segiddins: imho this change make this method a bit more difficult to follow

segiddins:
Needs a CHANGELOG entry, and I'd kinda prefer to not merge the `read_filesystem` changes :/

lib/cocoapods/sandbox/path_list.rb
```diff
@@ -47,15 +50,12 @@ def read_file_system
...
-        relative_dirs  = absolute_dirs.map  { |p| p[root_length..-1] }
-        absolute_paths = paths.reject { |p| p == "#{root}/." || p == "#{root}/.." }
-        relative_paths = absolute_paths.map { |p| p[root_length..-1] }
-        @files = relative_paths - relative_dirs
-        @dirs  = relative_dirs.map { |d| d.gsub(/\/\.\.?$/, '') }.reject { |d| d == '.' || d == '..' } .uniq
+        absolute_paths = Pathname.glob(escaped_root + '**/*', File::FNM_DOTMATCH)
+        dirs_and_files = absolute_paths.reject { |path| path.basename.to_s =~ /^\.\.?$/ }
+        dirs, files = dirs_and_files.partition(&:directory?)
+        @dirs  = dirs.map{ |dir| dir.relative_path_from(root).to_s }.sort_by(&:upcase)
+        @files = files.map{ |file| file.relative_path_from(root).to_s }.sort_by(&:upcase)
```
marcboquet: I gave it another go, see if you like this one :)

My main concerns with the existing implementation are:
- Relying on the root_length variable to do all operations.
- Not leveraging existing ruby classes for path manipulation.
- Too much checking for "." and ".." in various forms, each of those is removing some of the instances.

I won't push any more if you still prefer the old way, I just wanted to try and explain my thinking.

marcboquet:
What would be in the Changelog for this one? I thought maybe it could go in the same entry as #5324? Or do all PRs need its own entry?

spec/unit/sandbox/path_list_spec.rb
```diff
@@ -6,6 +6,15 @@ module Pod
       @path_list = Sandbox::PathList.new(fixture('banana-lib'))
     end
 
+    it 'supports unicode paths' do
+      # Load fixture("ü") fixture with chars ["u", "̈"] instead of ["ü"]
+      unicode_name = [117, 776].pack('U*')
+      @path_list = Sandbox::PathList.new(fixture(unicode_name))
+      puts @path_list.root.to_s.chars.inspect
```
segiddins: this needs to be removed

lib/cocoapods/sandbox/path_list.rb
```diff
@@ -47,15 +50,12 @@ def read_file_system
...
-        paths = Dir.glob(escaped_root + '**/*', File::FNM_DOTMATCH).sort_by(&:upcase)
-        absolute_dirs  = paths.select { |path| File.directory?(path) }
-        relative_dirs  = absolute_dirs.map  { |p| p[root_length..-1] }
-        absolute_paths = paths.reject { |p| p == "#{root}/." || p == "#{root}/.." }
-        relative_paths = absolute_paths.map { |p| p[root_length..-1] }
-        @files = relative_paths - relative_dirs
-        @dirs  = relative_dirs.map { |d| d.gsub(/\/\.\.?$/, '') }.reject { |d| d == '.' || d == '..' } .uniq
+        absolute_paths = Pathname.glob(escaped_root + '**/*', File::FNM_DOTMATCH)
+        dirs_and_files = absolute_paths.reject { |path| path.basename.to_s =~ /^\.\.?$/ }
+        dirs, files = dirs_and_files.partition(&:directory?)
```
segiddins: can we do the `relative_path_from(root)` and `sort_by` before the partition?

segiddins:
@marcboquet a single CHANGELOG entry for this is fine, since we're fixing the same bug in both

lib/cocoapods/sandbox/path_list.rb
```diff
@@ -47,15 +50,12 @@ def read_file_system
...
-        relative_dirs  = absolute_dirs.map  { |p| p[root_length..-1] }
-        absolute_paths = paths.reject { |p| p == "#{root}/." || p == "#{root}/.." }
-        relative_paths = absolute_paths.map { |p| p[root_length..-1] }
-        @files = relative_paths - relative_dirs
-        @dirs  = relative_dirs.map { |d| d.gsub(/\/\.\.?$/, '') }.reject { |d| d == '.' || d == '..' } .uniq
+        absolute_paths = Pathname.glob(escaped_root + '**/*', File::FNM_DOTMATCH)
+        dirs_and_files = absolute_paths.reject { |path| path.basename.to_s =~ /^\.\.?$/ }
+        relative_paths = dirs_and_files.map { |path| path.relative_path_from(root) }
+        sorted_paths = relative_paths.map(&:to_s).sort_by(&:upcase)
+        @dirs, @files = sorted_paths.partition { |path| File.directory?(root + path) }
```
marcboquet: Good call @segiddins, I like it with the partition at the end.

marcboquet: This can be `File.directory?(root + path)` or also `(root + path).directory?` if we want to stick with Pathname methods.

segiddins: either one is fine

segiddins:
👍

PR #5333, opened 2016-05-15
Update an info message for 1.x: Change xcodeproj → project

araile:
Fixes an informative message that appears when the project can't be found.

This is what was seeing before:

<pre>[!] Could not automatically select an Xcode project. Specify one in your Podfile like so:

    xcodeproj 'path/to/Project.xcodeproj'

[...]

[!] xcodeproj was renamed to `project`. Please use that from now on.</pre>


segiddins:
👍🏻

PR #5335, opened 2016-05-16
[Validator] Adds space to fix typo in output in Validator

samjarman:
Before: `The pattern includes header files that are not listedin source_files`
After: `The pattern includes header files that are not listed in source_files`


segiddins:
👍 thanks @samjarman !

PR #5371, opened 2016-05-19
[Repo Push] Help validation

nwest:
`super` should be first in the case that the help flag is passed in.
https://github.com/CocoaPods/CLAide/blob/master/lib/claide/command.rb#L543-L550


segiddins:
👍

Issue #5376, opened 2016-05-19
Use `git ls-remote` to avoid git clones of `:branch` dependencies


champo:
- [x] I've read and understood the [*CONTRIBUTING guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

Right now whenever you run `pod update` and have dependencies against a git branch, pods always clones the target repository. This happens even if the target commit is cached. Using `git ls-remote` you can query the repository for the current commit the target branch points to, and avoid cloning if that particular version is in pods cache.

I already have a proof of concept patch for this, but before sending it I'd like to know if this would be accepted. I also seriously doubt the patch would survive code review, so I'd have some questions regarding it if it's acceptable.

orta:
From my perspective it seems like a good idea

segiddins:
That'd be super welcome!

champo:
:D

About the implementation itself, I added a not-so-nice check in `Downloader::Cache` that if the `Request` object has `:git` and `:branch` symbols in its params. Looking at how most of the code is git agnostic this doesn't seem like a good place for it. 

Maybe adding a step where each concrete downloader implementation can preprocess the request before the cache does it's magic? I'm cleaning up the code a bit to send a WIP PR.

segiddins:
@champo I think that sort of functionality in cocoapods-downloader would be _awesome_

orta:
Looks like this is in 👍

PR #5386, opened 2016-05-20
[Cache] Use `git ls-remote` to skip full clones for branch dependencies

champo:
This adds the posibility of caches hits for git branch dependencies.
Previously, since pods didn't know the commit the branch points it
always did a full clone of the repo. Using `ls-remote` we can check the
commit the branch points to and reuse cached version of the pods if
available.

Related to https://github.com/CocoaPods/CocoaPods/issues/5376

## TODO

- [x] Merge https://github.com/CocoaPods/cocoapods-downloader/pull/57
- [x] Merge https://github.com/CocoaPods/cocoapods-integration-specs/pull/71
- [x] Update changelog
- [x] Tests

### DIFF
--- a/lib/cocoapods/downloader.rb
+++ b/lib/cocoapods/downloader.rb
@@ -33,11 +33,16 @@ module Pod
       cache_path: Config.instance.cache_root + 'Pods'
     )
       can_cache &&= !Config.instance.skip_download_cache
+
+      request = preprocess_request(request)
+
       if can_cache
         raise ArgumentError, 'Must provide a `cache_path` when caching.' unless cache_path
         cache = Cache.new(cache_path)
         result = cache.download_pod(request)
       else
+        raise ArgumentError, 'Must provide a `target` when caching is disabled.' unless target
+
         require 'cocoapods/installer/pod_source_preparer'
         result, = download_request(request, target)
         Installer::PodSourcePreparer.new(result.spec, result.location).prepare!
@@ -108,6 +113,21 @@ module Pod
       else
         downloader.checkout_options
       end
+    end
+
+    # Return a new request after preprocessing by the downloader
+    #
+    # @param  [Request] request
+    #         the request that needs preprocessing
+    #
+    # @return [Request] the preprocessed request
+    #
+    def self.preprocess_request(request)
+      Request.new(
+        :spec => request.spec,
+        :released => request.released_pod?,
+        :name => request.name,
+        :params => Downloader.preprocess_options(request.params))
     end
 
     public
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Peter Ryszkiewicz](https://github.com/pRizz)
   [#5414](https://github.com/CocoaPods/CocoaPods/pull/5414)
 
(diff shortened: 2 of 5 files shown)


lib/cocoapods/downloader/request.rb
```diff
@@ -21,7 +21,7 @@ class Request
 
       # @return [Hash<Symbol, String>] The download parameters for this request.
       #
-      attr_reader :params
+      attr_accessor :params
```
segiddins: i'd rather keep `Request` immutable and just construct a new instance instead of mutating?

segiddins:
🎆 you can actually change the Gemfile to point to your branch for now so we can try this out :D

lib/cocoapods/downloader/request.rb
champo: I was torn on this. I agree that adding mutation here isn't the nicest thing. But it seemed weird to me that the request instance that's held by the user of the downloader would stop being "truthful". Looking at the call sites for `Downloder.download_request` the request is never even saved to a variable so I think I'll change it to be immutable again.

champo:
@segiddins done!

champo:
@segiddins I'm not sure how to go about adding tests for this, do you have any ideas?

segiddins:
In https://github.com/CocoaPods/CocoaPods/blob/master/spec/unit/downloader/cache_spec.rb, maybe just make sure the options are preprocessed?

lib/cocoapods/downloader.rb
```diff
@@ -33,11 +33,16 @@ def self.download(
...
       can_cache &&= !Config.instance.skip_download_cache
+
+      request = preprocess_request(request)
+
       if can_cache
         raise ArgumentError, 'Must provide a `cache_path` when caching.' unless cache_path
         cache = Cache.new(cache_path)
         result = cache.download_pod(request)
       else
+        raise ArgumentError, 'Must provide a `target` when caching is disabled.' unless target
```
champo: @segiddins this is a bit out of scope in this PR but I stumbled across it writing the test. Does this seem right to you?

segiddins: If I recall this is because otherwise you don't know where to download to

champo: exactly. I stumbled upon it throwing a random error farther down the call chain so I copied the check for `cache_path`.

I've added a simple test btw. Considering the coverage over at `cocoapods-downloader` I think this is enough, but I can add more if you disagree.

segiddins:
👍🏻
:ship:

talesp:
I'm trying to use cocoapods right form the sources (setting my Gemfile as `gem 'cocoapods', :git => 'https://github.com/CocoaPods/CocoaPods.git'`) because I want try the fix of the merge from #5503, since I'm getting the error `The bundle “ishopMobileKitTests” couldn’t be loaded because it is damaged or missing necessary resources. Try reinstalling the bundle.` with version 1.0.1.

But now I'm getting an error on pod install, and looks like the problem is caused by this merge.

`NoMethodError - undefined method `preprocess_options' for Pod::Downloader:Module`

`preprocess_request` calls `preprocess_options`, but I didn't find this method definition...

talesp:
Ok, nevermind, I added `gem 'cocoapods-downloader', :git => 'https://github.com/CocoaPods/cocoapods-downloader.git'` and the install was completed. My fault :)

PR #5436, opened 2016-06-01
Add a license type to generated acknowledgements file in plist

naoty:
This pull request adds a license, such as `MIT` or `Commercial`, to generated `Pods-{Project}-acknowledgements.plist`. I send this PR because I want to filter some licenses by their types. It is useful to show licenses only for open libraries, not private libraries.

### DIFF
--- a/lib/cocoapods/generator/acknowledgements/plist.rb
+++ b/lib/cocoapods/generator/acknowledgements/plist.rb
@@ -37,6 +37,7 @@ module Pod
             :Type => 'PSGroupSpecifier',
             :Title => sanitize_encoding(spec.name),
             :FooterText => sanitize_encoding(license),
+            :License => spec.license[:type],
           }
         end
       end
--- a/lib/cocoapods/generator/acknowledgements/plist.rb
+++ b/lib/cocoapods/generator/acknowledgements/plist.rb
@@ -37,7 +37,7 @@ module Pod
             :Type => 'PSGroupSpecifier',
             :Title => sanitize_encoding(spec.name),
             :FooterText => sanitize_encoding(license),
-            :License => spec.license[:type],
+            :License => sanitize_encoding(spec.license[:type]),
           }
         end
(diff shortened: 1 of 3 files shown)


lib/cocoapods/generator/acknowledgements/plist.rb
```diff
@@ -37,6 +37,7 @@ def hash_for_spec(spec)
             :Type => 'PSGroupSpecifier',
             :Title => sanitize_encoding(spec.name),
             :FooterText => sanitize_encoding(license),
+            :License => spec.license[:type],
```
segiddins: This might also need a `sanitize_encoding`?

segiddins:
@orta thoughts?

orta:
I'm half-half leaving heavily towards 👍. 

We don't have any strong reasons to outright deprecate acknowledgements anytime soon, and it looks well tested. IMO we should be including licenses as much as possible in user's projects to make sure people aren't accidentally breaking them via tooling.

lib/cocoapods/generator/acknowledgements/plist.rb
naoty: I missed it. I'll add `sanitize_encoding` soon.

naoty:
If there are nothing wrong, would you merge this PR?

segiddins:
@naoty CI seems to have failed

naoty:
@segiddins OK. But, I don't know where I should fix, because I have never used this testing framework. So, would you tell me where to fix?

segiddins:
Ah looks like it's only the integration specs, we can update those posy-merge 👍🏻

naoty:
@segiddins Thanks! Would you update them?

segiddins:
Manually merged via https://github.com/CocoaPods/CocoaPods/commit/8447331639fa6a73eda3eecf8208d80704dfdb35

PR #5497, opened 2016-06-09
use `pr_title` instead of `pr.title` in Dangerfile

DanToml:
Danger was crashing because `pr` was undefined, looks like this was a typo for `pr_title`


orta:
Yep, looks correct 👍

segiddins:
👍🏻

Issue #4014, opened 2015-08-12
Should error during install when there are duplicate library names


samurai00:
I am building an app which uses cocoapods to handle my external dependencies.
Cocoapods version 0.37.2 this works fine.
But I got error after I upgrade Cocoapods to 0.38.2.
Xcode version: 6.4
My Podfile:
```
platform :ios, "7.0"
pod "Pingpp", "~> 2.1.0"
pod "AFNetworking"
```
error:
```
Undefined symbols for architecture arm64:
  "_OBJC_CLASS_$_PingppInternal", referenced from:
      l_OBJC_$_CATEGORY_PingppInternal_$_Alipay in libpingpp+alipay.a(PingppInternal+Alipay.o)
      l_OBJC_$_CATEGORY_PingppInternal_$_Bfb in libpingpp+bfb.a(PingppInternal+Bfb.o)
      l_OBJC_$_CATEGORY_PingppInternal_$_Up in libpingpp+up.a(PingppInternal+Up.o)
      l_OBJC_$_CATEGORY_PingppInternal_$_Wx in libpingpp+wx.a(PingppInternal+Wx.o)
  "_OBJC_CLASS_$_PingppUtil", referenced from:
      objc-class-ref in libpingpp+alipay.a(PingppInternal+Alipay.o)
      objc-class-ref in libpingpp+bfb.a(PingppViewController.o)
      objc-class-ref in libpingpp+bfb.a(PingppInternal+Bfb.o)
      objc-class-ref in libpingpp+up.a(PingppInternal+Up.o)
      objc-class-ref in libpingpp+wx.a(PingppInternal+Wx.o)
  "_PingppLocalizedString", referenced from:
      -[PingppInternal(Alipay) alipayResult:] in libpingpp+alipay.a(PingppInternal+Alipay.o)
      -[PingppInternal(Bfb) channelBfb:withViewController:] in libpingpp+bfb.a(PingppInternal+Bfb.o)
      ___53-[PingppInternal(Bfb) channelBfb:withViewController:]_block_invoke103 in libpingpp+bfb.a(PingppInternal+Bfb.o)
      -[PingppInternal(Up) channelUpmp:withViewController:] in libpingpp+up.a(PingppInternal+Up.o)
      -[PingppInternal(Wx) onResp:] in libpingpp+wx.a(PingppInternal+Wx.o)
ld: symbol(s) not found for architecture arm64
```

[...]

PR #5505, opened 2016-06-11
[Installer] Verify no duplicate static libs

DanToml:
Closes https://github.com/CocoaPods/CocoaPods/issues/4014

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Daniel Tomlinson](https://github.com/dantoml)
   [#5480](https://github.com/CocoaPods/CocoaPods/pull/5480)
 
-* Add the ability to inhibit swift warnings.
+* Add the ability to inhibit swift warnings.  
   [Peter Ryszkiewicz](https://github.com/pRizz)
   [#5414](https://github.com/CocoaPods/CocoaPods/pull/5414)
 
@@ -74,6 +74,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Boris Bügling](https://github.com/neonichu)
   [#5528](https://github.com/CocoaPods/CocoaPods/issues/5528)
 
+* Error during install when there are duplicate library names.  
+  [Daniel Tomlinson](https://github.com/dantoml)
+  [#4014](https://github.com/CocoaPods/CocoaPods/issues/4014)
 
 ## 1.0.1 (2016-06-02)
 
--- a/lib/cocoapods/installer.rb
+++ b/lib/cocoapods/installer.rb
@@ -110,7 +110,7 @@ module Pod
       prepare
       resolve_dependencies
       download_dependencies
-      verify_no_duplicate_framework_names
+      verify_no_duplicate_framework_and_library_names
       verify_no_static_framework_transitive_dependencies
       verify_framework_usage
       generate_pods_project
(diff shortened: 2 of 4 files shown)


lib/cocoapods/installer.rb
```diff
@@ -399,6 +400,21 @@ def verify_no_duplicate_framework_names
       end
     end
 
+    def verify_no_duplicate_library_names
```
manuyavuz: @DanToml what about adding this check into `verify_no_duplicate_framework_names` and rename it `verify_no_duplicate_framework_and_library_names`? Or moving common logic into a base method and use it inside these two methods? Currently these two functions are almost the same, and could be handled together in a method, or use a common base method.

CHANGELOG.md
```diff
@@ -26,6 +26,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Samuel Giddins](https://github.com/segiddins)
   [#5321](https://github.com/CocoaPods/CocoaPods/issues/5321)
 
+* Error during install when there are duplicate library names.
```
manuyavuz: Missing trailing two spaces.

lib/cocoapods/installer.rb
DanToml: Good question. I don't generally like to add layers of abstraction for fewer than three uses (especially on code that doesn't change often), but we could do something along the lines of:

```ruby
def verify_no_duplicate_frameworks_and_libraries_names
    aggregate_target.user_build_configurations.keys.each do |config|
      pod_targets = aggregate_target.pod_targets_for_build_configuration(config)
      verify_no_duplicate_framework_names_in_pod_targets(targets)
      verify_no_duplicate_library_names_in_pod_targets(targets)
    end
  end
end

def verify_no_duplicate_framework_names_in_pod_targets(targets)
  vendored_frameworks = pod_targets.flat_map(&:file_accessors).flat_map(&:vendored_frameworks).uniq
  frameworks = vendored_frameworks.map { |fw| fw.basename('.framework') }
  frameworks += pod_targets.select { |pt| pt.should_build? && pt.requires_frameworks? }.map(&:product_module_name)

  duplicates = frameworks.group_by { |f| f }.select { |_, v| v.size > 1 }.keys
  unless duplicates.empty?
    raise Informative, "The '#{aggregate_target.label}' target has " \
      "frameworks with conflicting names: #{duplicates.to_sentence}."
  end
end

def verify_no_duplicate_library_names_in_pod_targets(targets)
  libraries = pod_targets.flat_map(&:file_accessors).flat_map(&:vendored_libraries).uniq.map(&:basename)
  libraries += pod_targets.select { |pt| pt.should_build? && !pt.requires_frameworks? }.map(&:product_name)
  duplicates = libraries.group_by { |l| l.to_s.downcase }.select { |_, v| v.size > 1 }.keys
  unless duplicates.empty?
    raise Informative, "The '#{aggregate_target.label}' target has " \
      "libraries with conflicting names: #{duplicates.to_sentence}."
  end
end
```

but I'm not sure that we actually get too much from it

CHANGELOG.md
DanToml: Oops, was stripped by vim. Will fix now.

segiddins:
👍🏻

spec/unit/installer_spec.rb
```diff
@@ -305,6 +305,28 @@ def @installer.analyze(*)
 
     #-------------------------------------------------------------------------#
 
+    describe '#verify_no_duplicate_library_names' do
```
mrackwitz: This should either refer to an actual existing method or be put into natural case.

mrackwitz:
@DanToml: Could you rebase this branch and address my comment above, so that we can get this merged and :ship:ed?

DanToml:
@mrackwitz Yeah 👍, will get it sorted this evening.

orta:
This'll need nother rebase - sorry

DanToml:
@orta Done, should be good to :shipit: when CI is done.

PR #5575, opened 2016-06-28
[ISSUE_TEMPLATE] Make sections collapsible

mrackwitz:
/c @orta

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,7 +1,10 @@
-> ℹ Please fill out this template when filing an issue.
-> All lines beginning with an ℹ symbol instruct you with
-> what info we expect.  
-> Please remove this line and all above before submitting.
+<details><summary>
+ℹ Please fill out this template when filing an issue.
+</summary>
+All lines beginning with an ℹ symbol instruct you with
+what info we expect.  
+Please remove this line and all above before submitting.
+</details>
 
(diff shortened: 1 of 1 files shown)


orta:
Interesting, when I go to the page for this, it all seems really unreliable WRT what it shows

https://github.com/CocoaPods/CocoaPods/blob/5b55147aee9eec8b71dee2bf4ff3f9dad3b6a09b/.github/ISSUE_TEMPLATE.md

mrackwitz:
How do you mean unreliable? The first section with instructions should be collapsed by default, while all others are expanded by default.

orta:
The behavior of the open/closing is unpredictable:

orta:
I bet it needs the `<summary>`  - http://www.w3schools.com/tags/tag_details.asp / https://gist.github.com/orta/057eab0832773aa299a6ebd39187ab06

mrackwitz:
Okay, that seems to be more a general rendering issue by the browser, maybe that's caused by skipping the `<summary>` element? I'm unable to reproduce that locally. Which browser do you use?

orta:
I'm on Safari on the `10.11.6 Beta (15G19a)`

mrackwitz:
Well as that primarily affect us as developers, we should defer this change at least until the beta is fixed. I looked into the WebKit bugtracker, but didn't found anything related. Maybe worth a radar.

orta:
👍

orta:
Checked again, still inconsistent behaviour on on Safari

orta:
Took another look, still hanky - gonna close this, thanks for trying @mrackwitz 👍

PR #5269, opened 2016-05-07
[Travis] Update RubyGems

segiddins:
Hopefully this fixes things 😁

PR #5640, opened 2016-07-18
[Travis] Account for system ruby going from p645 -> p648

segiddins:
See https://github.com/CocoaPods/CocoaPods/pull/5269


DanToml:
:green_apple: - installing CocoaPods still works on system Ruby so this seems fine to me.

PR #5732, opened 2016-08-10
Set ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES when appropriate

benasher44:
It looks like we should be setting this new flag in Xcode 8:
- This new flag has made #4203 crop up again.
- According to research done in #5598, this deprecates the old `EMBEDDED_CONTENT_CONTAINS_SWIFT` flag.

This shouldn't be merged until the following is done:
- [x] Someone participating in #4203 has reported submitting to the App Store with this PR using Xcode 8 was successful

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -9,7 +9,7 @@
 	url = https://github.com/tonymillion/Reachability.git
 [submodule "spec/cocoapods-integration-specs"]
 	path = spec/cocoapods-integration-specs
-	url = https://github.com/benasher44/cocoapods-integration-specs.git
+	url = https://github.com/CocoaPods/cocoapods-integration-specs.git
 [submodule "spec/fixtures/spec-repos/test_repo"]
 	path = spec/fixtures/spec-repos/test_repo
 	url = https://github.com/CocoaPods/cocoapods-test-specs.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -9,7 +9,7 @@
 	url = https://github.com/tonymillion/Reachability.git
 [submodule "spec/cocoapods-integration-specs"]
 	path = spec/cocoapods-integration-specs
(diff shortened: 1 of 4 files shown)


benasher44:
Ack: will fix the xcconfig diff failures.

segiddins:
👍 once confirmed

benasher44:
Looks like I still missed a few xcconfig files (that didn't show up as failures locally somehow?). Will fix in the integration specs repo and update.

benasher44:
Okay this fixed is confirmed, so we're all set! The only Travis failure is the spurious `execution_output.txt` one, so CocoaPods/cocoapods-integration-specs#76 is ready to merge. I'll update this PR once that's done.

.gitmodules
```diff
@@ -9,7 +9,7 @@
 	url = https://github.com/tonymillion/Reachability.git
 [submodule "spec/cocoapods-integration-specs"]
 	path = spec/cocoapods-integration-specs
-	url = https://github.com/CocoaPods/cocoapods-integration-specs.git
+	url = https://github.com/benasher44/cocoapods-integration-specs.git
```
segiddins: This needs to be reverted

segiddins:
🚢 once the submodule ref is changed back to the CocoaPods repo

.gitmodules
benasher44: I should have called this out more specifically, but that's what I meant about waiting on CocoaPods/cocoapods-integration-specs#76. I wait on that to be merged, and then I update the submodule references/pointers here right?

segiddins: well you have push now so you can just make your branch in the CocoaPods repo, I thought you'd done that

benasher44: >< yep I'll do that now.

poohwifey (commented):
I

Issue #5737, opened 2016-08-11
regression in 1.1.0.beta.1: can't install with multiple targets


Coeur:
Note: could be a duplicate of either #5661 or #5700, but my Podfile looking quite different, I feel like to post it
The issue is that installing my targets with 1.1.0.beta.1 doesn't work. It is working with 1.0.1.
Note2: the target "MyApp-UnitTests" has no host: it's independent from the other two. The project is currently for Xcode 8 beta 4 (using Swift 2.3)

### Command

```
/usr/local/bin/pod install
```

### Stack

```
   CocoaPods : 1.1.0.beta.1
        Ruby : ruby 2.0.0p648 (2015-12-16 revision 53162) [universal.x86_64-darwin15]
    RubyGems : 2.0.14.1
        Host : Mac OS X 10.11.6 (15G31)
       Xcode : 7.3.1 (7D1014)
         Git : git version 2.7.4 (Apple Git-66)
Ruby lib dir : /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib
Repositories : example-pods-repo - ssh://git@stash.example.com:7999/mobile/pods-repo.git @ f974d4fe849d999f7aeaf7635f10b393a79c7e5a
               master - https://github.com/CocoaPods/Specs.git @ 0d40c4864f501eb51f6703884330cb2780ac6ac2
```

### Plugins

```
cocoapods-deintegrate : 1.0.0
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
cocoapods-trunk       : 1.0.0
cocoapods-try         : 1.1.0
```

### Podfile

```ruby
platform :ios, '8.0'
use_frameworks!

abstract_target 'default_pods' do
    # swift dependencies
    pod 'Alamofire', :git => 'https://github.com/Alamofire/Alamofire.git', :branch => 'swift2.3'
    
    target 'MyApp-Release'
    target 'MyApp-Debug'
    target 'MyApp-UnitTests'
end

post_install do |installer|
    installer.pods_project.targets.each do |target|
        target.build_configurations.each do |config|
            config.build_settings['SWIFT_VERSION'] = '2.3'
        end
    end
end
```

### Error

```
Errno::EEXIST - File exists - /Users/antoinecoeur/Development/MyApp/Pods/Target Support Files/Alamofire
[...]

PR #5747, opened 2016-08-13
Verify embedded target platform/swift matches its host

benasher44:
The new swift build settings are a bit confusing, and my coworkers and I were able to reproduce what we initially thought was the same issues as #5700 and #5737. It turned out to be slightly different. We don't validate that the swift version of the embedded target matches that of its host. This change adds that validation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 * Remove references to the pre-1.0 Migrator.  
   [Danielle Tomlinson](https://github.com/dantoml)
   [#5635](https://github.com/CocoaPods/CocoaPods/pull/5635)  
-  
+
 * Improve performance of dependency resolution.
   [yanzhiwei147](https://github.com/yanzhiwei147)
   [#5510](https://github.com/CocoaPods/CocoaPods/pull/5510)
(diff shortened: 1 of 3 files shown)


segiddins:
Is it a difference in ruby versions, per chance?

CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

```markdown
* Uniq pod targets by name during install 
  [benasher44](https://github.com/benasher44)
```

</p>

benasher44:
Double checked that we were both using El Cap system ruby

benasher44:
Tests coming soon, but with this update, you'll get the following message when there's this kind of issue:
```
[!] Unable to integrate the following embedded targets with their respective host targets (a host target is a "parent" target which embeds a "child" target like a framework or extension):

- Extension (3.0) and App () do not both use the same Swift version.
```

Only remaining issue here though is there seems to be another bug. `App` does use swift, but it's legacy swift version setting is set at the project level. When the legacy swift version setting is configured at the project level, `TargetDefition#swift_version` returns `nil`. @DanToml, I think you added `swift_version`?

benasher44:
@DanToml nvm. I see where to fix. I'll have a separate PR up soon.

benasher44:
Still missing specs. Expect to get to this by Friday.

benasher44:
Message now looks like this:

```
[!] Unable to integrate the following embedded targets with their respective host targets (a host target is a "parent" target which embeds a "child" target like a framework or extension):

- App (2.3) and Extension (3.0) do not both use the same Swift version.
```

benasher44:
Only failure is the `execution_output.txt` one

PR #5726, opened 2016-08-09
Added support for message extensions embedded in messages applications

benasher44:
Fixes #5594 

During analysis, we collect information on the product types of the host targets for the embedded targets. Then we use that to help determine if `requires_host_target?` is true or false. The addition here is that it's now false when the host target is a messages application because this type of target does not have a user-modifiable (in the Xcode 8 UI) target dependency list/build phase.

PR #5769, opened 2016-08-18
Improved clarity of title for #5726 summary entry.

zygoat:
Capitalized “Messages” to make clear that it is a product name reference (Mac/iOS Messages.app), and added a verb to clarify that the fix pertains to the end use of CocoaPods as opposed to its integration with something else.


orta:
👍

PR #5787, opened 2016-08-24
[Create] Pass the Pod::version as a ENV var for the configure script to read

orta:
The new pod-template isn't backwards compatible with older versions of CocoaPods due to linking to a framework, https://github.com/CocoaPods/pod-template/pull/187 - this gives it the ability to let people know that.

### DIFF
--- a/lib/cocoapods/command/lib/create.rb
+++ b/lib/cocoapods/command/lib/create.rb
@@ -75,7 +75,7 @@ module Pod
           UI.section("Configuring #{@name} template.") do
             Dir.chdir(@name) do
               if File.exist?('configure')
-                system('./configure', @name, *@additional_args)
+                system({ 'COCOAPODS_VERSION' => Pod::Version }, './configure', @name, *@additional_args)
               else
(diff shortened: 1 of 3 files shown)


lib/cocoapods/command/lib/create.rb
```diff
@@ -75,7 +75,7 @@ def configure_template
           UI.section("Configuring #{@name} template.") do
             Dir.chdir(@name) do
               if File.exist?('configure')
-                system('./configure', @name, *@additional_args)
+                system({ 'POD_VERSION' => Pod::Version }, './configure', @name, *@additional_args)
```
segiddins: `COCOAPODS_VERSION`?

orta: I'm cool with that 👍

orta:
These CIs are all passing, but I get 

```
Done. Your build exited with 0.
/Users/travis/build.sh: line 109: shell_session_update: command not found
```

at the end, is this happening to others?

Issue #5807, opened 2016-09-01
adding --sources to pod update (and possibly install)


maschall:
🌈
- [X] I've read and understood the [*CONTRIBUTING guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

I started looking into implementing this functionality and wanted to start the conversation here.

I work on a project where we share a lot of internal dependencies.  These dependencies (for better or worse) update frequently, such that we made it part of our workflow to `pod update` in Jenkins to make sure we have the latest.  However this means that 3rd party dependencies are also update frequently when not necessarily desired.

What I'd like to do is tell update to only update pods that are in a specific cocoapod repo.

The suggested syntax would be `pod update --sources=InternalRepo,master` and for our team, we'd just use `pod update --sources=InternalRepo` to skip any updates in master.

`--sources` would default to all repos currently added in the Podfile, like we do for `lint` and `repo push` commands.

Additionally, I'm thinking the same code could also be used in `pod install` to explicitly prevent dependencies not in a specific source.  For instance if you wanted to force all dependencies be stored in a InternalRepo.  (This may already be handled with the Podfile source syntax)

orta:
Interesting, I'm a 👍 for this, seems like a useful and well scoped idea

orta:
This is merged 👍

zhigangfang:
@maschall the '--sources' syntax didn't skip master update, using pod version 1.2.0

PR #5809, opened 2016-09-02
[update] adding --sources to specify to only update pods from a repo

maschall:
🌈  this is for #5807 

Again the primary example for the use of this is that, I want to update all the pods in my private Pods repo but not any from master.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
-* None.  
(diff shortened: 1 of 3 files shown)


dnkoutso:
Hey @maschall, we are also hitting cases where the master repo is very slow to update for us. Do you plan on finishing this? If no I can probably help take it over the finish line for you.

maschall:
@dnkoutso I can pick it back up, sorry for breaking the tests, I thought I was green originally.

Thanks for the bump

orta:
Fixes https://github.com/CocoaPods/CocoaPods/issues/5807

PR #5822, opened 2016-09-08
Fixed pod repo push validation and push issues related to Dir.chdir

### LINKED ISSUE #5805: 'pod spec lint' fails with message 'App.xcworkspace' does not exist. 🌈
* [X] I've read and understood the [*CONTRIBUTING guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?
Run 'pod spec lint --verbose' on 'https://github.com/psturm-swift/SwiftySignals/blob/Swift_3.0/SwiftySignals.podspec'

I expect that the validation passes.

## What happened instead?
Validation ends with an error saying that App.xcworkspace does not exist. However, if I run pod spec lint with option --no-clean the workspace is there but it is not found by pod spec lint.

### Stack

```
   CocoaPods : 1.1.0.beta.1
        Ruby : ruby 2.3.0p0 (2015-12-25 revision 53290) [x86_64-darwin15]
    RubyGems : 2.5.1
        Host : Mac OS X 10.11.6 (15G31)
       Xcode : 8.0 (8S201h)
         Git : git version 2.8.4 (Apple Git-73)
Ruby lib dir : /Users/psturm/.rvm/rubies/ruby-2.3.0/lib
Repositories : master - https://github.com/CocoaPods/Specs.git @ 34ac2d09edb2bf66b9cb6badcccdc32c99fd1192
```

### Installation Source

```
Executable Path: /Users/psturm/.rvm/rubies/ruby-2.3.0/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.1
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
cocoapods-trunk       : 1.0.0
cocoapods-try         : 1.1.0
```

https://github.com/psturm-swift/SwiftySignals/tree/Swift_3.0

benasher44:
Removes uses of `Dir.chdir` for this flow. It prefers using `git -C` and passing the full path to App.xcworkspace for validation. This fixes #5805

### DIFF
--- a/lib/cocoapods/command/repo/push.rb
+++ b/lib/cocoapods/command/repo/push.rb
@@ -96,7 +96,7 @@ module Pod
         # specs to the master repo.
         #
         def check_if_master_repo
-          remotes = Dir.chdir(repo_dir) { `git remote -v 2>&1` }
+          remotes = `git -C #{repo_dir} remote -v 2>&1`
           master_repo_urls = [
             'git@github.com:CocoaPods/Specs.git',
             'https://github.com/CocoaPods/Specs.git',
@@ -145,7 +145,7 @@ module Pod
         # @return [void]
         #
         def check_repo_status
-          clean = Dir.chdir(repo_dir) { `git status --porcelain  2>&1` } == ''
+          clean = `git -C #{repo_dir} status --porcelain  2>&1` == ''
           raise Informative, "The repo `#{@repo}` at #{UI.path repo_dir} is not clean" unless clean
         end
 
@@ -155,7 +155,7 @@ module Pod
         #
         def update_repo
           UI.puts "Updating the `#{@repo}' repo\n".yellow
-          Dir.chdir(repo_dir) { UI.puts `git pull 2>&1` }
+          UI.puts `git -C #{repo_dir} pull 2>&1`
         end
 
         # Commits the podspecs to the source, which should be a git repo.
@@ -189,15 +189,13 @@ module Pod
               FileUtils.cp(spec_file, output_path)
             end
 
-            Dir.chdir(repo_dir) do
-              # only commit if modified
-              if git!('status', '--porcelain').include?(spec.name)
-                UI.puts " - #{message}"
-                git!('add', spec.name)
-                git!('commit', '--no-verify', '-m', message)
-              else
-                UI.puts " - [No change] #{spec}"
-              end
+            # only commit if modified
+            if repo_git('status', '--porcelain').include?(spec.name)
+              UI.puts " - #{message}"
+              repo_git('add', spec.name)
+              repo_git('commit', '--no-verify', '-m', message)
+            else
+              UI.puts " - [No change] #{spec}"
             end
           end
         end
@@ -208,7 +206,7 @@ module Pod
         #
         def push_repo
           UI.puts "\nPushing the `#{@repo}' repo\n".yellow
-          Dir.chdir(repo_dir) { UI.puts `git push origin master 2>&1` }
+          UI.puts `git -C #{repo_dir} push origin master 2>&1`
         end
 
         #---------------------------------------------------------------------#
@@ -216,6 +214,12 @@ module Pod
         private
 
         # @!group Private helpers
+
+        # @return result of calling the git! with args in repo_dir
+        #
+        def repo_git(*args)
+          git!(["-C", repo_dir] + args)
+        end
 
         # @return [Pathname] The directory of the repository.
         #
--- a/lib/cocoapods/command/repo/push.rb
+++ b/lib/cocoapods/command/repo/push.rb
@@ -96,7 +96,7 @@ module Pod
         # specs to the master repo.
         #
         def check_if_master_repo
-          remotes = `git -C #{repo_dir} remote -v 2>&1`
+          remotes = `git -C "#{repo_dir}" remote -v 2>&1`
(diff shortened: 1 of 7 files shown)


benasher44:
Unsure where in ruby I should be quoting and not-quoting arguments for shell commands. Maybe someone could provide some guidance there?

lib/cocoapods/command/repo/push.rb
```diff
@@ -145,7 +145,7 @@ def validate_podspec_files
         # @return [void]
         #
         def check_repo_status
-          clean = Dir.chdir(repo_dir) { `git status --porcelain  2>&1` } == ''
+          clean = `git -C #{repo_dir} status --porcelain  2>&1` == ''
```
orta: these should need the quotations around the ` #{repo_dir}` -> ` "#{repo_dir}"`

benasher44:
One thing to note is that because of all of the `Dir.chdir` going on in tests, the tests aren't doing so well in an Xcode 8 environment at least for me. I can fix all of the `git` calls to use `git -C`, but that's not enough. There are many instances where it needs to `Dir.chdir` to run a working directory-dependent `pod` command.

segiddins:
@benasher44 all the stuff using `Executable` doesn't even use a shell at all, so no need to quote anything

benasher44:
@segiddins 👍. But I still need to use quotes for paths in back tick strings, so that it correctly handles paths with spaces, correct?

segiddins:
@benasher44 yup

benasher44:
Great. I think we're set there then. Any thoughts on the other `Dir.chdir` issue I mentioned? I'm basically holding out dumping Xcode 7 so that I can run `bundle exec rake` 😭. I can continue down the `git -C` path (bad pun) and convert more `git` commands, but this doesn't fix places where we have to `Dir.chdir` to run a `pod` command in tests.

segiddins:
@benasher44 given that the issue is now fixed in Xcodeproj, it shouldn't be necessary to?

benasher44:
Many `Dir.chdir`s seem to still be affected in the main CocoaPods project. Did we expect the patch in Xcodeproj to apply here as well? If so, that doesn't seem to be the case.

segiddins:
Did anyone actually `bundle update xcodeproj`?

DanToml:
@benasher44 when running CocoaPods specs against Xcodeproj head, they all pass for me. (including validation)

benasher44:
@DanToml huh interesting. I ran into this bug yesterday when using 1.1.0.beta.2 at work. I can double check, but then I was debugging I switched to CocoaPods master and was hitting it there too.

If I run just `bundle exec bacon spec/unit/validator_spec.rb` on the 1.1.0.beta.2 tag, one spec fails in Core's `ensure_in_repo!`. After running that whole spec, `git status` tells me I'm now in the CocoaPods specs repo.

benasher44:
I think it's important to do what with `xcodebuild` version 8 selected (by `sudo xcode-select --switch`ing to the right Xcode if you have 2 installed)

benasher44:
@DanToml After `bundle update xcodeproj`, the spec does pass without the changes in this PR. That said, `git -C` seems like a better alternative to `Dir.chdir`: saves a `Dir.chdir` (slow now but that could change) and `git -C` will error instead of do git ops on the wrong repo.

For the `xcodebuild` change, it seems like a minor improvement to pass the full path to the workspace since it will accept one.

In any case, these are now both just improvements, so they're probably not release blocking.

benasher44:
These tests should pass once this PR is updated to use the changes from CocoaPods/Core#352, since it's expecting more `git -C` from that change.

DanToml:
@benasher44 Yeah the Xcodeproj change was only merged yesterday so didn't go into b2. 

This is definitely just a nice to have now, so it's your call but I'd personally like this to go in the 1.1 release candidate (otherwise we can hold off till 1.2).

benasher44:
@DanToml I'm comfortable with this change in the RC 👍. Once the Core change is shipped, I'll update this one.

DanToml:
@benasher44 👍 SGTM.

benasher44:
This doesn't fully fix anything without either the `bundle update xcodeproj` bump or bumping the version of CP and therefore Core. Once Core is updated, there are validator specs that will need to be fixed. I'll revert the changes here that depend on the Core update, and then I think this is ready to go.

benasher44:
Still one failure in here that looks real. Working on that.

CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

```markdown
* Fixed pod repo push validation and push issues related to Dir.chdir 
  [benasher44](https://github.com/benasher44)
```

</p>

heshamsalman:
Heads up, this works to fix issues related to App.xcworkspace not being found, but in my testing it doesn't seem to fix the issue of the generated app target not having the proper swift version. 

Managed to hack together a situation-specific fix from this branch, but swift version of the generated app should be inherited from the pod.

benasher44:
The only failure left is the one that will be fixed by bumping CP's pointer to Core, so this is ready to go!

PR #5837, opened 2016-09-10
Cache result of uses_swift and should_build

dnkoutso:
We have a large project and our `pod install` time keeps increasing the more targets and pods we add.

Based on an old profile we did (http://danielribeiro.github.io/braintree_ios/ProfileForCallStackPrinter.html) it seems that during installation a lot of time is spent for checking if a target can be built or if it uses swift.

This small change speeds up our `pod install` by a lot (from 1m 40s to about 1m and 3s).

@segiddins the only concern is if that value should be invalidated at any point. For example, is there a chance the `prepare_command` of a pod generate sources that this change will miss?

### DIFF
--- a/lib/cocoapods/target/pod_target.rb
+++ b/lib/cocoapods/target/pod_target.rb
@@ -141,9 +141,12 @@ module Pod
     # A target should not be build if it has no source files.
     #
     def should_build?
-      source_files = file_accessors.flat_map(&:source_files)
-      source_files -= file_accessors.flat_map(&:headers)
-      !source_files.empty?
+      return @should_build if defined? @should_build
+      @should_build = begin
+        source_files = file_accessors.flat_map(&:source_files)
+        source_files -= file_accessors.flat_map(&:headers)
+        !source_files.empty?
+      end
     end
 
     # @return [Array<Specification::Consumer>] the specification consumers for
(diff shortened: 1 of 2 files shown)


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

```markdown
* Cache result of uses_swift and should_build 
```

</p>

lib/cocoapods/target/pod_target.rb
```diff
@@ -141,9 +141,12 @@ def product_module_name
     # A target should not be build if it has no source files.
     #
     def should_build?
-      source_files = file_accessors.flat_map(&:source_files)
-      source_files -= file_accessors.flat_map(&:headers)
-      !source_files.empty?
+      return @should_build if defined? @should_build
+      @should_build ||= begin
```
segiddins: the `||=` is redundant, it will never be used, this code will only ever be executed once per instance

segiddins:
Yes, the prepare_command could change this

segiddins:
What's the estimated savings of this change?

dnkoutso:
I don't have a new profile run to tell you the numbers, but as I said in the initial description, it took our pod install time down to 1 minute from 1 minute and ~40 seconds.

In another app it took it down from 23 seconds to 12.

Again this is mostly our environment that has 100+ pods being installed with a lot of files.

lib/cocoapods/target/pod_target.rb
dnkoutso: Will fix.

segiddins:
👍 looks good to me!

dnkoutso:
Awesome to hear on the prepare command! Thank you for verifying this I was just trying to :)

dnkoutso:
Hoping to get this landed for 1.1.0!

segiddins:
@DanToml I'm 👍 to back port this if you are

DanToml:
@segiddins 👍

PR #5856, opened 2016-09-13
[Travis] Limit building examples to system ruby

segiddins:
Let's make sure this actually works before 🚢 

\c @DanToml

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -147,8 +147,10 @@ begin
       title 'Running Integration tests'
       sh 'bundle exec bacon spec/integration.rb'
 
-      title 'Running examples'
-      Rake::Task['examples:build'].invoke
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -147,8 +147,10 @@ begin
       title 'Running Integration tests'
       sh 'bundle exec bacon spec/integration.rb'
 
-      title 'Running examples'
-      Rake::Task['examples:build'].invoke
+      unless ENV['TRAVIS'] && ENV['TRAVIS_RUBY_VERSION'] == 'system'
```
DanToml: shouldn't this be `if ENV...`?

DanToml:
🎉 🎉 🎉 🎉 🎉 🎉

Rakefile
segiddins: The logic is right, the ENV var is wrong

Issue #5864, opened 2016-09-14
Can't install pod with Xcode 8/Swift 3/Cocoapods 1.1.0 RC2


yannickl:
* [x] I've read and understood the [*CONTRIBUTING guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

```
/usr/local/bin/pod install
```

## What did you expect to happen?

Install all pod dependencies correctly.

## What happened instead?

The install failed and no workspace is created.

### Stack

```
   CocoaPods : 1.1.0.rc.2
        Ruby : ruby 2.3.0p0 (2015-12-25 revision 53290) [x86_64-darwin15]
    RubyGems : 2.5.1
        Host : Mac OS X 10.11.6 (15G1004)
       Xcode : 8.0 (8A218a)
         Git : git version 2.6.4
Ruby lib dir : /usr/local/Cellar/ruby/2.3.0/lib
Repositories : master - https://github.com/CocoaPods/Specs.git @ 90c40f1bd8fb3738b8e1b725c3ffe9ac284ac9bb
```

### Plugins

```
cocoapods-deintegrate : 1.0.1
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
cocoapods-trunk       : 1.0.0
cocoapods-try         : 1.1.0
```

### Podfile

```ruby
platform :ios, '9.0'

source 'https://github.com/CocoaPods/Specs.git'
use_frameworks!

def shared_pods
  pod 'PromiseKit', '~> 4'
end

target 'AwaitKitExample' do
  shared_pods
end

target 'AwaitKitTests' do
  shared_pods
end
```

### Error

```
Errno::EEXIST - File exists @ dir_s_mkdir - /Users/yannickloriot/Development/perso/AwaitKit/Example/Pods/Target Support Files/PromiseKit
/usr/local/lib/ruby/gems/2.3.0/gems/cocoapods-1.1.0.rc.2/lib/cocoapods/installer/xcode/pods_project_generator/target_installer.rb:92:in `mkdir'
/usr/local/lib/ruby/gems/2.3.0/gems/cocoapods-1.1.0.rc.2/lib/cocoapods/installer/xcode/pods_project_generator/target_installer.rb:92:in `mkdir'
[...]

Issue #5834, opened 2016-09-09
Errno::EEXIST - File exists with XCode 8 migrating to Swift 2.3


dimitryvin:
### Command

```
/usr/local/bin/pod install
```

### Report

* What did you do?

Run pod install

* What did you expect to happen?

For it to setup the project like before

* What happened instead?

Failed with Errno::EEXIST - File exists - /Users/dimitryvin/Developer/Repos/stem-ios/Stem/Pods/Target Support Files/AFNetworking

### Stack

```
   CocoaPods : 1.1.0.beta.2
        Ruby : ruby 2.0.0p648 (2015-12-16 revision 53162) [universal.x86_64-darwin16]
    RubyGems : 2.0.14.1
        Host : Mac OS X 10.12 (16A313a)
       Xcode : 8.0 (8A218a)
         Git : git version 2.8.4 (Apple Git-73)
Ruby lib dir : /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib
Repositories : master - https://github.com/CocoaPods/Specs.git @ 227e9778abe84e9a549ff67e86fd1baeea05f065
```

### Plugins

```
cocoapods-deintegrate : 1.0.1
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
cocoapods-trunk       : 1.0.0
cocoapods-try         : 1.1.0
```

### Podfile

```ruby
# Uncomment this line to define a global platform for your project
# platform :ios, '8.0'
# Uncomment this line if you're using Swift
# use_frameworks!

def shared

pod 'ContactsPicker', :git => 'https://github.com/stems/EPContactsPicker.git'
use_frameworks!
pod 'Fabric'
pod 'Crashlytics'
pod 'Alamofire'
pod 'Parse'
pod 'MaterialControls', :git => 'https://github.com/stems/Material-Controls-For-iOS.git'
pod 'FXPageControl'
pod 'Material', '1.37.3'
pod 'Charts'
pod 'libPhoneNumber-iOS', '~> 0.8'
pod 'Intercom'
pod 'FPPicker'
pod 'EasyTipView', :git => 'https://github.com/eyegroove/EasyTipView.git'
pod 'Google/Analytics'
pod 'Cartography'
end

target 'stem' do
    shared
end

target 'stem-staging' do
    shared
end

target 'stem-sandbox' do
    shared
end
```

### Error

```
[...]

PR #5917, opened 2016-09-20
Error if project targets have mismatched Swift versions

benasher44:
With this change, having a project that has one target that has a Swift version set and another that has it unset or set to something different, the user will get an error with a message like this during `pod install`

```
[!] The following pods are integrated into Swift targets that do not have the same Swift version:

- SomePod required by MyAppTarget1 (Swift 2.3), MyAppTarget2 (Swift 3.0)
```

This should help folks resolve issues related to #5864 and #5834.

### DIFF
--- a/lib/cocoapods/installer/analyzer.rb
+++ b/lib/cocoapods/installer/analyzer.rb
@@ -417,25 +417,32 @@ module Pod
       #
       def check_pod_target_swift_versions(specs_by_target)
         targets_by_spec = {}
+        swift_specs = Set.new
         specs_by_target.each do |target, specs|
           specs.each do |spec|
             (targets_by_spec[spec] ||= []) << target
           end
+          swift_specs.merge specs.select do |spec|
+            spec.consumer(target.platform).source_files.any? do |source_file|
+              source_file.end_with? '.swift'
+            end
+          end
         end
 
         target_msg = lambda do |target|
-          if target.swift_version.nil? || target.swift_version.empty?
-            "#{target.name} (Swift version missing)"
-          else
-            "#{target.name} (Swift #{target.swift_version})"
-          end
+          "#{target.name} (Swift #{target.swift_version})"
         end
 
         error_messages = []
         targets_by_spec.each do |spec, targets|
-          error_messages << "- #{spec.name} required by #{targets.map(&target_msg).join(', ')}" unless targets.uniq(&:swift_version).count == 1
+          swift_targets = targets.reject do |target|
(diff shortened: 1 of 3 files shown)


benasher44:
Hopefully with this update, it'll be more obvious that you have to modify your project to fix the issue.

orta:
👍

benasher44:
Good to go on my end! Looks like the only failure is the spurious `execution_output.txt` one.

benasher44:
Actually I'm going to wait on shipping this. I want to update this to better handle targets that don't specify a swift version.

benasher44:
@DanToml this is updated with master, and it's updated to cover the following scenarios:
1. Swift targets with different versions that share a Swift pod -> error
2. Swift targets with different versions that share only a non-Swift pod -> no error
3. 1 Swift target with 0 or more non-Swift targets that share only a Swift pod -> no error

lib/cocoapods/installer/analyzer.rb
```diff
@@ -406,6 +407,44 @@ def generate_target(target_definition, pod_targets)
...
+      #
+      def check_pod_target_swift_versions(specs_by_target)
+        targets_by_spec = {}
+        swift_specs = Set.new
+        specs_by_target.each do |target, specs|
+          specs.each do |spec|
+            (targets_by_spec[spec] ||= []) << target
+          end
+          swift_specs.merge specs.select do |spec|
+            spec.consumer(target.platform).source_files.any? do |source_file|
```
benasher44: @segiddins pointed out that this doesn't do what I think it does. Will have to fix this up.

DanToml: So, I've been looking into this today and it turns out that doing this reliably is pretty much impossible.

The only way to get the full file list is to get a `FileAccessor` for the given pod, which can only be done when we've actually downloaded it.

For now we may have to not support sharing any pods between different swift versions :/

DanToml: An alternative would be duplicating the pods, and changing Target to generate paths that include the associated Swift version?

benasher44: Support multiple Swift versions in the same project feels like an edge case, especially right now when Apple has said that Swift 2.3 will not be available in a future version of Xcode 8

DanToml: Yeah, I'm just about to push some commits that strip this down to just disallowing that, and strips swift version from PodVariant

benasher44: Okay, I'm going to close this, and maybe re-open it if it feels relevant after your `PodVariant` work

PR #5926, opened 2016-09-21
Remove EMBEDDED_CONTENT_CONTAINS_SWIFT as it is deprecated

codymoorhouse:
Previously was producing warnings when updating pods because of a deprecated flag.

### DIFF
--- a/lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
+++ b/lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
@@ -69,7 +69,6 @@ module Pod
           # cause an App Store rejection because frameworks cannot be embedded
           # in embedded targets.
           if !target.requires_host_target? && pod_targets.any?(&:uses_swift?)
-            config['EMBEDDED_CONTENT_CONTAINS_SWIFT'] = 'YES'
             config['ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES'] = 'YES'
           else
             config['ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES'] = 'NO'
--- a/lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
+++ b/lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
@@ -68,11 +68,25 @@ module Pod
           # that use Swift. Setting this for the embedded target would
           # cause an App Store rejection because frameworks cannot be embedded
           # in embedded targets.
-          if !target.requires_host_target? && pod_targets.any?(&:uses_swift?)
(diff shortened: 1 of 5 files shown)


benasher44:
@orta @dantoml do we still need to support Xcode 7?

benasher44:
If we do, it might be better to just set the appropriate flag based on your Xcode version

CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

```markdown
* Remove EMBEDDED_CONTENT_CONTAINS_SWIFT as it is deprecated 
  [codymoorhouse](https://github.com/codymoorhouse)
```

</p>

DanToml:
@benasher44 I don't think we should drop support for it immediately - especially as maintaining compatibility shouldn't be too difficult.

codymoorhouse:
I can add it in, just not sure how to check xcode version

benasher44:
@dantoml sounds good. @codymoorhouse my first thought would be just ,`xcodebuild -version`, but you could dig around in the xcodeproj gem to see if there's something more convenient there.

codymoorhouse:
sure i'll close this for now and reopen after i make those changes if i can figure it out

benasher44:
You're more than welcome to leave the PR open!

codymoorhouse:
@benasher44 Okay I think I got it? I highly encourage a double check though

lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
```diff
@@ -68,12 +68,25 @@ def generate
           # that use Swift. Setting this for the embedded target would
           # cause an App Store rejection because frameworks cannot be embedded
           # in embedded targets.
-          if !target.requires_host_target? && pod_targets.any?(&:uses_swift?)
-            config['EMBEDDED_CONTENT_CONTAINS_SWIFT'] = 'YES'
-            config['ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES'] = 'YES'
+
+	  # Extract xcodeversion number
+	  @xcodeversion = 'xcodebuild -version | cut -f 2 -d ' ' | head -n 1`
```
segiddins: doing this sort of check doesn't really work, as the user can be building with a different version of Xcode than is reported when just running `xcodebuild`

codymoorhouse: Ah good point, didn't think about that

codymoorhouse: Ah good point, didn't think about that

PR #5933, opened 2016-09-22
Change EMBEDDED_CONTENT_CONTAINS_SWIFT as it is deprecated

codymoorhouse:
SEE: #5926

Please double check my work! My experience in ruby is only of 2 days now.

@segiddins how does this look?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -101,6 +101,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Ben Asher](https://github.com/benasher44)
   [#5732](https://github.com/CocoaPods/CocoaPods/pull/5732)
 
+* Update EMBEDDED_CONTENT_CONTAINS_SWIFT flag behaviour based on xcode version.
+  [codymoorhouse](https://github.com/codymoorhouse)
+  [#5732](https://github.com/CocoaPods/CocoaPods/issues/5732)
+
 * Verify that embedded target platform and swift version matches the host.  
   [Ben Asher](https://github.com/benasher44)
   [#5747](https://github.com/CocoaPods/CocoaPods/pull/5747)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,6 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 * Update EMBEDDED_CONTENT_CONTAINS_SWIFT flag behaviour based on xcode version.
   [codymoorhouse](https://github.com/codymoorhouse)
-  [Ben Asher](https://github.com/benasher44)
   [#5732](https://github.com/CocoaPods/CocoaPods/issues/5732)
 
 ##### Bug Fixes
@@ -48,6 +47,7 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Stefan Pühringer](https://github.com/b-ray)
   [#5891](https://github.com/CocoaPods/CocoaPods/issues/5891)
 
+
 ## 1.1.0.rc.2 (2016-09-13)
 
 ##### Enhancements
@@ -62,6 +62,7 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 * Correctly pass Pod:VERSION in `pod lib create`.  
   [Danielle Tomlinson](https://github.com/dantoml)
   [#5840](https://github.com/CocoaPods/CocoaPods/issues/5840)
+
 
 ## 1.1.0.rc.1 (2016-09-10)
 
@@ -135,6 +136,7 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 * Fix analyzer checking the compatibility of an embedded target with a host that has not been added the Podfile.  
   [Ben Asher](https://github.com/benasher44)
   [#5783](https://github.com/CocoaPods/CocoaPods/issues/5783)
+
 
 ## 1.1.0.beta.1 (2016-07-11)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,11 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Ben Asher](https://github.com/benasher44)
   [#5959](https://github.com/CocoaPods/CocoaPods/pull/5959)
 
+* Update EMBEDDED_CONTENT_CONTAINS_SWIFT flag behaviour based on xcode version.
+  [codymoorhouse](https://github.com/codymoorhouse)
+  [Ben Asher](https://github.com/benasher44)
+  [#5732](https://github.com/CocoaPods/CocoaPods/issues/5732)
+
 ##### Bug Fixes
 
 * Remove special handling for messages apps  
@@ -57,7 +62,6 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 * Correctly pass Pod:VERSION in `pod lib create`.  
   [Danielle Tomlinson](https://github.com/dantoml)
   [#5840](https://github.com/CocoaPods/CocoaPods/issues/5840)
-
 
 ## 1.1.0.rc.1 (2016-09-10)
 
@@ -100,10 +104,6 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 * Set ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES when appropriate.  
   [Ben Asher](https://github.com/benasher44)
   [#5732](https://github.com/CocoaPods/CocoaPods/pull/5732)
-
-* Update EMBEDDED_CONTENT_CONTAINS_SWIFT flag behaviour based on xcode version.
-  [codymoorhouse](https://github.com/codymoorhouse)
-  [#5732](https://github.com/CocoaPods/CocoaPods/issues/5732)
 
 * Verify that embedded target platform and swift version matches the host.  
   [Ben Asher](https://github.com/benasher44)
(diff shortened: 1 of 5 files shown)


neonichu:
I think this should be conditional on the used version of Xcode, since we still want to support Xcode 7.3.1 for a while.

codymoorhouse:
Yeah, well originally what I did check the Xcode version using the command xcodeproject -version, but as @segiddins said, you could still be building with a different Xcode version then the one the command line is reporting. So I figured this was a good workaround because swift 3.0 is not available for Xcode 7.3.1 up to my knowledge. What do you think @neonichu?

See #5926

neonichu:
Sorry, I missed Sam's comment. Maybe we could actually just generating both build settings, I think Xcode 8 doesn't care about `EMBEDDED_CONTENT_CONTAINS_SWIFT` if both are present.

codymoorhouse:
True Xcode doesn't care, however when updating or installing pods it does create a warning for it that one is overriding another

Cody Moorhouse
cody.moorhouse@icloud.com
(780)938-3008

neonichu:
Ok, I guess your approach seems fine then. But we should take 2.3 into account as well.

codymoorhouse:
Okay which file are you referring to when I should I be checking for swift 2.3?

DenHeadless:
Having EMBEDDED_CONTENT_CONTAINS_SWIFT set in user defined settings is causing Xcode 8 to always display unfixable warning about need to update to recommended settings. This might be problematic for users, that treat warnings as errors ;)

lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
```diff
@@ -68,12 +68,26 @@ def generate
           # that use Swift. Setting this for the embedded target would
           # cause an App Store rejection because frameworks cannot be embedded
           # in embedded targets.
-          if !target.requires_host_target? && pod_targets.any?(&:uses_swift?)
-            config['EMBEDDED_CONTENT_CONTAINS_SWIFT'] = 'YES'
-            config['ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES'] = 'YES'
+
+	  # Extract swift_version number
+	  @swift_version = target.build_configurations.build_settings['SWIFT_VERSION']
```
benasher44: this should be something like `target.resolved_build_setting('SWIFT_VERSION').values.compact.uniq`. The important thing here is that it uses `resolved_build_setting`, so that it gets the build setting, even if it's set at the project level and not specifically on this target.

lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
```diff
@@ -68,12 +68,26 @@ def generate
...
+          @swift_version_check = 3.0
+
+	  # Check if swift_version is prior to v3.0
+	  if !target.requires_host_target? && pod_targets.any?(&:uses_swift?)
+	    if @swift_version < @swift_version_check
```
benasher44: nit: instead of doing this to decide the build setting in each branch, maybe pull this out above into something like:
```
if @swift_version < @swift_version_check
  embedded_swift_key = 'EMBEDDED_CONTENT_CONTAINS_SWIFT'
else
  embedded_swift_key = 'ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES'
end
```
then below you can just set `config[embedded_swift_key] = `…

lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
```diff
@@ -68,12 +68,26 @@ def generate
           # that use Swift. Setting this for the embedded target would
           # cause an App Store rejection because frameworks cannot be embedded
           # in embedded targets.
-          if !target.requires_host_target? && pod_targets.any?(&:uses_swift?)
-            config['EMBEDDED_CONTENT_CONTAINS_SWIFT'] = 'YES'
-            config['ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES'] = 'YES'
+
+	  # Extract swift_version number
+	  @swift_version = target.build_configurations.build_settings['SWIFT_VERSION']
+          @swift_version_check = 3.0
```
benasher44: this should be an all caps constant near the top of this file. See `lib/cocoapods/target/aggregate_target.rb` for an example (although that one is an array instead of a number)

spec/unit/generator/xcconfig/aggregate_xcconfig_spec.rb
```diff
@@ -241,20 +241,20 @@ def pod_target(spec, target_definition)
             @xcconfig.to_hash['OTHER_SWIFT_FLAGS'].should.include '$(inherited) "-D" "COCOAPODS"'
           end
 
-          it 'sets EMBEDDED_CONTENT_CONTAINS_SWIFT when there is swift' do
+          it 'sets ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES when there is swift' do
```
benasher44: For tests, I think there should be 2 sets: 1 set stubs the @swift_version to be 2.2, and the other stubs it to be 2.3. In one set of tests, it'll set `ALWAYS_EMBED_SWIFT_LIBRARIES`, and in the other it's set the older key. Since we're going to support both, we should probably have tests that check for both.

lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
benasher44: Also, this should be 2.3. 2.3 is the minimum version used in Xcode 8, and it also requires the newer flag.

codymoorhouse: For the swift version check, should it not be 3.0? I know that Xcode 8 can use version 2.3, but this is why we are using the target build settings as opposed to checking which version of Xcode is being used. The target would specify which swift version to use, setting the appropriate flag. Let me know what you think!

benasher44: Sorry I'm not sure I understand. Reading the Swift version from the build settings is different than checking `xcodebuild -version` because you could have multiple versions of `xcodebuild` available on the system, and whichever one is in use depends on the environment (specifically if you've used `xcode-select --switch`), but not necessarily the project.

But, I'm thinking about this more, and this check won't work if the project isn't a Swift project, since it won't have a swift version set. What if we just check for the existence of the `EMBEDDED_CONTENT_CONTAINS_SWIFT` swift setting? This'll allow the user to upgrade just the way we all did by click on the thing in Xcode and approving the new recommended settings. Then, we'll start being the new `ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES`, and then CocoaPods can switch to that one. If we don't see the `EMBEDDED_CONTENT_CONTAINS_SWIFT` flag, it won't matter in older Xcode versions, since Xcode will just ignore it. Does this make sense?

benasher44: So to recap: if you see `EMBEDDED_CONTENT_CONTAINS_SWIFT`, use that setting, otherwise, use the newer `ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES` setting.

codymoorhouse: Hmm okay, I think I see what your saying. I will try to work on something for that, I made all the changes thus far. Thanks for the feedback 👍

benasher44: No prob!

CHANGELOG.md
```diff
@@ -29,6 +29,11 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Ben Asher](https://github.com/benasher44)
   [#5959](https://github.com/CocoaPods/CocoaPods/pull/5959)
 
+* Update EMBEDDED_CONTENT_CONTAINS_SWIFT flag behaviour based on xcode version.
+  [codymoorhouse](https://github.com/codymoorhouse)
+  [Ben Asher](https://github.com/benasher44)
```
benasher44: I'm helping with the review, but this is your change! No need to include me, but I appreciate the sentiment :)

CHANGELOG.md
```diff
@@ -58,7 +63,6 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Danielle Tomlinson](https://github.com/dantoml)
   [#5840](https://github.com/CocoaPods/CocoaPods/issues/5840)
 
-
```
DanToml: We need to keep the double new line here (or it breaks the generated changelog)

lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
```diff
@@ -68,12 +73,23 @@ def generate
...
-            config['ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES'] = 'NO'
+
+	  # Extract swift_version number
+	  @swift_version = target.resolved_build_setting('SWIFT_VERSION').values.compact.uniq
+	  embed_default_value = 'NO'
+
+	  # Check if swift_version is prior to SWIFT_VERSION_CHECK
+	  if !target.requires_host_target? && pod_targets.any?(&:uses_swift?)
+	    embed_default_value = 'YES'
+	    if @swift_version < SWIFT_VERSION_CHECK
```
DanToml: Couple of points here:
1. This is doing a String comparison which would break if they eventually add patch versions into the mix due to being a lexicographic compare.

2. I think this should be `<=` to also use these keys on Xcode 8 Swift 2.3 projects

benasher44:
Just a heads up: it looks like you'll need to update your CHANGELOG with master, as well as update your cocoapods-integration-specs with master as well.

benasher44:
@codymoorhouse are you waiting on more feedback? Or should we wait for more commits?

DanToml:
Thanks for the contribution @codymoorhouse, I finished this up and merged it in https://github.com/CocoaPods/CocoaPods/pull/6016 for todays release.

codymoorhouse:
No worries @DanToml sounds great! Sorry @benasher44 I had to put this on pause because I have been super busy with school haha

PR #5934, opened 2016-09-23
Cache result of include_in_build_config and inhibit_warnings

dnkoutso:
We re-run the profiler against our project and noticed two more big improvements in `pod install` times.

If you wanted to see how long a project with 10+ targets and 130 pods takes to integrate I've attached the generated log here. You might be able to pick other areas of improvements but we noticed again that `pod_target` could be caching some of its results again. 

I do not believe caching these could cause any harm or throw a `pod_target` out of sync.

[profile.zip](https://github.com/CocoaPods/CocoaPods/files/488807/profile.zip) (it might take a bit of time to render in Chrome :))

This dropped our `pod install` time from 1m 8s to about 48seconds.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
+* Cache result of inhibit_warnings and include_in_build_config to speed up pod install.  
+  [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
+  [#5934](https://github.com/CocoaPods/CocoaPods/pull/5934)
+  
 * Improve performance of PathList.read_file_system 
   [Heath Borders](https://github.com/hborders)
   [#5890](https://github.com/CocoaPods/CocoaPods/issues/5890)
--- a/lib/cocoapods/target/pod_target.rb
+++ b/lib/cocoapods/target/pod_target.rb
@@ -47,6 +47,7 @@ module Pod
       @file_accessors = []
       @resource_bundle_targets = []
       @dependent_targets = []
+      @build_config_cache = {}
     end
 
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
+* Cache result of inhibit_warnings and include_in_build_config to speed up pod install.  
+  [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
+  [#5857](https://github.com/CocoaPods/CocoaPods/pull/5857)
```
orta: This should be 5934 👍

lib/cocoapods/target/pod_target.rb
```diff
@@ -243,13 +251,16 @@ def include_in_build_config?(target_definition, configuration_name)
     # @return [Bool]
     #
     def inhibit_warnings?
+      return @inhibit_warnings if defined? @inhibit_warnings
```
orta: interesting trick 👍

CHANGELOG.md
dnkoutso: Duh fixed ;)

DanToml:
This looks interesting, I'll need to dig in to the code a bit to see if there are any ways this can be invalidated (not too familiar with it off the top of my head), but provisional lgtm.

dnkoutso:
FWIW, we've been running with this change internally for over a week with 0 problems and its similar to https://github.com/CocoaPods/CocoaPods/pull/5837

segiddins:
👍 lgtm

dnkoutso:
🎉

DanToml:
Thanks @dnkoutso!

dnkoutso:
thank you!

PR #5947, opened 2016-09-24
[WIP] Correctly set direct and transitive dependency header search paths fo…

dnkoutso:
…r pod targets

closes https://github.com/CocoaPods/CocoaPods/issues/5941

@orta the solution/code might be a bit too junior as I am very junior in Ruby. In a large project we have I can still compile all the Pod dependencies.

This does not force podspecs to include their transitive dependencies. Its still much better than the "catch-all" approach than before.

screen shot 2016-09-24 at 8 36 39 am

### DIFF
--- a/lib/cocoapods/sandbox/headers_store.rb
+++ b/lib/cocoapods/sandbox/headers_store.rb
@@ -24,6 +24,7 @@ module Pod
         @sandbox       = sandbox
         @relative_path = relative_path
         @search_paths  = []
+        @cache = {}
       end
 
       # @param  [Platform] platform
@@ -39,6 +40,21 @@ module Pod
 
         headers_dir = root.relative_path_from(sandbox.root).dirname
         ["${PODS_ROOT}/#{headers_dir}/#{@relative_path}"] + platform_search_paths.uniq.map { |entry| "${PODS_ROOT}/#{headers_dir}/#{entry[:path]}" }
+      end
+
+      def search_paths_for_target(platform, target)
+        key = [platform, target.name]
+        if @cache.key?(key)
+          return @cache[key]
+        end
+        platform_search_paths = @search_paths.select do |entry|
+          entry[:platform] == platform.name and entry[:path].basename.to_s == target.name
+        end
+
+        headers_dir = root.relative_path_from(sandbox.root).dirname
+        result = platform_search_paths.uniq.map { |entry| "${PODS_ROOT}/#{headers_dir}/#{entry[:path]}" }
+        @cache[key] = result
+        result
       end
 
       # Removes the directory as it is regenerated from scratch during each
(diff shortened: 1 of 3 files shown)


lib/cocoapods/sandbox/headers_store.rb
```diff
@@ -41,6 +42,21 @@ def search_paths(platform)
         ["${PODS_ROOT}/#{headers_dir}/#{@relative_path}"] + platform_search_paths.uniq.map { |entry| "${PODS_ROOT}/#{headers_dir}/#{entry[:path]}" }
       end
 
+      def search_paths_for_target(platform, target)
```
dnkoutso: This can probably be merged with the method above and be smarter about it.

lib/cocoapods/sandbox/headers_store.rb
```diff
@@ -24,6 +24,7 @@ def initialize(sandbox, relative_path)
         @sandbox       = sandbox
         @relative_path = relative_path
         @search_paths  = []
+        @cache = {}
```
dnkoutso: Recursively searching for header search targets, I figured a cache will help `header_store`. It actually does as I saw almost minimal increase in `pod install` times, versus without the cache of 5-10 seconds.

dnkoutso:
This also doesnt affect `aggregate_targets` which btw all `Pods-<something` targets imo should have 0 header search paths as they compile nothing and contain only a dummy source.

dnkoutso:
@mjholgate you can manually apply this patch for you if you directly edit the files where the cocoapods gem is located to try it out.

orta:
ideally @segiddins / @alloy gonna need you two to look over this one, think it's a little bit about history as well as the technical aspect

mjholgate:
Giving this a whirl, thanks @dnkoutso

mjholgate:
Wow, this made a great improvement to our build time (ran each test twice):

CP 1.0.0
real   	23m52.380s
user   	41m24.358s
sys    	34m30.548s

CP 1.0.0
real   	24m43.586s
user   	42m24.748s
sys    	35m21.628s

CP 1.0.0 + patch
real   	19m2.669s
user   	33m41.174s
sys    	25m16.503s

CP 1.0.0 + patch
real   	18m34.129s
user   	33m13.066s
sys    	24m1.580s

mjholgate:
I can imagine the result would be even better if we could specify in the DSL which dependencies are required by the public headers of a pod (and thus cut out most transitive dependencies too).

dnkoutso:
@mjholgate crazy your build takes 18m even with this change. huge project seems like. I am looking from feedback from @segiddins @alloy and I think I need to clean up some stuff before this can land.

mjholgate:
Yeah, it is crazy :(. The project is way too big.

segiddins:
@dnkoutso are there any functional changes to this? If so, tests would be very helpful in determining what's going on

dnkoutso:
I haven't updated tests yet but I am planning to and also planning to add some new ones.

There should be 0 functional changes imo. Its just that pod targets will no longer be able to include header search paths from "above" them in the dependency tree.

segiddins:
That is a functional change that certainly warrants additional test coverage imo

dnkoutso:
I am not planning at the moment to continue this PR. I am going to close it since its marked as WIP.

PR #5950, opened 2016-09-25
Expand the usage of the GH Inspector to all of pod install

orta:
Previously it would only look at the issues if the error occurred inside a Podfile, or as a part of an informational error that we've created. The problem is, most of our errors right now are occurring outside of that scope. 

So this expands the scope to all of `pod install`. Only when it comes from Ruby, so it should hit a lot of the `Errno::EEXIST - File exists` kind of issues for example. 

Also fixes a minor spacing issue in the gemspec.


segiddins:
Why was this necessary? As-is, all non-imforamtive messages already have this done via the error report https://github.com/CocoaPods/CocoaPods/blob/master/lib/cocoapods/user_interface/error_report.rb#L116

orta:
Just double checked, yeah, this is redundant, will send a PR removing it

PR #5951, opened 2016-09-25
[Validator] Inform people of .swift-version on err

DanToml:
@orta - Would appreciate any thoughts on the message for this.

### DIFF
--- a/lib/cocoapods/validator.rb
+++ b/lib/cocoapods/validator.rb
@@ -142,7 +142,12 @@ module Pod
         reasons << 'all results apply only to public specs, but you can use ' \
                    '`--private` to ignore them if linting the specification for a private pod'
       end
-      reasons.to_sentence
+      if dot_swift_version.nil?
+        reasons.to_sentence + ".\n[!] If you are trying to validate a Swift 3.0 Pod, " \
+          'you need to have a `.swift-version` file. e.g `echo "3.0" > .swift-version`'
+      else
+        reasons.to_sentence
+      end
     end
 
     #-------------------------------------------------------------------------#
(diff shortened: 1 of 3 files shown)


DanToml:
screen shot 2016-09-25 at 18 45 35

lib/cocoapods/validator.rb
```diff
@@ -142,7 +142,12 @@ def failure_reason
         reasons << 'all results apply only to public specs, but you can use ' \
                    '`--private` to ignore them if linting the specification for a private pod'
       end
-      reasons.to_sentence
+      if dot_swift_version.nil?
```
orta: can we add some logic similar to the `pod_target.uses_swift` that see's if any of the related targets contain Swift to ensure that is actually a Swift pod?

lib/cocoapods/validator.rb
```diff
@@ -142,7 +142,12 @@ def failure_reason
...
+        reasons.to_sentence + ".\n[!] If you are trying to validate a Swift 3.0 Pod, " \
+          'you need to have a `.swift-version` file. e.g `echo "3.0" > .swift-version`'
+      else
+        reasons.to_sentence
+      end
     end
```
orta: Word wise, let's try aim for future proofing, e.g. explain why it is, then give the 3.0 example:

```

[!] The validator for Swift projects uses Swift 2.3 by default, if you are using a different version of swift
you can use a `.swift-version` file to set the version for your Pod. For example to use Swift 3.0, run:

`echo "3.0" > .swift-version` to create a version file in your repo.
```

DanToml:
@orta I tried to add a uses swift type thing but ran across some cases where installer or linter would be nil. Will take another look this evening.

lib/cocoapods/validator.rb
DanToml: done 👍

orta:
I saw the chatter about this - looks like we can't check for whether it's a swift lib - this is coo lthen 👍

PR #5984, opened 2016-10-01
Improve handling of mismatched Swift versions

DanToml:
\c @benasher44

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,14 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Ben Asher](https://github.com/benasher44)
   [#5959](https://github.com/CocoaPods/CocoaPods/pull/5959)
 
+* Error with helpful message when integrating a pod into targets that have mismatched Swift versions.  
+  [Ben Asher](https://github.com/benasher44)
+  [#5984](https://github.com/CocoaPods/CocoaPods/pull/5984)
+
+* Allow users to share pods between Objective-C and Swift targets.
+  [Danielle Tomlinson](https://github.com/dantoml)
+  [#5984](https://github.com/CocoaPods/CocoaPods/pull/5984)
+
 ##### Bug Fixes
 
 * Remove special handling for messages apps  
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -24,6 +24,14 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Ben Asher](https://github.com/benasher44)
   [#5959](https://github.com/CocoaPods/CocoaPods/pull/5959)
 
+* Error with helpful message when integrating a pod into targets that have mismatched Swift versions.  
+  [Ben Asher](https://github.com/benasher44)
```
benasher44: <3

lib/cocoapods/installer/analyzer.rb
```diff
@@ -406,6 +407,39 @@ def generate_target(target_definition, pod_targets)
...
+            (targets_by_spec[spec] ||= []) << target
+          end
+        end
+
+        error_message_for_target = lambda do |target|
+          "#{target.name} (Swift #{target.swift_version})"
+        end
+
+        error_messages = targets_by_spec.map do |spec, targets|
+          swift_targets = targets.reject { |target| target.swift_version.blank? }
```
benasher44: swift_version can be nil in some scenarios, and I think we want to reject that here as well. I made these notes in an early commit in my previous PR:
```
# when the swift version is unset at the project level, but set in one target, swift_version is nil
# when the swift version is set at the project level, but unset in one target, swift_version is empty
```

benasher44: Maybe `TargetDefinition` could normalize this?

DanToml: Blank?  Checks for nil and empty.  I think there is actually a semantic difference between them in Xcode so I'm not so sure that we should normalise?

benasher44: Ahhh. I had a suspicion that was the case. I couldn't find the docs for it, but now I se active support adds this. 👍

PR #5989, opened 2016-10-02
Allow passing --swift-version to pod spec/lib lint

DanToml:
Default remains as `2.3`. `.swift-version` file is still supported to prevent users from needing to always pass the flag. The flag takes precedence over the file.

For future people viewing this PR:

### Locally

`pod lib lint --swift-version=3.0`

or:

```
echo "3.0" > .swift-version
pod lib lint
```

### Online

`pod spec lint --swift-version=3.0`

or:

```
echo "3.0" > .swift-version
pod spec lint
```

### DIFF
--- a/lib/cocoapods/command/lib/lint.rb
+++ b/lib/cocoapods/command/lib/lint.rb
@@ -21,6 +21,7 @@ module Pod
              '(defaults to https://github.com/CocoaPods/Specs.git). ' \
              'Multiple sources must be comma-delimited.'],
             ['--private', 'Lint skips checks that apply only to public specs'],
+            ['--swift-version=VERSION', 'The SWIFT_VERSION that should be used to lint the spec'],
           ].concat(super)
         end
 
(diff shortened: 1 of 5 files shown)


DanToml:
@orta Gonna PR `cocoapods-trunk` when this is merged

orta:
👍

lib/cocoapods/command/lib/lint.rb
```diff
@@ -21,6 +21,7 @@ def self.options
              '(defaults to https://github.com/CocoaPods/Specs.git). ' \
              'Multiple sources must be comma-delimited.'],
             ['--private', 'Lint skips checks that apply only to public specs'],
+            ['--swift-version=VERSION', 'The SWIFT_VERSION that should be used to lint the spec'],
```
benasher44: Can we add: "This takes precedence over a .swift-version file containing SWIFT_VERSION."

This will include `.swift-version` in the help message, which will create the opportunity for people to be exposed to `.swift-version` before linting, since otherwise it's a message you get if the swift version isn't set.

DanToml: 👍

DanToml: @benasher44 updated

Issue #6032, opened 2016-10-13
"pod repo update" can return exit code 0 in case of failure


UkuLoskit:
# Report

## What did you do?

Ran:
```pod repo update```
and got the following output:
```
STDOUT for command pod repo update:
Updating spec repo `master`
  $ /usr/local/bin/git pull --ff-only

CocoaPods 1.1.0.rc.2 is available.
To update use: `gem install cocoapods --pre`
[!] This is a test version we'd love you to try.

For more information, see https://blog.cocoapods.org and the CHANGELOG for this version at https://github.com/CocoaPods/CocoaPods/releases/tag/1.1.0.rc.2


STDERR for command pod repo update:
  fatal: unable to access 'https://github.com/CocoaPods/Specs.git/': transfer closed with outstanding read data remaining

[!] CocoaPods was not able to update the `master` repo. If this is an unexpected issue and persists you can inspect it running `pod repo update --verbose`
```

## What did you expect to happen?
```pod repo update``` should have returned a non-zero exit code

## What happened instead?

The exit code for this process was 0

### Stack

```
   CocoaPods : 1.0.1
        Ruby : ruby 2.2.2p95 (2015-04-13 revision 50295) [x86_64-darwin15]
    RubyGems : 2.4.5
        Host : Mac OS X 10.11.6 (15G1004)
       Xcode : 8.0 (8A218a)
         Git : git version 2.10.0
Ruby lib dir : /Users/greenhouse/.rbenv/versions/2.2.2/lib
Repositories : master - https://github.com/CocoaPods/Specs.git @ e0683c231234ef55ca07a854234f003485120da4
```

### Installation Source

```
Executable Path: /Users/greenhouse/.rbenv/versions/2.2.2/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.1
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
cocoapods-trunk       : 1.0.0
cocoapods-try         : 1.1.0
```

### Project that demonstrates the issue

Not related to a specific project

orta:
Awesome, yeah, a PR adding this would be awesome

PR #6040, opened 2016-10-15
Exit with non-zero exit status if pod repo update fails

UkuLoskit:
**Edit (@dantoml)**
closes https://github.com/CocoaPods/CocoaPods/issues/6032


benasher44:
This looks good to me. I think this was a `UI.warn` before because at one point we would always update spec repos on `pod install`? Now you have to provide a flag or use `pod repo update`

DanToml:
@benasher44 Yeah I think that was the reason too.

DanToml:
@UkuLoskit Should be good with a rebase

PR #6069, opened 2016-10-23
Don't explicitly set ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES to NO

### LINKED ISSUE #6064: Swift standard libraries embedding explicitly set to NO by CocoaPods
# Report

## What did you do?

Include ObjC only pod libraries in your project, project uses a custom Swift framework not managed by CocoaPods.

CocoaPods should not explicitly set the `ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES` to `NO` inside the xcconfig files if it thinks no Swift runtime library is needed. Actually it should not set this option at all if it determines no Swift runtime is needed by the pods and thus leave it up to the project settings. Or there should at least be an option to tell CocoaPods that the Swift runtime libraries are needed even though it is not able to detect the requirement.

## What happened instead?

CocoaPods automatically sets the `ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES` to `NO` inside the xcconfig files  and shows a warning
```
[!] The `Test [Debug]` target overrides the `ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES` build setting defined in `Pods/Target Support Files/Pods-Test/Pods-Test.debug.xcconfig'. This can lead to problems with the CocoaPods installation
    - Use the `$(inherited)` flag, or
    - Remove the build settings from the target.
```

### Stack

```
   CocoaPods : 1.1.1
    RubyGems : 2.5.1
        Host : Mac OS X 10.12 (16A323)
       Xcode : 8.0 (8A218a)
         Git : git version 2.8.4 (Apple Git-73)
Ruby lib dir : /Users/Test/.rbenv/versions/2.3.1/lib
Repositories : master - https://github.com/CocoaPods/Specs.git @ cce89b3e3691f3ebe6841bf06b71639c3d105c0e
```

### Podfile

```ruby
# Uncomment this line to define a global platform for your project
# platform :ios, '9.0'

target 'Test' do
  # Uncomment this line if you're using Swift or would like to use dynamic frameworks
  use_frameworks!
end
```

benasher44:
This is the previous behavior for how we handled `EMBEDDED_CONTENT_CONTAINS_SWIFT`. I can't remember why I explicitly set the new one to NO. It's possible that it was related to an early Xcode beta issue, or I explicitly set it because that's what Xcode does when you apply the recommended build setting changes when updating to Xcode 8.

This fixes #6064.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 [Konrad Feiler](https://github.com/Bersaelor)
 [#6052](https://github.com/CocoaPods/CocoaPods/issues/6052)
 
+* Don't explicitly set `ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES` to NO
+  [Ben Asher](https://github.com/benasher44)
+  [#6064](https://github.com/CocoaPods/CocoaPods/issues/6064)
+
 
 ## 1.1.1 (2016-10-20)
 
--- a/lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
+++ b/lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
@@ -109,17 +109,14 @@ module Pod
 
           swift_version = Gem::Version.new(target_swift_version)
           should_embed = !target.requires_host_target? && pod_targets.any?(&:uses_swift?)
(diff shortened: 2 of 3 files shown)


benasher44:
This still needs an accompanying cocoapods-integration-specs PR, which is coming soon.

PR #6087, opened 2016-10-26
[CODE_OF_CONDUCT] clarify rationale

arielelkin:
🌈


orta:
Thanks for the PR, we spent a long time coming up with the current COC and are pretty happy with how it is. Appreciate the effort though.

arielelkin:
Hi @orta , thanks for your response 😀

Where could we learn more about how you came up with the current COC?

orta:
I'm afraid it was done on our private developers Slack, we looked through a lot of the existing COCs  at the time, discussed a lot of the ramifications and how we would handle them.

PR #6321, opened 2016-12-17
Removed extension integration errors about mismatched Swift version / platform

### LINKED ISSUE #6173: App extensions deployment target
> ℹ Please fill out this template when filing an issue.
> All lines beginning with an ℹ symbol instruct you with
> what info we expect.  
> Please remove this line and all above before submitting.

> Before you start, are you using the CocoaPods beta release?
> a lot has changed with Xcode 8 that is not backwards compatible.
>
> Xcode 8 Status: Only works with CocoaPods 1.1.0
> Swift 3: You need to add a `.swift-version` file, see the CHANGELOG
> China: Use https://gems.ruby-china.org/ instead of https://gems.ruby-china.org/ for the RC see #5910

# Report

## What did you do?
Created a pod file to install frameworks into 2 different targets. One of those targets is an NoticationServiceExtension that requires the deployment target to be set to 10, the other is the Main application with deployment target set to 9. Ran `pod install`

I tried working around this issue by setting the deployment target lower on the NotificaitonServiceExtension but it breaks deployment to older iOS devices with this error:
```
This app contains an app extension that specifies an extension
point identifier that is not supported on this version of iOS
for the value of the NSExtensionPointIdentifier key in its Info.plist
```

Pods to install.

## What happened instead?
Received this error:
```
[!] Unable to integrate the following embedded targets with their respective host targets (a host target is a "parent" target which embeds a "child" target like a framework or extension):

- Goat (iOS 9.0) and GoatServiceExtension (iOS 10.0) do not use the same platform.
```

## CocoaPods Environment

ℹ Please replace this with the output of `pod env`.

### Stack

```
   CocoaPods : 1.1.1
        Ruby : ruby 2.3.0p0 (2015-12-25 revision 53290) [x86_64-darwin15]
    RubyGems : 2.5.1
        Host : Mac OS X 10.11.6 (15G31)
       Xcode : 8.1 (8B62)
         Git : git version 2.10.1
Ruby lib dir : /usr/local/var/rbenv/versions/2.3.0/lib
Repositories : master - https://github.com/CocoaPods/Specs.git @ 456d4f3f7206e052befa18da85c1bfe287c4cd5c
```

### Installation Source

```
Executable Path: /usr/local/var/rbenv/versions/2.3.0/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.1
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
cocoapods-trunk       : 1.1.1
cocoapods-try         : 1.1.0
```

### Podfile

```ruby
use_frameworks!

project 'Goat/Goat.xcodeproj'

target 'Goat' do
  pod 'UrbanAirship-iOS-SDK'
end

target 'GoatServiceExtension' do
  pod 'UrbanAirship-iOS-AppExtensions'
end

```

benasher44:
This fixes #6173. Xcode should warn you about these issues, so trying to maintain code that replicates this at `pod install` time seems like less of a great idea now. `use_frameworks!` issues are still ones that would be fixed with your CP setup though, so that error will stay.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Ben Asher](https://github.com/benasher44)
   [#6123](https://github.com/CocoaPods/CocoaPods/issues/6123)
 
+* Remove errors that prevent host/extension target mismatches, which Xcode will warn about.
+  [Ben Asher](https://github.com/benasher44)
+  [#6173](https://github.com/CocoaPods/CocoaPods/issues/6173)
+
 
 ## 1.2.0.beta.1 (2016-10-28)
 
--- a/lib/cocoapods/installer/analyzer.rb
+++ b/lib/cocoapods/installer/analyzer.rb
@@ -324,19 +324,12 @@ module Pod
         end
 
         target_mismatches = []
-        check_prop = lambda do |target_definition1, target_definition2, attr, msg|
-          attr1 = target_definition1.send(attr)
-          attr2 = target_definition2.send(attr)
-          if attr1 != attr2
-            target_mismatches << "- #{target_definition1.name} (#{attr1}) and #{target_definition2.name} (#{attr2}) #{msg}."
-          end
-        end
         host_uuid_to_embedded_target_definitions.each do |uuid, target_definitions|
           host_target_definition = target_definitions_by_uuid[uuid]
           target_definitions.each do |target_definition|
-            check_prop.call(host_target_definition, target_definition, :platform, 'do not use the same platform')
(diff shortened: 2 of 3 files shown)


benasher44:
One of the reasons this was done in the first place was to prevent errors like in #5864, but we've since removed Swift version from `PodVariant`. I'm a bit worried this could bring back this error but for mismatched platforms. This seems a lot less likely though, and I don't think this is the best place to handle such an error.

benasher44:
I agree. I think this is the right change for now, but we can certainly revisit it.

PR #6435, opened 2017-01-26
Correctly set runtime search paths for OSX unit test bundles

dnkoutso:


lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
```diff
@@ -217,8 +217,9 @@ def generate_ld_runpath_search_paths
           ld_runpath_search_paths = ['$(inherited)']
           if target.platform.symbolic_name == :osx
             ld_runpath_search_paths << "'@executable_path/../Frameworks'"
+            symbol_type = target.user_targets.map(&:symbol_type).uniq.first
```
dnkoutso: Is it possible that an `aggregate_target` integrate more than one `user_target`? If so what is the scenario I can use to test this out and what should the expected behavior be?

dnkoutso:
Will need to bump integration specs.

benasher44:
My only thought is maybe abstract targets can have multiple user targets? I haven't verified that though.

benasher44:
I think in that case, you'd just skip this for abstract targets, for which I think there is a flag/method you can call to ask that.

dnkoutso:
Will update.

dnkoutso:
@benasher44 I simulated this scenario but `target.user_targets` never returns abstract targets:

```
abstract_target 'Abstract1' do
  abstract_target 'Abstract2' do
    target 'SampleApp Mac Tests' do
      platform :osx, '10.10'
      pod 'OCMock', '~> 3.1.2'
    end
  end 
end
```

benasher44:
I believe target itself would be abstract. The `user_targets` should always be `PBXNativeTarget` types.

dnkoutso:
Ah, looks like still that the `generate_ld_runpath_search_paths` is only invoked for non-abstract targets.

if I print out the `target.name` I see `Pods-Abstract1-Abstract2-SampleApp Mac Tests`

DanToml:
Yep that should all be handled in analyzer/variants. This change LGTM.

dnkoutso:
Btw creating a new target in xcode for a macOS unit test bundle automatically creates this rpath:

screen shot 2017-02-05 at 8 55 04 am

benasher44:
This change looks good to me. Just waiting on merge conflict resolution and green CI

dnkoutso:
@benasher44 the corresponding change for integration specs is here https://github.com/CocoaPods/cocoapods-integration-specs/pull/95

DanToml:
@dnkoutso looks like there is one failure left but I'm not sure why. Maybe needs a rebase?

dnkoutso:
I just did lets see :)

dnkoutso:
Can I merge? :)

DanToml:
Whoop, hadn't seen the ci pass, let ship it

PR #6393, opened 2017-01-10
Fix build phase resource references to point at PBXVariantGroups where relevant

Westacular:
Fixes issue reported at https://github.com/CocoaPods/CocoaPods/issues/6373

This just aligns the `add_files_to_build_phases` behaviour with what is
already (correctly) done in `add_resources_bundle_targets`.

PR #6462, opened 2017-02-04
Fix resources with PBXVariantGroups outside of a resource bundle

### LINKED ISSUE #6373: Target build phase references individual file_refs for localized resources instead of their PBXVariantGroups
* [ x ] I've read and understood the [*CONTRIBUTING guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

Localized resources in that are included via `.resources` (rather than `.resource_bundles`) are added to their target slightly incorrectly.

Specifically, what is added to the target are _individual_ file references to each component of the localized resource (e.g. Base.lproj/ViewController.xib and fr.lproj/ViewController.strings), rather than the PBXVariantGroup created to logically represent the overall resource.

This doesn't affect the built app — in that regard, this problem appears to be benign.

What it does affect:

- In Xcode's file inspector, these resources don't show as being members of their targets
- Xcode's "Export for localization..." feature ignores these resources entirely.

The latter is what led me to discover this issue. It appears that the "Export for localization" action traverses the targets, looking for PBXVariantGroups, as the basis for what it exports. Although appropriate PBXVariantGroups exist for resources, the target doesn't directly reference them, and they are ignored.

I believe the solution already exists in `lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb`:

In `add_files_to_build_phases`, line 114, it has:

```
resource_refs = file_accessor.resources.flatten.map do |res|
	project.reference_for_path(res)
end

# Some nested files are not directly present in the Xcode project, such as the contents
# of an .xcdatamodeld directory. These files will return nil file references.
resource_refs.compact!
```

Meanwhile, the corresponding code in `add_resources_bundle_targets`, line 141:

```
file_references = paths.map do |path|
  ref = project.reference_for_path(path)

  # Some nested files are not directly present in the Xcode project, such as the contents
  # of an .xcdatamodeld directory. These files are implicitly included by including their
  # parent directory.
  next if ref.nil?

  # For variant groups, the variant group itself is added, not its members.
  next ref.parent if ref.parent.is_a?(Xcodeproj::Project::Object::PBXVariantGroup)

  ref
end
file_references = file_references.uniq.compact
```

I believe the logic of the latter should also be applied in the former case, as well. I've tested with this change locally, and it had the desired effect: "Export for localization" created an XLIFF that included the formerly-missing files and their strings.

Versions used:
CocoaPods 1.1.1
Xcode 8.2.1

dnkoutso:
closes #6373 
completes #6393 by adding a test and extracting a method.

/cc @Westacular @benasher44

### DIFF
--- a/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
@@ -65,6 +65,31 @@ module Pod
             settings
           end
 
+          # Filters the given resource file references discarding empty paths which are
+          # added by their parent directory. This will also include references to the parent [PBXVariantGroup]
+          # for all resources underneath it.
+          #
+          # @param  Array[<Pathname>] resource_file_references
+          #         The array of all resource file references to filter.
+
+          # @return [Array<Pathname>] The filtered resource file references.
+          #
+          def filter_resource_file_references(resource_file_references)
+            resource_file_references.map do |resource_file_reference|
+              ref = project.reference_for_path(resource_file_reference)
+
+              # Some nested files are not directly present in the Xcode project, such as the contents
+              # of an .xcdatamodeld directory. These files are implicitly included by including their
+              # parent directory.
+              next if ref.nil?
+
+              # For variant groups, the variant group itself is added, not its members.
+              next ref.parent if ref.parent.is_a?(Xcodeproj::Project::Object::PBXVariantGroup)
+
+              ref
+            end
+          end
+
           #-----------------------------------------------------------------------#
 
           SOURCE_FILE_EXTENSIONS = Sandbox::FileAccessor::SOURCE_FILE_EXTENSIONS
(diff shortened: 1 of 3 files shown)


lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -65,6 +65,31 @@ def custom_build_settings
             settings
           end
 
+          # Filters the given resource file references discarding empty paths which are
```
dnkoutso: @benasher44 If you have better wording here, I'd like to hear it as I don't like it so much :)

benasher44: Hm. I don't have any specific suggestions. I would err on the side of verbosity, if you feel like a longer explanation might work better :)

dnkoutso: Nah I think its good, just wanted to double check with you.

Issue #6065, opened 2016-10-21
Duplicate classes in tests from generated xcconfigs for test targets with target dependencies


morganchen12:
🌈
- [x] I've read and understood the [CONTRIBUTING](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md) guidelines and have done my best effort to follow.
# Report
## What did you do?

I have a dynamic framework project with a handful of dependencies and a unit test target that lists the framework as a target dependency. My `Podfile` looks like this:

``` ruby
target 'FirebaseDatabaseUI' do
  use_frameworks!

  # Pods for Database
  pod 'Firebase/Database'

  target 'FirebaseDatabaseUITests' do
    inherit! :search_paths
  end
end
```

I noticed in one of the betas (definitely before 1.1.0.rc.2, I think around 1.1.0.beta.2) that duplicate classes from the dependencies were showing up in the unit tests runtime despite using `inherit! :search_paths`. After some hunting I found that CocoaPods was adding all of the test target's dependency's dependencies into the xcconfig file.

```
OTHER_LDFLAGS = $(inherited) -ObjC -l"c++" -l"icucore" -l"sqlite3" -l"z" 
  -framework "AddressBook" 
  -framework "CFNetwork" 
  -framework "CoreGraphics" 
  -framework "FirebaseAnalytics" 
  -framework "FirebaseCore" 
  -framework "FirebaseDatabase" 
  -framework "FirebaseInstanceID" 
  -framework "GoogleInterchangeUtilities" 
  -framework "GoogleSymbolUtilities" 
  -framework "GoogleUtilities" 
  -framework "Security" 
  -framework "StoreKit" 
  -framework "SystemConfiguration"
```

This issue is still present even if the tests target is moved out of its parent.

``` ruby
target 'FirebaseDatabaseUI' do
  use_frameworks!

  # Pods for Database
  pod 'Firebase/Database'
end

target 'FirebaseDatabaseUITests' do
  use_frameworks!
end
```

Deleting the `-framework` flags from the `xcconfig` file fixes the issue temporarily until the next `pod install`.

CocoaPods 1.0.1 doesn't have this issue, as it doesn't add any `ld` flags to the test target.

### Workaround

[...]

PR #5514, opened 2016-06-13
Add aggregated search paths targets to vendored build settings

plu:
Fix #5512 

I assume this will break some tests. Before fixing/adapting them I'd like to see your approval that this is the correct fix :).

PR #6477, opened 2017-02-09
Fix integration with vendored static frameworks and libraries

dnkoutso:
This fixes a few things:

* Search paths will now properly be added even for targets with vendored static frameworks or libraries (regression from https://github.com/CocoaPods/CocoaPods/pull/5514)
* Just like static frameworks, static libraries will now **not** be linked to targets that use `inherit! search_paths` (from https://github.com/CocoaPods/CocoaPods/issues/6065#issuecomment-276815528)
* Fixes an issue where if a target with `inherit! search_paths` explicitly defines a pod that brings in a static framework or library will not have its `other_ldflags` properly wired up, now it will.
* Also fixes an issue where transitive static dependencies would still get linked when they should not be.

/cc @plu @benasher44 @DanToml @segiddins

### DIFF
--- a/lib/cocoapods/generator/xcconfig/xcconfig_helper.rb
+++ b/lib/cocoapods/generator/xcconfig/xcconfig_helper.rb
@@ -103,13 +103,34 @@ module Pod
         #
         def self.add_static_dependency_build_settings(aggregate_target, pod_target, xcconfig, file_accessor)
           file_accessor.vendored_static_frameworks.each do |vendored_static_framework|
-            if aggregate_target.nil? || aggregate_target.target_definition.inheritance != 'search_paths'
-              XCConfigHelper.add_framework_build_settings(vendored_static_framework, xcconfig, pod_target.sandbox.root)
-            end
+            adds_other_ldflags = XCConfigHelper.links_static_dependency?(aggregate_target, pod_target)
+            XCConfigHelper.add_framework_build_settings(vendored_static_framework, xcconfig, pod_target.sandbox.root, adds_other_ldflags)
           end
           file_accessor.vendored_static_libraries.each do |vendored_static_library|
-            XCConfigHelper.add_library_build_settings(vendored_static_library, xcconfig, pod_target.sandbox.root)
+            adds_other_ldflags = XCConfigHelper.links_static_dependency?(aggregate_target, pod_target)
+            XCConfigHelper.add_library_build_settings(vendored_static_library, xcconfig, pod_target.sandbox.root, adds_other_ldflags)
           end
+        end
+
+        # @param  [AggregateTarget] aggregate_target
+        #         The aggregate target, may be nil.
+        #
+        # @param  [PodTarget] pod_target
+        #         The pod target to link or not.
+        #
+        # @return [Boolean] Whether static dependency should be added to the 'OTHER_LDFLAGS'
+        #         of the aggregate target. Aggregate targets that inherit search paths will only link
+        #         if the target has explicitly declared the pod dependency.
+        #
+        def self.links_static_dependency?(aggregate_target, pod_target)
+          return true if aggregate_target.nil? || aggregate_target.target_definition.inheritance != 'search_paths'
+          # Only link this static  dependency if its explicitly defined for this target.
+          aggregate_target.target_definition.non_inherited_dependencies.each do |d|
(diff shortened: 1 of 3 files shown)


lib/cocoapods/generator/xcconfig/xcconfig_helper.rb
```diff
@@ -103,15 +103,23 @@ def self.add_settings_for_file_accessors_of_target(aggregate_target, pod_target,
...
+            XCConfigHelper.add_framework_build_settings(vendored_static_framework, xcconfig, pod_target.sandbox.root, should_add_other_ldflags)
           end
           file_accessor.vendored_static_libraries.each do |vendored_static_library|
-            XCConfigHelper.add_library_build_settings(vendored_static_library, xcconfig, pod_target.sandbox.root)
+            should_add_other_ldflags = XCConfigHelper.should_link_static_framework_or_library(aggregate_target, pod_target)
+            XCConfigHelper.add_library_build_settings(vendored_static_library, xcconfig, pod_target.sandbox.root, should_add_other_ldflags)
           end
         end
 
+        # TODO
```
dnkoutso: will take care of this.

lib/cocoapods/generator/xcconfig/xcconfig_helper.rb
```diff
@@ -103,15 +103,23 @@ def self.add_settings_for_file_accessors_of_target(aggregate_target, pod_target,
...
+        #
+        def self.should_link_static_framework_or_library(aggregate_target, pod_target)
+          return true if aggregate_target.nil? || aggregate_target.target_definition.inheritance != 'search_paths'
+          return aggregate_target.pod_targets.include?(pod_target)
```
dnkoutso: @benasher44 one of the flaws this logic has is that transitive dependencies with pre-built frameworks/libraries are still included in the `other_ldflags`. I might need a bit of help to tackle this.

lib/cocoapods/generator/xcconfig/xcconfig_helper.rb
```diff
@@ -165,13 +173,13 @@ def self.add_spec_build_settings_to_xcconfig(consumer, xcconfig)
         #
         # @return [void]
         #
-        def self.add_framework_build_settings(framework_path, xcconfig, sandbox_root)
+        def self.add_framework_build_settings(framework_path, xcconfig, sandbox_root, include_other_ldflags)
```
segiddins: this either needs to be a default argument or the other callsites need to be updated -- see the test failures

spec/unit/generator/xcconfig/xcconfig_helper_spec.rb
```diff
@@ -124,7 +124,7 @@ module XCConfig
           it 'adds the build settings of a framework to the given xcconfig' do
             framework_path = config.sandbox.root + 'Parse/Parse.framework'
             xcconfig = Xcodeproj::Config.new
-            @sut.add_framework_build_settings(framework_path, xcconfig, config.sandbox.root)
+            @sut.add_framework_build_settings(framework_path, xcconfig, config.sandbox.root, true)
```
dnkoutso: went with the default `true` parameter so no need to add this

dnkoutso:
This should be ready now

dnkoutso:
Awesome!

I still think that when a target uses `inherit! search_paths` none of its source dependencies should be linked either. This is out of the scope of this pull request though.

plu:
Thank you for taking the time to fix this!

About source dependencies: Might be necessary if your test target does not use a host app. Didn't  verify that assumption though, so I might be wrong.

dnkoutso:
@plu correct. I was under the assumption the use of a host app is the default rather than not.

justinseanmartin:
If the test target doesn't use a host app, the Podfile shouldn't use `inherit! search_paths` to  actually link the other libraries into the xctest bundle.

dnkoutso:
@plu I think I take it back. It *does* correctly skip linking sources, but there is a case with development pods that it might actually not. Seems like a separate bug.

FlyingNoob:
@dnkoutso I do the test for this issue 'transitive static dependencies'. My private pod lib dependency a vendor lib which include '.a' static lib .  pod 'my private pod' , execute pod install . I get the error 'target has transitive dependencies that include static binaries'

PR #6497, opened 2017-02-20
Provide installation option to disable multiple pod sources warnings

dnkoutso:
We have 100+ pods where we publish both as sources and as binary. Sometimes these warnings can become extremely long during `pod install`. 

Resolving this issue also requires to re-enter these pods in the `Podfile` and provide an explicit `:source` to silence this warning. This in turn forces the pods to be linked to a target even though they do not necessarily have to.

The best (long term solution) is to provide:

```ruby
source 'https://github.com/CocoaPods/Specs' do
  pod ...
  pod ...
end
```

But this option here is a good balance to prevent the warning from happening. Please note the default still is to warn so this shouldn't affect anyone.

Usage:

```ruby
install! 'cocoapods', warn_for_multiple_pod_sources: false
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
   [#6534](https://github.com/CocoaPods/CocoaPods/issues/6534)
   
+* Provide installation option to disable multiple pod sources warnings.  
+  [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
+  [#6497](https://github.com/CocoaPods/CocoaPods/pull/6497)
+
 * Use the colored2 gem instead of colored.  
   [Orta Therox](https://github.com/orta)
   [xcodeproj#463](https://github.com/CocoaPods/Xcodeproj/pull/463)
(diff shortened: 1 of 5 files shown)


dnkoutso:
There are no tests for `lazy_specification.rb` :|

orta:
Seems 👍 to me

dnkoutso:
@orta not sure if I should land this. Thinking that maybe its better to compact the message instead....after talking to @DanToml. Not sure what to do or if its that important...

benasher44:
@dnkoutso compacting the warning sounds like a nice improvement, and as a bonus it sounds like there wouldn't need to be new API?

dnkoutso:
@benasher44 @DanToml with your permission even if we compact this I still think we can merge this. It's relatively harmless and I am not sure how much more compact the message can become. If we make it super compact it might become unreadable.

DanToml:
:ship:

Whirlwind:
If a dependency is from a Pod, and I hope all dependency search my private source default, I must write all dependencies in my Podfile!

I could not list all dependencies from my private source at first time, so I must repeat try to add they to my Podfile. I think it is a bad way.

benasher44:
@Whirlwind this PR adds a new installation option, so there should be no change to default behavior. I'm not entirely sure what issue you're experiencing, but we'd appreciate a new open issue with repro steps. Thanks!

PR #6533, opened 2017-03-03
Updated colored2 gem

benasher44:
It was locked to 3.1.1 before, which appears to have been removed from rubygems.org. This updates it to 3.1.2. The diff looks like mostly travis and readme changes in their repo: https://github.com/kigster/colored2/compare/v3.1.1...v3.1.2.


orta:
The fail looks to just be a travis hiccup, not related to the changes.

benasher44:
@orta thanks!

Issue #6033, opened 2016-10-13
Exit with a special error code when `pod install` fails due to an out of date specs repo


juli1quere:
🌈
## What did you do?

`pod install` on a computer (a CI) who has not done a `pod repo update` recently with a `PodFile.lock` referencing a brand new version of a pod as dependancy. 
## What did you expect to happen?

I know that `pod repo update` is not automatic anymore to reduce pressure on Github servers. I know that I could pass the `--repo-update` parameter to `pod`. 

But It would be great if `pod` could detect the situation and say "Oh, hey, I see that PodFile.lock require a version that I don't have on my repo local copy. Let's try to do a `pod repo update` and `pod install` again !" 

I think it would be a good balance between the need to have something that works every time and the execution time / servers charge. Especially when you use Cocoapods on a CI.

At least, it could be an option like `--repo-update-when-needed` (or something less verbose :/) 
## What happened instead?

```
[13:00] onejjy@Mac-Julien-3:Demo $ pod install
Analyzing dependencies
[!] Unable to satisfy the following requirements:

- `RealmSwift` required by `Podfile`
- `RealmSwift (= 2.0.1)` required by `Podfile.lock`

None of your spec sources contain a spec satisfying the dependencies: `RealmSwift, RealmSwift (= 2.0.1)`.

You have either:
 * out-of-date source repos which you can update with `pod repo update`.
 * mistyped the name or version.
 * not added the source repo that hosts the Podspec to your Podfile.

Note: as of CocoaPods 1.0, `pod repo update` does not happen on `pod install` by default.
```
### Stack

```
   CocoaPods : 1.0.1
        Ruby : ruby 2.0.0p648 (2015-12-16 revision 53162) [universal.x86_64-darwin15]
    RubyGems : 2.4.8
        Host : Mac OS X 10.11.6 (15G1004)
       Xcode : 7.3 (7D175)
         Git : git version 2.7.2
Ruby lib dir : /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib
Repositories : master - https://github.com/CocoaPods/Specs.git @ 3f0c8ff9097ab484f65e1b4ba3982c55f664e434
```
### Installation Source

```
Executable Path: /Users/onejjy/.gemhome/bin/pod
```
### Plugins

```
cocoapods-deintegrate : 1.0.1
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
[...]

PR #6553, opened 2017-03-13
Return different exit status (31) when spec not found

adellibovi:
Hi,

As part of #6033, this returns a different exit code when spec not found (i.e. master repo out-dated).
I couldn't find any other examples in the project of exit codes, so I'm not sure if we have any convention to follow on this, do you have any hint?
I have few other questions:
Should we document exit codes somewhere?
Is having a NoSpecFound class the right approach, if so where is the right place? Or is good enough to just set @exit_status?

### DIFF
--- a/lib/cocoapods/command/install.rb
+++ b/lib/cocoapods/command/install.rb
@@ -21,6 +21,10 @@ module Pod
         This will configure the project to reference the Pods static library,
         add a build configuration file, and add a post build script to copy
         Pod resources.
+
+        This may return one of several error codes if it encounters problems.
+        * `1` Generic error code
+        * `31` Spec not found (i.e out-of-date source repos, mistyped Pod name etc...)
       DESC
 
       def self.options
--- a/lib/cocoapods/resolver.rb
+++ b/lib/cocoapods/resolver.rb
@@ -2,6 +2,12 @@ require 'molinillo'
 require 'cocoapods/resolver/lazy_specification'
 
 module Pod
+  class NoSpecFoundError < Informative
+    def exit_status
+      @exit_status ||= 31
+    end
+  end
+
   # The resolver is responsible of generating a list of specifications grouped
   # by target for a given Podfile.
   #
@@ -389,6 +395,7 @@ module Pod
     #
     def handle_resolver_error(error)
       message = error.message.dup
+      type = Informative
       case error
       when Molinillo::VersionConflict
         error.conflicts.each do |name, conflict|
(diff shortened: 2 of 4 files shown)


dnkoutso:
@adellibovi can you add a test for this?

adellibovi:
@dnkoutso I was thinking of updating resolver_spec: 'raises if no such version of a dependency exists' and 'raises with a list of dependencies if there are many dependencies but no versions of a dependency exists' to check that it raise the new class and that the exit_status is the one expected. What do you think?

orta:
Looking at the "standards-ish" page here: http://tldp.org/LDP/abs/html/exitcodes.html

I think perhaps the closest mapping we have is `127`?

With respect to documentation, we could get them put in https://guides.cocoapods.org/terminal/commands.html#pod_install if you wanted? (the docs are derived from the source, so that can come in with this PR)

adellibovi:
127 LGTM 😊

adellibovi:
Actually, maybe we should use one that is not "reserved"? What do you think?

lib/cocoapods/command/install.rb
```diff
@@ -21,6 +21,10 @@ class Install < Command
         This will configure the project to reference the Pods static library,
         add a build configuration file, and add a post build script to copy
         Pod resources.
+
```
adellibovi: @orta, what do you think about the formatting?

benasher44:
Maybe the `EX_NOINPUT` sys exit (66) makes sense? I couldn't find a constant defined like this for ruby (python has one).

adellibovi:
Mmh, yes, @benasher44 , I agree with that. `EX_NOINPUT` looks more inline with this type of error. But what I would like to question is: what if we will have other cases that could fit`EX_NOINPUT`? One of the main benefits of this PR is enabling CI to redo a `pod install --repo-update` based on the error code. That's why I was thinking of having a _very_ specific exit code. We could go with the `EX_NOINPUT` but then I suggest to not reuse for other cases. Thoughts?

lib/cocoapods/command/install.rb
orta: :+1:

benasher44:
Hm okay very specific sounds good!

benasher44:
If you could define some constants for these and include code comments about the decision, that'd be helpful too!

adellibovi:
Ok @benasher44, which is the best place for those constants?

benasher44:
hmm. Maybe command.rb? Thoughts @dantoml?

adellibovi:
Another option would be to create an exceptions.rb and move there the new error class and the constant, that's an approach also used by others

DanToml:
@benasher44 @adellibovi I would probably leave this here and follow the way we do it for stuff like DownloaderError (keeping it defined in the relevant namespace). It should be renamed to `NoSpecFoundError` tho.

adellibovi:
@DanToml renamed 👍

lib/cocoapods/resolver.rb
```diff
@@ -2,6 +2,13 @@
 require 'cocoapods/resolver/lazy_specification'
 
 module Pod
+
```
DanToml: rm this line to make Rubocop :smile: (can also run `bundle exec rubocop --auto-correct lib`)

Issue #6403, opened 2017-01-12
`lint`-ing lib including CoreData CodeGen leads to `file not found` error


CharlesThierry:
🌈

* [x] I've read and understood the [*CONTRIBUTING guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?
I tried to lint a lib (created using the `pod create lib` template) that uses AutoGenerated CoreData managed object categories.

## What did you expect to happen?
To be amazed by my lib being lint'd without an issue.

## What happened instead?
Was stricken down by the compiler informing me that autogenerated code was apparently not generated 
(`DRTestEntity+CoreDataProperties.m: No such file or directory` using the exemple below).  
It could be related to #6259 

### Stack

```
   CocoaPods : 1.1.1
        Ruby : ruby 2.3.0p0 (2015-12-25 revision 53290) [x86_64-darwin15]
    RubyGems : 2.5.1
        Host : Mac OS X 10.11.6 (15G1212)
       Xcode : 8.2 (8C38)
         Git : git version 2.10.1 (Apple Git-78)
Ruby lib dir : /Users/charles/.rvm/rubies/ruby-2.3.0/lib
               master - https://github.com/CocoaPods/Specs.git @ a4160c198470400b7e7fd469a636d975b9d2a2ca
```

### Installation Source

```
Executable Path: /Users/charles/.rvm/gems/ruby-2.3.0/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.1
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
cocoapods-trunk       : 1.1.1
cocoapods-try         : 1.1.0
```

## Project that demonstrates the issue
Problem reproduced with a minimal example code that can be found here: [https://github.com/DeRunco/DRCoreDataTest](https://github.com/DeRunco/DRCoreDataTest)
```
git clone git@github.com:DeRunco/DRCoreDataTest.git
```

benasher44:
Yep! This is a dup. Thanks for filing though! Your demo project is really useful, and I'll be using it as I work on this one. Cheers!

PR #6560, opened 2017-03-15
Added special handling for xcdatamodeld resources

### LINKED ISSUE #6259: Autogenerated NSManagedObject Subclasses for Core Data can lead to linker error

# Report

## What did you do?

Building & running an app (written in Objective-C) that uses pods and has autogenerated NSManagedObject subclasses for core data in its target. Those classes were freshly created using the `Create NSManagedObject Subclasses...` command in the editor.
I am using Xcode 8.1.

I expected the build to succeed normally with the new subclasses.

## What happened instead?

The linker failed to link the .o files. On further inspection it seems this happened during one of the build phases defined by cocoapods (i.e. in a `Pods-...-.xcconfig` file).

### Stack

```
   CocoaPods : 1.1.1
        Ruby : ruby 2.0.0p648 (2015-12-16 revision 53162) [universal.x86_64-darwin16]
    RubyGems : 2.6.3
        Host : Mac OS X 10.12.1 (16B2555)
       Xcode : 8.1 (8B62)
         Git : git version 2.9.3 (Apple Git-75)
Ruby lib dir : /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib
Repositories : master - https://github.com/CocoaPods/Specs.git @ fd21d0686bee5dc036a305c3c420b2f0c9957214
```

### Installation Source

```
Executable Path: /usr/local/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.1
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
cocoapods-trunk       : 1.1.1
cocoapods-try         : 1.1.0
slather               : 2.3.0
```

### Podfile

```ruby
# Uncomment the next line to define a global platform for your project
platform :ios, '9.0'

target 'GHPodsAndCDAutogenCode' do
  # Uncomment the next line if you're using Swift or would like to use dynamic frameworks
  # use_frameworks!

  # Pods for GHPodsAndCDAutogenCode
  pod 'AFNetworking', '~> 3.1'

  target 'GHPodsAndCDAutogenCodeTests' do
    inherit! :search_paths
    # Pods for testing
    pod 'Specta'
    pod 'Expecta'
    pod 'OCMock'
    pod 'OHHTTPStubs'
  end

end
```

[Demo project](https://github.com/GeroHerkenrath/GHPodsAndCDAutogenCode). Please see its `README-autogenerated-files` to see how to switch the issue on and off. The above `pod env` stats are from this example project.

## Workaround

After creating the NSManagedObject subclasses with Xcode, select their corresponding entities and set their `Codegen` setting to `Manual/None`. This will not delete the freshly created classes from your project, but resolve the issue for the linker. If you need to re-create the classes again later, revert this setting (the default is `Class Definition`). 
I documented and explained this workaround a bit in my demo project above as well.

benasher44:
This fixes #6259. This adds xcdatamodeld files to the sources build phases, so that Xcode-generated headers (categories that add nice accessors for the models) can be built in time to be used in the pod. This is not enabled for resource bundles, since these kinds of models don't seem to build properly (error) in any build phase for resource bundles.

### DIFF
--- a/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
@@ -80,7 +80,7 @@ module Pod
           #
           # @note   Core Data model directories (.xcdatamodeld) used to be added to the
           #         `Copy Resources` build phase like all other resources, since they would
-          #         compile correctly in either the resources or compile phaes. In recent
+          #         compile correctly in either the resources or compile phase. In recent
           #         versions of xcode, there's an exception for data models that generate
           #         headers. These need to be added to the compile sources phase of a real
           #         target for the headers to be built in time for code in the target to
@@ -101,7 +101,7 @@ module Pod
 
               ref
             end.compact.uniq
-            compile_phase_matcher = lambda {|ref| !(ref.path =~ /.*\.xcdatamodeld/i).nil? }
+            compile_phase_matcher = lambda { |ref| !(ref.path =~ /.*\.xcdatamodeld/i).nil? }
             resources_phase_refs = file_references.reject(&compile_phase_matcher)
             compile_phase_refs = file_references.select(&compile_phase_matcher)
             yield resources_phase_refs, compile_phase_refs
@@ -170,7 +170,6 @@ module Pod
           def add_resources_bundle_targets
             target.file_accessors.each do |file_accessor|
               file_accessor.resource_bundles.each do |bundle_name, paths|
-
                 label = target.resources_bundle_target_label(bundle_name)
                 bundle_target = project.new_resources_bundle(label, file_accessor.spec_consumer.platform_name)
                 bundle_target.product_reference.tap do |bundle_product|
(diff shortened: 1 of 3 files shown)


lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -69,13 +69,26 @@ def custom_build_settings
...
+          #
+          # @yield_param  [Array<PBXFileReference>} The filtered resource file references to be installed
+          #               in the copy resources phase.
+          #
+          # @yield_param  [Array<PBXFileReference>} The filtered resource file references to be installed
+          #               in the compile sources phase.
+          #
+          # @note   Core Data model directories (.xcdatamodeld) used to be added to the
+          #         `Copy Resources` build phase like all other resources, since they would
+          #         compile correctly in either the resources or compile phaes. In recent
```
dnkoutso: nit: typo `phaes`

dnkoutso:
Nice. Have you tested this with #6259 specifically that it works properly?

benasher44:
Yep!

benasher44:
Er actually I was testing with a different sample project for some reason called DRCoreDataTest. Don't remember where I found it, but it's the same deal where it imports the category file generated by Xcode and used the generated property from the data model. It fails to build before this PR, and it works after.

benasher44:
Ah it's from a dup: #6403

benasher44:
I should also note that I tried just reordering the build phases so that the resources phase would build first, which would theoretically build the data model and related headers in time. Unfortunately, this causes the build to fail the first time and pass the 2nd. Not sure why, but maybe an Xcode bug.

PR #6563, opened 2017-03-16
[Resolver] Improve error message when specs not found

adellibovi:
Hi,

As part of #6033, this improves the error message when specs not found (i.e. source repo out-dated), and there is also a small fix in the tests.

Another thought, I think it would be nice to hide the message about "try `pod install --repo-update`" if the error is raised while executing that command, since it means that user have updated the repos but the error is probably one of the other two (mistyped or no source repo).
Should I change the analyser init to pass if it's called within a repo update? Or do you have any other ideas on how I can access that property?

### DIFF
--- a/lib/cocoapods/installer/analyzer.rb
+++ b/lib/cocoapods/installer/analyzer.rb
@@ -156,6 +156,11 @@ module Pod
 
       private
 
+      # @return [Bool] Whether the analysis has updated sources repositories.
+      #
+      attr_accessor :specs_updated
+      alias_method :specs_updated?, :specs_updated
+
       def validate_podfile!
         validator = Installer::PodfileValidator.new(podfile)
         validator.validate
@@ -222,6 +227,7 @@ module Pod
             UI.message "Skipping `#{source.name}` update because the repository is not a git source repository."
           end
         end
+        @specs_updated = true
       end
 
       private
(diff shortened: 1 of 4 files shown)


adellibovi:
@benasher44 thoughts on the error message and about hiding it when not needed? :)

lib/cocoapods/resolver.rb
```diff
@@ -439,7 +439,7 @@ def handle_resolver_error(error)
               message << "\n\nNone of your spec sources contain a spec satisfying "\
                 "the #{dependencies}: `#{conflicts.join(', ')}`." \
                 "\n\nYou have either:" \
-                "\n * out-of-date source repos which you can update with `pod repo update`." \
+                "\n * out-of-date source repos, try to run `pod install --repo-update` which will update them." \
```
benasher44: Can we include the previous command but mention the `pod install --repo-update` option as an alternative?

benasher44:
Hiding that part of the message if we know they're using the `--repo-update` flag sounds good to me!

adellibovi:
Hi @benasher44, I edited the message having both commands is that ok? I also implemented the hiding logic, do you think is the correct way to pass repo_updated?

I still have to add the spec for testing this last thing.

lib/cocoapods/installer/analyzer.rb
```diff
@@ -156,6 +156,11 @@ def update_mode
 
       private
 
+      # @return [Bool] Whether the analysis has updated sources repositories.
+      #
+      attr_accessor :repo_updated
```
benasher44: nit: can we call this `specs_updated`? That `repo` is a bit generic imo.

adellibovi: You are definitely right! Good point 👍

benasher44:
Thanks @adellibovi!

Issue #6457, opened 2017-02-02
Using specific git repository with Podspec that has syntax error will say no Podspec found


Legoless:
# Report

## What did you do?

Trying to install a pod from a specific git repository, that stores the Podspec file. This is the entry in Podfile:

```
pod 'Kvitto', :git => 'https://github.com/cocoanetics/kvitto.git'
```

After running `pod install`, this is the output:
```
Pre-downloading: `Kvitto` from `https://github.com/cocoanetics/kvitto.git`
[!] Unable to find a specification for 'Kvitto'.
```

## What did you expect to happen?

The specification is available on the URL, but the Podspec file has syntax error. The error output by Cocoapods, should clearly state that Podspec was found, but was unable to be parsed.

## What happened instead?
Cocoapods error was to unable to find specification. Since Podspec file is there, I was looking for error in wrong places for quite a while.

### Stack

```
   CocoaPods : 1.2.0
        Ruby : ruby 2.0.0p648 (2015-12-16 revision 53162) [universal.x86_64-darwin16]
    RubyGems : 2.0.14.1
        Host : Mac OS X 10.12.3 (16D32)
       Xcode : 8.2.1 (8C1002)
         Git : git version 2.10.1 (Apple Git-78)
Ruby lib dir : /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib
Repositories : 
               master - https://github.com/CocoaPods/Specs.git @ a0eeb5b2f70ec81c193486cf869f6ca29ed38293
```

### Installation Source

```
Executable Path: /usr/local/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.1
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
cocoapods-trunk       : 1.1.2
cocoapods-try         : 1.1.0
```

### Podfile

```ruby
platform :ios, '9.0'

inhibit_all_warnings!
use_frameworks!

source 'https://github.com/CocoaPods/Specs.git'

abstract_target 'Application' do
    pod 'Kvitto', :git => 'https://github.com/legoless/Kvitto.git'

    target 'FirstBlubs'

end
```

DanToml:
@Legoless There is a syntax error on line 4 of the podspec https://github.com/Cocoanetics/Kvitto/blob/develop/Kvitto.podspec#L4 (see double, double quotes for spec.summary =)

DanToml:
[...]

PR #6571, opened 2017-03-19
Fix regression from #6457 to ensure a correct error message is given …

dnkoutso:
…when a spec is not found.

/cc @benasher44

### DIFF
--- a/lib/cocoapods/external_sources/abstract_external_source.rb
+++ b/lib/cocoapods/external_sources/abstract_external_source.rb
@@ -116,9 +116,11 @@ module Pod
           rescue Pod::DSLError => e
             raise Informative, "Failed to load '#{name}' podspec: #{e.message}"
           rescue => _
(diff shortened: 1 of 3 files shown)


lib/cocoapods/external_sources/abstract_external_source.rb
```diff
@@ -116,10 +116,12 @@ def pre_download(sandbox)
           rescue Pod::DSLError => e
             raise Informative, "Failed to load '#{name}' podspec: #{e.message}"
           rescue => _
-            raise Informative, "Unable to find a specification for '#{name}'."
+            raise Informative, "Failed to download '#{name}'."
           end
           spec = download_result.spec
 
+          raise Informative, "Unable to find a specification for '#{name}'." unless spec
```
dnkoutso: There is a possibility that the pre-download "works" but the podspec is not found in the source. This should not have been removed!

PR #6576, opened 2017-03-21
Avoid unneeded master specs cloning for pod lib lint

### LINKED ISSUE #6154: Pod lib lint downloads master repo for specs with no dependencies
🌈

> Before you start, are you using the CocoaPods beta release?
> a lot has changed with Xcode 8 that is not backwards compatible.

no 

> Xcode 8 Status: Only works with CocoaPods 1.1.0
> Swift 3: You need to add a `.swift-version` file, see the CHANGELOG
> China: Use https://gems.ruby-china.org/ instead of https://gems.ruby-china.org/ for the RC see #5910

* [X ] I've read and understood the [*CONTRIBUTING guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

Run `pod lib lint` as a step on CircleCi

Build the pod and make sure the current version of the pod spec is valid

## What happened instead?

CocoaPods first runs `Setting up CocoaPods master repo`, which takes almost five minutes and isn't necessary. I can work around it by supplying a git repo to `pod lib lint --sources=<repo>`, but really I'd like to be able to specify that no external sources should be used.

### Stack

```
   CocoaPods : 1.1.1
        Ruby : ruby 2.0.0p648 (2015-12-16 revision 53162) [universal.x86_64-darwin16]
    RubyGems : 2.0.14.1
        Host : Mac OS X 10.12.1 (16B2555)
       Xcode : 8.1 (8B62)
         Git : git version 2.9.3 (Apple Git-75)
Ruby lib dir : /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib
Repositories : raizlabs - https://github.com/raizlabs/Swiftilities @ 74035d7415e564281a1c32a0a3232e498163265e
```

### Installation Source

```
Executable Path: /Users/mskiba/.gem/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.1
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
cocoapods-trunk       : 1.1.1
cocoapods-try         : 1.1.0
```

https://github.com/Raizlabs/Swiftilities

adellibovi:
This fixes #6154.

It avoids creating or updating the default master specs repo while executing `pod lib lint` if there aren't any dependency.

### DIFF
--- a/lib/cocoapods/installer.rb
+++ b/lib/cocoapods/installer.rb
@@ -71,6 +71,7 @@ module Pod
       @lockfile = lockfile
 
       @use_default_plugins = true
+      @has_dependencies = true
     end
 
     # @return [Hash, Boolean, nil] Pods that have been requested to be
@@ -79,6 +80,11 @@ module Pod
     #         not taken into account for deciding what Pods to install.
     #
     attr_accessor :update
+
+    # @return [Boolean] Whether it has dependencies. Defaults to true.
+    #
+    attr_accessor :has_dependencies
+    alias_method :has_dependencies?, :has_dependencies
 
     # @return [Boolean] Whether the spec repos should be updated.
     #
@@ -240,6 +246,7 @@ module Pod
     def create_analyzer
       Analyzer.new(sandbox, podfile, lockfile).tap do |analyzer|
         analyzer.installation_options = installation_options
+        analyzer.has_dependencies = has_dependencies?
       end
     end
 
--- a/lib/cocoapods/installer/analyzer.rb
+++ b/lib/cocoapods/installer/analyzer.rb
@@ -45,6 +45,7 @@ module Pod
 
         @update = false
         @allow_pre_downloads = true
+        @has_dependencies = true
       end
 
       # Performs the analysis.
@@ -151,6 +152,15 @@ module Pod
       #
       attr_accessor :allow_pre_downloads
       alias_method :allow_pre_downloads?, :allow_pre_downloads
+
+      # @return [Bool] Whether the analysis has dependencies and thus
+      #         sources must be configured.
+      #
+      # @note   This is used by the `pod lib lint` command to prevent
+      #         update of specs when not needed.
+      #
+      attr_accessor :has_dependencies
+      alias_method :has_dependencies?, :has_dependencies
 
       #-----------------------------------------------------------------------#
 
@@ -819,11 +829,11 @@ module Pod
 
           # Add any sources specified using the :source flag on individual dependencies.
           dependency_sources = podfile.dependencies.map(&:podspec_repo).compact
-
           all_dependencies_have_sources = dependency_sources.count == podfile.dependencies.count
+
           if all_dependencies_have_sources
             sources = dependency_sources
-          elsif sources.empty?
+          elsif has_dependencies? && sources.empty?
             sources = ['https://github.com/CocoaPods/Specs.git']
           else
             sources += dependency_sources
--- a/lib/cocoapods/validator.rb
+++ b/lib/cocoapods/validator.rb
@@ -28,8 +28,12 @@ module Pod
     #         the Source URLs to use in creating a {Podfile}.
     #
     def initialize(spec_or_path, source_urls)
-      @source_urls = source_urls.map { |url| config.sources_manager.source_with_name_or_url(url) }.map(&:url)
       @linter = Specification::Linter.new(spec_or_path)
+      @source_urls = if @linter.spec && @linter.spec.dependencies.empty?
+                       []
+                     else
+                       source_urls.map { |url| config.sources_manager.source_with_name_or_url(url) }.map(&:url)
+                     end
     end
 
     #-------------------------------------------------------------------------#
@@ -405,6 +409,7 @@ module Pod
       sandbox = Sandbox.new(config.sandbox_root)
       @installer = Installer.new(sandbox, podfile)
       @installer.use_default_plugins = false
+      @installer.has_dependencies = !spec.dependencies.empty?
       %i(prepare resolve_dependencies download_dependencies).each { |m| @installer.send(m) }
       @file_accessor = @installer.pod_targets.flat_map(&:file_accessors).find { |fa| fa.spec.name == consumer.spec.name }
     end
--- a/spec/unit/validator_spec.rb
+++ b/spec/unit/validator_spec.rb
@@ -372,13 +372,44 @@ module Pod
         end
       end
 
-      it 'repects the source_urls parameter' do
+      it 'empties sources when no dependencies' do
         sources = %w(master https://github.com/CocoaPods/Specs.git)
         Command::Repo::Add.any_instance.stubs(:run)
         validator = Validator.new(podspec_path, sources)
         validator.stubs(:validate_url)
         podfile = validator.send(:podfile_from_spec, :ios, '5.0')
+        podfile.sources.should == %w()
+      end
+
+      it 'repects the source_urls parameter when there are dependencies' do
+        podspec = stub_podspec(/.*name.*/, '"name": "SBJson",').gsub(/.*version.*/, '"version": "3.2",')
+        file = write_podspec(podspec, 'SBJson.podspec.json')
+        spec = Specification.from_file(file)
+        set = mock
+        set.stubs(:all_specifications).returns([spec])
+        Source::Aggregate.any_instance.stubs(:search).with(Dependency.new('SBJson', '~> 3.2')).returns(set)
+
+        podspec = stub_podspec(/.*name.*/, '"name": "ZKit",')
+        podspec.gsub!(/.*requires_arc.*/, '"dependencies": { "SBJson": [ "~> 3.2" ] }, "requires_arc": false')
+        file = write_podspec(podspec, 'ZKit.podspec.json')
+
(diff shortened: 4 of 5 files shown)


dnkoutso:
@adellibovi please add tests for this :)

CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

```markdown
* No specs update when not needed doing pod lib lint 
  [adellibovi](https://github.com/adellibovi)
```

</p>

adellibovi:
@dnkoutso yep :) I was preparing the PR, changelog and tests are missing, will do soon

lib/cocoapods/installer/analyzer.rb
```diff
@@ -803,10 +803,15 @@ def sources
 
           # Add any sources specified using the :source flag on individual dependencies.
           dependency_sources = podfile.dependencies.map(&:podspec_repo).compact
-
           all_dependencies_have_sources = dependency_sources.count == podfile.dependencies.count
+
+          external_sources = podfile.dependencies.map(&:external_source).compact
+          all_dependencies_have_external_sources = external_sources.count == podfile.dependencies.count
```
adellibovi: This is actually not correct, I would like to get also if the external source have dependencies, because in that case we should still use the default master repo. Ideas on how to do it @dnkoutso?

spec/unit/validator_spec.rb
```diff
@@ -372,12 +372,28 @@ def podspec_path(name = 'JSONKit', version = '1.4')
...
+      it 'check if source_urls is empty when podspec has no depedencies' do
         sources = %w(master https://github.com/CocoaPods/Specs.git)
         Command::Repo::Add.any_instance.stubs(:run)
         validator = Validator.new(podspec_path, sources)
         validator.stubs(:validate_url)
         podfile = validator.send(:podfile_from_spec, :ios, '5.0')
+        podfile.sources.should == %w()
+      end
+
+      it 'repects the source_urls parameter when podspec has dependencies' do
```
dnkoutso: `respects`

dnkoutso:
I think this is good, I'd prefer someone else also gives the 👍 perhaps @benasher44 or @DanToml

lib/cocoapods/validator.rb
```diff
@@ -28,8 +28,12 @@ class Validator
     #         the Source URLs to use in creating a {Podfile}.
     #
     def initialize(spec_or_path, source_urls)
-      @source_urls = source_urls.map { |url| config.sources_manager.source_with_name_or_url(url) }.map(&:url)
       @linter = Specification::Linter.new(spec_or_path)
+      @source_urls = if !@linter.spec || @linter.spec.dependencies.empty?
```
benasher44: This doesn't seem like the right fix to me because we're emptying out the `source_urls` parameter with the goal of short circuiting a `pod repo update` in the future knowing that it can't do anything if there are no repos to update. What if we just set `repo_update = false` on `@installer` in `download_pod` if we know there are no dependencies (and this is `pod lib lint` and not `pod spec lint`? What do you think?

lib/cocoapods/installer.rb
```diff
@@ -71,6 +71,7 @@ def initialize(sandbox, podfile, lockfile = nil)
       @lockfile = lockfile
 
       @use_default_plugins = true
+      @has_dependencies = true
```
benasher44: While we're at it, can we also set `repo_update` to `true` here, since that's the default? i'm not sure how I missed that in the other PR. Thanks!

lib/cocoapods/validator.rb
adellibovi: Thanks @benasher44, as always you are being very helpful :)
I understand what you are saying and I agree with the fact that setting the source_urls as an empty array is kind of a hack. The problem is that just setting what you are suggesting won't work, because calling `find_or_create_source_with_url` automatically set up the spec repo and clone it. Do you have any other idea, or am I missing something?

lib/cocoapods/installer.rb
adellibovi: Agree, done 👍

lib/cocoapods/validator.rb
benasher44: Hm. I see. Would it make sense to check `repo_update` in the `sources` method in `analyzer.rb`? If repo_update is false, just return `podfile.sources`.

lib/cocoapods/installer.rb
adellibovi: I'll have a better look tomorrow, seems that breaks other things

benasher44: 🙌

CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

```markdown
* [WIP] No specs update when not needed doing pod lib lint 
  [adellibovi](https://github.com/adellibovi)
```

</p>

adellibovi:
@benasher44 I reverted most of the changes and left the set up of repo update as true by default. I am having some issue with a test. I believe when `includes sources from source provider plugins` is called I receive something like:
`fatal: Cannot change to '/Repositories/CocoaPods/https:/github.com/artsy/Specs.git': No such file or directory`
The tests are passed, but I don't think this is correct...

lib/cocoapods/validator.rb
adellibovi: @benasher44 if I understood correctly, this is something already done with `has_dependencies?` and probably it's not enough, because map starts from `source_urls` (which defaults to master specs repo) and down the road will clone the master repo anyway, no matter what sources returns. Think about someone that wants to lint a Pod without dependencies and from a clean install (i.e. CI)

adellibovi: It's actually more about shortcutting `pod setup` btw

adellibovi: Thinking a bit more about the issue document in #6154, is that even setting an empty --source was going to clone the master repo and that because, like you were saying, the sources methods. So the shortcut is not really needed, a user could set an empty --source and that would work. I'll remove it.

benasher44:
Ah you're right. My mistake!

benasher44:
Looking good! I think now we just need a spec/test?

adellibovi:
@benasher44 here you go :)

spec/unit/validator_spec.rb
```diff
@@ -381,6 +381,16 @@ def podspec_path(name = 'JSONKit', version = '1.4')
         podfile.sources.should == %w(https://github.com/CocoaPods/Specs.git)
       end
 
+      it 'avoids creation of sources when source_urls is empty and no dependencies' do
```
benasher44: Hm. I thought the idea here was that even if there were sources, we didn't need to update any of the sources if there are no dependencies, since the only spec you need is the one being linted. Maybe I'm missing something? Thanks again for all of your work on this!

adellibovi: Yes, that was my original idea too, but the thing is that if the sources are not empty the initialise of the validator create/clone the repo within `source_with_name_or_url`. That was the reason of shortcutting it. Another solution would be to pass a new parameter to source_with_name_or_url, but if that's the case I'll need to go down to modify `pod repo add` to support a "light clone", ATM we cannot avoid creating the repo because is required by `source_with_url`.
This is my understanding, I may be wrong. Also I don't know if this would be better then shortcutting. Thoughts? :)

adellibovi: I.e. for "light clone" I mean doing a git init and adding the remote, instead of cloning.

benasher44: Hm okay. Maybe we should go back to the shortcutting solution if that's the most straightforward. I was under the impression this would be easier to bypass. What do you think?

adellibovi: Done, maybe I should have explained better at the beginning :) sorry about that

benasher44:
@adellibovi sorry for the delay! I went out of town, and then I got busy when I got back. It appears you have a lint error. If you could run `bundle exec rake rubocop`, fix the issue, and then update with master and resolve the conflict, I think this is ready to go once CI is green. Thanks for your patience!

adellibovi:
@benasher44 absolutely no need to be sorry :) will fix the PR during the next hours! Thanks

adellibovi:
@benasher44 updated the PR 👍

dnkoutso:
@adellibovi will wait for @benasher44 to approve and merge this. Thanks for the PR!

benasher44:
Hey sorry for the delay here! Happened to go on vacation shortly after my last update, and I'll have a final look at this when I'm back at the beginning of next week! Thanks again for your patience!

benasher44:
@adellibovi I'm back now! I think this is ready once CI is green!

benasher44:
@adellibovi thanks for this, and thanks again for your patience here!

adellibovi:
@benasher44 thanks to you and @dnkoutso for the great support :)

PR #6666, opened 2017-04-19
pass actual spec name to PodSourcePreparer

jpsim:
and set `COCOAPODS_SPEC` environment variable when running `prepare_command`.

Because prepare commands can't be set on subspecs, this is the least-bad way I could come up with to allow having custom behavior in prepare commands depending on which subspecs are being installed.

My main motivation for this is to add subspecs to Realm Objective-C and Realm Swift to allow building entirely from source, including the core database engine.

I have a PR towards Realm that does this (realm/realm-cocoa#4851) but it requires specifying an environment variable when installing: e.g. `REALM_DISABLE_SYNC=1 pod install`

I did this to test the feasibility and "test the waters", so to speak, to see if the CocoaPods maintainers would be amenable to changes along these lines. If so, I can add tests, update docs and see if there's a way to simplify these changes.

I'm quite open to other ideas on how to accomplish my high-level goal of providing source-only builds for Realm Objective-C & Realm Swift. Either in CocoaPods directly, or in our podspecs.

Thanks for your consideration!

### DIFF
--- a/lib/cocoapods/installer/pod_source_installer.rb
+++ b/lib/cocoapods/installer/pod_source_installer.rb
@@ -58,7 +58,7 @@ module Pod
       #
       def install!
         download_source unless predownloaded? || local?
-        PodSourcePreparer.new(root_spec, root).prepare! if local?
+        PodSourcePreparer.new(root_spec, root, specs.first.name).prepare! if local?
       end
 
       # Cleans the installations if appropriate.
--- a/lib/cocoapods/installer/pod_source_preparer.rb
+++ b/lib/cocoapods/installer/pod_source_preparer.rb
@@ -12,15 +12,20 @@ module Pod
       #
       attr_reader :path
 
+      # @return [String] the name of the specification (root or not) of the Pod.
+      attr_reader :spec_name
+
       # Initialize a new instance
       #
       # @param [Specification] spec the root specification of the Pod.
       # @param [Pathname] path the folder where the source of the Pod is located.
+      # @param [String] spec_name the name of the specification of the Pod.
       #
-      def initialize(spec, path)
+      def initialize(spec, path, spec_name = nil)
         raise "Given spec isn't a root spec, but must be." unless spec.root?
         @spec = spec
         @path = path
+        @spec_name = spec_name || spec.name
       end
 
       #-----------------------------------------------------------------------#
@@ -61,11 +66,13 @@ module Pod
             begin
               ENV.delete('CDPATH')
               ENV['COCOAPODS_VERSION'] = Pod::VERSION
+              ENV['COCOAPODS_SPEC'] = spec_name
               prepare_command = spec.prepare_command.strip_heredoc.chomp
               full_command = "\nset -e\n" + prepare_command
               bash!('-c', full_command)
             ensure
               ENV.delete('COCOAPODS_VERSION')
+              ENV.delete('COCOAPODS_SPEC')
             end
           end
         end
--- a/lib/cocoapods/downloader.rb
+++ b/lib/cocoapods/downloader.rb
@@ -30,7 +30,8 @@ module Pod
       request,
       target,
       can_cache: true,
-      cache_path: Config.instance.cache_root + 'Pods'
+      cache_path: Config.instance.cache_root + 'Pods',
+      spec_name: nil
     )
       can_cache &&= !Config.instance.skip_download_cache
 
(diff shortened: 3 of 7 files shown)


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

```markdown
* pass actual spec name to PodSourcePreparer 
  [jpsim](https://github.com/jpsim)
```

</p>

segiddins:
Why not just set the pod name in the spec's prepare command?

jpsim:
Because you can't override a `prepare_command` in subspecs, and I need to distinguish between the `Realm` root pod and the `Realm/OSSCore` subspec.

segiddins:
Given that the prepare_command is run before caching, and we don't cach each combination of subspecs separately, I'm not entirely sure what you're trying to achieve?

jpsim:
> I'm not entirely sure what you're trying to achieve?

I'm trying to change the behavior of `Realm` & `RealmSwift`'s podspecs' prepare commands based on whether or not their respective `OSSCore` subspecs are being pulled in, as defined by this PR: https://github.com/realm/realm-cocoa/pull/4851

This patch accomplishes that, as if `pod 'RealmSwift/OSSCore'` is specified in the podfile, that podspec name (as opposed to the root `RealmSwift`) will be passed as the `COCOAPODS_SPEC` env variable, allowing Realm's `build.sh cocoapods-setup` command to detect that it should avoid downloading Realm's "core engine" binary dependency.

Alternatively, if you have a different suggestion as to how to accomplish my high-level goal of providing both closed-source sync-enabled builds of the `Realm` & `RealmSwift` pods, as well as 100% open-source variants, I'm all ears.

Another option we have that wouldn't require any changes to CocoaPods would be to have two new specs entirely: say `RealmOSS` & `RealmSwiftOSS`. This way we could specify different prepare commands as those would be root specs.

segiddins:
I don't think this patch really implements what you want though, because of pod caching

segiddins:
The real fix would probably to pull realm core into its own podspec

jpsim:
That is being done: https://github.com/realm/realm-core/blob/master/RealmCore.podspec

And yes, this patch does what I need it to, I'm just not sure the implementation direction is acceptable to you or other maintainers. Which I understand if that's the case, it's just that I can't think of a better way to do it.

segiddins:
Does it work when the spec is being pulled from a specs repo rather than a git repo (and thus hits the spec caching)?

jpsim:
Yes because the name is still pulled from the podfile. Not the repo.

jpsim:
Do other maintainers also have an opinion on this? If I complete this approach by adding tests for both cached & non-cached installations, complete docs updates & see if I can simplify the implementation, am I wasting my time or is it likely to be accepted?

orta:
I don't see a problem with it 👍  @benasher44 / @dnkoutso

dnkoutso:
Same for me, @jpsim want to add tests and complete this?

dnkoutso:
ping @jpsim if interested in landing this for a 1.3.0.beta.1

jpsim:
I'm still interested in completing this, but I can't commit to a timeframe at the moment. Don't hold back on my account.

dnkoutso:
@jpsim I might be able to pick this one up from you. What was the remaining tasks just tests?

jpsim:
Tests, and seeing if there's a way to make this change less intrusive. Currently it modifies more method signatures than I'd like, but it may not be practical to avoid that.

dnkoutso:
Cool thanks!

dnkoutso:
Adding:
```
def fetch(sandbox, spec_name = nil)
...
end
```
To `path_source.rb` should fix most of the tests.

lib/cocoapods/installer/pod_source_installer.rb
```diff
@@ -58,7 +58,7 @@ def name
       #
       def install!
         download_source unless predownloaded? || local?
-        PodSourcePreparer.new(root_spec, root).prepare! if local?
+        PodSourcePreparer.new(root_spec, root, specs.first.name).prepare! if local?
```
dnkoutso: @jpsim can you explain why is this `specs.first.name`?

lib/cocoapods/installer/pod_source_preparer.rb
```diff
@@ -12,15 +12,20 @@ class PodSourcePreparer
       #
       attr_reader :path
 
+      # @return [String] the name of the specification (root or not) of the Pod.
```
dnkoutso: nit add line with `#` below this one.

dnkoutso:
The more I review this change the more it still makes me comfortable making it. I think having to split a subspec out in order to get a `prepare_command` might be too much to ask. 

I think if we get it green we could land it.

lib/cocoapods/installer/pod_source_installer.rb
segiddins: I think it could just be `root_spec.name`?

dnkoutso:
@jpsim support for script phases was built into CocoaPods 1.4.0 (see https://github.com/CocoaPods/CocoaPods/pull/7092)

Does this solve your issue? If yes I'd like to close this PR. I will also close it if there is no further activity.

jpsim:
I might never get back to this, so let's close.

PR #6694, opened 2017-05-03
Delete extraneous files in rsync destination

jgavris:
We often do incremental builds, and have noticed that the pods resources are synced without cleaning the destination. Often this is fine, but in our use case, it's important that only the current set of files are present.

`--delete                delete extraneous files from dest dirs`

Looks scary maybe, but should be exactly the same as doing a clean build, which is what your CI is hopefully doing anyway.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Louis D'hauwe](https://github.com/louisdh)
   [#6629](https://github.com/CocoaPods/CocoaPods/pull/6629)
   
+* Delete extraneous files in rsync destination 
+  [jgavris](https://github.com/jgavris)
+  [#6694](https://github.com/CocoaPods/CocoaPods/pull/6694)
+  
 ## 1.2.1 (2017-04-11)
 
 ##### Enhancements
--- a/lib/cocoapods/generator/copy_resources_script.rb
+++ b/lib/cocoapods/generator/copy_resources_script.rb
@@ -150,8 +150,8 @@ EOM
     *.framework)
       echo "mkdir -p ${TARGET_BUILD_DIR}/${FRAMEWORKS_FOLDER_PATH}" || true
       mkdir -p "${TARGET_BUILD_DIR}/${FRAMEWORKS_FOLDER_PATH}"
-      echo "rsync -av $RESOURCE_PATH ${TARGET_BUILD_DIR}/${FRAMEWORKS_FOLDER_PATH}" || true
-      rsync -av "$RESOURCE_PATH" "${TARGET_BUILD_DIR}/${FRAMEWORKS_FOLDER_PATH}"
+      echo "rsync --delete -av $RESOURCE_PATH ${TARGET_BUILD_DIR}/${FRAMEWORKS_FOLDER_PATH}" || true
+      rsync --delete -av "$RESOURCE_PATH" "${TARGET_BUILD_DIR}/${FRAMEWORKS_FOLDER_PATH}"
       ;;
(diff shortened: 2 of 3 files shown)


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

```markdown
* Delete extraneous files in rsync destination 
  [jgavris](https://github.com/jgavris)
```

</p>

jgavris:
@dnkoutso @postmechanical any feedback on this?

postmechanical:
@jgavris I defer to @dnkoutso for this.

dnkoutso:
@jgavris from [this](https://superuser.com/questions/156664/what-are-the-differences-between-the-rsync-delete-options) I think this is a good change.

It means that if once you had a few resources added, then updating a pod that no longer has the same resources it will delete the diff. Is this what you are trying to accomplish?

jgavris:
@dnkoutso Yes, exactly. The files that are not present anymore in the source are removed from the destination.

jgavris:
I've always found it interesting that `rsync` doesn't do this by default, it doesn't do a very good job of 'sync' without it :)

dnkoutso:
This is 👍 then to me. @benasher44 @DanToml if you have any objections this can proceed imo.

jgavris:
I gave a shot at updating the integration specs, I'm not sure how you all handle that. FWIW, submodules have proved to be more trouble than they're worth. Would you ever consider just bringing those specs into this main repo?

dnkoutso:
It might be a bigger change and different discussion outside of this issue. But yes overall agreed that submodules can be a pain.

jgavris:
*especially* with forks / a public project

dnkoutso:
@jgavris you've tried this locally and it works fine? Any reason to believe this will have more significant performance implications when the script runs?

jgavris:
Yes, we've been dogfooding this for the past week. It will definitely be slower than not deleting, but I  don't think anyone will notice unless they have many thousands of files.

dnkoutso:
> they have many thousands of files.

We have a pretty large project internally just wanted to see if there is negligible impact or high impact on a small project.

Either way I am approving this.

dnkoutso:
@jgavris please bump the submodule, I merged the integration specs PR

jgavris:
@dnkoutso Bumped the submodule back to `master`, thanks for the review!

dnkoutso:
@jgavris from testing CocoaPods 1.3.0.beta.2 we are seeing some red builds due to Xcode parallelizing multiple targets using the same frameworks and executing multiple embed framework scripts at the same time.

The --delete flag causes a file to be deleted in between targets and could race and cause a failure.

We should provide a locking mechanism for this.

jgavris:
@dnkoutso could you describe the race a little bit more? `--delete` will only affect the destination. If there is a race it would be there without `--delete` and the script may be copying an inconsistent tree as is.

dnkoutso:
@jgavris check this out https://github.com/CocoaPods/CocoaPods/pull/6873

jgavris:
That looks good to me! I will test it out shortly.

dnkoutso:
@jgavris did you get a chance to test it out?

jgavris:
Yeah, I did a few builds, no issues. I've never seen that issue myself though, and we've been running with `--delete` since that PR was merged.

dnkoutso:
👍  we also added 150 builds with this change and all went green, some were red for unrelated reasons so I think I will proceed with this.

PR #6700, opened 2017-05-08
Use `git!` when executing `push` command in order to raise informativ…

dnkoutso:
…e and set exit code.

We had a false positive build in which our CI failed to `push` to the repo but it never threw an exception to and set the exit code.

This PR will change `push` to use `git!` (via `repo_git`) in order to do that.


dnkoutso:
@benasher44 @DanToml no rush, whenever you get a chance.

dnkoutso:
This should be easy to merge too.

PR #6726, opened 2017-05-18
Aggregate targets should not include pod targets only used by tests

dnkoutso:


lib/cocoapods/installer.rb
```diff
@@ -212,7 +212,9 @@ def generate_pods_project(generator = create_generator)
     #         generated as result of the analyzer.
     #
     def pod_targets
-      aggregate_targets.map(&:pod_targets).flatten.uniq
+      aggregate_target_pod_targets = aggregate_targets.map(&:pod_targets).flatten
```
dnkoutso: the `installer` always took for granted that the set of aggregate targets include the full set of pod targets. This is no longer true because we can have pod targets that are only used for tests and therefore should not be included in the aggregate targets. 

Added a test for this and also checked existing tests still work here.

lib/cocoapods/generator/xcconfig/pod_xcconfig.rb
```diff
@@ -72,8 +72,8 @@ def generate
           @xcconfig.merge! XCConfigHelper.settings_for_dependent_targets(target, target.recursive_dependent_targets)
           if @test_xcconfig
             @xcconfig.merge! XCConfigHelper.settings_for_dependent_targets(target, target.test_dependent_targets)
-            XCConfigHelper.generate_vendored_build_settings(nil, target.test_dependent_targets, @xcconfig)
-            XCConfigHelper.generate_other_ld_flags(nil, target.test_dependent_targets, @xcconfig)
+            XCConfigHelper.generate_vendored_build_settings(nil, [target, *target.test_dependent_targets], @xcconfig)
```
dnkoutso: we no longer manually inject the target itself as a `test_dependent_target` so this will need to be added manually when we wire up the xcodeproj.

tests still pass, no need for a new one.

lib/cocoapods/installer/analyzer.rb
```diff
@@ -360,7 +360,7 @@ def generate_targets
         check_pod_target_swift_versions(specs_by_target)
         pod_targets = generate_pod_targets(specs_by_target)
         aggregate_targets = specs_by_target.keys.map do |target_definition|
-          generate_target(target_definition, pod_targets)
+          generate_aggregate_target(target_definition, pod_targets)
```
dnkoutso: just rename of method

lib/cocoapods/installer/analyzer.rb
```diff
@@ -419,12 +419,30 @@ def generate_target(target_definition, pod_targets)
         end
 
         target.pod_targets = pod_targets.select do |pod_target|
-          pod_target.target_definitions.include?(target_definition)
+          pod_target.target_definitions.include?(target_definition) &&
```
dnkoutso: this is the important change. the `pod_targets` of an aggregate target used to be pretty much all pods included in the target definition.

This is no longer true, we can have pod targets used only by tests and therefore we have to filter them out.

This line:

```ruby
target_definition.non_inherited_dependencies.map(&:name).include?(pod_target.name)
```

Allows someone to explicitly specify a pod like OCMock which would have been considered a test only pod and it would have failed to be linked.

lib/cocoapods/installer/analyzer.rb
```diff
@@ -496,8 +514,6 @@ def generate_pod_targets(specs_by_target)
             test_dependencies = transitive_dependencies_for_specs(target.specs.select(&:test_specification?), target.platform, all_specs).group_by(&:root)
             target.dependent_targets = filter_dependencies(dependencies, pod_targets_by_name, target)
             target.test_dependent_targets = filter_dependencies(test_dependencies, pod_targets_by_name, target)
-            # Test dependendent targets include our own target as a test dependency.
```
dnkoutso: this is now manually handled instead because this causing confusion on whether a pod is a test only pod or not.

spec/unit/installer/xcode/pods_project_generator_spec.rb
```diff
@@ -264,8 +264,11 @@ def test_extension_target(symbol_type)
               mock_native_target = mock('CoconutLib')
               mock_test_native_target = mock('CoconutLib-Unit-Tests')
 
-              dependent_target = mock('dependent-target', :should_build? => true, :native_target => 'DependentNativeTarget')
-              test_dependent_target = mock('dependent-test-target', :should_build? => true, :native_target => 'TestDependentNativeTarget')
+              dependent_native_target = mock('DependentNativeTarget')
```
dnkoutso: no real changes here, just updated test to actually use a mock object as a representation of the native target instead of just a string.

dnkoutso:
@DanToml damn CircleCI build is fast and awesome.

dnkoutso:
@DanToml @benasher44 @orta this is 👍 to go.

dnkoutso:
rebased.

lib/cocoapods/generator/xcconfig/pod_xcconfig.rb
DanToml: :+1:

lib/cocoapods/installer/analyzer.rb
DanToml: :+1:

lib/cocoapods/installer/analyzer.rb
```diff
@@ -406,7 +406,7 @@ def generate_targets
       #
       # @return [AggregateTarget]
       #
-      def generate_target(target_definition, pod_targets)
+      def generate_aggregate_target(target_definition, pod_targets)
```
dnkoutso: 👍 will revert.

lib/cocoapods/installer/analyzer.rb
```diff
@@ -428,12 +428,30 @@ def generate_target(target_definition, pod_targets)
         end
 
         target.pod_targets = pod_targets.select do |pod_target|
-          pod_target.target_definitions.include?(target_definition)
+          pod_target.target_definitions.include?(target_definition) &&
+              (target_definition.non_inherited_dependencies.map(&:name).include?(pod_target.name) || !pod_target_test_only?(pod_target, pod_targets))
```
dnkoutso: Will do.

dnkoutso:
Will rebase, address feedback and merge

dnkoutso:
Feedback addressed and rebased.

dnkoutso:
@DanToml shall I merge?

PR #6733, opened 2017-05-22
Only check for valid Swift version for pod targets that use Swift

dnkoutso:
The previous check does not make sense when consuming pods that do not have any Swift sources in them.

We have a local project I tried this on and it worked just fine.

Existing specs pass too and updated one.

### DIFF
--- a/lib/cocoapods/installer/analyzer.rb
+++ b/lib/cocoapods/installer/analyzer.rb
@@ -367,7 +367,6 @@ module Pod
       #
       def generate_targets
         specs_by_target = result.specs_by_target.reject { |td, _| td.abstract? }
-        check_pod_target_swift_versions(specs_by_target)
         pod_targets = generate_pod_targets(specs_by_target)
         aggregate_targets = specs_by_target.keys.map do |target_definition|
           generate_target(target_definition, pod_targets)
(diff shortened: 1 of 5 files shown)


lib/cocoapods/installer/analyzer.rb
```diff
@@ -367,8 +367,8 @@ def analyze_host_targets_in_podfile(aggregate_targets, embedded_aggregate_target
       #
       def generate_targets
         specs_by_target = result.specs_by_target.reject { |td, _| td.abstract? }
-        check_pod_target_swift_versions(specs_by_target)
         pod_targets = generate_pod_targets(specs_by_target)
+        check_pod_target_swift_versions(pod_targets)
```
dnkoutso: @DanToml I know you've had some reservations about this but I don't think it breaks anything in terms of running the check after pod targets have been created.

dnkoutso:
This was wrong. By the time I am querying `uses_swift?` none of the file accessors have been set.

Instead I moved the check inside `TargetValidator` which seems the right spot to check this correctly excluding pods that contain no Swift and preventing a `pod install`

spec/unit/installer/xcode/target_validator_spec.rb
```diff
@@ -143,6 +143,86 @@ def create_validator(sandbox, podfile, lockfile)
...
+            podfile.target_definitions['TestRunner'].stubs(:swift_version).returns(nil)
+
+            orangeframework_pod_target = stub(:name => 'OrangeFramework', :uses_swift? => true, :target_definitions => [podfile.target_definitions['SampleProject'], podfile.target_definitions['TestRunner']])
+            matryoshka_pod_target = stub(:name => 'matryoshka', :uses_swift? => true, :target_definitions => [podfile.target_definitions['SampleProject'], podfile.target_definitions['TestRunner']])
+
+            @validator = TargetValidator.new([], [orangeframework_pod_target, matryoshka_pod_target])
+            lambda { @validator.validate! }.should.not.raise
+          end
+
+          it 'does not raise when targets integrate the same pod but none of the pod targets use swift' do
```
dnkoutso: new test that catches specifically the case where 2 pod targets link to multiple target definitions with different swift versions but they both dont use Swift so it should not raise.

lib/cocoapods/installer/xcode/target_validator.rb
```diff
@@ -34,6 +34,7 @@ def initialize(aggregate_targets, pod_targets)
         def validate!
           verify_no_duplicate_framework_and_library_names
           verify_no_static_framework_transitive_dependencies
+          verify_no_pods_used_with_multiple_swift_versions
```
dnkoutso: This needs to move here since calling `uses_swift?` must ensure the `file_accessors` are set.

dnkoutso:
Green!

dnkoutso:
<3

PR #6744, opened 2017-05-23
Disable version check from bundler

dnkoutso:


segiddins:
Let's add a comment explaining why we're doing this, but 👍

dnkoutso:
Agreed, and done.

PR #6745, opened 2017-05-24
[GitHub] Add probot/stale configuration

DanToml:
Cc @orta and @dnkoutso any thoughts on wording?

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,29 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 90
+
+# Number of days of inactivity before a stale Issue or Pull Request is closed
+daysUntilClose: 7
+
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,29 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 90
+
+# Number of days of inactivity before a stale Issue or Pull Request is closed
+daysUntilClose: 7
```
dnkoutso: I think 7 is a bit low, should we do 14 or 21? We do not very actively look at issues and this will give us enough time to respond.

DanToml: This is 7 days after it has been marked as stale, so a total of 97 days, but we could up this to 14?

dnkoutso: Ah then it's good!

orta:
👍

orta:
Wording seems fine, it's easier to fix once we've seen a few and feel it needs to change later.

PR #6758, opened 2017-05-26
Use unique temp folder during lint for parallel execution

### LINKED ISSUE #5117: Two lint actions running simultaneously break each other
- [x] I've read and understood the [*CONTRIBUTING guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).
# Report

🌈 - Two lint actions running simultaneously break each other
## What did you do?

I have a CI server with two slots for processing jobs.
I have two distinct jobs to lint my pod spec.
One job does:
`pod lib lint SDK.podspec --allow-warnings --private`
On open merge requests to the podspec

The other job does:
`pod repo push podspecs 'SDK.podspec' --allow-warnings`
When a merge request is accepted to lint and push pod spec to my private pod specs repo`

I expect that both lint processes run in parallel.
## What happened instead?

Instead, when one process is running and the other starts at the same time, they both do their operations on the same temp folder. Causing one of the processes to fail cause it can't find given source files.
## Suggestion

Both lint actions were using the same folder in `/tmp` perhaps it could be as easy as not using the same folder for every lint action.

dnkoutso:
closes #5117

### DIFF
--- a/lib/cocoapods/validator.rb
+++ b/lib/cocoapods/validator.rb
@@ -242,7 +242,7 @@ module Pod
     # @return [Pathname] the temporary directory used by the linter.
     #
     def validation_dir
-      Pathname(Dir.tmpdir) + 'CocoaPods/Lint'
+      @validation_dir ||= Pathname(Dir.mktmpdir(['CocoaPods-Lint-', "-#{spec.name}"]))
     end
 
     # @return [String] the SWIFT_VERSION to use for validation.
--- a/spec/unit/validator_spec.rb
+++ b/spec/unit/validator_spec.rb
@@ -88,7 +88,7 @@ module Pod
         validator.quick = true
         validator.stubs(:validate_url)
         validator.validate
-        validator.validation_dir.should.be == Pathname.new(Dir.tmpdir) + 'CocoaPods/Lint'
+        validator.validated?.should.be.true
       end
 
       describe '#only_subspec' do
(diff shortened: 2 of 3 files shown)


spec/unit/validator_spec.rb
```diff
@@ -88,7 +88,7 @@ def podspec_path(name = 'JSONKit', version = '1.4')
         validator.quick = true
         validator.stubs(:validate_url)
         validator.validate
-        validator.validation_dir.should.be == Pathname.new(Dir.tmpdir) + 'CocoaPods/Lint'
```
dnkoutso: I do not 100% understand this test............but this validation won't work.

segiddins: I think it originated with making sure we used `realpath` but I agree, it's not great. I'm ok with removing it

lib/cocoapods/validator.rb
```diff
@@ -242,7 +242,7 @@ def result_color
     # @return [Pathname] the temporary directory used by the linter.
     #
     def validation_dir
-      Pathname(Dir.tmpdir) + 'CocoaPods/Lint'
+      @validation_dir ||= Pathname(Dir.mktmpdir) + 'CocoaPods/Lint'
```
segiddins: Maybe `Pathname(Dir.mktmpdir(["CocoaPods-Lint-", "-#{spec.name}"])` ?

dnkoutso: ha totally thought to add the `spec.name` too!

dnkoutso: updated.

dnkoutso:
le green.

PR #6762, opened 2017-05-30
Do not use the same product name for test targets

dnkoutso:


lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -214,7 +214,7 @@ def add_test_targets
               language = target.uses_swift? ? :swift : :objc
               native_test_target = project.new_target(product_type, name, platform, deployment_target, nil, language)
 
-              product_name = target.product_name
+              product_name = name
```
segiddins: Should this have `.xctest` at the end?

dnkoutso: I do not think so. I created a standard mac unit test target and it's not there. 

I think `xctest` is appended from `Wrapper Extension` value which is `xctest`

segiddins:
👍🏻

PR #6766, opened 2017-05-31
Ensure product for tests is not overridden by custom build settings

dnkoutso:


segiddins:
👍 with a test, if one is possible

dnkoutso:
There is already one covering it :) but I can amend to add something more.

dnkoutso:
Will merge on le green

PR #6787, opened 2017-06-07
Use a cache when figuring out if a pod target is test only

dnkoutso:
As of CP 1.3.0.beta.1 this method would add 7-10 seconds to out pod install time.

This adds a cache to the result but also prevents a number of calls to this method if they are not needed by short circuiting earlier.

There are no tests for this PR because the logic is the same. I hot patched CocoaPods for us and the `pod install` now remains the same.

### DIFF
--- a/lib/cocoapods/installer/analyzer.rb
+++ b/lib/cocoapods/installer/analyzer.rb
@@ -46,6 +46,8 @@ module Pod
         @update = false
         @allow_pre_downloads = true
         @has_dependencies = true
+        @test_pod_target_analyzer_cache = {}
+        @test_pod_target_key = Struct.new(:name, :pod_targets)
       end
 
       # Performs the analysis.
@@ -430,8 +432,9 @@ module Pod
         target.pod_targets = pod_targets.select do |pod_target|
           target_definition_included = pod_target.target_definitions.include?(target_definition)
           explicitly_defined_in_target_definition = target_definition.non_inherited_dependencies.map(&:name).include?(pod_target.name)
-          test_pod_target_only = pod_target_test_only?(pod_target, pod_targets)
-          target_definition_included && (explicitly_defined_in_target_definition || !test_pod_target_only)
+          next false unless target_definition_included
+          next true if explicitly_defined_in_target_definition
+          !pod_target_test_only?(pod_target, pod_targets)
         end
 
         target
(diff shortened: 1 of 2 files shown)


lib/cocoapods/installer/analyzer.rb
```diff
@@ -449,9 +451,14 @@ def generate_target(target_definition, pod_targets)
       # @return [Boolean] if the pod target is only referenced from test dependencies.
       #
       def pod_target_test_only?(pod_target, pod_targets)
-        source = pod_targets.any? { |pt| pt.dependent_targets.map(&:name).include?(pod_target.name) }
-        test = pod_targets.any? { |pt| pt.test_dependent_targets.map(&:name).include?(pod_target.name) }
-        !source && test
+        name = pod_target.name
+        key = [name, pod_targets]
```
dnkoutso: thoughts welcome on how to combine this perhaps and make it faster.

lib/cocoapods/installer/analyzer.rb
```diff
@@ -449,9 +451,14 @@ def generate_target(target_definition, pod_targets)
...
+        if @analyzer_cache.key?(key)
+          return @analyzer_cache[key]
+        end
+        source = pod_targets.any? { |pt| pt.dependent_targets.map(&:name).include?(name) }
```
dnkoutso: thoughts welcome on how to combine this perhaps and make it faster.

lib/cocoapods/installer/analyzer.rb
```diff
@@ -46,6 +46,7 @@ def initialize(sandbox, podfile, lockfile = nil)
         @update = false
         @allow_pre_downloads = true
         @has_dependencies = true
+        @analyzer_cache = {}
       end
```
segiddins: Please name this after what it is caching

lib/cocoapods/installer/analyzer.rb
segiddins: Depending on how often this is hit, it might be worth using a struct as the key instead of an array

dnkoutso: Interesting. Will attempt it.

lib/cocoapods/installer/analyzer.rb
dnkoutso: +1 fixing

dnkoutso:
Fixed the naming problem and rubocop. Looking intro struct way but I think it might be OK anyway.

lib/cocoapods/installer/analyzer.rb
dnkoutso: @segiddins done! Still works :)

dnkoutso:
✅

PR #6789, opened 2017-06-09
[WIP] Remove redundant `recursive_dependent_targets` method from `pod_target`

dnkoutso:
The analyzer already sets all `dependent_targets` recursively. This method was unnecessary to invoke to recalculate dependent targets.

Keeping it WIP since I need to double triple check it :)

From https://github.com/CocoaPods/CocoaPods/blob/master/lib/cocoapods/installer/analyzer.rb#L498-L503

```ruby
pod_targets.each do |target|
  dependencies = transitive_dependencies_for_specs(target.specs.reject(&:test_specification?),
  target.platform, all_specs).group_by(&:root)
  test_dependencies = transitive_dependencies_for_specs(target.specs.select(&:test_specification?),
  target.platform, all_specs).group_by(&:root)
  target.dependent_targets = filter_dependencies(dependencies, pod_targets_by_name, target)
  target.test_dependent_targets = filter_dependencies(test_dependencies, pod_targets_by_name, target)
end
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
+* Remove redundant `recursive_dependent_targets` method from `pod_target`
+  [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
+  [#6789](https://github.com/CocoaPods/CocoaPods/pull/6789)
+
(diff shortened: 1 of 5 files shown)


dnkoutso:
Hm maybe i have this wrong in my mind and need to spend more time with it to see if its true. Otherwise, I would probably need a `recusrive_test_dependent_targets` method

dnkoutso:
Will decline this for now to focus on one thing at a time. My primary concern is header search paths.

PR #6803, opened 2017-06-16
Simplify logic around framework input and output paths

dnkoutso:
This adds tests and cleans up a bit the method for returning all framework input and output paths. It was getting hard to follow but I think now its much better.


dnkoutso:
Will add resources script input outputs next

dnkoutso:
damn, green first run!

PR #6806, opened 2017-06-16
Add inputs and outputs for resources script phase

dnkoutso:


lib/cocoapods/installer/user_project_integrator/target_integrator.rb
```diff
@@ -164,7 +164,11 @@ def add_copy_resources_script_phase
             phase = create_or_update_build_phase(native_target, phase_name)
             script_path = target.copy_resources_script_relative_path
             phase.shell_script = %("#{script_path}"\n)
-            # TODO: Add input and output paths for the copy resources script phase.
+            resource_paths_by_config = target.resource_paths_by_config
+            unless resource_paths_by_config.empty?
+              phase.input_paths = [target.copy_resources_script_relative_path, *resource_paths_by_config.values.flatten.uniq]
+              phase.output_paths = ['${TARGET_BUILD_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}']
```
dnkoutso: @segiddins I'll be honest this is slightly lazy. It seems the copy resources script is rather complicated in where all these files are manually compiled but I think the catch all of the folder should do the trick for most of the cases.

Let me know if we should actually go deeper for this.

lib/cocoapods/target/aggregate_target.rb
```diff
@@ -222,17 +222,17 @@ def framework_paths_by_config
...
+    def resource_paths_by_config
       library_targets = pod_targets.reject do |pod_target|
         pod_target.should_build? && pod_target.requires_frameworks?
       end
       user_build_configurations.keys.each_with_object({}) do |config, resources_by_config|
         resources_by_config[config] = library_targets.flat_map do |library_target|
           next [] unless library_target.include_in_build_config?(target_definition, config)
           resource_paths = library_target.file_accessors.flat_map do |accessor|
-            accessor.resources.flat_map { |res| res.relative_path_from(sandbox.project.path.dirname) }
+            accessor.resources.flat_map { |res| "${PODS_ROOT}/#{res.relative_path_from(sandbox.project.path.dirname)}" }
```
dnkoutso: This was slightly a mistake. This was returning `Pathname` versus the one on line 238 was returning Strings.

Furthermore, we should be prepending `${PODS_ROOT}` here.

dnkoutso:
And the integration specs https://github.com/CocoaPods/cocoapods-integration-specs/pull/115/files

lib/cocoapods/installer/user_project_integrator/target_integrator.rb
segiddins: I don't really know how the input files feature works, so I can't give you much help here, sorry

dnkoutso: This one is for the output but the idea is that giving a folder its contents will be monitored by Xcode.

If any file changes within this folder then this will re-trigger the task.

The "lazy" part is that I am not explicitly defining the list of resources to monitor and instead I use the entire folder. This means that if other files change in there (outside of the ones in the scripts concerns) it will re-trigger the script.

Based on executing this locally it has been fine and the script is avoided. In our flagship a target takes 8 seconds to run this script and with this change its completely avoided!

dnkoutso: I chose the lazy way because different resources produce different types of artifacts when compiled. Also I do not believe resources change that often when someone is compiling their code and making changes to sources so it should be fine.

dnkoutso:
Added specs, fixed to properly handle edge case when frameworks or resources by config are empty.

lib/cocoapods/installer/user_project_integrator/target_integrator.rb
jgavris: @dnkoutso This looks fantastic, I'm trying it out. Unfortunately, it looks like this script is being executed on a no-op build (build twice in a row). What little I can find on this feature is here https://lists.apple.com/archives/xcode-users/2009/Oct/msg00316.html

>Adding the directory as an input works for me, but only if the files in that directory are 1) in the Xcode project, and 2) modified from within Xcode, rather than from the terminal or another application.

dnkoutso: @jgavris how do you tell its re-executing? Does the phase you are running it takes a long time? Note that if you go through the "All" tab in your build log you will see the script but it doesnt show up on "Recent".

Did embed pods frameworks script and check pods manifest lock run again for you also? 

Admittedly the "Copy resources script" is the one that has the highest chance of re-running because of the output folder is an entire folder instead of each individual file unlike "Embed Pods Frameworks" script.

Do you have a sample app for me to test this on?

dnkoutso: I verified by adding an artificial `sleep 10` on the script itself so I can visually see its running for a long time or its skipping it. Can you attempt the same?

jgavris: Wow, TIL. I put `date` at the top of the script and it appears to be working! Xcode prints out the output of the last time it ran the script...I feel like that's a bit misleading!

dnkoutso: Yes! That was weird for me too! The artificial `sleep 10` will be your definitive answer :)

keith:
@dnkoutso this is a great change, I noticed in our project, where we have some targets that have no resources in their CP based dependencies, that we still add this build phase, with now input or output files, in that case should we consider not adding the script phase at all? Currently these get re-run each build, even though they don't do anything.

dnkoutso:
@keith yes we can definitely skip adding this script phase entirely if we do not have anything to copy, both for embed frameworks and copy resources.

keith:
Here's a first attempt at that change https://github.com/CocoaPods/CocoaPods/pull/7181

PR #6810, opened 2017-06-17
Do not use target definition label for key caching

dnkoutso:
This method can be slow when the key is being generated as it compares the platforms of the `target_definition`.

screen_shot_2017-06-16_at_10_57_42_pm

If you do not like this change, we could perhaps cache the `label` method within `target_definition.rb`.

Also `include_in_build_config?` is invoked in multiple places, not just in the screenshot above.

### DIFF
--- a/lib/cocoapods/target/pod_target.rb
+++ b/lib/cocoapods/target/pod_target.rb
@@ -290,7 +290,7 @@ module Pod
     #         The name of the build configuration.
     #
     def include_in_build_config?(target_definition, configuration_name)
-      key = [target_definition.label, configuration_name]
+      key = [target_definition, configuration_name]
       if @build_config_cache.key?(key)
         return @build_config_cache[key]
       end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,9 +22,13 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
+* Do not use target definition label for key caching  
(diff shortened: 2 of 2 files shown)


lib/cocoapods/target/pod_target.rb
```diff
@@ -290,7 +290,7 @@ def recursive_dependent_targets
     #         The name of the build configuration.
     #
     def include_in_build_config?(target_definition, configuration_name)
-      key = [target_definition.label, configuration_name]
+      key = [target_definition, configuration_name]
```
segiddins: are you sure that target definitions properly implement `hash` and `eql?`

dnkoutso: Let me double check.

dnkoutso: Caching still works. `target_definition.rb` does not implement `hash` and `eql` but it works due to the fact that the target definition instances are the same.

dnkoutso: @segiddins another way around this is we cache the `label` function in `target_definition.rb` and we keep this the same.

dnkoutso: ...which might actually speed up more places?!

dnkoutso:
@segiddins I will close this and open one in core that caches the label function. It seems it gets called ~25k times from the screenshot above so that should make it faster in many other places.

dnkoutso:
https://github.com/CocoaPods/Core/pull/387/files

Issue #6772, opened 2017-06-01
Add source static library framework CocoaPods


paulb777:
🌈
* [X] I've read and understood the [*CONTRIBUTING guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## Introduction

This is a proposal and specification to add support for static library framework CocoaPods. For a combination of performance and dependency constraints, the currently supported dynamic library frameworks do not support some use cases. 

## Background

Currently CocoaPods requires static library frameworks with module maps to be binary (via vendored_frameworks). This is an impediment to providing source CocoaPods, since dynamic frameworks are not always viable:
  * If a pod has dependencies that are static library frameworks, the pod itself must be static.
  * The slower cold start time of dynamic libraries is not acceptable for some apps. We've measured around a 600 millisecond impact of each dynamic library on an iPhone 6.

The specific issue is that we'd like to incrementally open source Firebase and enable the open source build process to be inter-operable with the closed source build process. Firebase has historically release as several inter-dependent closed-source static library pods.  CocoaPods does not currently support building open source library pods that depend on closed-source static library pods, without hacks like **def installer.verify_no_static_framework_transitive_dependencies;** Also CocoaPods does not support building static library frameworks from source which is necessary to meet our performance requirements if we want to provide source CocoaPod support to our Swift users.

Xcode has built in support for creating dynamic library frameworks, but leaves static library framework construction to external scripts. CocoaPods already generates constructs to manage static libraries. The following section describes how to extend and/or replace the static library management with static frameworks.

## Specification 

Currently the static library build for a library Foo consists of the following components that are referenced by the CocoaPods generated xcconfig files:
  * A static library in _"$PODS_CONFIGURATION_BUILD_DIR/MyFoo/libFoo.a"_ 
    * referenced it by adding _"$PODS_CONFIGURATION_BUILD_DIR/MyFoo"_ to **LIBRARY_SEARCH_PATHS** in the generated xcconfig files
    *  and _-l"Foo"_ to **OTHER_LDFLAGS** 
  * The header files are copied into _"${PODS_ROOT}/Headers/Public/MyLibrary"_
    * The header path is added to **HEADER_SEARCH_PATHS**

Instead, a static library framework should be constructed at _"$PODS_CONFIGURATION_BUILD_DIR/MyFoo/Foo.framework"_. 
[...]

PR #6811, opened 2017-06-17
Add support for source static library frameworks

paulb777:
Address #6772 

https://github.com/CocoaPods/Core/pull/386 adds the static_framework podspec attribute.

This PR uses that attribute to build a source pod into a static library framework. All dependents and dependencies should interact with the source pod just as if it were a vendored_framework static library framework CocoaPod.

### DIFF
--- a/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
@@ -25,7 +25,9 @@ module Pod
               create_xcconfig_file
               create_test_xcconfig_files if target.contains_test_specifications?
               if target.requires_frameworks?
-                create_info_plist_file
+                unless target.static_framework?
+                  create_info_plist_file
+                end
                 create_module_map
                 create_umbrella_header do |generator|
                   file_accessors = target.file_accessors
@@ -39,6 +41,9 @@ module Pod
                                        end
                 end
                 create_build_phase_to_symlink_header_folders
+                if target.static_framework?
+                  create_build_phase_to_move_static_framework_archive
+                end
               end
               create_prefix_header
               create_dummy_source
@@ -385,6 +390,22 @@ module Pod
           base="$CONFIGURATION_BUILD_DIR/$WRAPPER_NAME"
           ln -fs "$base/${PUBLIC_HEADERS_FOLDER_PATH\#$WRAPPER_NAME/}" "$base/${PUBLIC_HEADERS_FOLDER_PATH\#\$CONTENTS_FOLDER_PATH/}"
           ln -fs "$base/${PRIVATE_HEADERS_FOLDER_PATH\#\$WRAPPER_NAME/}" "$base/${PRIVATE_HEADERS_FOLDER_PATH\#\$CONTENTS_FOLDER_PATH/}"
+            eos
+          end
+
+          # Creates a build phase to put the static framework archive in the appropriate framework location
+          # Since Xcode does not provide template support for static library frameworks, we've built a static library
+          # of the form lib{LibraryName}.a. We need to move that to the framework location -
+          # {LibraryName}.framework/{LibraryName}.
+          #
+          # @return [void]
+          #
+          def create_build_phase_to_move_static_framework_archive
+            build_phase = native_target.new_shell_script_build_phase('Setup Static Framework Archive')
+            build_phase.shell_script = <<-eos.strip_heredoc
+          mkdir -p "${BUILT_PRODUCTS_DIR}/${PRODUCT_NAME}.framework/Modules"
+          cp "${BUILT_PRODUCTS_DIR}/lib${PRODUCT_NAME}.a" "${BUILT_PRODUCTS_DIR}/${PRODUCT_NAME}.framework/${PRODUCT_NAME}"
+          cp "${MODULEMAP_FILE}" "${BUILT_PRODUCTS_DIR}/${PRODUCT_NAME}.framework/Modules/module.modulemap"
             eos
           end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   dependency resolution.  
   [Samuel Giddins](https://github.com/segiddins)
 
+* Add support for source static library frameworks
+  [Paul Beusterien](https://github.com/paulb777)
+  [#6811](https://github.com/CocoaPods/CocoaPods/pull/6811)
+
 ##### Bug Fixes
 
 * None.
@@ -111,7 +115,6 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 ## 1.3.0.beta.2 (2017-06-22)
 
 ##### Enhancements
-
 * Add inputs and outputs for resources script phase  
   [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
   [#6806](https://github.com/CocoaPods/CocoaPods/pull/6806)
--- a/lib/cocoapods/installer/xcode/pods_project_generator.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator.rb
@@ -225,9 +225,10 @@ module Pod
               aggregate_target.native_target.add_dependency(pod_target.native_target)
               configure_app_extension_api_only_for_target(pod_target) if is_app_extension
 
-              add_dependent_targets_to_native_target(pod_target.dependent_targets, pod_target.native_target, is_app_extension, pod_target.requires_frameworks?, frameworks_group)
-
-              add_pod_target_test_dependencies(pod_target, frameworks_group)
+              unless pod_target.static_framework?
+                add_dependent_targets_to_native_target(pod_target.dependent_targets, pod_target.native_target, is_app_extension, pod_target.requires_frameworks?, frameworks_group)
+                add_pod_target_test_dependencies(pod_target, frameworks_group)
+              end
             end
           end
         end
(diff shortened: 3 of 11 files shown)


1amageek:
Cool!!

最高だな！おい

dpassage:
Just for clarity on this enhancement - is this support for `.podspec`s which say "please build me as a static framework", or is it support for someone in their `Podfile` to say "please build this pod as a static framework instead of dynamic"?

paulb777:
@dpassage This is support for a podspec to say "please build me as a static framework" - AND allow me to have dependencies that are static framework pods.

dpassage:
Thanks for the clarity @paulb777 . I think at least some people are hoping for the latter - is there another issue open for that use case?

paulb777:
@dpassage Not that I know of specifically - although there are many issues in github and stackoverflow from people struggling at varied levels of understanding of dynamic and static libraries.

 BTW, I considered the Podfile approach and rejected it because it has dramatically higher levels of complexity, disruption and risk.

The podspec approach only impacts podspec creators who choose to use it. From a Podfile perspective, it should work exactly as static framework vendored_frameworks always have.

dnkoutso:
Dont want to hijack this PR or diverge it to a bigger discussion but @paulb777 is right that overall there is a lot of confusion around packaging libraries and specifically static libraries, static frameworks and dynamic frameworks.

CocoaPods currently makes it a hard division between "all static libraries" or "all dynamic frameworks" for all the pods, something of which in a later version we could re-think and allow complete mix and match of static libraries/frameworks and dynamic frameworks driven by a combination of the Podfile settings and podspec requirements.

dpassage:
Thanks, I'll take a look for those issues, and not muddy this issue further. 

Certainly don't want to say you shouldn't be building the feature that you are building, just want to be clear about what it is and is not. :)

DanToml:
@paulb777 If we can get this rebased after the core pr is merged then it should be good to go

paulb777:
Rebased.  

Should something in the Gemfile also be changed to get the master Core instead of the 1.3 Core?

paulb777:
I updated Gemfile.lock to the right Core version, fixed lint issues, and rebased. 

Should be good to go ...

CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

samuelsainz:
@DanToml do you have an estimation of when are releasing this feature?

saumitrabhave:
Hi @paulb777 , Thanks for this life saving PR.

I just tried this change in my local environment, but I think the change doesn't seem to work when there are subspecs defined in podspecs. following sample might make the issue clearer.

This Works just fine with `pod-dev lib lint`
```
# After the usual attributes
s.ios.deployment_target = '8.0'
s.static_framework = true
s.source_files = 'MyLib/Classes/**/*'
s.public_header_files = 'Pod/Classes/**/*.h'
s.dependency 'GoogleTagManager', '~> 5.0'
```

but this will still give the same transitivity error with `pod-dev lib lint`
```
s.ios.deployment_target = '8.0'
s.static_framework = true
s.source_files = 'MyLib/Classes/**/*'
s.public_header_files = 'Pod/Classes/**/*.h'
s.dependency 'GoogleTagManager', '~> 5.0'

s.subspec 'Xcode8' do |xc8|
end

s.default_subspec = 'Xcode8'
```

I also tried setting the static_framework flag within subspec but I found out that's not allowed as well.

is this an issue in my understanding or code? :) .

paulb777:
@samuelsainz See https://github.com/CocoaPods/CocoaPods/issues/6863#issuecomment-331605940

@saumitrabhave Thanks for the report. I reproduced the issue and will see if I can come up with a fix tomorrow.

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -388,6 +393,22 @@ def create_build_phase_to_symlink_header_folders
...
 
+          # Creates a build phase to put the static framework archive in the appropriate framework location
+          # Since Xcode does not provide template support for static library frameworks, we've built a static library
+          # of the form lib{LibraryName}.a. We need to move that to the framework location -
+          # {LibraryName}.framework/{LibraryName}.
+          #
+          # @return [void]
+          #
+          def create_build_phase_to_move_static_framework_archive
+            build_phase = native_target.new_shell_script_build_phase('Setup Static Framework Archive')
```
dnkoutso: @paulb777 somehow I came back to this PR from a Twitter link.

Should we:

- Prefix this script phase with `[CP]` that designates all CocoaPods script phases?
- Do we need to add input/output paths to ensure this is not running over and over again?

paulb777: @dnkoutso Yes and Probably.  I'll try to take a look today, but might not get to it until later next week.

paulb777: @dnkoutso Actually it looks like the pod build phases aren't prefixed, presumably because they're for the Pod target as opposed to modifications to the developer's existing targets. See an example here: https://github.com/CocoaPods/CocoaPods/blob/master/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb#L397

dnkoutso: ah I see. I guess those are fine and dont take that much time. It might be fine for now.

ronak2121:
Hi paulb777,

Thanks for working on this and showing us how Swift "static" libraries should be made. Apple has been quite lacking in their documentation.

I tried out the support you have above and found something weird:

My setup was:

1. Library A depends on Library B & are both written entirely in Swift
2. App C depends on A & B directly

I noticed in this setup that the Pods.framework was a Static Mach-O Type, but library A and B are actually dylibs. If you look at the generated App product, I've also noticed that you end up having to package the A & B dylibs also.

Why is this? Is this a limitation in how "static" frameworks can be linked together?

Ronak

paulb777:
@ronak2121 Actually there is nothing special about #6811 for Swift - or really for static libraries. It's about wrapping static libraries into frameworks. #6966 might be more what you're looking for. 

If you're seeing an issue with the static_framework attribute, please create an issue that includes a reproducible example.

ronak2121:
Thanks for the redirection. It looks like #6966 is what I was actually looking for. Thanks a lot!

mukeshthawani:
@paulb777 In our framework I want to add GoogleMaps as a dependency (which is a statically built), after adding `s.static_framework = true` I am not getting the transitive dependencies error. The issue is that now I am getting these Objective-C framework errors
screen shot 2017-10-09 at 4 31 32 pm

This is how I have added in the podspec file:
```
s.static_framework = true
s.dependency 'Kingfisher', '~> 3.6.2'
s.dependency 'GoogleMaps', '~> 2.4.0'
```
Do I need to change something?

paulb777:
@mukeshthawani There may be an issue with Swift integration, since testing has been primarily Objective-C. If you can share a reproducible example in another [Issue](https://github.com/CocoaPods/CocoaPods/issues), I'll investigate further.

mukeshthawani:
Thanks for replying. I will share a reproducible example.

RobertFiorentino:
@paulb777 I'd like to see if this change will help with a framework I'm putting together. App A would like to use Framework B which among other things has one .a static library. Currently it appears that bringing in Framework B via cocoapods does not include the static library. With this change will the static library now be included?

paulb777:
@RobertFiorentino This PR enables the scenario you describe, assuming that Framework B is wrapped in a CocoaPod.

ghugues:
Simple question: Since `static_framework` results in building a static library (packaged as a framework), why does it requires `use_frameworks!` to be specified in the podfile ? Couldn't it work as well for projects that use only static libraries (and do not use `use_frameworks!`) ?

morganchen12:
@ghugues IIRC that's the default flow if one does not specify `use_frameworks!`: CocoaPods will build a static library that's not packaged as a framework.

This flag exists to distinguish the linking behavior in projects using `use_frameworks!`.

ghugues:
Yes I know. But by packaging the static library as a framework in the default behaviour you can use it as a module. In particular you can use it as a Swift module without having to create a custom module map and without using a bridging header.

morganchen12:
You should open another issue, since changing the default behavior of CocoaPods for objc binaries will likely warrant a lot of discussion.

A less significant variant could be enabling framework packaging on a case-by-case basis.

paulb777:
@ghugues It could be possible to extend `static_framework` to work without a `use_frameworks!` specification, but it might be considered inconsistent to want a framework usage pattern when `use_frameworks!` is not in the Podfile.

The current implementation is intended to allow pod creators to ship source static frameworks as a transparently drop-in replacement for what had previously been binary static vendored frameworks.

StoneStoneStoneWang:
I used RxSwift and RxCocoa in my spec ,my i
1. update cocoaspod to 1.6.0.beta.2  :::gem install cocoapods -v 1.6.0.beta.2
2.in Podfile replace  user_frameworks  with use_modular_headers! 
You can use pod 'RxAtomic' ,:modular_headers => true  
3.clear DerivedData 
4.remove Podfile.lock and workspace and Pods 
5.pod install again or pod update  
6. in podspec file add 
 s.pod_target_xcconfig = { 'DEFINES_MODULE' => 'YES' }
 s.static_framework = true

7.pod lib lint WLReqKit.podspec --allow-warnings --verbose --no-clean without --use-libraries
8.pod trunk push WLReqKit.podspec --verbose --allow-warnings without --use-libraries
 
Then i resolved this issue

Punitbansal:
s.static_framework = true worked during pod install. But while building the project the library is not loaded for framework. I am getting below error 
Screen Shot 2019-05-03 at 11 28 06 AM

zhiqingchen:
> s.static_framework = true worked during pod install. But while building the project the library is not loaded for framework. I am getting below error

@Punitbansal Do you know why?

Issue #6815, opened 2017-06-18
generator: header.rb: umbrella always imports UIKit on iOS, even if the pod doesn't depend on it.


r-mckay:
* [x] I've read and understood the [*CONTRIBUTING guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

I wanted to install pods that do not depend on UIKit, RxSwift for example.

## What did you expect to happen?

I expected that these pods that do not depend on UIKit didn't import it.

## What happened instead?

The pods imported UIKit.

## Project that demonstrates the issue

Adding pod RxSwift, for example, demonstrates it. It is confirmed that it does not depend on UIKit, but if I'm not mistaken it imports it when installed via CocoaPods.
Please see: https://github.com/ReactiveX/RxSwift/issues/1298

keith:
This one hasn’t been resolved bot 😛


--
Keith Smiley

paulb777:
What's a good way to fix this without breaking all of the existing podspecs that are unknowingly depending upon the CocoaPods generated pch file?

dnkoutso:
@paulb777 not sure......need to think about this. Perhaps a new `#if COCOAPODS_1_5`  that gets set only on a newer version? Wonder if it's also considered a breaking change and must be `2.0`

keith:
In the meantime, I've done something horrible and created a CP plugin that just always uses "Foundation" in these headers. https://github.com/keith/cocoapods-foundation-headers

This could end up being a problem if someone developed a library only using CocoaPods and is somehow relying on this behavior. If that comes up I guess I'll take arguments and allow exclusions. But in the meantime this works on our project!

paulb777:
One possible solution would be adding to `s.pch = false` to the podspec to disable the pch file and option generation.

Would such Core and CocoaPod implementation PRs be accepted?

paulb777:
I discovered that I was conflating two related issues. Both the generated umbrella header and the generated pch header have the issue of bringing in an import of UIKit.h. More at https://stackoverflow.com/questions/44401464/is-there-a-way-to-disable-the-default-pch-file-in-cocoapods

keith:
FWIW this same codepath is responsible for both of those.

paulb777:
@keith I saw that, but the umbrella case is a bit more complex because the UIKit import is just a part of it instead of the whole thing.

[...]

PR #6816, opened 2017-06-18
generator: prevent header from importing UIKit on iOS & tvOS

r-mckay:
This commit replaces the UIKit import made in the header generator
by a Foundation import when targetting iOS and tvOS.

This is done because we don't know if the framework that is built
depends or not on UIKit. For example, frameworks that do not use UI
do not need to import UIKit.

Issue: #6815

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
+* Generator: prevent header from importing UIKit on iOS & tvOS 
+  [Romain Bertozzi](https://github.com/r-mckay)
+  [#6816](https://github.com/CocoaPods/CocoaPods/pull/6816)
+
 * Add inputs and outputs for resources script phase  
   [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
   [#6806](https://github.com/CocoaPods/CocoaPods/pull/6806)
--- a/lib/cocoapods/generator/header.rb
+++ b/lib/cocoapods/generator/header.rb
@@ -92,8 +92,8 @@ module Pod
       #
       def generate_platform_import_header
         case platform.name
-        when :ios then "#import <UIKit/UIKit.h>\n"
-        when :tvos then "#import <UIKit/UIKit.h>\n"
+        when :ios then "#import <Foundation/Foundation.h>\n"
+        when :tvos then "#import <Foundation/Foundation.h>\n"
         when :osx then "#import <Cocoa/Cocoa.h>\n"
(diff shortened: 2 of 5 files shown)


paulb777:
This looks like a step in the right direction -- although it will likely cause a bunch of build failures for libraries with implicit dependencies on UIKit, including the Travis test failures.

However, why is CocoaPods doing an implicit pch at all? it makes the library build fragile and less portable - https://stackoverflow.com/questions/44401464/is-there-a-way-to-disable-the-default-pch-file-in-cocoapods

r-mckay:
Integration specs bump CocoaPods/cocoapods-integration-specs#117

r-mckay:
Hello @paulb777 and thank you for your comment!

Yes, I do agree that it will create the build failures you mentioned, and I'm still uncomfortable with the idea of breaking stuff.
Furthermore and just like you said, the PCH files will be impacted. Examples of this are available in the integration tests from the PR linked above. It's quite a side effect of this PR (remove the UIKit import, and replace it with Foundation for the FOUNDATION_EXPORT, but just for the umbrella files).

Of course, if there's a better way to avoid this UIKit forced import I'd be glad to talk about it and implement it.

segiddins:
This is a backwards incompatible change that is liable to break existing specifications

r-mckay:
Hello @segiddins.
Do you see another implementation that could satisfy the linked issue and that I could write? If this one is not good enough I will then be glad to implement it differently thanks to the community advices.
Or is it the opened issue that is problematic? If yes then we could talk about it and see what can be done and what can't be.
Thank you very much.

segiddins:
I haven't looked at the issue, all I know is that we purposefully didn't make this change for 1.0 because of the potential for massive breakage of existing specs.

paulb777:
How about a podspec option to disable the creation and inclusion of the pch file?

Then the change could only be made by those that know it doesn't break their specs and existing specs would be fine.

r-mckay:
I agree with @paulb777 on the idea. It could be useful to have an option that allows the user to safely not use the default umbrella / pch couple that is provided with CocoaPods. He/she could use data he or she provides instead.

@segiddins, when you and the community studied this case for the 1.0, what options/alternatives did you explore? Maybe we could try to see if some of them could fit now.

What do you think?

Thanks!

segiddins:
> @segiddins, when you and the community studied this case for the 1.0, what options/alternatives did you explore? Maybe we could try to see if some of them could fit now.

This was almost two years ago, I don't remember, you'd have to search through the issues / PRs.

DanToml:
I'm going to close the PR in its current form as it would be a major breaking change for older pods, in favour of continuing discussion about the issue itself in #6815.

PR #6828, opened 2017-06-24
Integrate test targets to embed frameworks and resources

dnkoutso:
- [x] Specs
- [x] Extract pod target integrator to its own class
- [x] Cleanup

PR #6829, opened 2017-06-25
Ensure resource bundle and test dependencies are set for test native targets

dnkoutso:
predates https://github.com/CocoaPods/CocoaPods/pull/6828

If a pre-built binary pod provides test specs then its dependencies were not being added.

The same goes for resource bundles which are now handled as well.

This is required to complete shipping tests specs for 1.3.0.

### DIFF
--- a/lib/cocoapods/installer/xcode/pods_project_generator.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator.rb
@@ -198,10 +198,8 @@ module Pod
               configure_app_extension_api_only_for_target(aggregate_target) if is_app_extension
 
               unless pod_target.should_build?
-                pod_target.resource_bundle_targets.each do |resource_bundle_target|
-                  aggregate_target.native_target.add_dependency(resource_bundle_target)
-                end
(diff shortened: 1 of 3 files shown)


lib/cocoapods/installer/xcode/pods_project_generator.rb
```diff
@@ -198,10 +198,8 @@ def set_target_dependencies
               configure_app_extension_api_only_for_target(aggregate_target) if is_app_extension
 
               unless pod_target.should_build?
-                pod_target.resource_bundle_targets.each do |resource_bundle_target|
-                  aggregate_target.native_target.add_dependency(resource_bundle_target)
-                end
-
+                add_resource_bundles_to_native_target(pod_target, aggregate_target.native_target)
+                add_pod_target_test_dependencies(pod_target, frameworks_group)
                 next
```
dnkoutso: This `next` was causing to ignore integrating the test native targets for any pod target. We now ensure we do the right thing even in pods that they should not be built.

PR #6800, opened 2017-06-15
Add inputs and outputs to check manifest lock and embed framework script phases

dnkoutso:
closes #6797, #6252

- [x] Tests
- [x] Documentation / TODOs

PR #6870, opened 2017-07-13
[embed_frameworks_script] Added condition to install framework only when necessary

aizatomar:
Previously:
- Cocoapods will try to install all libraries again whenever the workspace is being build

Now:
- This change will check if frameworks need to be install again by first checking if the framework already exist. 2ndly, checking if the framework built has a different version than the source codes. This way, build time can be cut by avoiding unnecessary build.

### DIFF
--- a/lib/cocoapods/generator/embed_frameworks_script.rb
+++ b/lib/cocoapods/generator/embed_frameworks_script.rb
@@ -45,6 +45,15 @@ module Pod
 
           install_framework()
           {
+            if [ -e "$1" ]; then
+              version=$(defaults read "${PODS_ROOT}/Target Support Files/$(basename -s .framework "$1")/Info.plist" CFBundleShortVersionString)
+              current_version=$(defaults read "$1/Info.plist" CFBundleShortVersionString)
+              
+              if [ ${version} == ${current_version} ] ; then
+                return 0
+              fi
+            fi
+            
             if [ -r "${BUILT_PRODUCTS_DIR}/$1" ]; then
               local source="${BUILT_PRODUCTS_DIR}/$1"
             elif [ -r "${BUILT_PRODUCTS_DIR}/$(basename "$1")" ]; then
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Eric Amorde](https://github.com/amorde)
   [#6814](https://github.com/CocoaPods/CocoaPods/pull/6814)
 
+* Added condition to install framework only when necessary
+	[Aizat Omar](https://github.com/aizatomar)
+  [#6870](https://github.com/CocoaPods/CocoaPods/pull/6870)
+
 ##### Bug Fixes
 
 * Ensure resource bundle and test dependencies are set for test native targets  


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

```markdown
* added condition to build library only when there are new changes  
  [aizatomar](https://github.com/aizatomar)
```

</p>

lib/cocoapods/generator/embed_frameworks_script.rb
```diff
@@ -45,6 +45,15 @@ def script
 
           install_framework()
           {
+            if [ -e "$1" ]; then
+              version=$(defaults read "${PODS_ROOT}/Target Support Files/$(basename -s .framework "$1")/Info.plist" CFBundleShortVersionString)
```
dnkoutso: how does this work for local development pods in which the version hasn't changed but the contents of the framework have changed?

amorde: I had the same concern as well. This would break that functionality as far as I can tell

aizatomar: Hello guys! Thank you for the feedbacks! Ok I did not take into account that scenario. So my suggestion will be; what if we compare the modified timestamp of the source folder with the framework folder created in the project's derived data folder?

dnkoutso: I think we might be entering a bigger discussion here, what does this PR exactly try to resolve? In CocoaPods 1.3.0 we added input and output paths to the xcode phase and this script will be skipped from running unless the contents have changed.

Are you trying to achieve a performance improvement? If so have you measured the impact of your change?

burnflare: To add context, links to the input/output paths change in 1.3.0: #6800 #6806.

aizatomar: Hmm I apologise 1st as I did not check the new changes that have been approved in the current beta version. My bad. I took a look at #6800 #6806 . I think I had the same idea as you @dnkoutso on skipping the frameworks from being rebuilt.

Just more context on what I was trying to achieved.
@dnkoutso Trying to avoid the frameworks from rebuilding when they have already been build in the derived data. So I first checked if the framework folder already existed. 2nd, if the framework folder already existed, I compare the version of the framework with the version of the source codes (from both info.plist) to see if I need to reinstall the framework again. Avoiding unnecessary framework reinstall saves me a lot of build time when building the main project workspace. "A lot" here is just referring to cutting the build time from about a minute to a few seconds.

dnkoutso: Have you tried CocoaPods 1.3.0.beta.2?

aizatomar: Trying it out now. :)

aizatomar: @dnkoutso Exactly what I was looking for! :) 👍 Great job and thanks! Sorry again for missing out on checking this beta version. Will be looking forward to version 1.3.0.

Also can I proceed with closing this PR?

dnkoutso: sure!

PR #6873, opened 2017-07-14
Protect rsync tmp files from being deleted if two targets sync at the same time

justinseanmartin:
🌈

We started recently seeing two Embed Frameworks scripts failing with something along the following lines:

> rsync: stat "...DERIVED_DATA.../InternalFramework.framework.dSYM/Contents/Resources/DWARF/.InternalFramework.1cYxXV" failed: No such file or directory (2)

When I look a little higher in one of the two script phase logs, I find this:

> deleting InternalFramework.framework.dSYM/Contents/Resources/DWARF/.InternalFramework.1cYxXV

It turns out that there is a remote chance of a collision when two targets both integrate the same library dependency and happen to overlap in timing of then the rsync's for those libraries are being performed. 

We just moved an extra dependency into our app this morning which apparently started hitting this condition surprisingly often (~25% of the time).

I'm leveraging a workaround for this issue that I found proposed here:
https://lists.samba.org/archive/rsync/2008-February/020158.html

I've run our builds ~150 times and it hasn't hit the issue once, and nothing else failed, so I feel pretty good about this addressing the race condition without causing side effects.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
+* Protect rsync tmp files from being deleted if two targets sync at the same time  
+  [Justin Martin](https://github.com/justinseanmartin)
+  [#6873](https://github.com/CocoaPods/CocoaPods/pull/6873)
+
 * Include test schemes within library schemes  
   [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
   [#6765](https://github.com/CocoaPods/CocoaPods/issues/6765)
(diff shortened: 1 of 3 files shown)


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

```markdown
* Protect rsync tmp files from being deleted if two targets sync at the same time  
  [justinseanmartin](https://github.com/justinseanmartin)
```

</p>

justinseanmartin:
Yea, the alternative would've been to use a `--temp-dir` that is outside of the destination, so that it wouldn't get picked up by the `--delete` of the second `rsync` task that is running at the same time. I think that would probably also work, but I wasn't sure where a good place to put these files such that they work consistently and aren't messing with anything else in the build system. 

The "--filter protect" approach seemed like a more straightforward solution IMO.

dnkoutso:
👍  we can go ahead with this, @jgavris also tried it and nothing appears to regress.

PR #6883, opened 2017-07-16
[WIP] Block network access when running CI

DanToml:


DanToml:
unshare doesn't wanna do the thing, will look into this again in another day.

PR #6893, opened 2017-07-21
Cache result of resource and framework paths

dnkoutso:


lib/cocoapods/target/aggregate_target.rb
```diff
@@ -188,26 +188,30 @@ def uses_swift?
...
-          pod_target.include_in_build_config?(target_definition, config)
+      @framework_paths_by_config ||= begin
+        framework_paths_by_config = {}
+        user_build_configurations.keys.each do |config|
+          relevant_pod_targets = pod_targets.select do |pod_target|
+            pod_target.include_in_build_config?(target_definition, config)
+          end
+          framework_paths_by_config[config] = relevant_pod_targets.flat_map(&:framework_paths)
         end
-        framework_paths_by_config[config] = relevant_pod_targets.flat_map(&:framework_paths)
```
dnkoutso: diff looks a bit weird here, there are no changes except wrapping the whole thing with:
```ruby
@framework_paths_by_config ||= begin
```

PR #6915, opened 2017-07-31
Properly install pod targets with test specs within subspecs

dnkoutso:


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

```markdown
* Properly install pod targets that are part of subspecs  
```

</p>

lib/cocoapods/installer/analyzer.rb
```diff
@@ -429,17 +429,28 @@ def generate_target(target_definition, pod_targets)
...
         end
 
-        target.pod_targets = pod_targets.select do |pod_target|
-          target_definition_included = pod_target.target_definitions.include?(target_definition)
-          explicitly_defined_in_target_definition = target_definition.non_inherited_dependencies.map(&:name).include?(pod_target.name)
-          next false unless target_definition_included
-          next true if explicitly_defined_in_target_definition
-          !pod_target_test_only?(pod_target, pod_targets)
-        end
+        target.pod_targets = filter_pod_targets_for_target_definition(pod_targets, target_definition)
```
dnkoutso: extracted into a method so I can test it easier.

lib/cocoapods/installer/analyzer.rb
```diff
@@ -457,8 +468,12 @@ def pod_target_test_only?(pod_target, pod_targets)
         if @test_pod_target_analyzer_cache.key?(key)
           return @test_pod_target_analyzer_cache[key]
         end
-        source = pod_targets.any? { |pt| pt.dependent_targets.map(&:name).include?(name) }
-        test = pod_targets.any? { |pt| pt.test_dependent_targets.map(&:name).include?(name) }
+        source = pod_targets.any? do |pt|
+          pt.dependent_targets.map(&:name).include?(name)
+        end
+        test = pod_targets.any? do |pt|
+          pt.test_dependent_targets.reject { |dpt| dpt.name == pt.name }.map(&:name).include?(name)
```
dnkoutso: We intentionally ignore test dependencies to ourselves. This was the bug. A test spec technically always depends on the parent pod target so when we added a dependency to an aggregate target that has 0 other dependencies it was mistakenly considered as test only.

Test case below to cover for this.

lib/cocoapods/installer/analyzer.rb
```diff
@@ -429,17 +429,28 @@ def generate_target(target_definition, pod_targets)
...
+      #        the array of pod targets to check against
+      #
+      # @param [TargetDefinition] target_definition
+      #        the target definition to use as the base for filtering
+      #
+      # @return [Array<PodTarget>] the filtered list of pod targets.
+      #
+      def filter_pod_targets_for_target_definition(pod_targets, target_definition)
+        pod_targets.select do |pod_target|
+          pod_target.target_definitions.include?(target_definition) && !pod_target_test_only?(pod_target, pod_targets)
```
dnkoutso: a pod target should be part of an aggregate target if:

1) It's included in its target definition (as it has always been before test specs)
2) It is not used only by tests.

For example, 'OCMock' IS part of the target definition BUT it is only referenced by only pod targets as their test dependency therefore it should NOT be part of the aggregate target.

dnkoutso:
@DanToml with this you can release 1.3.0

lib/cocoapods/installer/analyzer.rb
DanToml: nice :+1:

lib/cocoapods/installer/analyzer.rb
DanToml: good catch :+1:

PR #6948, opened 2017-08-11
Wire dependencies for pod targets not part of any aggregate target

dnkoutso:


lib/cocoapods/installer/xcode/pods_project_generator.rb
```diff
@@ -225,10 +227,22 @@ def set_target_dependencies
               aggregate_target.native_target.add_dependency(pod_target.native_target)
               configure_app_extension_api_only_for_target(pod_target) if is_app_extension
 
-              add_dependent_targets_to_native_target(pod_target.dependent_targets, pod_target.native_target, is_app_extension, pod_target.requires_frameworks?, frameworks_group)
-
+              add_dependent_targets_to_native_target(pod_target.dependent_targets,
```
dnkoutso: reformat here as the line was a bit too long.

lib/cocoapods/installer/xcode/pods_project_generator.rb
```diff
@@ -225,10 +227,22 @@ def set_target_dependencies
...
+                                                     pod_target.native_target, is_app_extension,
+                                                     pod_target.requires_frameworks?, frameworks_group)
+              add_pod_target_test_dependencies(pod_target, frameworks_group)
+            end
+          end
+          # Wire up remaining pod targets used only by tests and are not used by any aggregate target.
+          test_only_pod_targets.each do |pod_target|
```
dnkoutso: this is the fix. After we've processed all aggregate targets and their pod targets we loop through remaining pod targets that could be used only by tests.

PR #6966, opened 2017-08-18
Swift Static Library support

### LINKED ISSUE #6899: Please support static libraries for Swift
The [release notes for Xcode 9 beta 4](https://download.developer.apple.com/Developer_Tools/Xcode_9_beta_4/Xcode_9_beta_4_Release_Notes.pdf) note that

> The new build system supports static library targets which contain Swift code. Debugging applications which use Swift static libraries may require having a complete set of build artifacts (in their original location) available. (33297067)

Also a [tweet from Daniel Dunbar](https://twitter.com/daniel_dunbar/status/889533883338838016) of Apple's dev tools team notes that this also applies to the old build system.

To date CocoaPods has (by necessity) required `use_frameworks!` for Pods written in Swift. This is no longer necessary. Since [using a lot of frameworks can slow down app launch times](https://blog.automatic.com/how-we-cut-our-ios-apps-launch-time-in-half-with-this-one-cool-trick-7aca2011e2ea), it would be really nice to see a future version of CocoaPods build with static libraries for Swift Pods.

DanToml:
Closes #6899.

- [x] Examples that build static swift
- [x] New integration specs
- [x] Validator to ensure no swift pod targets depend upon built targets that don't define modules
- [x] CHANGELOG entry
- [x] Allow enabling/disabling module map generation for all pods in a target from the Podfile
- [x] Allow enabling/disabling module map generation for one pod in a target from the Podfile
- [x] Move the generated umbrella header to be in the public headers directory rather than private headers?

### DIFF
--- a/examples/AFNetworking Example/AFNetworking Mac Example.xcodeproj/project.pbxproj
+++ b/examples/AFNetworking Example/AFNetworking Mac Example.xcodeproj/project.pbxproj
@@ -152,8 +152,6 @@
 				F8129BF71591061B009BFE23 /* Sources */,
 				F8129BF81591061B009BFE23 /* Frameworks */,
 				F8129BF91591061B009BFE23 /* Resources */,
-				CC1632E381344B54BA63986B /* [CP] Copy Pods Resources */,
-				F0DC30B6798CF042746A5614 /* [CP] Embed Pods Frameworks */,
 			);
 			buildRules = (
 			);
@@ -218,36 +216,6 @@
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
 			shellScript = "diff \"${PODS_PODFILE_DIR_PATH}/Podfile.lock\" \"${PODS_ROOT}/Manifest.lock\" > /dev/null\nif [ $? != 0 ] ; then\n    # print error to STDERR\n    echo \"error: The sandbox is not in sync with the Podfile.lock. Run 'pod install' or update your CocoaPods installation.\" >&2\n    exit 1\nfi\n# This output is used by Xcode 'outputs' to avoid re-running this script phase.\necho \"SUCCESS\" > \"${SCRIPT_OUTPUT_FILE_0}\"\n";
-			showEnvVarsInLog = 0;
-		};
-		CC1632E381344B54BA63986B /* [CP] Copy Pods Resources */ = {
-			isa = PBXShellScriptBuildPhase;
-			buildActionMask = 2147483647;
-			files = (
-			);
-			inputPaths = (
-			);
-			name = "[CP] Copy Pods Resources";
-			outputPaths = (
-			);
-			runOnlyForDeploymentPostprocessing = 0;
-			shellPath = /bin/sh;
-			shellScript = "\"${SRCROOT}/Pods/Target Support Files/Pods-AFNetworking Example/Pods-AFNetworking Example-resources.sh\"\n";
-			showEnvVarsInLog = 0;
-		};
-		F0DC30B6798CF042746A5614 /* [CP] Embed Pods Frameworks */ = {
-			isa = PBXShellScriptBuildPhase;
-			buildActionMask = 2147483647;
-			files = (
-			);
-			inputPaths = (
-			);
-			name = "[CP] Embed Pods Frameworks";
-			outputPaths = (
-			);
-			runOnlyForDeploymentPostprocessing = 0;
-			shellPath = /bin/sh;
-			shellScript = "\"${SRCROOT}/Pods/Target Support Files/Pods-AFNetworking Example/Pods-AFNetworking Example-frameworks.sh\"\n";
 			showEnvVarsInLog = 0;
 		};
 /* End PBXShellScriptBuildPhase section */
--- a/lib/cocoapods/generator/module_map.rb
+++ b/lib/cocoapods/generator/module_map.rb
@@ -67,7 +67,7 @@ module Pod
       #
       def generate
         <<-MODULE_MAP.strip_heredoc
-#{module_specificier_prefix}module #{target.product_module_name} {
+#{module_specifier_prefix}module #{target.product_module_name} {
   #{headers.join("\n  ")}
 
   export *
@@ -81,7 +81,7 @@ module Pod
       # The prefix to `module` to prepend in the module map.
       # Ensures that only framework targets have `framework` prepended.
       #
-      def module_specificier_prefix
+      def module_specifier_prefix
         if target.requires_frameworks?
           'framework '
         else
--- a/lib/cocoapods/generator/module_map.rb
+++ b/lib/cocoapods/generator/module_map.rb
@@ -41,7 +41,9 @@ module Pod
       #
       def initialize(target)
         @target = target
-        @headers = [Header.new(target.umbrella_header_path.basename, true)]
+        @headers = [
+          Header.new(target.umbrella_header_path.basename, true),
+        ]
       end
 
       # Generates and saves the Info.plist to the given path.
@@ -64,12 +66,12 @@ module Pod
       #
       def generate
         <<-MODULE_MAP.strip_heredoc
-          #{module_specificier_prefix}module #{target.product_module_name} {
-            #{headers.join("\n  ")}
+#{module_specificier_prefix}module #{target.product_module_name} {
+  #{headers.join("\n  ")}
 
-            export *
-            module * { export * }
-          }
+  export *
+  module * { export * }
+}
         MODULE_MAP
       end
 
--- a/lib/cocoapods/generator/module_map.rb
+++ b/lib/cocoapods/generator/module_map.rb
@@ -10,12 +10,38 @@ module Pod
       #
       attr_reader :target
 
+      attr_reader :headers
+
+      Header = Struct.new(:path, :umbrella, :private, :textual, :size, :mtime) do
+        alias private? private
+        def to_s
+          [
+            (:private if private?),
+            (:textual if textual),
+            (:umbrella if umbrella),
+            'header',
+            %("#{path}"),
+            attrs,
+        ].compact.join(' ')
+        end
+
+        def attrs
+          attrs = {
+            'size' => size,
+            'mtime' => mtime,
+          }.reject {|k,v| v.nil? }
+          return nil if attrs.empty?
+          attrs.to_s
+        end
+      end
+
       # Initialize a new instance
       #
       # @param  [PodTarget] target @see target
       #
       def initialize(target)
         @target = target
+        @headers = [Header.new(target.umbrella_header_path.basename, true)]
       end
 
       # Generates and saves the Info.plist to the given path.
@@ -39,7 +65,7 @@ module Pod
       def generate
         <<-MODULE_MAP.strip_heredoc
           #{module_specificier_prefix}module #{target.product_module_name} {
-            umbrella header "#{target.umbrella_header_path.basename}"
+            #{headers.join("\n  ")}
 
             export *
             module * { export * }
--- a/lib/cocoapods/generator/module_map.rb
+++ b/lib/cocoapods/generator/module_map.rb
@@ -67,7 +67,7 @@ module Pod
       #
       def generate
         <<-MODULE_MAP.strip_heredoc
-#{module_specifier_prefix}module #{target.product_module_name} {
+#{module_specifier_prefix}module #{target.product_module_name}#{module_declaration_attributes} {
   #{headers.join("\n  ")}
 
   export *
@@ -87,6 +87,15 @@ module Pod
         else
           ''
         end
+      end
+
+      # The suffix attributes to `module`.
+      # Ensures that library module maps are treated as `system` modules,
+      # supressing warnings when imported, as is done for header imports given via `-isystem`.
+      #
+      def module_declaration_attributes
+        return '' if target.requires_frameworks?
+        ' [system]'
       end
     end
   end
--- a/lib/cocoapods/generator/module_map.rb
+++ b/lib/cocoapods/generator/module_map.rb
@@ -13,7 +13,7 @@ module Pod
       attr_reader :headers
 
       Header = Struct.new(:path, :umbrella, :private, :textual, :size, :mtime) do
-        alias private? private
+        alias_method :private?, :private
         def to_s
           [
             (:private if private?),
@@ -22,14 +22,14 @@ module Pod
             'header',
             %("#{path}"),
             attrs,
-        ].compact.join(' ')
+          ].compact.join(' ')
         end
 
         def attrs
           attrs = {
             'size' => size,
             'mtime' => mtime,
-          }.reject {|k,v| v.nil? }
+          }.reject { |_k, v| v.nil? }
           return nil if attrs.empty?
           attrs.to_s
         end
--- a/lib/cocoapods/generator/module_map.rb
+++ b/lib/cocoapods/generator/module_map.rb
@@ -38,13 +38,26 @@ module Pod
       #
       def generate
         <<-MODULE_MAP.strip_heredoc
-          framework module #{target.product_module_name} {
+          #{module_specificier_prefix}module #{target.product_module_name} {
             umbrella header "#{target.umbrella_header_path.basename}"
 
             export *
             module * { export * }
           }
         MODULE_MAP
+      end
+
+      private
+
+      # The prefix to `module` to prepend in the module map.
+      # Ensures that only framework targets have `framework` prepended.
+      #
+      def module_specificier_prefix
+        if target.requires_frameworks?
+          'framework '
+        else
+          ''
+        end
       end
     end
   end
--- a/lib/cocoapods/generator/module_map.rb
+++ b/lib/cocoapods/generator/module_map.rb
@@ -12,13 +12,14 @@ module Pod
 
       attr_reader :headers
 
-      Header = Struct.new(:path, :umbrella, :private, :textual, :size, :mtime) do
+      Header = Struct.new(:path, :umbrella, :private, :textual, :exclude, :size, :mtime) do
         alias_method :private?, :private
         def to_s
           [
             (:private if private?),
             (:textual if textual),
             (:umbrella if umbrella),
+            (:exclude if exclude),
             'header',
             %("#{path}"),
             attrs,
--- a/lib/cocoapods/installer/xcode/pods_project_generator/aggregate_target_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/aggregate_target_installer.rb
@@ -20,6 +20,9 @@ module Pod
                 create_info_plist_file(target.info_plist_path, native_target, target.version, target.platform)
                 create_module_map
                 create_umbrella_header
+              elsif target.uses_swift?
+                create_module_map
+                create_umbrella_header
               end
               # Because embedded targets live in their host target, CocoaPods
               # copies all of the embedded target's pod_targets to its host
--- a/lib/cocoapods/installer/xcode/pods_project_generator/aggregate_target_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/aggregate_target_installer.rb
@@ -63,6 +63,8 @@ module Pod
               'PODS_ROOT'                          => '$(SRCROOT)',
               'PRODUCT_BUNDLE_IDENTIFIER'          => 'org.cocoapods.${PRODUCT_NAME:rfc1034identifier}',
               'SKIP_INSTALL'                       => 'YES',
+
+              'ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES' => 'NO',
             }
             super.merge(settings)
           end
--- a/lib/cocoapods/installer/xcode/pods_project_generator/aggregate_target_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/aggregate_target_installer.rb
@@ -64,6 +64,10 @@ module Pod
               'PRODUCT_BUNDLE_IDENTIFIER'          => 'org.cocoapods.${PRODUCT_NAME:rfc1034identifier}',
               'SKIP_INSTALL'                       => 'YES',
 
+              # Needed to ensure that static libraries won't try to embed the swift stdlib,
+              # since there's no where to embed in for a static library.
+              # Not necessary for dynamic frameworks either, since the aggregate targets are never shipped
+              # on their own, and are always further embedded into an app target.
               'ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES' => 'NO',
             }
             super.merge(settings)
--- a/lib/cocoapods/installer/xcode/pods_project_generator/target_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/target_installer.rb
@@ -106,6 +106,7 @@ module Pod
           # @return [Void]
           #
           def update_changed_file(generator, path)
+            path.dirname.mkpath
             if path.exist?
               generator.save_as(support_files_temp_dir)
               unless FileUtils.identical?(support_files_temp_dir, path)
--- a/lib/cocoapods/installer/xcode/pods_project_generator/target_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/target_installer.rb
@@ -74,9 +74,7 @@ module Pod
           # @return [Hash{String => String}]
           #
           def custom_build_settings
-            settings = {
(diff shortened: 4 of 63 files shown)


keith:
This is fantastic! I'm testing it out now, the first big issue I've noticed is that Swift pods with Swift dependencies need their dependencies' build paths in their `SWIFT_INCLUDE_PATHS`.

keith:
Here's a sample project for this [TestSwiftNestedDependencies.zip](https://github.com/CocoaPods/CocoaPods/files/1235647/TestSwiftNestedDependencies.zip)

keith:
The next issue I've noticed is that if you have a Swift pod that depends on an Objective-C pod, it has trouble building the module:

image

Here's an example project [TestObjectiveCDependency.zip](https://github.com/CocoaPods/CocoaPods/files/1235673/TestObjectiveCDependency.zip)

In this project I've added the `SWIFT_INCLUDE_PATHS` mentioned above. I've solved a similar problem like this before by making the path in the module map absolute, but that doesn't seem to work in this case. Either way I assume there's a better solution for that 🤔

DanToml:
@keith Both of those projects build for me now 👍.

keith:
Nice! Looks like that works on my end too. I noticed that specs using `module_name` don't seem to set the `PRODUCT_NAME` correctly. It looks like you explicitly commented that out though? If you want a test pod for that `ModelMapper` defines the module name as `Mapper`

CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

```markdown
* [WIP] Swift Static Library support  
  [DanToml](https://github.com/DanToml)
```

</p>

DanToml:
@keith I wasn't sure if it was going to be needed, but looks like it will be 👍

keith:
Next thing I ran into is it seems that the Facebook SDK fails to build statically because of some non modular include warnings. I don't see this when using frameworks today. This obviously could 100% be on them, but figured it was worth mentioning. Here's a sample project: [TestFacebookSDK.zip](https://github.com/CocoaPods/CocoaPods/files/1236200/TestFacebookSDK.zip)

keith:
Hmm actually it looks like it's happening with other Objective-C pods we're using as well. I can give another project example if needed.

DanToml:
@keith Good to know - a smaller example than FBSDK would be handy, but I think that's because I'm not preserving header paths properly yet.

keith:
Here's a slightly smaller example. It appears to be an issue with imports referencing nested Objective-C dependencies, so this example also has 2 dependencies: [TestLightstep.zip](https://github.com/CocoaPods/CocoaPods/files/1236209/TestLightstep.zip)

keith:
Ah the problem appears to be the imports in the umbrella header generated by CP. In both of these sample projects, commenting all the imports in that file out makes them compile.

fabb:
Is there a `pre` version of cocoapods where static swift libs can be tried out?

chenxiao0228:
@DanToml I tried your branch and applied it to our large scale project. In addition to your changes, I also made following changes for our project to build:
- module_map.rb "framework module" -> "module" for static lib
- pod_xcconfig.rb add dependent_targets' configuration_build_sir to HEADER_SEARCH_PATHS for Xcode to be able to find modulemap and umbrella headers
- xcconfig_helper.rb add "-import-underlying-module" for swift/objc hybrid libs
- tweaks in pod_target_installer.rb

There is however one issue left: make a swift class visible to ObjC class which depends on the lib. In theory this requires @import module -> modulemap -> umbrella -> module-Swift.h inclusion chain. I can copy it from derived-object folder but I am not sure if it's the best approach.

I attached my [diff](https://github.com/CocoaPods/CocoaPods/files/1282908/diff.txt). It's just a quick hack and really messy. Let me know if you wanna collaborate.

Pearapps:
Is this branch still being updated? Is there anything we can do to help move this along?

dnkoutso:
@Pearapps yes! if you feel you can move this faster than the current pace feel free to pick it up and get it past the finish line.

As a reminder, most of the work in CocoaPods is done from people's free time. There is never a deadline or strict timeline to get things done.

Pearapps:
I opened up a PR against this that uses that diff that @chenxiao0228 put up in a previous comment https://github.com/CocoaPods/CocoaPods/pull/7067

DanToml:
Sorry for the silence on this, I've been super busy lately, and CocoaPods feature work has taken a back seat.

I'm taking another look at this today.

DanToml:
@chenxiao0228 Thanks for that diff, coercing module maps into the right format was one of the things I was wondering about.

lib/cocoapods/installer/xcode/pods_project_generator/target_installer.rb
```diff
@@ -88,6 +88,9 @@ def custom_build_settings
                 settings['PRIVATE_HEADERS_FOLDER_PATH'] = framework_name + '.framework' + '/PrivateHeaders'
               end
             else
+              # if target.uses_swift?
```
segiddins: commented code

dnkoutso:
Punted to 1.5.0.

dnkoutso:
Reviving this for 1.5.0 now that 1.4.0 shipped.

segiddins:
I've taken over this work -- just push a rebase & a bunch of fixes. Will commit an example project shortly, but swift static libraries are now working!

keith:
Ran this on our project and found 1 issue on `pod install`. I'm trying to track it down, but the general issue is that we have static library targets, that depend on [Mapper](https://github.com/lyft/mapper/) who's CocoaPod is ModelMapper and defines `s.module_name               = "Mapper"`

This results in this error:

```
[!] The 'Pods-APITests' target has libraries with conflicting names: libmodelmapper.a.
```

I'm trying to isolate this in a smaller project.

segiddins:
My guess is the issue is that you're compiling Mapper into both your app and test target, instead of using `inherit! :search_paths`? But I'm not sure

keith:
In our case we definitely have distinct targets in the Podfile that depend on ModelMapper:

```ruby
target :Foo do
  pod "ModelMapper"

  target :FooTests do
    inherit! :search_paths
  end 
end

target :Bar do
  pod "ModelMapper"

  target :BarTests do
    inherit! :search_paths
  end 
end
```

Just testing this setup in a sample project doesn't reproduce this error, so it must be a bit more than this.

lib/cocoapods/installer/xcode/target_validator.rb
```diff
@@ -35,7 +35,6 @@ def validate!
           verify_no_duplicate_framework_and_library_names
           verify_no_static_framework_transitive_dependencies
           verify_no_pods_used_with_multiple_swift_versions
-          verify_framework_usage
```
dnkoutso: adios!

examples/AFNetworking Example/AFNetworking Mac Example.xcodeproj/project.pbxproj
```diff
@@ -220,36 +218,6 @@
 			shellScript = "diff \"${PODS_PODFILE_DIR_PATH}/Podfile.lock\" \"${PODS_ROOT}/Manifest.lock\" > /dev/null\nif [ $? != 0 ] ; then\n    # print error to STDERR\n    echo \"error: The sandbox is not in sync with the Podfile.lock. Run 'pod install' or update your CocoaPods installation.\" >&2\n    exit 1\nfi\n# This output is used by Xcode 'outputs' to avoid re-running this script phase.\necho \"SUCCESS\" > \"${SCRIPT_OUTPUT_FILE_0}\"\n";
 			showEnvVarsInLog = 0;
 		};
-		CC1632E381344B54BA63986B /* [CP] Copy Pods Resources */ = {
```
dnkoutso: so much cleaner, if there is nothing to run just dont add the script at all.

segiddins:
This should now be ready for review!

keith:
Getting this crash on this branch now:

```
ArgumentError - wrong number of arguments (given 4, expected 3)
/path/to/repo/vendor/bundle/ruby/2.3.0/gems/cocoapods-core-1.4.0/lib/cocoapods-core/lockfile.rb:378:in `generate'
/path/to/repo/vendor/bundle/ruby/2.3.0/bundler/gems/CocoaPods-dd8ffe6a7821/lib/cocoapods/installer.rb:553:in `write_lockfiles'
/path/to/repo/vendor/bundle/ruby/2.3.0/bundler/gems/CocoaPods-dd8ffe6a7821/lib/cocoapods/installer.rb:189:in `block in generate_pods_project'
/path/to/repo/vendor/bundle/ruby/2.3.0/bundler/gems/CocoaPods-dd8ffe6a7821/lib/cocoapods/user_interface.rb:64:in `section'
/path/to/repo/vendor/bundle/ruby/2.3.0/bundler/gems/CocoaPods-dd8ffe6a7821/lib/cocoapods/installer.rb:183:in `generate_pods_project'
/path/to/repo/vendor/bundle/ruby/2.3.0/bundler/gems/CocoaPods-dd8ffe6a7821/lib/cocoapods/installer.rb:119:in `install!'
/path/to/repo/vendor/bundle/ruby/2.3.0/bundler/gems/CocoaPods-dd8ffe6a7821/lib/cocoapods/command/install.rb:41:in `run'
/path/to/repo/vendor/bundle/ruby/2.3.0/gems/claide-1.0.2/lib/claide/command.rb:334:in `run'
/path/to/repo/vendor/bundle/ruby/2.3.0/bundler/gems/CocoaPods-dd8ffe6a7821/lib/cocoapods/command.rb:52:in `run'
/path/to/repo/vendor/bundle/ruby/2.3.0/bundler/gems/CocoaPods-dd8ffe6a7821/bin/pod:55:in `<top (required)>'
/path/to/repo/vendor/bundle/ruby/2.3.0/bin/pod:22:in `load'
/path/to/repo/vendor/bundle/ruby/2.3.0/bin/pod:22:in `<top (required)>'
/Users/ksmiley/.gem/ruby/2.3.3/gems/bundler-1.16.0/lib/bundler/cli/exec.rb:75:in `load'
/Users/ksmiley/.gem/ruby/2.3.3/gems/bundler-1.16.0/lib/bundler/cli/exec.rb:75:in `kernel_load'
/Users/ksmiley/.gem/ruby/2.3.3/gems/bundler-1.16.0/lib/bundler/cli/exec.rb:28:in `run'
/Users/ksmiley/.gem/ruby/2.3.3/gems/bundler-1.16.0/lib/bundler/cli.rb:424:in `exec'
/Users/ksmiley/.gem/ruby/2.3.3/gems/bundler-1.16.0/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
/Users/ksmiley/.gem/ruby/2.3.3/gems/bundler-1.16.0/lib/bundler/vendor/thor/lib/thor/invocation.rb:126:in `invoke_command'
/Users/ksmiley/.gem/ruby/2.3.3/gems/bundler-1.16.0/lib/bundler/vendor/thor/lib/thor.rb:387:in `dispatch'
/Users/ksmiley/.gem/ruby/2.3.3/gems/bundler-1.16.0/lib/bundler/cli.rb:27:in `dispatch'
/Users/ksmiley/.gem/ruby/2.3.3/gems/bundler-1.16.0/lib/bundler/vendor/thor/lib/thor/base.rb:466:in `start'
/Users/ksmiley/.gem/ruby/2.3.3/gems/bundler-1.16.0/lib/bundler/cli.rb:18:in `start'
/Users/ksmiley/.gem/ruby/2.3.3/gems/bundler-1.16.0/exe/bundle:30:in `block in <top (required)>'
/Users/ksmiley/.gem/ruby/2.3.3/gems/bundler-1.16.0/lib/bundler/friendly_errors.rb:122:in `with_friendly_errors'
/Users/ksmiley/.gem/ruby/2.3.3/gems/bundler-1.16.0/exe/bundle:22:in `<top (required)>'
/Users/ksmiley/.gem/ruby/2.3.3/bin/bundle:22:in `load'
/Users/ksmiley/.gem/ruby/2.3.3/bin/bundle:22:in `<main>'
```

dnkoutso:
@keith you might need to use `cocoapods-core` master too since it depends on updated lockfile changes to include spec repos

keith:
For some reason I was thinking bundler might handle that for me, sorry, the install works after that!

keith:
We have a slightly different error here then we did last time I posted it:

```
[!] The 'Pods-APITests' target has libraries with conflicting names: libmodelmapper-ios.a.
```

But I assume this is the same cause

segiddins:
@keith can you share something that reproduces that error?

keith:
Here's a sample project with the error: [CocoaPodsStatic.zip](https://github.com/CocoaPods/CocoaPods/files/1690131/CocoaPodsStatic.zip)

Now that I've reduced this it makes a bit more sense. This happens when the app depends on 2 static libraries, that also share a pods dependency. It looks like I can work around this by not including the shared pods dependency in the Podfile for both library targets, although I guess ideally CP would still make them all depend on it, but only link it to one final target.

keith:
Here's another sample project with I think shows a different issue. I have a Swift static library that depends on a random Objective-C dependency, but at build time it fails because it can't find the module [CocoaPodsStaticObjCCompile.zip](https://github.com/CocoaPods/CocoaPods/files/1690184/CocoaPodsStaticObjCCompile.zip)

lib/cocoapods/generator/module_map.rb
```diff
@@ -38,14 +67,27 @@ def save_as(path)
       #
       def generate
         <<-MODULE_MAP.strip_heredoc
-          framework module #{target.product_module_name} {
-            umbrella header "#{target.umbrella_header_path.basename}"
+#{module_specificier_prefix}module #{target.product_module_name} {
```
paulb777: specifier spelling

segiddins:
> Now that I've reduced this it makes a bit more sense. This happens when the app depends on 2 static libraries, that also share a pods dependency. It looks like I can work around this by not including the shared pods dependency in the Podfile for both library targets, although I guess ideally CP would still make them all depend on it, but only link it to one final target.

This issue seems to be present on `master` with static libs, I don't think it's related to the changes in this PR?

keith:
> This issue seems to be present on master with static libs, I don't think it's related to the changes in this PR?

Ah yea, I should've clarified that I was just testing with our swift project so we weren't able to use static libs before to compare.

segiddins:
> Here's another sample project with I think shows a different issue. I have a Swift static library that depends on a random Objective-C dependency, but at build time it fails because it can't find the module CocoaPodsStaticObjCCompile.zip

We can't generate module maps by default for ObjC static libs (it could break them), so they need to opt-in by adding `'DEFINES_MODULE' => 'YES'` to their `pod_target_xcconfig` for now

segiddins:
> Ah yea, I should've clarified that I was just testing with our swift project so we weren't able to use static libs before to compare.

Care to open another issue / PR to track that? I don't think we want it to be a blocker for landing this PR, since it's an existing issue

lib/cocoapods/installer/xcode/target_validator.rb
```diff
@@ -108,6 +109,31 @@ def verify_no_pods_used_with_multiple_swift_versions
...
+              non_module_dependencies << dependent_target.name
+            end
+
+            next if non_module_dependencies.empty?
+
+            error_messages << "The swift pod `#{pod_target.name}` depends upon #{non_module_dependencies.to_sentence}, " \
+                              'which do not define modules. ' \
+                              'To opt into those targets generating module maps '\
+                              '(which is necessary to import them from swift when building as static libraries), ' \
+                              "you may add `'DEFINES_MODULE' => 'YES'` to those pods' `pod_target_xcconfig`."
```
Pearapps: I haven't tried this yet, however - a couple questions:

1. What if you don't control the pod in question
2. What about the module map file, in this case?

segiddins: > What if you don't control the pod in question

It needs to be updated to opt-in, for now. We can explore making this an option in the Podfile down the road, but as that's purely new funcitonality I dont consider it a blocker

> What about the module map file, in this case?
 
The module map file will only be generated if:

1) The pod is consumed as a framework
2) The pod contains swift
3) `'DEFINES_MODULE' => 'YES'`

This is because adding a modulemap for a pod that has not been updated has the potential to break imports of that pods headers

SSheldon:
I've been testing this out on our project and ran into some issues. I'm not sure if they're fixable from the CocoaPods side or if these are just bugs in the Swift compiler, but I figured I'd report them here in case y'all can identify a CocoaPods fix that I can't see 😊

This first repro case I reduced, [CocoaPodsReproImport.zip](https://github.com/CocoaPods/CocoaPods/files/1692074/CocoaPodsReproImport.zip), will compile with `use_frameworks!`, but when compiling the same project with static libs it fails with:
```
<unknown>:0: error: module 'ReproKit' in AST file '~/Library/Developer/Xcode/DerivedData/ModuleCache/S9FE7KWZ7VGL/ReproKit-6DJJ8U61AGCP.pcm' (imported by AST file '~/Library/Developer/Xcode/DerivedData/Repro-bvtdxgvmkdshgmfjyoameesjmbfg/Build/Intermediates.noindex/PrecompiledHeaders/Repro-Bridging-Header-swift_1SVLZE1Z6S6TD-clang_S9FE7KWZ7VGL.pch') is not defined in any loaded module map file; maybe you need to load '~/Library/Developer/Xcode/DerivedData/Repro-bvtdxgvmkdshgmfjyoameesjmbfg/Build/Products/Debug-iphonesimulator/ReproKit/ReproKit.modulemap'?
<unknown>:0: note: imported by '~/Library/Developer/Xcode/DerivedData/Repro-bvtdxgvmkdshgmfjyoameesjmbfg/Build/Intermediates.noindex/PrecompiledHeaders/Repro-Bridging-Header-swift_1SVLZE1Z6S6TD-clang_S9FE7KWZ7VGL.pch'
<unknown>:0: error: clang importer creation failed
```
It's somehow triggered by having a bridged header that `@import`s a Swift static lib module. Interestingly, if you add the `-disable-bridging-pch` flag, this error disappears, which might suggest it's a compiler bug instead.

The second case I reduced, [CocoaPodsReproSegfault.zip](https://github.com/CocoaPods/CocoaPods/files/1692076/CocoaPodsReproSegfault.zip), is even more of a long shot, probably just a Swift compiler bug. It compiles fine with `use_frameworks!`, but when compiling it with static libs the compiler segfaults with:
```
*** DESERIALIZATION FAILURE (please include this section in any bug report) ***
top-level value not found
Cross-reference to module 'ReproKit'
... Foo
```
This one is somehow caused by `@import`ing a Swift static lib in a bridged header and also using a typealias from that module defined in a class that inherits from an Objective-C class whose designated initializer references another type from the module... phew. Seems unlikely that an issue this specific could be caused by the project configuration, but figured I'd share here before filing a Swift bug.

segiddins:
@SSheldon thanks for checking this out! I'm going to go out on a limb and say that both of those are _probably_ compiler bugs. The first because my guess is the pre-compilation of the bridging header ignores the `-Xcc -fmodule-map-file=...` flag we're setting, and the second because its a segfault.

lib/cocoapods/installer/xcode/pods_project_generator/aggregate_target_installer.rb
```diff
@@ -60,6 +63,8 @@ def custom_build_settings
               'PODS_ROOT'                          => '$(SRCROOT)',
               'PRODUCT_BUNDLE_IDENTIFIER'          => 'org.cocoapods.${PRODUCT_NAME:rfc1034identifier}',
               'SKIP_INSTALL'                       => 'YES',
+
+              'ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES' => 'NO',
```
dnkoutso: nit: either remove the space above or add a comment explaining why this is needed. I assume static libs cannot "embed" anything, but also dynamic frameworks don't need to embed the standard libs since the user/app target will?

spec/unit/installer/xcode/target_validator_spec.rb
```diff
@@ -352,31 +352,62 @@ def create_validator(sandbox, podfile, lockfile, integrate_targets = false)
...
+                                              :should_build? => true, :defines_module? => true, :dependent_targets => [matryoshka_pod_target], :spec_swift_version => '4.0')
 
-        describe '#verify_framework_usage' do
-          it 'raises when Swift pods are used without explicit `use_frameworks!`' do
+            @validator = TargetValidator.new([], [orangeframework_pod_target, matryoshka_pod_target])
+            e = lambda { @validator.validate! }.should.raise Informative
+            e.message.should.include <<-EOS.strip_heredoc.strip
+              [!] The following swift pods cannot yet be integrated as static libraries:
+
+              The swift pod `OrangeFramework` depends upon matryoshka, which do not define modules. To opt into those targets generating module maps (which is necessary to import them from swift when building as static libraries), you may add `'DEFINES_MODULE' => 'YES'` to those pods' `pod_target_xcconfig`.
```
dnkoutso: can we wrap depends upon `matryoshka` with backticks like `OrangeFramework`?

dnkoutso:
Caught an issue with test specs but it was fixed. This looks good. I say merge it! Let's slowly proceed perhaps with a 1.5.0.beta.1 in the very near future for people to start using it.

🎉

DanToml:
@segiddins This has merge conflicts but lgtm

segiddins:
One last thing I'd appreciate some feedback on -- should `defines_module?` be set from the Podfile instead of from the Podspec, similar to how any Podfile can opt a pod into building as a framework?

paulb777:
Supporting from the Podfile is better to distribute control as much as possible, but it increases the quality requirements of the implementation since it becomes trickier for a pod provider to ensure the pod works in any combination in which their pod may be used. 

It could make sense to start with a podspec-only implementation and add Podfile support later - like we've discussed for `static_framework`.

segiddins:
> It could make sense to start with a podspec-only implementation and add Podfile support later - like we've discussed for static_framework.

I'd really like there to be 1 way to flip settings, otherwise it gets super confusing (what happens when settings in different places conflict, etc)

dnkoutso:
There are only 7 pods in the master repo that use `DEFINES_MODULE`:

```
Specs/0/0/0/HCBaseSystem/0.1.2/HCBaseSystem.podspec.json:    "DEFINES_MODULE": "YES",
Specs/0/2/8/JackpotRising/2.10/JackpotRising.podspec.json:    "DEFINES_MODULE": "YES"
Specs/6/5/9/hccoren/0.1.1/hccoren.podspec.json:    "DEFINES_MODULE": "YES"
Specs/8/6/5/HCPlayer/0.0.1/HCPlayer.podspec.json:    "DEFINES_MODULE": "YES",
Specs/9/f/0/HCMinizip/1.2.6/HCMinizip.podspec.json:    "DEFINES_MODULE": "YES"
Specs/b/b/a/HCAudioUnit/0.0.4/HCAudioUnit.podspec.json:    "DEFINES_MODULE": "YES",
Specs/e/5/8/HCSharePush/0.2.0/HCSharePush.podspec.json:    "DEFINES_MODULE": "YES"
```

My gut feeling is that pod authors have not done significant testing on their header search paths. 

The incremental way that doesn't feel like breaking the CocoaPods world is to allow app developers to control this via the Podfile. Remember this only affects static libraries instead of dynamic framework imports which I think already work as we expect.

You *could* support both and error out if `DEFINES_MODULE => 'YES'` but the app developer asked for no modular header paths for this pod.

paulb777:
This may be less true for static libraries, but here's what I've thought about for static frameworks - In most cases I know of, `static_framework` in a podspec implies static linking is required.  Therefore adding user control to Podfile's could be added and `static_framework` = true in a podspec could continue to mean override any Podfile options. The podspec option could be incrementally removed from pods that support both types of linking.

chenxiao0228:
+1 to @dnkoutso 
For ObjC libs, DEFINES_MODULE enables the library to be used by Swift. App developers need the control either from Podfile or post_install hook.

segiddins:
> I strongly think this should be in the Podfile - since we don't have access to every Pod's podspec.

If we do it in the podfile, does it need to be per-target? Per-podfile? Per-pod?

dnkoutso:
per pod is best and most granular

dnkoutso:
But could provide "global" flag like `inhibit_all_warnings`

segiddins:
Quick update: running into _major_ issues caused by generating module maps for static library pods. I don't know if they're fixable.

chenxiao0228:
@segiddins do you have more details for the issues?

segiddins:
Rebased, this is basically ready to go!

arithma:
Question: Will this resolve the issue with transitive dependency stuff?

PR #6983, opened 2017-08-25
Update travis ruby versions

dnkoutso:
See https://github.com/travis-ci/travis-ci/issues/8302

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,20 +8,15 @@ git:
 
 matrix:
   include:
-  - rvm: 2.4.0
+  - rvm: 2.4.1
     env: COCOAPODS_CI_TASKS=LINT
     git:
       submodules: false
 
 rvm:
-  # OS X 10.9.5-10.10.0 (2.0.0-p481)
-  - 2.0.0-p481
-  # OS X 10.9.3-10.9.4
-  - 2.0.0-p451
-  # OS X 10.9.0-10.9.2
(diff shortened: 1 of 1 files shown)


dnkoutso:
...or we decide to use another ruby version that works with Trusty from here http://rubies.travis-ci.org/

segiddins:
We test that ruby version specifically since it used to come pre-installed on OS X. Also note that precise is deprecated?

dnkoutso:
Looks like it. But I dont think it will go away...

> By the end of September 2017, we hope to have 100% of incoming configurations with dist: precise routed to the sudo-enabled infrastructure, and to retire the container-based infrastructure. We have no immediate plans to discontinue support for jobs that specify sudo: required and dist: precise, although we do not plan to update the images for these jobs any more.

.travis.yml
```diff
@@ -14,13 +14,13 @@ matrix:
...
-  - 2.0.0-p481
+  # OS X 10.9.5-10.10.0 (2.0.0-p647)
+  - 2.0.0-p647
   # OS X 10.9.3-10.9.4
-  - 2.0.0-p451
+  - 2.0.0-p647
   # OS X 10.9.0-10.9.2
   # TODO: currently unavailable: https://github.com/travis-ci/travis-ci/issues/2918
-  # - 2.0.0-p247
+  # - 2.0.0-p647
```
segiddins: no need to test it twice, and let's remove the outdated comments

PR #6988, opened 2017-08-30
Fix OTHER_LDFLAGS for projects including source static frameworks

paulb777:
- Include source static framework library dependencies in aggregate target builds
- Set -framework option for dependent targets in aggregate target builds when there are source static frameworks

I suspect that the -framework option change should be made more broadly and that -framework should be set in concert with updating the FRAMEWORK_SEARCH_PATH, but I'm making the conservative change of only impacting projects that include source static frameworks.

In addition to the unit tests, these changes have been tested with [Firebase](https://github.com/firebase/firebase-ios-sdk/tree/pb-future).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,13 +16,17 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   dependency resolution.  
   [Samuel Giddins](https://github.com/segiddins)
 
-* Add support for source static library frameworks
+* Add support for source static library frameworks  
   [Paul Beusterien](https://github.com/paulb777)
   [#6811](https://github.com/CocoaPods/CocoaPods/pull/6811)
 
-* Add Private Header support to static frameworks
+* Add Private Header support to static frameworks  
   [Paul Beusterien](https://github.com/paulb777)
   [#6969](https://github.com/CocoaPods/CocoaPods/pull/6969)
+
+* For source static frameworks, include frameworks from dependent targets and libraries in OTHER_LDFLAGS  
+  [paulb777](https://github.com/paulb777)
+  [#6988](https://github.com/CocoaPods/CocoaPods/pull/6988)
 
 ##### Bug Fixes
 
--- a/lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
+++ b/lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
@@ -170,11 +170,11 @@ module Pod
         #  - `@import …;` / `import …`
         #
         def generate_settings_to_import_pod_targets
-          @xcconfig.merge! XCConfigHelper.settings_for_dependent_targets(target, pod_targets)
+          @xcconfig.merge! XCConfigHelper.search_paths_for_dependent_targets(target, pod_targets)
           @xcconfig.merge!(settings_to_import_pod_targets)
           target.search_paths_aggregate_targets.each do |search_paths_target|
             generator = AggregateXCConfig.new(search_paths_target, configuration_name)
-            @xcconfig.merge! XCConfigHelper.settings_for_dependent_targets(nil, search_paths_target.pod_targets)
+            @xcconfig.merge! XCConfigHelper.search_paths_for_dependent_targets(nil, search_paths_target.pod_targets)
             @xcconfig.merge!(generator.settings_to_import_pod_targets)
           end
         end
--- a/lib/cocoapods/generator/xcconfig/xcconfig_helper.rb
+++ b/lib/cocoapods/generator/xcconfig/xcconfig_helper.rb
@@ -270,7 +270,7 @@ module Pod
         #
         # @return [Hash<String, String>] the settings
         #
-        def self.settings_for_dependent_targets(target, dependent_targets, test_xcconfig = false)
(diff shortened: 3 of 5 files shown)


lib/cocoapods/generator/xcconfig/xcconfig_helper.rb
```diff
@@ -288,15 +288,23 @@ def self.settings_for_dependent_targets(target, dependent_targets, test_xcconfig
...
+                if dependent_targets.any?(&:static_framework?)
+                  frameworks += %( -framework "#{dependent_target.product_basename}")
+                end
               else
                 library_search_paths << dependent_target.configuration_build_dir(CONFIGURATION_BUILD_DIR_VARIABLE)
               end
             end
             build_settings['FRAMEWORK_SEARCH_PATHS'] = XCConfigHelper.quote(framework_search_paths.uniq)
             build_settings['LIBRARY_SEARCH_PATHS']   = XCConfigHelper.quote(library_search_paths.uniq)
+            build_settings['OTHER_LDFLAGS']          = frameworks
```
dnkoutso: I think you may want to be invoking `generate_other_ld_flags` instead that actually sets these flags. You should be setting this within `aggregate_xcconfig.rb` instead.

This method here probably needs to be renamed to just specify adding `search_paths_for_dependent_targets` but it doesn't have to be in this PR.

dnkoutso: interestingly though the `generate_other_ld_flags` only adds the flags for pod targets that require building (have sources included) which in your case I dont think thats true so maybe look into:

```
        def self.generate_vendored_build_settings(aggregate_target, pod_targets, xcconfig)
          pod_targets.each do |pod_target|
            unless pod_target.should_build? && pod_target.requires_frameworks?
              XCConfigHelper.add_settings_for_file_accessors_of_target(aggregate_target, pod_target, xcconfig)
            end
          end
        end
```

paulb777:
@dnkoutso Thanks for the quick review. I went done that path in my investigation, but couldn't figure out how to make it generate for the Tests_ios.*.xcconfig file. I'll try again.

Why is the logic for setting the -framework option separate from updating the FRAMEWORK_SEARCH_PATH? It seems like they should be always in sync?

dnkoutso:
> Why is the logic for setting the -framework option separate from updating the FRAMEWORK_SEARCH_PATH? It seems like they should be always in sync?

Agreed. I think things were just like that before my time.

paulb777:
@dnkoutso How's this?

lib/cocoapods/generator/xcconfig/xcconfig_helper.rb
```diff
@@ -321,10 +321,20 @@ def self.settings_for_dependent_targets(target, dependent_targets, test_xcconfig
...
-            unless pod_target.should_build? && pod_target.requires_frameworks?
+            unless pod_target.should_build? && pod_target.requires_frameworks? && !pod_target.static_framework?
               XCConfigHelper.add_settings_for_file_accessors_of_target(aggregate_target, pod_target, xcconfig)
             end
           end
+          # Make sure -framework option gets added for the search paths when static_frameworks are involved.
+          # Otherwise test targets won't link in their primary target's dependencies.
+          unless aggregate_target.nil?
+            dependent_targets = aggregate_target.search_paths_aggregate_targets
+            dependent_targets.each do |dependent_target|
```
dnkoutso: can do `dependent_target.select(&:static_framework?)`?

lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
```diff
@@ -170,11 +170,11 @@ def settings_to_import_pod_targets
         #  - `@import …;` / `import …`
         #
         def generate_settings_to_import_pod_targets
-          @xcconfig.merge! XCConfigHelper.settings_for_dependent_targets(target, pod_targets)
+          @xcconfig.merge! XCConfigHelper.search_paths_for_dependent_targets(target, pod_targets)
```
dnkoutso: 👍

lib/cocoapods/generator/xcconfig/xcconfig_helper.rb
paulb777: No, because all the pod_targets need to be added whether or not they're a static_framework. 

I don't think the static_framework? check is needed at all, but I want to avoid changing projects that aren't using it to be conservative.

paulb777:
@dnkoutso Rebased and green. Anything else needed to merge?

dnkoutso:
@paulb777 I will take another look over the weekend and update you.

dnkoutso:
Needs a rebase. Otherwise LGTM. Also wouldn't mind a second reviewer.

paulb777:
Thanks @dnkoutso! - Rebased and green again. Anyone else able to review?

CHANGELOG.md
```diff
@@ -24,6 +24,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Paul Beusterien](https://github.com/paulb777)
   [#6969](https://github.com/CocoaPods/CocoaPods/pull/6969)
 
+* Fix OTHER_LDFLAGS for projects involving source static frameworks
```
segiddins: missing two trailing spaces and doesn't describe what was wrong before

PR #6994, opened 2017-08-31
Use the resolver to identify which pod targets are test only

dnkoutso:
`resolver.rb` is the best (and only) place to identify if a specification for a given target definition is only used by tests or not.

This PR effectively moves the logic into the resolver because trying to figure out which pod target is a test only pod target *after* the resolver has executed is not possible.

We now wrap the `specs_by_target` result with a `ResolvedSpecification` container that includes the necessary metadata (such as if the resolved spec was only used by tests).

Everything else just percolates through.

This effectively fixes all edge cases and problems around mistakenly identifying (and therefore linking) pod targets that are only used by tests.


lib/cocoapods/installer/analyzer.rb
```diff
@@ -429,59 +433,33 @@ def generate_target(target_definition, pod_targets)
           end
         end
 
-        target.pod_targets = filter_pod_targets_for_target_definition(pod_targets, target_definition)
+        target.pod_targets = filter_pod_targets_for_target_definition(target_definition, pod_targets)
```
dnkoutso: swapped the two parameters for symmetry since all other methods use `target_definition` first and `pod_targets` second.

lib/cocoapods/installer/analyzer.rb
```diff
@@ -429,59 +433,33 @@ def generate_target(target_definition, pod_targets)
...
-        key = @test_pod_target_key.new(name, pod_targets)
-        if @test_pod_target_analyzer_cache.key?(key)
-          return @test_pod_target_analyzer_cache[key]
-        end
-        source = pod_targets.any? do |pt|
-          pt.dependent_targets.map(&:name).include?(name)
-        end
-        test = pod_targets.any? do |pt|
-          pt.test_dependent_targets.reject { |dpt| dpt.name == pt.name }.map(&:name).include?(name)
+      def filter_pod_targets_for_target_definition(target_definition, pod_targets)
```
dnkoutso: This method now becomes very simple. A pod target is part of an aggregate only if:

1) It's part of its target definition
2) The resolver did not mark this as a test only pod target.

lib/cocoapods/installer/analyzer.rb
```diff
@@ -522,12 +500,12 @@ def generate_pod_targets(specs_by_target)
           specs_by_target.flat_map do |target_definition, specs|
             grouped_specs = specs.group_by(&:root).values.uniq
             pod_targets = grouped_specs.flat_map do |pod_specs|
-              generate_pod_target([target_definition], pod_specs).scoped(dedupe_cache)
+              generate_pod_target([target_definition], pod_specs.map(&:spec), pod_specs.all?(&:used_by_tests_only?)).scoped(dedupe_cache)
             end
 
             pod_targets.each do |target|
-              dependencies = transitive_dependencies_for_specs(target.specs, target.platform, specs).group_by(&:root)
-              test_dependencies = transitive_dependencies_for_specs(target.specs.select(&:test_specification?), target.platform, all_specs).group_by(&:root)
```
dnkoutso: When `deduplicate_targets` is OFF this logic was completely broken. `all_specs` is not even defined here.

Goes to show how many people actually turn `deduplicate_targets` OFF

lib/cocoapods/resolver.rb
```diff
@@ -91,12 +124,12 @@ def specs_by_target
           dependencies = {}
           specs = target.dependencies.map(&:name).flat_map do |name|
             node = @activated.vertex_named(name)
-            valid_dependencies_for_target_from_node(target, dependencies, node) << node
+            (valid_dependencies_for_target_from_node(target, dependencies, node) << node).map { |s| [s, node.payload.test_specification?] }
```
dnkoutso: this is the "meat" of the change. We basically wrap all paylods with a `ResolvedSpecification` container and as we walk through the tree we ensure all nodes are marked as to whether this was for a test spec or not.

lib/cocoapods/installer/analyzer.rb
```diff
@@ -429,59 +433,33 @@ def generate_target(target_definition, pod_targets)
...
-      # @param [PodTarget] pod_target
-      #        the pod target being queried
-      #
       # @param [Array<PodTarget>] pod_targets
       #        the array of pod targets to check against
       #
-      # @return [Boolean] if the pod target is only referenced from test dependencies.
+      # @return [Array<PodTarget>] the filtered list of pod targets.
       #
-      def pod_target_test_only?(pod_target, pod_targets)
```
dnkoutso: Method was removed and so did the caching. The resolver now takes care of this much better.

spec/unit/installer/analyzer_spec.rb
```diff
@@ -672,84 +672,22 @@ module Pod
 
       #-------------------------------------------------------------------------#
 
-      it 'handles test only pod targets' do
```
dnkoutso: tests no longer applicable as `pod_target_test_only?` was removed as a method of the `analyzer.rb`

dnkoutso:
Added more tests in `resolver_spec`.

This should be good to go.

dnkoutso:
Uh wat? I dont remember merging this.

PR #7000, opened 2017-09-01
Exclude test resource and framework paths from aggregate targets

dnkoutso:


lib/cocoapods/target/pod_target.rb
```diff
@@ -200,56 +200,59 @@ def supported_test_types
     # @return [Array<Hash{Symbol => [String]}>] The vendored and non vendored framework paths
     #         this target depends upon.
     #
-    def framework_paths
-      @framework_paths ||= begin
```
dnkoutso: removed the `@framework_paths ||= begin` this is why the diff looks weird.

dnkoutso: Easier to review with https://github.com/CocoaPods/CocoaPods/pull/7000/files?w=1

lib/cocoapods/target/pod_target.rb
```diff
@@ -200,56 +200,59 @@ def supported_test_types
...
-            framework[:dsym_output_path] = "${DWARF_DSYM_FOLDER_PATH}/#{dsym_name}"
-          end
-          frameworks << framework
-        end
-        if should_build? && requires_frameworks? && !static_framework?
-          frameworks << { :name => product_name,
-                          :input_path => build_product_path('${BUILT_PRODUCTS_DIR}'),
-                          :output_path => "${TARGET_BUILD_DIR}/${FRAMEWORKS_FOLDER_PATH}/#{product_name}" }
+    def framework_paths(include_test_framework_paths = true)
+      accessors = file_accessors.flat_map
```
justinseanmartin: What's the `flat_map` here with no block for?

dnkoutso:
Just amended the commit message to be more accurate. No code changes on the last push.

lib/cocoapods/target/pod_target.rb
DanToml: What's the purpose of removing the caching? This is kinda expensive

dnkoutso: So now we are parameterized whether we return paths including the ones from tests or not.

I can actually cache it where the key includes the boolean if you'd like. WDYT?

dnkoutso:
All green!

lib/cocoapods/target/pod_target.rb
dnkoutso: Cache is now added back and it depends on the parameter.

Issue #6967, opened 2017-08-19
CocoaPods generates wrong frameworks search path when merging subspec dependencies from multiple podspec's #


paulb777:
🌈
* [X] I've read and understood the [*CONTRIBUTING guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

Reopen #6584, since I'm still seeing it with Xcode9 and CocoaPods 1.3.1.

I'm noticing that the frameworks search path is not generated correctly in an Xcode workspace is generated from two podspecs with separate subspec dependencies from the same external pod.

Reproduce in two minutes with:

1. `git clone git@github.com:paulb777/CocoapodsSubspecBug.git`
1. `cd CocoaPodsSubspecBug/Example/`
1. `pod update`
1. `open CocoaPodsSubspecBug.xcworkspace/`
1. Command-B to run the CocoaPodsSubspecBug-Example scheme. Note that it succeeds.
1. Change the scheme to Root-Example. Command-B and note success.
1. Change the scheme back to CocoaPodsSubspecBug-Example
1. Shift-Command-K to clean
1. Command-B and note compiler failure:

`/Users/paulbeusterien/demo/CocoaPodsSubspecBug/CocoaPodsSubspecBug/Classes/ReplaceMe.m:2:9: fatal error: 'GoogleToolboxForMac/GTMNSDictionary+URLArguments.h' file not found
#import <GoogleToolboxForMac/GTMNSDictionary+URLArguments.h>
        ^
1 error generated.`

## What did you expect to happen?

I expected the framework search path to point to the corresponding version of GoogleToolboxForMac.framework to the scheme.

## What happened instead?

Two GoogleToolboxForMac.frameworks paths were passed to clang and clang used the first, which led to it not finding the right header file.

### Stack

```
   CocoaPods : 1.3.1
        Ruby : ruby 2.0.0p648 (2015-12-16 revision 53162) [universal.x86_64-darwin16]
    RubyGems : 2.6.12
        Host : Mac OS X 10.12.6 (16G29)
       Xcode : 8.3.3 (8E3004b)
         Git : git version 2.14.1.480.gb18f417b89-goog
Ruby lib dir : /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib
[...]

PR #7002, opened 2017-09-02
Filter out subset dependent targets from FRAMEWORK_SEARCH_PATHS

paulb777:
When multiple build targets in a workspace have dependencies on different subspecs of another pod, builds can fail if the subset build has already been done and the framework search path occurs first on the command line.

This PR fixes the issue by removing the redundant subset subspecs from FRAMEWORK_SEARCH_PATHS. See the fixed issue #6967 for more details and a reproducible example.


dnkoutso:
LGTM. @DanToml want to double check?

PR #7006, opened 2017-09-04
Propagate HEADER_SEARCH_PATHS settings from search paths

### LINKED ISSUE #6891: Search path not propagated to test target

# Report

## What did you do?
1. Create a Swift project with unit tests
2. `pod init` to create a `Podfile`
3. Add `pod 'Firebase'` to `Podfile` under `use_frameworks!`
4. Run `pod update`
5. Open the generated xcworkspace
6. Add `import Firebase` to the `AppDelegate.swift`
7. Add `@testable import <ProjectName>` into `<ProjectName>Tests.swift`

Header imports from the Firebase pod are added to the testing target, causing
tests to build and run successfully.

## What happened instead?

A "Missing required module 'Firebase'" build error will appear. The directory
from the Firebase pod is **not** added to the testing target's
`Header Search Paths` even though `inherit! :search_paths` is set.

### Stack

```
   CocoaPods : 1.2.1
        Ruby : ruby 2.1.3p242 (2014-09-19 revision 47630) [x86_64-darwin16.0]
    RubyGems : 2.2.2
        Host : Mac OS X 10.12.5 (16F73)
       Xcode : 9.0 (9M174d)
         Git : git version 2.13.0.rc1.294.g07d810a77f-goog
Ruby lib dir : /Users/wilsonryan/.rbenv/versions/2.1.3/lib
Repositories : master - https://github.com/CocoaPods/Specs.git @ 92958bc76be648b458ed2cdf50ea7a5185aef134
```

(This does happen on previous Xcode versions as well).

### Installation Source

```
Executable Path: /Users/wilsonryan/.rbenv/versions/2.1.3/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.1
cocoapods-playgrounds : 1.2.2
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
cocoapods-trunk       : 1.2.0
cocoapods-try         : 1.1.0
```

Example project can be found [here](https://github.com/ryanwilson/BrokenFirebase).

## Discussion

This is fixable on the user's side by explicitly adding
`"${PODS_ROOT}/Firebase/Core/Sources"` to the test target's
`Header Search Paths`, but it shouldn't be necessary. There may be a fix for the
Podspec that can alleviate this issue in which case we're happy to do so!

The current podspec for the `Firebase` pod can be found [here](https://github.com/CocoaPods/Specs/blob/master/Specs/0/3/5/Firebase/4.0.3/Firebase.podspec.json)
and the high level structure for all of Firebase is as follows:

- Each Firebase product ships their own CocoaPod separate from the podspec above: [`FirebaseCore`](https://cocoapods.org/pods/FirebaseCore), [`FirebaseAuth`](https://cocoapods.org/pods/FirebaseAuth), [`FirebaseDatabase`](https://cocoapods.org/pods/FirebaseDatabase), and more.
- `Firebase` is an umbrella pod that contains multiple subspecs, all parts of Firebase:
  - `Core` is the default subspec that provides the umbrella header that imports any Firebase SDK that is available to the user.
  - There are multiple other subspecs based on products including `Auth`, `Database`, `Messaging`, etc.
  - Each of these subspecs depend on their equivalent external pods
    - Subspec `Auth` depends on `FirebaseAuth`
    - Subspec `Database` depends on `FirebaseDatabase`
    - etc.

At the moment, the spec relies on the `user_target_xcconfig` flag in the `Core`
subspec to explicitly add `${PODS_ROOT}/Firebase/Core/Sources` to
`HEADER_SEARCH_PATHS`. This isn't recommended, and the regular
target's case can be solved by explicitly adding
`"public_header_files": [ "Core/Sources/Firebase.h" ],` to the podspec, removing
`preserve_paths` and replacing it with `module_map` as seen below.

### Current Core Subspec
    {
      "dependencies": {
        "FirebaseAnalytics": "4.0.2",
        "FirebaseCore": "4.0.3"
      },
      "name": "Core",
      "preserve_paths": [
        "Core/Sources/module.modulemap"
      ],
      "source_files": [
        "Core/Sources/Firebase.h"
      ],
      "user_target_xcconfig": {
        "HEADER_SEARCH_PATHS": "$(inherited) ${PODS_ROOT}/Firebase/Core/Sources"
      }
    },

### "Fixed" Core Subspec
    {
      "dependencies": {
         "FirebaseAnalytics": "4.0.2",
         "FirebaseCore": "4.0.3"
       },
       "name": "Core",
       "module_map": "Core/Sources/module.modulemap",
       "public_header_files": [
         "Core/Sources/Firebase.h"
       ],
       "source_files": [
         "Core/Sources/Firebase.h"
       ]
    },

This is cleaner but unfortunately it doesn't solve the issue of the test target
not working out of the box.

If this is a simple `Podspec` fix we're happy to make the change, otherwise I
believe this is currently a bug.

paulb777:
For targets that  _inherit! :search_paths_, propagate HEADER_SEARCH_PATHS from _user_target_xcconfig_, in addition to propagating the FRAMEWORK_SEARCH_PATHS.

This fixes #6891 and https://github.com/firebase/firebase-ios-sdk/issues/58

### DIFF
--- a/lib/cocoapods/generator/xcconfig/xcconfig_helper.rb
+++ b/lib/cocoapods/generator/xcconfig/xcconfig_helper.rb
@@ -321,6 +321,32 @@ module Pod
           build_settings
         end
 
+        # Updates xcconfig with the HEADER_SEARCH_PATHS from the search_paths.
+        #
+        # @param  [Target] search_paths_target
+        #         The target.
+        #
+        # @param  [Xcodeproj::Config] xcconfig
+        #         The xcconfig to edit.
+        #
+        def self.propagate_header_search_paths_from_search_paths(search_paths_target, xcconfig)
+          header_search_paths_list = []
+          search_paths_target.pod_targets.each do |target|
+            target.spec_consumers.each do |spec_consumer|
+              paths = spec_consumer.user_target_xcconfig['HEADER_SEARCH_PATHS']
+              header_search_paths_list <<= paths unless paths.nil?
+            end
+            unless header_search_paths_list == []
+              header_search_paths = header_search_paths_list.join(' ')
+              unless header_search_paths.include? '$(inherited)'
+                header_search_paths = '$(inherited) ' + header_search_paths
+              end
+              build_settings = { 'HEADER_SEARCH_PATHS' => header_search_paths }
+              xcconfig.merge!(build_settings)
+            end
+          end
+        end
+
         # Add custom build settings and required build settings to link to
         # vendored libraries and frameworks.
         #
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Paul Beusterien](https://github.com/paulb777)
   [#7002](https://github.com/CocoaPods/CocoaPods/pull/7002)
 
+* Propagate HEADER_SEARCH_PATHS settings from search paths  
+  [Paul Beusterien](https://github.com/paulb777)
+  [#7006](https://github.com/CocoaPods/CocoaPods/pull/7006)
+
 ## 1.3.1 (2017-08-02)
 
 ##### Enhancements
--- a/lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
+++ b/lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
@@ -176,6 +176,9 @@ module Pod
             generator = AggregateXCConfig.new(search_paths_target, configuration_name)
             @xcconfig.merge! XCConfigHelper.search_paths_for_dependent_targets(nil, search_paths_target.pod_targets)
             @xcconfig.merge!(generator.settings_to_import_pod_targets)
+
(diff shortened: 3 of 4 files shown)


dnkoutso:
LGTM, would like one more additional review just to be sure.

lib/cocoapods/generator/xcconfig/xcconfig_helper.rb
```diff
@@ -312,6 +312,32 @@ def self.settings_for_dependent_targets(target, dependent_targets, test_xcconfig
...
 
+        # Updates xcconfig with the HEADER_SEARCH_PATHS from the search_paths.
+        #
+        # @param  [Target] search_paths_target
+        #         The target.
+        #
+        # @param  [Xcodeproj::Config] xcconfig
+        #         The xcconfig to edit.
+        #
+        def self.header_search_paths_from_search_paths(search_paths_target, xcconfig)
```
segiddins: this method probably should have a verb, since it updates/adds the search paths rather than returning them

DanToml:
This is good with a rebase

PR #7014, opened 2017-09-06
Use the resolver to identify which pod targets are test only

dnkoutso:
`resolver.rb` is the best (and only) place to identify if a specification for a given target definition is only used by tests or not.

This PR effectively moves the logic into the resolver because trying to figure out which pod target is a test only pod target after the resolver has executed is not possible.

We now wrap the `specs_by_target` result with a `ResolverSpecification` container that includes the necessary metadata (such as if the resolved spec was only used by tests).

Everything else just percolates through.

This effectively fixes all edge cases and problems around mistakenly identifying (and therefore linking) pod targets that are only used by tests.


justinseanmartin:
See description in #6994 for more info

lib/cocoapods/installer/analyzer.rb
```diff
@@ -365,13 +369,16 @@ def analyze_host_targets_in_podfile(aggregate_targets, embedded_aggregate_target
 
       # Creates the models that represent the targets generated by CocoaPods.
       #
+      # @param  [Array<AggregateTarget>] resolver_specs_by_target mapping of targets to resolved specs (containing information about test usage)
```
dnkoutso: @justinseanmartin this type is wrong will fix.

spec/unit/installer/analyzer_spec.rb
```diff
@@ -672,84 +672,30 @@ module Pod
 
       #-------------------------------------------------------------------------#
 
-      it 'handles test only pod targets' do
```
dnkoutso: tests are deleted because `pod_target_test_only?` is deleted. Similar tests moved to `resolver_spec.rb`

lib/cocoapods/installer/analyzer.rb
```diff
@@ -75,11 +75,15 @@ def analyze(allow_fetches = true)
...
-        @result.specs_by_target = validate_platforms(resolve_dependencies)
-        @result.specifications  = generate_specifications
-        @result.targets         = generate_targets
+        @locked_dependencies = generate_version_locking_dependencies
+        resolver_specs_by_target = resolve_dependencies
+        validate_platforms(resolver_specs_by_target)
+        @result.specifications  = generate_specifications(resolver_specs_by_target)
+        @result.targets         = generate_targets(resolver_specs_by_target)
         @result.sandbox_state   = generate_sandbox_state
+        @result.specs_by_target = resolver_specs_by_target.each_with_object({}) do |rspecs_by_target, hash|
```
dnkoutso: The result outside of the `analyzer` remains the same as it used to. A hash of specs by target.

DanToml:
Good with a rebase

dnkoutso:
Should go green now.

dnkoutso:
For posterity I pasted the description from #6994 here as well.

dnkoutso:
🎉

PR #7015, opened 2017-09-06
Ensure a unique ID is generated for each resource bundle

justinseanmartin:
🌈

Resource bundles can be processed multiple times across different pod targets with the same spec activated, but different subspec dependencies.

Example:
  Spec/Subspec1 has a resource bundle SubspecResources
  Spec/Subspec2 has a dependency on Spec/Subspec1
  Target1 depends on Spec/Subspec1
  Target2 depends on Spec/Subspec2 (which activates Spec/Subspec1 as a dependency)
  Both Target1 & Target2 try to integrate SubspecResources

Tested with and without the change and verified that the count of `FooResources.bundle` was 2 without the fix. This causes warnings to appear when using deterministic UUIDs for duplicate references in a valid scenario.

### DIFF
--- a/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
@@ -226,7 +226,6 @@ module Pod
                 bundle_target.product_reference.tap do |bundle_product|
                   bundle_file_name = "#{bundle_name}.bundle"
                   bundle_product.name = bundle_file_name
-                  bundle_product.path = bundle_file_name
                 end
 
                 filter_resource_file_references(paths) do |resource_phase_refs, compile_phase_refs|
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
+* Ensure a unique ID is generated for each resource bundle
+  [Justin Martin](https://github.com/justinseanmartin)
+  [#7015](https://github.com/CocoaPods/CocoaPods/pull/7015)
+
 * Do not include settings from file accessors of test specs into aggregate xcconfigs  
   [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
   [#7019](https://github.com/CocoaPods/CocoaPods/pull/7019)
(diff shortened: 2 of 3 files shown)


dnkoutso:
LGTM. Will let it soak for a bit for others to review.

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -221,10 +221,26 @@ def add_test_targets
...
+                # but different subspec dependencies. Example:
+                #   Spec/Subspec1 has a resource bundle SubspecResources
+                #   Spec/Subspec2 has a dependency on Spec/Subspec1
+                #   Target1 depends on Spec/Subspec1
+                #   Target2 depends on Spec/Subspec2 (which activates Spec/Subspec1 as a dependency)
+                #   Both Target1 & Target2 try to integrate SubspecResources
+                bundle_file_reference = project.products_group.find_file_by_path(bundle_file_name)
+
+                if bundle_file_reference
+                  bundle_target = bundle_file_reference.referrers.find { |t| t.is_a?(Xcodeproj::Project::Object::PBXNativeTarget) }
```
dnkoutso: style nit: `|r|` instead of `|t|`? Or `|referrer`?

justinseanmartin:
There is an issue with this, please do not merge. We are investigating.

justinseanmartin:
We ended up taking a little bit of a different tactic to solving this issue. Rather than trying to get rid of the duplicate resource targets, it was much more straightforward to get Xcodeproj to generate a unique UUID for the resource bundle based on having a different `path` value. 

The only thing this seems to impact is possibly whether the link shows up as red or not in the Xcode file navigator, and if all references have been built it does show up black instead.

I think more correct fixes would be to:
1. Get the Xcodeproj UUID generation to take into account the `referrers` and their UUIDs, rather than simply relying on the path/name of the FileReference itself
2. Deduplicate the resource bundles to only have 1 per spec. Both of the `PodTarget` objects that are generated for different sets of activated subspecs would then both point to the same resource bundle instead of two separate instances. 

Both of the above solutions would get rid of the warnings that are generated in this scenario.

#2 was the path that my original fix was heading towards and I believe is "more correct", but I believe there'd be much more downstream repercussions of doing this (handling different platforms, etc). This is why I've scaled it back and gone with the current approach.

justinseanmartin:
Just rebased and updated the changelog.

PR #7019, opened 2017-09-08
Do not include settings from file accessors of test specs into aggregate xcconfigs

dnkoutso:


lib/cocoapods/generator/xcconfig/xcconfig_helper.rb
```diff
@@ -76,13 +76,15 @@ def self.default_ld_flags(target, include_objc_flag = false)
         #
         def self.add_settings_for_file_accessors_of_target(aggregate_target, pod_target, xcconfig)
           pod_target.file_accessors.each do |file_accessor|
-            XCConfigHelper.add_spec_build_settings_to_xcconfig(file_accessor.spec_consumer, xcconfig)
-            XCConfigHelper.add_static_dependency_build_settings(aggregate_target, pod_target, xcconfig, file_accessor)
+            if aggregate_target.nil? || !file_accessor.spec.test_specification?
```
dnkoutso: this is the check thats important here and in the other methods...

spec/unit/generator/xcconfig/xcconfig_helper_spec.rb
```diff
@@ -241,6 +241,93 @@ module XCConfig
 
         #---------------------------------------------------------------------#
 
+        describe 'concerning settings for file accessors' do
```
dnkoutso: lol @paulb777 I think we wrote almost the same tests (https://github.com/CocoaPods/CocoaPods/pull/6988)...I think the change we both did is different however :)

dnkoutso: This change basically ensures test spec paths dont end up in aggregates

paulb777: :) Yeah, if I understand correctly, this change is preventing test spec paths from ending up in aggregates, while #6988 was about making sure that paths were properly propagated aggregates including static frameworks.

dnkoutso: Yeap exactly.

PR #7013, opened 2017-09-05
Perform code signing on xctest bundles in the Pods project

justinseanmartin:
🌈

We hit an issue where a xctest targets can't be launched with Xcode when they depend on a prebuilt binary framework. The error that was being yielded was:

```
2017-09-05 09:00:19.578 xctest[91346:821576] (dlopen_preflight(/Users/me/Library/Developer/Xcode/DerivedData/App-sdlkgjsdlkgjsdlkjsgdlksdglkj/Build/Products/Debug-iphonesimulator/TestTarget-Unit-Tests.xctest/TestTarget-Unit-Tests): Library not loaded: @rpath/VendoredFramework.framework/VendoredFramework
  Referenced from: /Users/me/Library/Developer/Xcode/DerivedData/App-sdlkgjsdlkgjsdlkjsgdlksdglkj/Build/Products/Debug-iphonesimulator/TestTarget-Unit-Tests.xctest/TestTarget-Unit-Tests
  Reason: no suitable image found.  Did find:
    /Users/me/Library/Developer/Xcode/DerivedData/App-sdlkgjsdlkgjsdlkjsgdlksdglkj/Build/Products/Debug-iphonesimulator/TestTarget-Unit-Tests.xctest/Frameworks/VendoredFramework.framework/VendoredFramework: required code signature missing for '/Users/me/Library/Developer/Xcode/DerivedData/App-sdlkgjsdlkgjsdlkjsgdlksdglkj/Build/Products/Debug-iphonesimulator/TestTarget-Unit-Tests.xctest/Frameworks/VendoredFramework.framework/VendoredFramework'
)
```

Specifically, the important part is `required code signature missing for '<path>'` at the end.

PR #7027, opened 2017-09-13
Don't codesign :osx test targets

justinseanmartin:
🌈

It turns out that we need to only sign iOS XCTest bundles for them to be launched within the simulator, but not OSX ones being run on the local machine. I've verified now that both iOS and OSX test bundles appear to work correctly in the same conditions.

This regression was introduced by #7013.


dnkoutso:
👍

PR #7036, opened 2017-09-14
Deduplicate test specs correctly from pod variants and targets

dnkoutso:
Given the following scenario:

```ruby
target 'MyApp'
  pod 'PodA', :testspecs => ['Tests']
end

target 'MyApp2'
  pod 'PodA'
end
```

This generates two variants of `PodA` one called `PodA` and one more called `PodA.root` but with the exact same sources. This is because testspecs were taken into consideration for the uniqueness of a `pod_variant.rb` within `analyzer.rb`

This is no longer the case. The number of test specs asked should not matter for the generation of a pod variant. Pod variants should vary across platform, subspecs and whether they require frameworks or not.

Tase case and change depends on https://github.com/CocoaPods/cocoapods-test-specs/pull/1

### DIFF
--- a/lib/cocoapods/installer/analyzer/pod_variant.rb
+++ b/lib/cocoapods/installer/analyzer/pod_variant.rb
@@ -7,6 +7,10 @@ module Pod
         #
         attr_accessor :specs
 
+        # @return [Array<Specification>] the test specs for the target
+        #
+        attr_accessor :test_specs
+
         # @return [Platform] the platform
         #
         attr_accessor :platform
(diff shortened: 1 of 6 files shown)


amorde:
Noticed this in our project as well, would love to see a fix for it 👍

dnkoutso:
This is the fix :) I will make sure it goes green and gets merged.

dnkoutso:
Should go green now

dnkoutso:
There we go.

dnkoutso:
@amorde can you use this patch and confirm it fixes your issue?

amorde:
@dnkoutso yup it works 👍

dnkoutso:
<3

lib/cocoapods/installer/analyzer/pod_variant.rb
```diff
@@ -7,6 +7,10 @@ class PodVariant
         #
         attr_accessor :specs
 
+        # @return [Array<Specification>] the test specs for the target
+        #
+        attr_accessor :testspecs
```
segiddins: `test_specs`

lib/cocoapods/installer/analyzer/pod_variant.rb
```diff
@@ -24,16 +28,20 @@ def root_spec
...
-        # @param [Array<String>] specs       @see #specs
-        # @param [Platform] platform         @see #platform
-        # @param [Bool] requires_frameworks  @see #requires_frameworks?
+        # @param [Array<Specification>] specs @see #specs
+        # @param [Platform] platform          @see #platform
+        # @param [Bool] requires_frameworks   @see #requires_frameworks?
         #
         def initialize(specs, platform, requires_frameworks = false)
           self.specs = specs
+          self.testspecs = []
```
segiddins: why not add this to the initializer?

dnkoutso: hm it might actually be OK now that I look at it.

dnkoutso:
thanks @segiddins added an extra test to ensure test specs are not included in the equality of two pod variants.

orta:
👍

Issue #1691, opened 2013-12-18
Precompiled headers of all projects using CocoaPods contain import of UIKit.


stanislaw:
Hello,

Here is example of file Pods-<mytarget>-AFNetworkingMeter-prefix.pch:

```
#ifdef __OBJC__
#import <UIKit/UIKit.h>
#endif

#import "Pods-<mytarget>-environment.h"
```

All my other CocoaPods-based projects have as well this import of UIKit included. I believe it is some kind of default behavior of CocoaPods but I want to avoid this import for the non-UI-projects.

How should I have my podspecs written, so they don't import UIKit?

P.S. I have such import even in projects, whose podspecs have explicit:

``` ruby
  s.framework = 'Foundation'
```

Thanks and sorry if I am missing something very simple here.

PR #7044, opened 2017-09-17
Add support for preventing pch file generation

paulb777:
See also https://github.com/CocoaPods/Core/pull/406.

Prevent pch file generation with the skip_pch podspec attribute.

Address #1691 and #1746. More context at https://stackoverflow.com/questions/44401464/is-there-a-way-to-disable-the-default-pch-file-in-cocoapods

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
-* None.  
+* Add support for preventing pch file generation with the skip_pch podspec attribute  
+  [Paul Beusterien](https://github.com/paulb777)
+  [#7044](https://github.com/CocoaPods/CocoaPods/pull/7044)
 
 ##### Bug Fixes
 
@@ -42,7 +44,7 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [#6969](https://github.com/CocoaPods/CocoaPods/pull/6969)
 
 * For source static frameworks, include frameworks from dependent targets and libraries in OTHER_LDFLAGS  
-  [paulb777](https://github.com/paulb777)
+  [Paul Beusterien](https://github.com/paulb777)
   [#6988](https://github.com/CocoaPods/CocoaPods/pull/6988)
 
 ##### Bug Fixes
--- a/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
@@ -45,12 +45,20 @@ module Pod
                   create_build_phase_to_move_static_framework_archive
                 end
               end
-              create_prefix_header
+              unless skip_pch?
+                create_prefix_header
+              end
               create_dummy_source
             end
           end
 
           private
+
+          # @return [Boolean] Whether the target should build a pch file.
+          #
+          def skip_pch?
+            target.specs.any? { |spec| spec.prefix_header_file.is_a?(FalseClass) }
+          end
 
           # Remove the default headers folder path settings for static library pod
           # targets.
--- a/spec/unit/installer/xcode/pods_project_generator/pod_target_installer_spec.rb
+++ b/spec/unit/installer/xcode/pods_project_generator/pod_target_installer_spec.rb
@@ -417,9 +417,6 @@ module Pod
                   'BananaLib-Pods-SampleProject-prefix.pch',
                   'BananaLib-Pods-SampleProject.xcconfig',
                 ]
-                group.children.map(&:display_name).sort.should.not == [
-                  'BananaLib-Pods-SampleProject-prefix.pch',
-                ]
               end
 
               it 'verifies disabling prefix header generation' do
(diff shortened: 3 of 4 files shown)


paulb777:
The corresponding Core PR (docs only) is https://github.com/CocoaPods/Core/pull/412

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -45,13 +45,21 @@ def install!
...
               create_dummy_source
             end
           end
 
           private
 
+          # @return [Boolean] Whether the target should build a pch file.
+          #
+          def skip_pch?
+            target.file_accessors.any? { |file_accessor| file_accessor.prefix_header == ':skip' }
```
dnkoutso: any reason to query each file accessor? I believe that method asks the spec itself which i think you have this information here.

I am only saying this because file accessors tend to be slow generally...

paulb777: You're right. Changed. Thanks!

paulb777:
Now requires https://github.com/CocoaPods/Core/pull/412 to go green

spec/unit/installer/xcode/pods_project_generator/pod_target_installer_spec.rb
```diff
@@ -476,6 +503,19 @@ class PodsProjectGenerator
                 ]
               end
 
+              it 'verifies disabling prefix header generation' do
```
dnkoutso: nit: same test name as the one above, is this a pattern across this file?

paulb777: yes, there's the same test name for within and without a scoped target

CHANGELOG.md
```diff
@@ -45,6 +45,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [paulb777](https://github.com/paulb777)
   [#6988](https://github.com/CocoaPods/CocoaPods/pull/6988)
 
+* Add support for preventing pch file generation with the skip_pch podspec attribute
```
dnkoutso: probably want to move up to the master section since 1.4.0.beta.1 shipped!

spec/unit/installer/xcode/pods_project_generator/pod_target_installer_spec.rb
```diff
@@ -408,6 +408,33 @@ class PodsProjectGenerator
...
+              it 'verifies keeping prefix header generation' do
+                @pod_target.specs.first.stubs(:prefix_header_file).returns(true)
+                @installer.install!
+                group = @project['Pods/BananaLib/Support Files']
+                group.children.map(&:display_name).sort.should == [
+                  'BananaLib-Pods-SampleProject-dummy.m',
+                  'BananaLib-Pods-SampleProject-prefix.pch',
+                  'BananaLib-Pods-SampleProject.xcconfig',
+                ]
+                group.children.map(&:display_name).sort.should.not == [
```
dnkoutso: what does assertion actually check for this specific test? I would imagine that there should be a `.pch` file if the value is set to `true` which is verified right above.

spec/unit/installer/xcode/pods_project_generator/pod_target_installer_spec.rb
```diff
@@ -408,6 +408,33 @@ class PodsProjectGenerator
...
+
+              it 'verifies disabling prefix header generation' do
+                @pod_target.specs.first.stubs(:prefix_header_file).returns(false)
+                @installer.install!
+                group = @project['Pods/BananaLib/Support Files']
+                group.children.map(&:display_name).sort.should == [
+                  'BananaLib-Pods-SampleProject-dummy.m',
+                  'BananaLib-Pods-SampleProject.xcconfig',
+                ]
+                group.children.map(&:display_name).sort.should.not == [
```
dnkoutso: maybe use `should.not.include`?

spec/unit/installer/xcode/pods_project_generator/pod_target_installer_spec.rb
```diff
@@ -476,6 +503,19 @@ class PodsProjectGenerator
...
 
+              it 'verifies disabling prefix header generation' do
+                @pod_target.specs.first.stubs(:prefix_header_file).returns(false)
+                @installer.install!
+                group = @project['Pods/BananaLib/Support Files']
+                group.children.map(&:display_name).sort.should == [
+                  'BananaLib-dummy.m',
+                  'BananaLib.xcconfig',
+                ]
+                group.children.map(&:display_name).sort.should.not == [
```
dnkoutso: same `should.not.include`?

spec/unit/installer/xcode/pods_project_generator/pod_target_installer_spec.rb
paulb777: The `should` test should be sufficient since it is testing for an exact match of the list. Deleting the `not`'s

dnkoutso:
:shipit:

PR #7051, opened 2017-09-20
Warn when a symlink is found during globbing

### LINKED ISSUE #6603: Pathname.glob() doesn't traverse symlinks
* [x] Using latest cocoapods (1.2.0)
* [x] Using Xcode 8.3 latest

# Report

Pathname.glob() in ruby doesn't traverse symlinks and thus results in broken integrations with Pods that use symlinked sources.

I can't confirm if this is an issue for remotely downloaded pods but this is an issue for local pods referenced via `pod 'SnapKit', path: 'Podspecs/'`.

We use a mono-repo and include all dependencies locally so I've created a custom Podspec for every dependency. We use `Podspecs/<ModuleName>.podspec` to house the specs while we keep the sources in `Vendor/<ModuleName>`

Because Cocoapods requires the `source_files` path to not be absolute and not traverse upwards in the file system we symlink the `Vendor` directory into the `Podspecs` directory.

With the symlink no pod will install because Cocoapods never traverses the sources. If we use a copy it works fine.

I got to digging in the Ruby code and found the exact place where the difference in path resolving occurs. In `path_list.rb` around line 50 there is `Pathname.glob(escaped_root + '**/*', File::FNM_DOTMATCH).lazy`. For some reason the evaluation of the globbing results in a shallow traversal for symlinks.

https://github.com/CocoaPods/CocoaPods/blob/master/lib/cocoapods/sandbox/path_list.rb#L55

## What did you do?

`pod install`

Pods to be installed

## What happened instead?

Cocoapods finds no sources and installs no pods

### Stack

```
   CocoaPods : 1.2.0
    RubyGems : 2.5.1
        Host : Mac OS X 10.12.2 (16C67)
       Xcode : 8.3 (8E162)
         Git : git version 2.11.0 (Apple Git-81)
Ruby lib dir : /usr/local/var/rbenv/versions/2.3.1/lib
Repositories : artsy - https://github.com/artsy/Specs.git @ 17ac6640821fe1fed0ef5159d7765662bd0a3653
               master - https://github.com/CocoaPods/Specs.git @ 6d1a8a522209bc650a62b09d7b69521419b9c14e
```

### Installation Source

```
Executable Path: /usr/local/var/rbenv/versions/2.3.1/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.1
cocoapods-keys        : 2.0.0
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
cocoapods-trunk       : 1.1.2
cocoapods-try         : 1.1.0
```

### Podfile

```ruby
platform :ios, '9.0'
use_frameworks!

target 'SymlinksFailure' do
    pod 'SnapKit', path: 'Podspecs/'
end
```

[SymlinksFailure.zip](https://github.com/CocoaPods/CocoaPods/files/880752/SymlinksFailure.zip)

amorde:
Closes #6603

[SymlinksFailure.zip](https://github.com/CocoaPods/CocoaPods/files/1323650/SymlinksFailure.zip)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
+* Warn when a symlink is found during globbing  
+  [Eric Amorde](https://github.com/amorde)
+  [#7051](https://github.com/CocoaPods/CocoaPods/pull/7051)
+
 * Clear input/output paths if they exceed an arbitrary limit  
   [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
   [#7362](https://github.com/CocoaPods/CocoaPods/issues/7362)
--- a/lib/cocoapods/sandbox/path_list.rb
+++ b/lib/cocoapods/sandbox/path_list.rb
@@ -57,6 +57,11 @@ module Pod
         root_length = root.cleanpath.to_s.length + File::SEPARATOR.length
         Find.find(root.to_s) do |f|
           directory = File.directory?(f)
+          path = Pathname.new(f)
+          # File.symlink? does not detect folder symlinks
+          if File.symlink?(f) || (path.realpath != path)
+            UI.warn "Symlink `#{f}` found while traversing `#{root}`. Symlinks are not traversed."
+          end
           f = f.slice(root_length, f.length - root_length)
           next if f.nil?
 
--- a/spec/unit/sandbox/path_list_spec.rb
+++ b/spec/unit/sandbox/path_list_spec.rb
@@ -192,6 +192,16 @@ module Pod
         path_list = Sandbox::PathList.new(fixture(unicode_name))
         path_list.files.should == ['README']
       end
+
+      it 'warns when symlinks are found' do
+        root = @path_list.root
+        Find.stubs(:find).multiple_yields(
+          "#{root}/SymlinkDirectory",
+        )
+        Pathname.any_instance.stubs(:realpath).returns(Pathname.new('/Some/different/path/SymlinkDirectory'))
+        @path_list.read_file_system
+        UI.warnings.should.include "Symlink `#{root}/SymlinkDirectory` found while traversing `#{root}`. Symlinks are not traversed"
+      end
     end
 
     #-------------------------------------------------------------------------#


lib/cocoapods/sandbox/path_list.rb
```diff
@@ -52,9 +52,18 @@ def read_file_system
...
-        dirs_and_files = absolute_paths.reject { |path| path.basename.to_s =~ /^\.\.?$/ }
-        dirs, files = dirs_and_files.partition { |path| File.directory?(path) }
+        absolute_paths = Pathname.glob(escaped_root + '**/*', File::FNM_DOTMATCH)
+        absolute_paths = absolute_paths.reject { |path| path.basename.to_s =~ /^\.\.?$/ }.sort_by(&:to_s)
+        symbolic_paths, absolute_paths = absolute_paths.partition { |path| File.symlink?(path) }
+        expanded_symbolic = symbolic_paths.map do |path|
+          begin
+            path.realpath
+          rescue Errno::ENOENT
+            raise Informative, "Unable to find reference for symbolic link #{path} pointing to #{File.readlink(path)}"
```
amorde: Unfortunately the prior error message from the project was able to include the target name, but here there's no way to access that. I added the destination of the symlink hoping it would alleviate that, but open to suggestions here

dnkoutso:
Want to rebase/squash?

amorde:
Not sure why this closed, that was weird. Rebased and squashed though

CHANGELOG.md
```diff
@@ -41,9 +41,12 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 * Do not merge `pod_target_xcconfig` from test specs into non test xcconfigs  
   [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
   [#7037](https://github.com/CocoaPods/CocoaPods/pull/7037)
+* Fix globbing to traverse symlinks
```
dnkoutso: nit: press enter

amorde: Oops, think that got messed up during the rebase

lib/cocoapods/sandbox/path_list.rb
```diff
@@ -52,9 +52,18 @@ def read_file_system
         end
         escaped_root = escape_path_for_glob(root)
 
-        absolute_paths = Pathname.glob(escaped_root + '**/*', File::FNM_DOTMATCH).lazy
-        dirs_and_files = absolute_paths.reject { |path| path.basename.to_s =~ /^\.\.?$/ }
-        dirs, files = dirs_and_files.partition { |path| File.directory?(path) }
+        absolute_paths = Pathname.glob(escaped_root + '**/*', File::FNM_DOTMATCH)
+        absolute_paths = absolute_paths.reject { |path| path.basename.to_s =~ /^\.\.?$/ }.sort_by(&:to_s)
```
dnkoutso: any reason for the `sort_by`?

lib/cocoapods/sandbox/path_list.rb
```diff
@@ -52,9 +52,18 @@ def read_file_system
         end
         escaped_root = escape_path_for_glob(root)
 
-        absolute_paths = Pathname.glob(escaped_root + '**/*', File::FNM_DOTMATCH).lazy
```
dnkoutso: also `.lazy` part no longer needed?

amorde: I think the lazy broke some of the next lines, but I don't know enough about Ruby to know why. I initially put it back but then removed it again

dnkoutso: Lets make sure we understand the intention here before we merge this.

http://ruby-doc.org/core-2.0.0/Enumerator/Lazy.html

dnkoutso: And http://blog.honeybadger.io/using-lazy-enumerators-to-work-with-large-files-in-ruby/

so I think it was for performance reasons which might be important especially for very large projects like ours.

lib/cocoapods/sandbox/path_list.rb
amorde: Just checked this and the `.lazy` and it seems to work without it - this may have been something else I was trying and forgot to remove, so I'll remove it

lib/cocoapods/sandbox/path_list.rb
amorde: Yup, totally agree. Mentioned in the other comment that I put it back and everything seems okay, I had forked robert's repo which had it removed

dnkoutso:
@amorde thanks, one more thing have you tried the patch into the project from #6603 to ensure it works?

amorde:
Just tried it now, the project by itself wouldn't work even if symlinks were traversed because the Podspec is looking for `'SnapKit/'` but it should be looking for `Vendor/SnapKit/`

Making that change causes an error in the `sorted_relative_paths_from_full_paths` block:

```
NoMethodError - undefined method `upcase' for nil:NilClass
/Users/eric/dev/CocoaPods/lib/cocoapods/sandbox/path_list.rb:81:in `yield'
/Users/eric/dev/CocoaPods/lib/cocoapods/sandbox/path_list.rb:81:in `each'
/Users/eric/dev/CocoaPods/lib/cocoapods/sandbox/path_list.rb:81:in `each'
/Users/eric/dev/CocoaPods/lib/cocoapods/sandbox/path_list.rb:81:in `each'
/Users/eric/dev/CocoaPods/lib/cocoapods/sandbox/path_list.rb:81:in `each'
/Users/eric/dev/CocoaPods/lib/cocoapods/sandbox/path_list.rb:81:in `sort_by'
/Users/eric/dev/CocoaPods/lib/cocoapods/sandbox/path_list.rb:81:in `block in read_file_system'
/Users/eric/dev/CocoaPods/lib/cocoapods/sandbox/path_list.rb:84:in `read_file_system'
/Users/eric/dev/CocoaPods/lib/cocoapods/sandbox/path_list.rb:34:in `files'
/Users/eric/dev/CocoaPods/lib/cocoapods/sandbox/path_list.rb:144:in `relative_glob'
/Users/eric/dev/CocoaPods/lib/cocoapods/sandbox/path_list.rb:106:in `glob'
```

@dnkoutso Could you take a look at it? At this point I'm not very comfortable merging it in since it seems I'm missing something here. I'll upload an updated version of that example project here

dnkoutso:
@amorde OK let's keep this open and I might get to it. It's good to have the test project to verify those :)

dnkoutso:
@amorde would love to land this for 1.4.0 if you want to take another stab at. 

Needs a rebase also.

amorde:
Sure, I'll take another look at it and see if I can get that sample project working

dnkoutso:
thank you!

amorde:
@dnkoutso I was able to get the sample project to work, but doing so breaks several different areas in the project. It looks like this will require a much larger change than anticipated, and I don't think I'll be able to complete it quickly

dnkoutso:
@amorde perhaps since this is not going to be solved now we intentionally raise an exception if any of the contents include symlinks?

amorde:
I can do that if that's what the team wants! I'm out of other ideas unfortunately

dnkoutso:
I think its better for us not to crash and provide a helpful message, even if it sucks overall for the experience.

amorde:
I'll take another look at this during the weekend and if I can't figure it out I'll open a separate PR for raising an error on symlinks

amorde:
@dnkoutso Made some progress on this, the last thing I need to do is remove any symlinks which are already accounted for by other file patterns so its not added twice

lib/cocoapods/sandbox/path_list.rb
```diff
@@ -111,7 +105,7 @@ def glob(patterns, options = {})
       #
       # @return [Array<Pathname>]
       #
-      def relative_glob(patterns, options = {})
+      def glob(patterns, options = {})
```
amorde: Because symlinks could point to somewhere that is not relative to the Podspec, the notion of relative globbing sort of had to be thrown out. Glob will now always return absolute paths

lib/cocoapods/sandbox/path_list.rb
```diff
@@ -133,18 +127,29 @@ def relative_glob(patterns, options = {})
...
+
+          pattern = (root + pattern).to_s
           expanded_patterns = dir_glob_equivalent_patterns(pattern)
           full_list.select do |path|
             expanded_patterns.any? do |p|
-              File.fnmatch(p, path, File::FNM_CASEFOLD | File::FNM_PATHNAME)
+              # Add the root to the LHS of the pattern so that symlinks match correctly with fnmatch
+              expanded = (root + p).to_s
+              result = File.fnmatch(expanded, path, File::FNM_CASEFOLD | File::FNM_PATHNAME)
+              # if result
```
amorde: Forgot about this, will remove it

spec/unit/project_spec.rb
```diff
@@ -235,7 +235,7 @@ def settings_for_root_configs(key)
           end.message.should.match /Paths must be absolute/
         end
 
-        it 'uses realpath for resolving symlinks' do
+        it 'retains original symlink paths' do
```
amorde: Symlinks added to the project will now retain the symlink path rather than the real path

spec/unit/sandbox/path_list_spec.rb
```diff
@@ -6,17 +6,25 @@ module Pod
...
-        expected = %w(
+        expected = expand_relative(%w(
           Banana.modulemap
           BananaLib.podspec
+          Bananalib.framework/Headers/Bananalib.h
+          Bananalib.framework/Headers/SubDir/SubBananalib.h
           Bananalib.framework/Versions/A/Headers/Bananalib.h
           Bananalib.framework/Versions/A/Headers/SubDir/SubBananalib.h
+          Bananalib.framework/Versions/Current/Headers/Bananalib.h
+          Bananalib.framework/Versions/Current/Headers/SubDir/SubBananalib.h
```
amorde: Need to remove these duplicates in an efficient way, open to suggestions

dnkoutso:
@amorde sorry I havent had time to take a deeper look here, just catching up slowly.

amorde:
@dnkoutso Have you gotten the chance to take a look at this? I'm unsure of the repercussions of the 1 integration test that failed, so I'm not sure if there's something that I need to fix or if I need to update the test

dnkoutso:
@amorde not yet, I owe you the full review will try to get on this this week.

CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

```markdown
* Warn when a symlink is found during globbing  
```

</p>

lib/cocoapods/sandbox/path_list.rb
```diff
@@ -57,6 +57,11 @@ def read_file_system
         root_length = root.cleanpath.to_s.length + File::SEPARATOR.length
         Find.find(root.to_s) do |f|
           directory = File.directory?(f)
+          path = Pathname.new(f)
+          # File.symlink? does not detect folder symlinks
+          if File.symlink?(f) || (path.realpath != path)
```
amorde: Would like a confirmation to make sure this logic is correct

segiddins: My guess is that only the latter part of the conditional is needed

amorde: I think you're right, I had the left hand side for files but the latter should catch both files and directories

lib/cocoapods/sandbox/path_list.rb
```diff
@@ -57,6 +57,11 @@ def read_file_system
...
+            UI.warn "Symlink `#{f}` found while traversing `#{root}`. Symlinks are not traversed."
```
segiddins: it might help to say what it symlinks to? (and maybe just keep a dictionary of what's found and do one big warning at the end?)

I'm also just generally worried about the performance of this as a whole, and wonder whether it might be better delegated to the linter?

lib/cocoapods/sandbox/path_list.rb
segiddins: I would also like to verify that this is needed at all? Since it looks like `find` may very well be traversing symlinks

lib/cocoapods/sandbox/path_list.rb
amorde: Yeah, this also appears to be triggering a lot of warnings for vendored frameworks so I don't think this is the right way to go.

lib/cocoapods/sandbox/path_list.rb
amorde: The attached project (on this PR) doesn't appear to be traversing symlinks

amorde:
Going to close this for now, since the various implementations I tried all had drawbacks.

Issue #7040, opened 2017-09-15
Test spec stalling with no output for a while (consider xcodebuild verbose flag)


abbeycode:
> Before you start,  are you using the latest CocoaPods release?

Yes, 1.3.1

* [x] I've read and understood the [*CONTRIBUTING guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

From the [issue59 branch of UnzipKit](https://github.com/abbeycode/UnzipKit/tree/issue59), run `./Scripts/cocoapod-validate.sh`, which basically calls `pod lib lint --verbose`, but sets up some environment variables first.

## What did you expect to happen?

Expected output to be written throughout the unit tests' run.

## What happened instead?

After the unit tests are started, there is no output for a long time. On Travis-CI ((example build)[https://travis-ci.org/abbeycode/UnzipKit/jobs/274427308])), it's going longer than 30 minutes without any output, causing the build to fail. When I run locally, it's going 6 minutes, which is still noticeable. This is the last line before output stalls:

> xcodebuild clean test -workspace /var/folders/bb/n7t3rs157850byt_jfdcq9k80000gn/T/CocoaPods-Lint-20170915-1321-1bl59r0-UnzipKit/App.xcworkspace -scheme UnzipKit-Unit-Tests -configuration Debug CODE_SIGN_IDENTITY=

### Stack
```
   CocoaPods : 1.3.1
        Ruby : ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-darwin16]
    RubyGems : 2.6.13
        Host : Mac OS X 10.12.5 (16F73)
       Xcode : 8.3.3 (8E3004b)
         Git : git version 2.13.1
Ruby lib dir : /Users/travis/.rvm/rubies/ruby-2.4.1/lib
Repositories : master - https://github.com/CocoaPods/Specs.git @ 2fe61865cbe4e17db78b41a0141558755e5d6de1
```
### Installation Source
```
Executable Path: /Users/travis/.rvm/gems/ruby-2.4.1/bin/pod
```
### Plugins
```
cocoapods-deintegrate : 1.0.1
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
cocoapods-trunk       : 1.2.0
cocoapods-try         : 1.1.0
```

## Project that demonstrates the issue

[...]

PR #7052, opened 2017-09-20
Added '-verbose' flag to xcodebuild for validation

abbeycode:
This will provide more diagnostic output for when validation is failing, or in my case, hopefully prevent the CI build to fail due to no output for a certain interval (Issue #7040).

### DIFF
--- a/lib/cocoapods/validator.rb
+++ b/lib/cocoapods/validator.rb
@@ -857,6 +857,10 @@ module Pod
         command += Fourflusher::SimControl.new.destination(:oldest, 'tvOS', deployment_target)
       end
 
+      if config.verbose
+        command += ' -verbose'
+      end
+
       output, status = _xcodebuild(command)
 
       unless status.success?
--- a/lib/cocoapods/validator.rb
+++ b/lib/cocoapods/validator.rb
@@ -857,7 +857,7 @@ module Pod
         command += Fourflusher::SimControl.new.destination(:oldest, 'tvOS', deployment_target)
       end
 
-      command += ' -verbose' if config.verbose?
+      command += [' -verbose'] if config.verbose?
 
       output, status = _xcodebuild(command)
 
--- a/lib/cocoapods/validator.rb
+++ b/lib/cocoapods/validator.rb
@@ -857,9 +857,7 @@ module Pod
         command += Fourflusher::SimControl.new.destination(:oldest, 'tvOS', deployment_target)
       end
 
-      if config.verbose
-        command += ' -verbose'
-      end
+      command += ' -verbose' if config.verbose?
 
       output, status = _xcodebuild(command)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [paulb777](https://github.com/paulb777)
   [#6988](https://github.com/CocoaPods/CocoaPods/pull/6988)
 
+* Added '-verbose' flag to xcodebuild for validation  
+  [abbeycode](https://github.com/abbeycode)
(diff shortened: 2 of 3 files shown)


dnkoutso:
I think you might need to update `validator_spec.rb` as well.

CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

```markdown
* Added '-verbose' flag to xcodebuild for validation  
  [abbeycode](https://github.com/abbeycode)
```

</p>

lib/cocoapods/validator.rb
```diff
@@ -857,6 +857,10 @@ def xcodebuild(action, scheme, configuration)
         command += Fourflusher::SimControl.new.destination(:oldest, 'tvOS', deployment_target)
       end
 
+      if config.verbose
+        command += ' -verbose'
```
dnkoutso: nit: I think this can be `command += ' -verbose' if config.verbose?`

Also use `verbose?` with question mark instead.

abbeycode: Thanks, Ruby's not exactly my specialty :)

abbeycode:
@dnkoutso Should I update the validator spec to check that the `-verbose` flag gets passed? If so, which case should I look to as an example? `runs xcodebuild with correct arguments for code signing` looks like it might be a possibility.

dnkoutso:
Either expand on an existing test or add a new one. Either should be fine for this.

orta:
👍

dnkoutso:
Good stuff! can you also squash into a single commit?

abbeycode:
@dnkoutso I'll squash when I'm done - all commits, or just the test spec commits?

dnkoutso:
All commits into a single one but yeah when you are done.

lib/cocoapods/validator.rb
```diff
@@ -857,6 +857,8 @@ def xcodebuild(action, scheme, configuration)
         command += Fourflusher::SimControl.new.destination(:oldest, 'tvOS', deployment_target)
       end
 
+      command += ' -verbose' if config.verbose?
```
dnkoutso: This needs to be `command += ['-verbose']` instead.

dnkoutso:
@abbeycode what does `-verbose` in xcodebuild actually do? We've tried your patch locally and it doenst seem it outputs any more information than what we have today.

dnkoutso:
@abbeycode they problem lies that in the current logic the whole output is captured via:

```ruby
    # Executes the given command in the current working directory.
    #
    # @return [(String, Status)] The output of the given command and its
    #         resulting status.
    #
    def _xcodebuild(command)
      UI.puts 'xcodebuild ' << command.join(' ') if config.verbose
      Executable.capture_command('xcodebuild', command, :capture => :merge)
    end
```

So the actual output of xcodebuild is not streamed as you'd expect and verbose won't make any difference.

abbeycode:
A ha... that makes sense then. What is the intention of doing it this way?

dnkoutso:
@abbeycode uncertain, predates me :)

abbeycode:
@orta? :)

abbeycode:
It looks like we can call `execute_command` instead of `capture_command`, and it'll only stream the output if `--verbose` is passed, which would (I think) satisfy my problem. But there may be a reason output was captured in the first place. Thoughts?

orta:
I don't _think_ there is any special reason for it

abbeycode:
I'm going to close this PR and open a new one taking the approach of replacing `capture_command` with `execute_command`.

Issue #6953, opened 2017-08-14
App host support for test specs


DanToml:
I couldn't find an issue tracking this, so I filed this for posterity.

dnkoutso:
DSL https://github.com/CocoaPods/Core/pull/399

dnkoutso:
https://github.com/CocoaPods/CocoaPods/pull/7054

PR #7054, opened 2017-09-22
Add app host support for test specs

dnkoutso:
closes https://github.com/CocoaPods/CocoaPods/issues/6953

- [x] generate Info.plist for app host
- [x] properly wire target to use the app host
- [x] Fix all documentation TODOs
- [x] add missing tests


spec/unit/validator_spec.rb
```diff
@@ -581,81 +581,6 @@ def podspec_path(name = 'JSONKit', version = '1.4')
           Xcodeproj::Project.schemes(project.path).should == %w(App)
         end
 
-        describe 'creating the importing file' do
```
dnkoutso: tests moved to `app_target_helper_spec.rb`, they were not deleted.

lib/cocoapods/generator/app_target_helper.rb
```diff
@@ -0,0 +1,136 @@
+module Pod
```
dnkoutso: this is extracted out from `validator.rb` and provides a bunch of helper methods, similar to `xcconfig_helper.rb`.

Easier to test and I can reuse this logic when I create app host target support below.

lib/cocoapods/generator/info_plist_file.rb
```diff
@@ -39,21 +44,6 @@ def save_as(path)
         end
       end
 
-      # The version associated with the current target
```
dnkoutso: This moved to `target.rb` and `pod_target.rb`.

spec/unit/generator/info_plist_file_spec.rb
```diff
@@ -2,64 +2,21 @@
 
 module Pod
   describe Generator::InfoPlistFile do
-    describe '#target_version' do
```
dnkoutso: These tests moved into `pod_target.rb`

lib/cocoapods/generator/info_plist_file.rb
```diff
@@ -108,15 +98,15 @@ def info
...
+          'CFBundleShortVersionString' => version,
           'CFBundleSignature' => '????',
           'CFBundleVersion' => '${CURRENT_PROJECT_VERSION}',
-          'NSPrincipalClass' => '',
           'CFBundleDevelopmentRegion' => 'en',
         }
 
         info['CFBundleExecutable'] = '${EXECUTABLE_NAME}' if bundle_package_type != :bndl
         info['CFBundleVersion'] = '1' if bundle_package_type == :bndl
+        info['NSPrincipalClass'] = 'NSApplication' if bundle_package_type == :appl && platform == :osx
```
dnkoutso: According to this https://developer.apple.com/library/content/documentation/General/Reference/InfoPlistKeyReference/Articles/CocoaKeys.html#//apple_ref/doc/uid/TP40009251-SW11

`NSPrincipalClass` is only present for macOS targets and must be set when generating an app host for macOS targets.

dnkoutso: Btw this changed because generating an app host for macOS requires this value to be set to `NSApplication` so I fixed it with this PR.

lib/cocoapods/generator/app_target_helper.rb
```diff
@@ -0,0 +1,226 @@
...
+      #
+      # @param  [Symbol] platform
+      #         the platform of the target. Can be `:ios` or `:osx`.
+      #
+      # @param  [String] name
+      #         The name of the folder to use and save the generated main file.
+      #
+      # @return [Pathname] the new source file that was generated.
+      #
+      def self.create_app_host_main_file(project, platform, name = 'App')
```
dnkoutso: this method and the constants below are new here, the rest is the same and copied from `validator.rb`

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -172,6 +175,43 @@ def add_files_to_build_phases
...
+              # Wire all test native targets with the app host.
+              native_test_target = target.native_target_for_spec(test_spec)
+              native_test_target.build_configurations.each do |configuration|
+                test_host = "$(BUILT_PRODUCTS_DIR)/#{name}.app/"
+                test_host << 'Contents/MacOS/' if target.platform == :osx
+                test_host << name.to_s
+                configuration.build_settings['TEST_HOST'] = test_host
+              end
+              # TODO THIS IS UGLY.
+              root_attributes = project.root_object.attributes['TargetAttributes'] || {}
```
dnkoutso: @segiddins it appears that when wiring up the app host if I don't do this Xcode will mark my app host as "Custom". Granted this is ugly do you have any ideas/thoughts perhaps to hide this within Xcodeproj?

spec/unit/generator/info_plist_file_spec.rb
```diff
@@ -74,10 +31,46 @@ module Pod
...
+          'CFBundleName' => '${PRODUCT_NAME}',
+          'CFBundlePackageType' => 'APPL',
+          'CFBundleShortVersionString' => '1.0.0',
+          'CFBundleSignature' => '????',
+          'CFBundleVersion' => '${CURRENT_PROJECT_VERSION}',
+          'NSPrincipalClass' => '',
+        }
+      end
+
+      it 'adds NSPrincipalClass for OSX platform' do
```
dnkoutso: new test that verifies that the Info.plist includes `NSApplication` as the value of `NSPrincipalClass` that is required by app targets for macOS.

dnkoutso:
A few more tests and this is done.

lib/cocoapods/installer/xcode/pods_project_generator.rb
```diff
@@ -285,11 +285,16 @@ def development_pod_targets
 
         def add_pod_target_test_dependencies(pod_target, frameworks_group)
           test_dependent_targets = pod_target.all_test_dependent_targets
-          pod_target.test_native_targets.each do |test_native_target|
+          pod_target.test_specs.each do |test_spec|
```
dnkoutso: here we loop through each test spec instead because we query the test spec if it requires an app host. If it does we wire up the dependency to the app host so the app host gets built.

dnkoutso:
**Update**

This is now complete! Added missing tests for `pod_target_installer` and now feeling really good about this. 

Tested it on a few internal projects also and until I added the flag `requires_app_host` the tests were failing. I added the flag and the whole thing worked out of the box.

<3 1.4.0!

dnkoutso:
Rebased onto 1.4.0.beta.1.

dnkoutso:
Looks a bit large but its fairly easy to review. The biggest diff is the extraction of helper methods from Validator for dealing with app target generation and main or import file generation.

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -172,6 +175,43 @@ def add_files_to_build_phases
...
+          # appropriate build configurations.
+          #
+          # @return [void]
+          #
+          def add_test_app_host_targets
+            target.test_specs.each do |test_spec|
+              next unless test_spec.consumer(target.platform).requires_app_host?
+              name = target.app_host_label(test_spec.test_type)
+              platform_name = target.platform.name
+              app_host_target = !project.targets.find { |t| t.name == name }
```
dnkoutso: this is wrong. There should be no `!` there. Internal testing caught this. fixing.

dnkoutso:
Rebased

lib/cocoapods/validator.rb
```diff
@@ -439,71 +439,24 @@ def download_pod
...
-      pod_target = @installer.pod_targets.find { |pt| pt.pod_name == spec.root.name }
-
-      source_file = write_app_import_source_file(pod_target)
-      source_file_ref = app_project.new_group('App', 'App').new_file(source_file)
       app_target = app_project.targets.first
-      app_target.add_file_references([source_file_ref])
-      add_swift_version(app_target)
-      add_xctest(app_target) if @installer.pod_targets.any? { |pt| pt.spec_consumers.any? { |c| c.frameworks.include?('XCTest') } }
+      pod_target = @installer.pod_targets.find { |pt| pt.pod_name == spec.root.name }
+      Pod::Generator::AppTargetHelper.add_app_project_import(app_project, app_target, pod_target, consumer.platform_name, use_frameworks)
```
dnkoutso: This was wrong.

```ruby
od::Generator::AppTargetHelper.add_app_project_import(app_project, app_target, pod_target, consumer.platform_name, deployment_target)
```
Should be
```ruby
od::Generator::AppTargetHelper.add_app_project_import(app_project, app_target, pod_target, consumer.platform_name, use_frameworks)
```

dnkoutso: Fixed.

lib/cocoapods/installer/xcode/pods_project_generator/target_installer.rb
```diff
@@ -138,10 +138,9 @@ def support_files_temp_dir
           #
```
justinseanmartin: NIT: params comments

lib/cocoapods/installer/xcode/pods_project_generator/target_installer.rb
```diff
@@ -138,10 +138,9 @@ def support_files_temp_dir
...
           # @return [void]
           #
-          def create_info_plist_file
-            path = target.info_plist_path
+          def create_info_plist_file(path, native_target, version, platform, bundle_package_type = :fmwk)
```
justinseanmartin: NIT: I'd just remove the default parameter for `bundle_package_type ` and force it to be explicit to have less optional parameters. I think it makes understanding callers more clear IMO.

lib/cocoapods/target.rb
```diff
@@ -180,6 +180,12 @@ def dummy_source_path
       support_files_dir + "#{label}-dummy.m"
     end
 
+    # @return [String] The version associated with this target
+    #
+    def version
+      '1.0.0'
```
justinseanmartin: NIT: Seems like maybe this should be an attr_reader initialized on creating the target, instead of a method with a hardcoded value?

justinseanmartin: Or make this a constant at the top and call it `DEFAULT_VERSION` or something?

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -180,6 +183,43 @@ def add_files_to_build_phases
...
+              end
+              # Wire all test native targets with the app host.
+              native_test_target = target.native_target_for_spec(test_spec)
+              native_test_target.build_configurations.each do |configuration|
+                test_host = "$(BUILT_PRODUCTS_DIR)/#{name}.app/"
+                test_host << 'Contents/MacOS/' if target.platform == :osx
+                test_host << name.to_s
+                configuration.build_settings['TEST_HOST'] = test_host
+              end
+              # TODO: This is kinda ugly, consider moving it to Xcodeproj instead?
```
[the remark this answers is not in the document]
dnkoutso: yes will remove.

lib/cocoapods/target.rb
dnkoutso: will make it a const

lib/cocoapods/generator/app_target_helper.rb
```diff
@@ -0,0 +1,226 @@
...
+        source_file = project.path.dirname.+("#{name}/main.m")
+        source_file.parent.mkpath
+        source_file.open('w') do |f|
+          case platform
+          when :ios, :tvos
```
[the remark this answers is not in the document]
dnkoutso: Works on `:tvos` but generally not for `:watchos` but I think this will be fixed by limiting app hosts at the DSL level. I will open a PR for this after this lands.

dnkoutso:
Merging.

PR #7058, opened 2017-09-24
Build subspecs in static frameworks without error

paulb777:
podspecs with subspecs that included the `static_framework` attribute were generating transitive errors even though they would build correctly.

This PR fixes that issue by checking the static_framework attribute from the outermost spec instead of the subspec itself.

The issue was reported at https://github.com/CocoaPods/CocoaPods/pull/6811#issuecomment-331628583.

### DIFF
--- a/spec/fixtures/matryoshka/matryoshka.podspec
+++ b/spec/fixtures/matryoshka/matryoshka.podspec
@@ -7,6 +7,7 @@ Pod::Spec.new do |s|
   s.homepage         = "http://httpbin.org/html"
   s.source           = { :git => "http://malyutin.local/matryoshka.git", :tag => s.version.to_s }
   s.license          = 'MIT'
+  s.static_framework = true
 
   s.source_files = 'Outmost.{h,m}'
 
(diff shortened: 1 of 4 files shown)


orta:
I don't know all the details, but this feels reasonable to me - is it worth having a test?

dnkoutso:
+1 on adding a test

paulb777:
I thought you might say that. :P Thanks for keeping from being lazy :)

I added a test that demonstrates the fix.

spec/fixtures/matryoshka/matryoshka.podspec
```diff
@@ -7,6 +7,7 @@ Pod::Spec.new do |s|
   s.homepage         = "http://httpbin.org/html"
   s.source           = { :git => "http://malyutin.local/matryoshka.git", :tag => s.version.to_s }
   s.license          = 'MIT'
+  s.static_framework = true
```
dnkoutso: surprised this change didnt break other tests :)

We could potentially provide another spec dedicated to static framework testing but I won't push hard on that.

paulb777: I was going to clone it if other test broke, but since it didn't, it might be better to have fewer fixtures.

PR #7068, opened 2017-09-26
Fix framework and resources paths caching

dnkoutso:
I followed https://www.justinweiss.com/articles/4-simple-memoization-patterns-in-ruby-and-one-gem/ for memoizing with parameters which does not seem to work.

I didn't want to spend enough time so I went back to a more "standard" way. 

No need for specs since this is already covered and the existing tests still pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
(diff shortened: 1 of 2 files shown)


dnkoutso:
I completely understand it now and this memoization technique was wrong. This is better.

dnkoutso:
Easier to review https://github.com/CocoaPods/CocoaPods/pull/7068/files?w=1

PR #7093, opened 2017-10-01
Add podspec, readme, license, docs, etc. to development pods

amorde:
This adds support for editing the Podspec (and other files) of local development pods.

It's a small thing that I've been wanting for our project, as we divide our app into frameworks using local development pods.

The following files are supported:

- Podspec
- Docs (in a folder which matches the glob pattern `doc{s}{*,.*}`)
- Custom module map (using the Podspec's `module_map` attribute)
- Prefix header (using the Podspec's `prefix_header_file` attribute)
- License (using either the Podspec's `license` attribute, or automatically found with the glob pattern `licen{c,s}e{*,.*}`)
- README (matching the glob pattern `readme{*,.*}`

Requires https://github.com/CocoaPods/cocoapods-integration-specs/pull/128


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

```markdown
* [WIP] Add support for editing podspecs of local development pods  
```

</p>

lib/cocoapods/external_sources/path_source.rb
```diff
@@ -27,6 +27,13 @@ def description
         "from `#{declared_path}`"
       end
 
+      # @return [Pathname] The absolute path of the podspec.
+      #
+      def podspec_path
```
amorde: I changed this to public access with the intention of using it in the installer, but ended up changing the implementation. Will change it back

lib/cocoapods/sandbox/file_accessor.rb
```diff
@@ -291,6 +291,15 @@ def module_map
         end
       end
 
+      # @return [Array<Pathname>] Paths to include for local pods to assist in development
+      #
+      def developer_files
+        if file = spec_consumer.license[:file]
```
amorde: I would have used the file accessor's `license` function but that will apply a glob to the directory of the podspec to search for the license, and wanted to avoid false positives for when multiple podspecs are stored in the same directory

spec/unit/generator/xcconfig/pod_xcconfig_spec.rb
```diff
@@ -86,7 +86,7 @@ module XCConfig
           it 'sets the PODS_TARGET_SRCROOT build variable for local pod' do
             @pod_target.sandbox.store_local_path(@pod_target.pod_name, @spec.defined_in_file)
             @xcconfig = @generator.generate
-            @xcconfig.to_hash['PODS_TARGET_SRCROOT'].should == '${PODS_ROOT}/../../spec/fixtures/banana-lib/BananaLib.podspec'
+            @xcconfig.to_hash['PODS_TARGET_SRCROOT'].should == '${PODS_ROOT}/../../spec/fixtures/banana-lib'
```
amorde: This seems like the original implementation was incorrect, as I don't think the full podspec path should be the value of `PODS_TARGET_SRCROOT`

lib/cocoapods/sandbox/file_accessor.rb
orta: This is pretty cool, I added this to `pod lib create` back in the day - I think it's also worth putting the README in there, it's a useful reference file and it's pretty common to change it.

screen shot 2017-10-01 at 8 02 31 am

If you were feeling bold, you could also check for:

* A CHANGELOG
* A folder called docs and include the folder reference

amorde: That's actually where I got the idea from :D 

I had the same thought about adding other files like README, CHANGELOG, etc. but was worried about situations where multiple podspecs are kept in a single folder - they would all have the same readme, changelog, etc. where that might not be the intended result. Do you think that wouldn't be an issue?

orta: I *think* it could be an issue because we have deterministic uuids, and including the same file _may_ cause duplicates. Someone more familiar with that code may know. 

In those cases I think it's fine to either not show them, you could see if there is more than 1 podspec in the folder and only add the files then,

dnkoutso: That's correct regarding deterministic uuids. we've had cases recently in which adding the same file twice would cause warnings. Make sure to try this on a multi platform pod as you might need to namespace the group according to the platform to avoid this.

lib/cocoapods/installer/xcode/pods_project_generator/file_references_installer.rb
```diff
@@ -109,6 +110,21 @@ def add_resources
...
+            UI.message '- Adding development pod helper files' do
+              file_accessors.each do |file_accessor|
+                pod_name = file_accessor.spec.name
+                next unless sandbox.local?(pod_name)
+                root_name = Specification.root_name(pod_name)
+                paths = file_accessor.developer_files
+                paths.each do |path|
+                  group = pods_project.group_for_spec(root_name, :developer)
+                  pods_project.add_file_reference(path, group, false)
+                end
```
amorde: I had this as 

```ruby
add_file_accessors_paths_to_pods_group(:developer_files, :developer, true)
```

but that caused some duplicate groups when subspecs were specified. I really just needed `group_for_spec` to use the root name instead of `file_accessor.spec.name`

amorde:
@dnkoutso @orta I think I've finished this, could you take another look at let me know what you think?

orta:
[Tests seem to be crashing inside](https://travis-ci.org/CocoaPods/CocoaPods/jobs/290722287#L2643)  `Pod::Generator::XCConfig` - have you been seeing this locally?

lib/cocoapods/sandbox.rb
```diff
@@ -352,20 +350,24 @@ def remove_checkout_source(name)
...
     # @param  [#to_s] path
-    #         The local path where the Pod is stored.
+    #         The path to the local Podspec
     #
     # @param  [Bool] was_absolute
     #         True if the specified local path was absolute.
     #
+    # @param  [AbstractExternalSource] source
+    #         The source which provided the podspec
+    #
```
orta: You may have removed this param at a later date

spec/unit/sandbox/file_accessor_spec.rb
```diff
@@ -211,6 +211,46 @@ module Pod
         @accessor.license.should == @root + 'LICENSE'
       end
 
+      it 'returns the docs of the specification' do
+        @accessor.docs.should == [
+          @root + 'docs/guide1.md',
+          @root + 'docs/subdir/guide2.md',
+        ]
```
orta: 👏 - this stuff is definitely worthy of going into the release post

lib/cocoapods/sandbox.rb
amorde: Indeed, good catch!

amorde:
@orta No was not seeing that locally, but the latest build seems to just have errors that will be fixed with the integration specs PR

amorde:
@paulb777 I think I could get a unit test in for that, not sure where to put it though

amorde:
Rebased

amorde:
Are there any changes I need to make before this is ready to merge? Other than the integration specs PR https://github.com/CocoaPods/cocoapods-integration-specs/pull/128

dnkoutso:
@amorde I dont think so, will review as well and inform you.

spec/unit/generator/xcconfig/pod_xcconfig_spec.rb
dnkoutso: I think this was just a bad test, the actual generated value is correct, right?

amorde:
@dnkoutso Ok, thanks! :)

spec/unit/sandbox/file_accessor_spec.rb
```diff
@@ -211,6 +211,46 @@ module Pod
...
+      it 'returns the matching podspec of the specification' do
+        @accessor.stubs(:specs).returns([@root + 'BananaLib.podspec', @root + 'OtherLib.podspec'])
+        @accessor.send(:podspec_file).should == @root + 'BananaLib.podspec'
+      end
+
+      it 'returns the developer files of the specification' do
+        @accessor.developer_files.should == [
+          @root + 'Banana.modulemap',
+          @root + 'BananaLib.podspec',
+          @root + 'Classes/BananaLib.pch',
```
dnkoutso: aren't `.pch` files already added in the project references? I checked a project and I see them there, are they being double added now?

lib/cocoapods/sandbox.rb
```diff
@@ -352,7 +350,7 @@ def remove_checkout_source(name)
     #         The name of the Pod.
     #
     # @param  [#to_s] path
```
dnkoutso: does this need to change into `[Pathname, String]`?

amorde: yeah I think so, good catch

spec/unit/generator/xcconfig/pod_xcconfig_spec.rb
amorde: Not sure, how was the testing passing before then?

spec/unit/sandbox/file_accessor_spec.rb
amorde: I thought it was copied to the `Support Files` folder, so editing it would not affect the original file. Did I misunderstand that?

amorde: Indeed, just double checked and the contents of the file will be appended to `{POD_NAME}-prefix.pch` which is stored in `Pods/Target Support Files/{POD_NAME}-{PLATFORM}/`. Adding it here allows the dev to edit it and run `pod install` to update the generated file

orta:
@amorde - these CI fails look legit

```

 -> BananaLib (1.0)
    - ERROR | [iOS] unknown: Encountered an unknown error (No such file or directory @ realpath_rec - PROJECT/LICENSE) during validation.
```

amorde:
@orta That should be fixed by the specs PR, see https://github.com/CocoaPods/cocoapods-integration-specs/pull/128/files#diff-4092bb2c94d1e2a5fe3721ec660903d3

dnkoutso:
@amorde merged the integration specs! Update the submodule reference here for green and we merge.

orta:
Niccceee

amorde:
All green woooooo

dnkoutso:
:shipit:

amorde:
Thank you @dnkoutso and @orta for your help on this!

dnkoutso:
@amorde one cool thing to also add is the `podspec` file being added we can change its language specification provider to be ruby, this will give xcode syntax highlighting. See https://github.com/CocoaPods/CocoaPods/blob/master/lib/cocoapods/project.rb#L222-L228 on how its done for the Podfile.

Just a minor but good improvement :)

amorde:
@dnkoutso I noticed that as well and couldn't figure out how to do it - Thank you for pointing me in the right direction!

francoisrouault:
Hi all !
I was wondering if there is a way to add helper files from another directory than `docs` or `doc` : 
I am using SwiftGen in my Development Pod (by adding build script phase in Podfile's `post_install`) and I would like my SwiftGen configuration file and input resource files (`Colors.txt`, `Fonts/**/*`) to  be in `Pods.xcproj/Development Pods/NameOfMyPod/Pod` or equivalent.
Thanks

amorde:
There isn't currently a way to do that

PR #7103, opened 2017-10-05
Do not set a `CODE_SIGN_IDENTITY` for macOS app hosts or xctest bundles

dnkoutso:
`validator.rb` sets `CODE_SIGN_IDENTITY` to nothing when lint occurs. This matches this case and fixes issues with code signing app hosts or macOS bundles.


dnkoutso:
Rebased, will merge on green.

PR #7109, opened 2017-10-06
Avoid 'Unspecified' value for SWIFT_VERSION build settings

heyzooi:
* Avoid 'Unspecificed' value for SWIFT_VERSION build settings
* Making Swift 3.2 the default value since this is the minimum version supported by Xcode 9

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
   [#7081](https://github.com/CocoaPods/CocoaPods/issues/7081)
 
+* Avoid 'Unspecificed' value for SWIFT_VERSION build settings and make 'Swift 3.2' the default value
+  [Victor Hugo Barros](https://github.com/heyzooi)
+  [#7109](https://github.com/CocoaPods/CocoaPods/issues/7109)
+
 ## 1.4.0.beta.1 (2017-09-24)
 
 ##### Enhancements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,7 +50,7 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
   [#7081](https://github.com/CocoaPods/CocoaPods/issues/7081)
 
-* Avoid 'Unspecificed' value for SWIFT_VERSION build settings and make 'Swift 3.2' the default value
+* Avoid 'Unspecified' value for SWIFT_VERSION build settings and make 'Swift 3.2' the default value
   [Victor Hugo Barros](https://github.com/heyzooi)
   [#7109](https://github.com/CocoaPods/CocoaPods/issues/7109)
 
--- a/examples/TestInclusions/TestInclusions/Base.lproj/Main.storyboard
+++ b/examples/TestInclusions/TestInclusions/Base.lproj/Main.storyboard
@@ -1,8 +1,12 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="10116" systemVersion="15D21" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" initialViewController="BYZ-38-t0r">
+<?xml version="1.0" encoding="UTF-8"?>
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="13196" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" colorMatched="YES" initialViewController="BYZ-38-t0r">
+    <device id="retina4_7" orientation="portrait">
+        <adaptation id="fullscreen"/>
+    </device>
     <dependencies>
-        <deployment identifier="iOS"/>
-        <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="10085"/>
+        <deployment version="1792" identifier="iOS"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="13173"/>
+        <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
     </dependencies>
     <scenes>
         <!--View Controller-->
@@ -14,9 +18,9 @@
                         <viewControllerLayoutGuide type="bottom" id="wfy-db-euE"/>
                     </layoutGuides>
                     <view key="view" contentMode="scaleToFill" id="8bC-Xf-vdC">
-                        <rect key="frame" x="0.0" y="0.0" width="600" height="600"/>
+                        <rect key="frame" x="0.0" y="0.0" width="375" height="667"/>
                         <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
-                        <color key="backgroundColor" white="1" alpha="1" colorSpace="custom" customColorSpace="calibratedWhite"/>
+                        <color key="backgroundColor" red="1" green="1" blue="1" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
                     </view>
                 </viewController>
                 <placeholder placeholderIdentifier="IBFirstResponder" id="dkx-z0-nzr" sceneMemberID="firstResponder"/>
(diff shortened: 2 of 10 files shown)


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

```markdown
* Avoid 'Unspecificed' value for SWIFT_VERSION build settings  
  [heyzooi](https://github.com/heyzooi)
```

</p>

CHANGELOG.md
```diff
@@ -50,6 +50,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
   [#7081](https://github.com/CocoaPods/CocoaPods/issues/7081)
 
+* Avoid 'Unspecificed' value for SWIFT_VERSION build settings and make 'Swift 3.2' the default value
```
dnkoutso: typo, `Unspecified`

heyzooi: thanks... fixing it!

lib/cocoapods/validator.rb
```diff
@@ -470,7 +470,7 @@ def install_pod
           next unless (native_target = pod_target.native_target)
           native_target.build_configuration_list.build_configurations.each do |build_configuration|
             (build_configuration.build_settings['OTHER_CFLAGS'] ||= '$(inherited)') << ' -Wincomplete-umbrella'
-            build_configuration.build_settings['SWIFT_VERSION'] = swift_version if pod_target.uses_swift?
+            build_configuration.build_settings['SWIFT_VERSION'] = swift_version if pod_target.uses_swift? && !swift_version.nil?
```
dnkoutso: is there a test that covers the `!swift_version.nil?` part? If not can you please add one?

lib/cocoapods/generator/app_target_helper.rb
```diff
@@ -100,7 +100,7 @@ def self.add_xctest_search_paths(target)
       #
       def self.add_swift_version(target, swift_version)
         target.build_configurations.each do |configuration|
-          configuration.build_settings['SWIFT_VERSION'] = swift_version
+          configuration.build_settings['SWIFT_VERSION'] = !swift_version.nil? ? swift_version : Validator::DEFAULT_SWIFT_VERSION
```
dnkoutso: needs a test

lib/cocoapods/project.rb
```diff
@@ -25,6 +25,9 @@ def initialize(path, skip_initialization = false,
       @pods = new_group('Pods')
       @development_pods = new_group('Development Pods')
       self.symroot = LEGACY_BUILD_ROOT
+      root_object.build_configuration_list.build_configurations.each do |config|
```
dnkoutso: why do we care for the entire project to set this value? It seems that each and every other pod target that uses swift should use a value now.

lib/cocoapods/validator.rb
heyzooi: i'm not a ruby developer, i will try my best to write unit tests

dnkoutso: not a problem! There are plenty of examples around the `*_spec.rb` for the class you are writing a test for you can cargo cult from.

spec/unit/validator_spec.rb
```diff
@@ -894,10 +894,10 @@ def test_swiftpod_with_dot_swift_version(version = '3.1.0')
       end
 
       describe '#swift_version' do
-        it 'defaults to Swift 3.0' do
+        it 'defaults to Swift 3.2' do
```
amorde: There is a test right above this that contains a warning message starting with `Swift projects uses Swift 3.0 by default` - could that test (and possibly the code that generates that warning) be updated as well?

lib/cocoapods/validator.rb
heyzooi: i'm trying to figured out how to run unit tests in ruby. reading this tutorial here: https://code.tutsplus.com/tutorials/ruby-for-newbies-testing-with-rspec--net-21297
but i'm getting this error:

```
/Users/victor/.rvm/gems/ruby-2.0.0-p648/gems/rspec-core-3.6.0/lib/rspec/core/formatters.rb:206:in `built_in_formatter': cannot load such file -- rspec/core/formatters/progress_formatter (LoadError)
```

googling it to figured out how to fix it. if know the solution, let me know please.

lib/cocoapods/generator/app_target_helper.rb
amorde: Nit pick here, rather than checking for nil in each loop, could the desired value be extracted to before the loop? Something like

```
version_to_set = swift_version || Validator::DEFAULT_SWIFT_VERSION
target.build_configurations.each do |configuration|
    configuration.build_settings['SWIFT_VERSION'] = version_to_set
end
```

lib/cocoapods/validator.rb
heyzooi: nvm... i figured how to run `bundle exec bacon ...`

amorde: you can also run `bundle exec rake spec` or `bundle exec rake spec:unit` for only unit tests

heyzooi:
hey guys, how do I link the `cocoapods-integration-specs` submodule to this PR https://github.com/CocoaPods/cocoapods-integration-specs/pull/125?

amorde:
@heyzooi That PR needs to be accepted before this one, then update the git submodule reference to point to the latest commit

lib/cocoapods/generator/app_target_helper.rb
heyzooi: sure! let me do that

lib/cocoapods/project.rb
heyzooi: if you don't specify a value for the `SWIFT_VERSION` build settings, Xcode will not compile, so the idea here was to set a default version and allow each pod to override the default version as needed.

dnkoutso: @heyzooi doesnt each pod that uses swift automatically get assigned a version? I am still not 100% convinced we need to be changing the project configuration in this PR.

examples/TestInclusions/TestInclusions/Base.lproj/Main.storyboard
```diff
@@ -1,8 +1,12 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
dnkoutso: does this file need to change?

lib/cocoapods/validator.rb
dnkoutso: @heyzooi can you provide a case where the part `!swift_version.nil?` can be true?

dnkoutso:
This change we should overall land but there are still a few questions remaining before we do so.

lib/cocoapods/generator/app_target_helper.rb
```diff
@@ -99,8 +99,9 @@ def self.add_xctest_search_paths(target)
       # @return [void]
       #
       def self.add_swift_version(target, swift_version)
+        version_to_set = !swift_version.nil? ? swift_version : Validator::DEFAULT_SWIFT_VERSION
```
dnkoutso: not sure this belongs here, and if it does we need to update the documentation.

I still need to clearly understand a case in which in `validator.rb`

```ruby
    # @return [String] the SWIFT_VERSION to use for validation.
    #
    def swift_version
      return @swift_version if defined?(@swift_version)
      if version = dot_swift_version
        @swift_version = version
      else
        DEFAULT_SWIFT_VERSION
      end
    end
```

Can return `nil`? Is it if the contents of the `.swift-version` file are empty?

DanToml: if `.swift-version` is empty, it returns nil, and goes to DEFAULT_SWIFT_VERSION

dnkoutso: So yes I don't see the reason why we are doing a `nil?` check here and in the other place, @heyzooi

dnkoutso:
@heyzooi I am going to decline this PR in favor of #7136. I gave you credit in this other PR.

I verified that with Xcode 9 I no longer see "Unsupported" version for Swift and instead defaults to 3.2.

I do not think the remaining changes in this PR as necessary.

piemonte:
as of recent, i've encountered this issue with the following project:

📎 https://github.com/piemonte/player

even though i have a .swift-verision and properly specified project file, linting still seems to fail. i haven't been able to push the latest release.

dnkoutso:
@piemonte are you on 1.4.0.beta.1? There was a regression in setting Swift version entirely for the project which was fixed in https://github.com/CocoaPods/CocoaPods/pull/7131 but it has not been released yet.

piemonte:
@dnkoutso ah, yes, i am. you're awesome, thanks for the link.

PR #7110, opened 2017-10-06
Warn when a pod that was added or changed includes script phases

dnkoutso:
Users must be aware if pods are adding script phases during installation. This is intended not to be a solution for protecting the user's system but provide visibility to the user who must inspect the script phase.


lib/cocoapods/installer.rb
```diff
@@ -626,19 +645,6 @@ def run_podfile_post_install_hook
 
     #-------------------------------------------------------------------------#
 
-    public
-
-    # @return [Array<Library>] The targets of the development pods generated by
-    #         the installation process.
-    #
-    def development_pod_targets
```
dnkoutso: unused method.

jmkk: FYI be careful when saying "unused method" - it might have been used by client scripts, not necessarily from within Cocoapods. This commit broke my scripting because I was depending on `development_pod_targets` being exposed.

ob: Likewise this broke our scripts. I think changes like this should go in the ChangeLog.

jmkk: @dnkoutso ^^ this might be more impactful for others that switch to 1.4.0

dnkoutso: Ill add it back and include a comment. Will be in rc.2

dnkoutso: Ill add it back and include a comment. Will be in rc.2

PR #6971, opened 2017-08-22
Fix validation when using --swift-version

DanToml:
Closes #6941

PR #7131, opened 2017-10-12
Better warning message for which Swift version was used during validation

### LINKED ISSUE #7121: Document/Improve Swift language version support

# Report

## What did you do?

The docs for publishing a pod don't really talk at all about `.swift-version`, but trying to run `lint` or `push` the pod can kick out warnings about it.  But if one trying to find out more about the file, there is no clear info in the official docs.  Even if you use the `--swift-version` command line arg, you still get warnings telling you to create the file, and the warning goes no indication that the command line arg was honored.

If the pod is really for a Swift project, odds are they are going support multiple versions of Swift (since the language is stabilizing and there are `#swift()` conditionals, it is very doable), having the file can make some developers think that version is required.

So better docs about what this file is would help for people who thing this implies something about the pod; likewise, docs in generally acknowledging how Swift versions come into play with pods would be welcomed.

On https://github.com/apple/swift-protobuf we're support back "one full major version" (so Swift 3.0.1 thru 4 currently).

dnkoutso:
closes #7121, fixes regression from https://github.com/CocoaPods/CocoaPods/pull/6971 in CocoaPods 1.4.0.beta.1

/cc @thomasvl The validator will only output this message if the user has not provided a swift version via a parameter or via a `.swift-version` file. 

This is how it looks now:

```
 -> CannonPodder (1.0.5)
    - WARN  | [iOS] swift_version: The validator for Swift projects used Swift 3.0 by default 
because no Swift version was specified. If you want to use a different version of Swift during 
validation, then either use the `--swift-version` parameter or use a `.swift-version` 
file to set the version of Swift to use for your Pod. For example to use Swift 2.3, run: 
`echo "2.3" > .swift-version`
    - ERROR | [iOS] xcodebuild: Returned an unsuccessful exit code.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
+* Better warning message for which Swift version was used during validation  
+  [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
+  [#7121](https://github.com/CocoaPods/CocoaPods/issues/7121)
+
 * Strip vendored dSYMs during embed script phase  
   [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
   [#7111](https://github.com/CocoaPods/CocoaPods/issues/7111)
@@ -139,7 +143,7 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 * Fix validation warnings when using --swift-version  
   [Danielle Tomlinson](https://github.com/dantoml)
-  [#6971](https://github.com/CocoaPods/CocoaPods/issue/6971)
+  [#6971](https://github.com/CocoaPods/CocoaPods/pull/6971)
 
 * Fix xcconfig boolean merging when substrings include yes or no  
   [Paul Beusterien](https://github.com/paulb777)
--- a/lib/cocoapods/generator/app_target_helper.rb
+++ b/lib/cocoapods/generator/app_target_helper.rb
@@ -99,6 +99,7 @@ module Pod
       # @return [void]
       #
       def self.add_swift_version(target, swift_version)
+        raise 'Cannot set empty Swift version to target.' if swift_version.blank?
         target.build_configurations.each do |configuration|
           configuration.build_settings['SWIFT_VERSION'] = swift_version
         end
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -139,7 +143,7 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 * Fix validation warnings when using --swift-version  
   [Danielle Tomlinson](https://github.com/dantoml)
-  [#6971](https://github.com/CocoaPods/CocoaPods/issue/6971)
+  [#6971](https://github.com/CocoaPods/CocoaPods/pull/6971)
```
dnkoutso: Fixed link

lib/cocoapods/generator/app_target_helper.rb
```diff
@@ -99,6 +99,7 @@ def self.add_xctest_search_paths(target)
       # @return [void]
       #
       def self.add_swift_version(target, swift_version)
+        raise 'Cannot empty Swift version to target.' if swift_version.nil?
```
dnkoutso: just being defensive here :)

lib/cocoapods/validator.rb
```diff
@@ -256,7 +256,7 @@ def validation_dir
     # @return [String] the SWIFT_VERSION to use for validation.
     #
     def swift_version
-      return @swift_version if defined?(@swift_version)
+      return @swift_version unless @swift_version.nil?
```
dnkoutso: this is the regression fix.

dnkoutso:
Should go green.

amorde:
Looks like this will cause conflicts with #7109

dnkoutso:
That's OK whoever wins the race gets to resolve the conflict :)

lib/cocoapods/generator/app_target_helper.rb
```diff
@@ -99,6 +99,7 @@ def self.add_xctest_search_paths(target)
       # @return [void]
       #
       def self.add_swift_version(target, swift_version)
+        raise 'Cannot set empty Swift version to target.' if swift_version.nil?
```
DanToml: `blank?` might be what you are looking for here?

dnkoutso: sure

dnkoutso:
Updated.

PR #7166, opened 2017-10-25
Prevent multiple script phases from stripping vendored dSYM

dnkoutso:
After upgrading to 1.4.0.beta.2 we noticed a race condition regarding dSYM architecture stripping. This is because multiple targets can execute the "Embed Pods Framework" script phase in parallel and strip the same dSYM file which can yield to a race and error.

This prevents that by moving the file into the intermediates folder of the scoped target, stripping it and then moving it to the final destination.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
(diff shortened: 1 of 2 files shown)


dnkoutso:
Verified fixed in our large project with 10+ targets sharing the same `dSYM` file each with embed pods frameworks.

dnkoutso:
Will merge on green.

Issue #7170, opened 2017-10-26
Pure Objective-C test_spec fails to compile or run if it depends on Swift pod


adc-amatosov:
* [x] I've read and understood the [*CONTRIBUTING guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report
Pure Objective-C test_spec fails to compile or run if it depends on Swift pod (direct or transient dependency)

## What did you do?
Create lib with `pod lib create TestSpecs`
Configured test_spec for it using the Tests.m generated by `lib create`
Tried to compile and run tests
Complete project available at https://github.com/adc-amatosov/TestSpecs.git

## What did you expect to happen?
test_spec should generate compilable Test target even if it has only Objective-C files

## What happened instead?

In case of transient dependency build fails with error:
```
ld: library not found for -lswiftSwiftOnoneSupport
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

In case of direct dependency build succeeds but fails to run with error:
```
2017-10-26 12:25:50.737707-0700 xctest[85233:2589292] The bundle “TestSpecs-Unit-Tests.xctest” couldn’t be loaded because it is damaged or missing necessary resources. Try reinstalling the bundle.
2017-10-26 12:25:50.737829-0700 xctest[85233:2589292] (dlopen_preflight(~/Library/Developer/Xcode/DerivedData/TestSpecs-getnbftqhnqcyjaymnbrfnmdtskn/Build/Products/Debug-iphonesimulator/TestSpecs-Unit-Tests.xctest/TestSpecs-Unit-Tests): Library not loaded: @rpath/libswiftCore.dylib
  Referenced from: ~/Library/Developer/Xcode/DerivedData/TestSpecs-getnbftqhnqcyjaymnbrfnmdtskn/Build/Products/Debug-iphonesimulator/ReactiveCocoa/ReactiveCocoa.framework/ReactiveCocoa
  Reason: image not found)
Program ended with exit code: 82
```

## Workaround 
Adding an empty `dummy.swift` file to test_spec solves both issues

### Stack

```
   CocoaPods : 1.3.1
        Ruby : ruby 2.2.5p319 (2016-04-26 revision 54774) [x86_64-darwin15]
    RubyGems : 2.6.13
        Host : Mac OS X 10.13 (17A405)
       Xcode : 9.0 (9A235)
         Git : git version 2.10.2
[...]

PR #7171, opened 2017-10-26
Set language to Swift for test native targets if any dependencies use Swift

dnkoutso:
closes https://github.com/CocoaPods/CocoaPods/issues/7170

Test native targets actually build and link multiple dependencies. If any of these dependencies require Swift, we should be setting the test native target language to Swift.


dnkoutso:
🙏

PR #7180, opened 2017-10-29
Add set -u to frameworks and resources copy scripts

keith:
This adds `set -u` to both the frameworks and resources copy shell
scripts. This means that if the scripts attempt to reference any
variable that is undefined, they will fail.

### DIFF
--- a/lib/cocoapods/generator/embed_frameworks_script.rb
+++ b/lib/cocoapods/generator/embed_frameworks_script.rb
@@ -37,6 +37,7 @@ module Pod
         script = <<-SH.strip_heredoc
           #!/bin/sh
           set -e
+          set -u
 
           echo "mkdir -p ${CONFIGURATION_BUILD_DIR}/${FRAMEWORKS_FOLDER_PATH}"
           mkdir -p "${CONFIGURATION_BUILD_DIR}/${FRAMEWORKS_FOLDER_PATH}"
--- a/lib/cocoapods/generator/embed_frameworks_script.rb
+++ b/lib/cocoapods/generator/embed_frameworks_script.rb
@@ -131,7 +131,7 @@ module Pod
 
           # Signs a framework with the provided identity
           code_sign_if_enabled() {
-            if [ -n "${EXPANDED_CODE_SIGN_IDENTITY}" -a "${CODE_SIGNING_REQUIRED}" != "NO" -a "${CODE_SIGNING_ALLOWED}" != "NO" ]; then
+            if [ -n "${EXPANDED_CODE_SIGN_IDENTITY}" -a "${CODE_SIGNING_REQUIRED:-}" != "NO" -a "${CODE_SIGNING_ALLOWED}" != "NO" ]; then
               # Use the current code_sign_identitiy
               echo "Code Signing $1 with Identity ${EXPANDED_CODE_SIGN_IDENTITY_NAME}"
               local code_sign_cmd="/usr/bin/codesign --force --sign ${EXPANDED_CODE_SIGN_IDENTITY} ${OTHER_CODE_SIGN_FLAGS} --preserve-metadata=identifier,entitlements '$1'"
--- a/lib/cocoapods/generator/embed_frameworks_script.rb
+++ b/lib/cocoapods/generator/embed_frameworks_script.rb
@@ -38,6 +38,7 @@ module Pod
           #!/bin/sh
           set -e
           set -u
+          set -o pipefail
 
           echo "mkdir -p ${CONFIGURATION_BUILD_DIR}/${FRAMEWORKS_FOLDER_PATH}"
           mkdir -p "${CONFIGURATION_BUILD_DIR}/${FRAMEWORKS_FOLDER_PATH}"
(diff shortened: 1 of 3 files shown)


dnkoutso:
👍

this will require integration specs bump

CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

```markdown
* Add set -u to frameworks and resources copy scripts  
  [keith](https://github.com/keith)
```

</p>

keith:
It also looks like before this is viable, some changes would have to be made. For example if you have a static library target that has pods dependencies, CocoaPods adds the copy resources build phase, but `UNLOCALIZED_RESOURCES_FOLDER_PATH` isn't set. Example project including this change: [PodsStaticLibraries.zip](https://github.com/CocoaPods/CocoaPods/files/1424780/PodsStaticLibraries.zip)

dnkoutso:
@keith any plans on continuing this PR?

keith:
Yes, I just need to figure out how to update the tests

segiddins:
@keith if you'd like, I can handle bumping the integration specs / rebasing

keith:
If you want to I'd totally appreciate it! I do plan to come back to it otherwise. I'm still a bit worried there are some default cases I'm missing depending on project configurations

dnkoutso:
> It also looks like before this is viable, some changes would have to be made. For example if you > have a static library target that has pods dependencies, CocoaPods adds the copy resources build > phase, but UNLOCALIZED_RESOURCES_FOLDER_PATH isn't set. Example project including this > change: PodsStaticLibraries.zip

@keith @segiddins we have an internal monkey patch for this method to avoid adding this script phase for static library targets. The intention is for people to convert these libraries into pods but the reality of things is that CocoaPods does not disallow to integrate a static library target. 

We could upstream our fix to prevent from adding this phase for static library targets.

lib/cocoapods/generator/embed_frameworks_script.rb
```diff
@@ -120,10 +122,10 @@ def script
 
           # Signs a framework with the provided identity
           code_sign_if_enabled() {
-            if [ -n "${EXPANDED_CODE_SIGN_IDENTITY}" -a "${CODE_SIGNING_REQUIRED}" != "NO" -a "${CODE_SIGNING_ALLOWED}" != "NO" ]; then
+            if [ -n "${EXPANDED_CODE_SIGN_IDENTITY}" -a "${CODE_SIGNING_REQUIRED:-}" != "NO" -a "${CODE_SIGNING_ALLOWED}" != "NO" ]; then
               # Use the current code_sign_identitiy
               echo "Code Signing $1 with Identity ${EXPANDED_CODE_SIGN_IDENTITY_NAME}"
-              local code_sign_cmd="/usr/bin/codesign --force --sign ${EXPANDED_CODE_SIGN_IDENTITY} ${OTHER_CODE_SIGN_FLAGS} --preserve-metadata=identifier,entitlements '$1'"
+              local code_sign_cmd="/usr/bin/codesign --force --sign ${EXPANDED_CODE_SIGN_IDENTITY} ${OTHER_CODE_SIGN_FLAGS-:} --preserve-metadata=identifier,entitlements '$1'"
```
segiddins: should this be `:-`?

keith: fixed

segiddins:
@keith what would you think of something like these guards for the script phases:

```diff
diff --git a/lib/cocoapods/generator/copy_resources_script.rb b/lib/cocoapods/generator/copy_resources_script.rb
index f00c84ca..4dc2b0cc 100644
--- a/lib/cocoapods/generator/copy_resources_script.rb
+++ b/lib/cocoapods/generator/copy_resources_script.rb
@@ -98,6 +98,12 @@ module Pod
 set -e
 set -u
 
+if [ -z ${UNLOCALIZED_RESOURCES_FOLDER_PATH+x} ]; then
+    # If UNLOCALIZED_RESOURCES_FOLDER_PATH is not set, then there's no where for us to copy
+    # resources to, so exit 0 (signalling the script phase was successful).
+    exit 0
+fi
+
 mkdir -p "${TARGET_BUILD_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}"
 
 RESOURCES_TO_COPY=${PODS_ROOT}/resources-to-copy-${TARGETNAME}.txt
diff --git a/lib/cocoapods/generator/embed_frameworks_script.rb b/lib/cocoapods/generator/embed_frameworks_script.rb
index f7fd2110..2378a8fd 100644
--- a/lib/cocoapods/generator/embed_frameworks_script.rb
+++ b/lib/cocoapods/generator/embed_frameworks_script.rb
@@ -39,6 +39,12 @@ module Pod
           set -e
           set -u
 
+          if [ -z ${FRAMEWORKS_FOLDER_PATH+x} ]; then
+              # If FRAMEWORKS_FOLDER_PATH is not set, then there's no where for us to copy
+              # frameworks to, so exit 0 (signalling the script phase was successful).
+              exit 0
+          fi
+
           echo "mkdir -p ${CONFIGURATION_BUILD_DIR}/${FRAMEWORKS_FOLDER_PATH}"
           mkdir -p "${CONFIGURATION_BUILD_DIR}/${FRAMEWORKS_FOLDER_PATH}"
 
```

This way, CocoaPods does not have to make assumptions about when to copy things, as the user can then set `FRAMEWORKS_FOLDER_PATH` / `UNLOCALIZED_RESOURCES_FOLDER_PATH` for static library targets?

keith:
What do you think about trying to finish some change like this to solve that instead https://github.com/CocoaPods/CocoaPods/pull/7181

segiddins:
I think both solutions can/should be implemented, together.

keith:
Sounds good. If possible I think we should `exit 1` wherever possible if it really is a mis-configuration of some sort

segiddins:
@keith updated!

dnkoutso:
Still LGTM.

keith:
Nice! Thanks for taking this over!

PR #7247, opened 2017-11-30
[XCConfigHelper] Quote framework names in OTHER_LDFLAGS

### LINKED ISSUE #7185: Spaces in a vendored framework path breaks OTHER_LDFLAGS
# Report
I'm creating and testing a private pod on my local machine.  I've got a vendored framework and a vendored static lib.  The framework has a space in the name and the generated xcconfig OTHER_LDFLAGS gets mangled.  `pod lib lint` also fails.

## What did you do?

In the podspec:
`s.vendored_libraries  = 'build/libcef_dll_wrapper/Debug/libcef_dll_wrapper.a'`
`s.vendored_frameworks = 'Debug/Chromium Embedded Framework.framework'`

NOTE that I tested escaping those spaces with backslashes and that did not help.

In the Podfile:
`pod 'CEF', :path => ".. path to where it is .."`

`$ pod install`

`OTHER_LDFLAGS = $(inherited) -ObjC -l"cef_dll_wrapper" -framework "Chromium Embedded Framework"`

## What happened instead?

`OTHER_LDFLAGS = $(inherited) -ObjC Embedded Framework -l"cef_dll_wrapper" -framework "Chromium"`

## CocoaPods Environment

`Tested with 1.4.0 Beta 2, and 1.3.1
High Sierra beta 17B46a`

AquaGeek:
Fixes #7185

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
-* Quote framework names in OTHER_LDFLAGS
+* Quote framework names in OTHER_LDFLAGS  
   [Tyler Stromberg](https://github.com/AquaGeek)
   [#7185](https://github.com/CocoaPods/CocoaPods/issues/7185)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
+* Quote framework names in OTHER_LDFLAGS
+  [Tyler Stromberg](https://github.com/AquaGeek)
+  [#7185](https://github.com/CocoaPods/CocoaPods/issues/7185)
+
 * Fix static framework archive regression from #7187  
   [Paul Beusterien](https://github.com/paulb777)
   [#7225](https://github.com/CocoaPods/CocoaPods/issues/7225)
--- a/lib/cocoapods/generator/xcconfig/xcconfig_helper.rb
+++ b/lib/cocoapods/generator/xcconfig/xcconfig_helper.rb
@@ -219,7 +219,7 @@ module Pod
           build_settings = {
             'FRAMEWORK_SEARCH_PATHS' => quote([dirname]),
           }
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
+* Quote framework names in OTHER_LDFLAGS
```
dnkoutso: nit: add two empty spaces after each entry `Quote framework names in OTHER_LDFLAGS  `

lib/cocoapods/generator/xcconfig/xcconfig_helper.rb
```diff
@@ -219,7 +219,7 @@ def self.add_framework_build_settings(framework_path, xcconfig, sandbox_root, in
           build_settings = {
             'FRAMEWORK_SEARCH_PATHS' => quote([dirname]),
           }
-          build_settings['OTHER_LDFLAGS'] = "-framework #{name}" if include_ld_flags
+          build_settings['OTHER_LDFLAGS'] = "-framework \"#{name}\"" if include_ld_flags
```
dnkoutso: look at line 250 below, does it also need the same change for libraries with spaces?

dnkoutso:
This might require cocoapods integration test changes but I can help you with that.

AquaGeek:
FWIW, Frameworks that have `DEFINES_MODULE = YES` aren't allowed to use spaces in the name:

> Warning: My Awesome Stuff is not a valid PRODUCT_NAME for use with framework targets enabling DEFINES_MODULE (name is not a valid C99 extended identifier)

Also, the way Xcode seems to be handling this (in cases where `DEFINES_MODULE = NO`) is to use backslashes: `-framework My\ Awesome\ Stuff`

lib/cocoapods/generator/xcconfig/xcconfig_helper.rb
AquaGeek: Done

dnkoutso:
Want to match Xcode's behavior? I am indifferent and can merge this as is.

As for `DEFINES_MODULE` I dont think we need to get super smart about it. Let the pod author do the right thing.

Issue #7238, opened 2017-11-24
podspec files allow unencrypted http links 


KrauseFx:
Hey CocoaPods team,

Please let me know if that's the right place to start a discussion around this. I took a look at the `CocoaPods/Specs` repo and did a quick search on the sources of the available SDKs to see if there is any potential security vulnerability for the user, so I search for:

```
"http": "http://"
```

and I got 1484 matches, however man of them are just different versions from the same SDKs.

I totally understand that this is not something that can be changed instantly and without any side effects. I believe it might make sense to not only warn the user about those SDKs when running `pod update` / `pod install`, but maybe even have the user manually opt-in on unsafe binary and source downloads. Additionally I'd think it might be a good idea to update the `pod lint` command to reject `.podspec` files that include unsafe sources.

Let me know what you think 🌈

orta:
I think we could negate a lot of the more negative aspects of using HTTP as a transport method pretty transparently (and without any additional work for library authors). On `pod trunk push`, we can generate a checksum of the HTTP resource, and store that as a part of the podspec JSON in trunk,  and then verified again on the other side. 

We do this for the [version of Swift](https://github.com/CocoaPods/Specs/blob/master/Specs/8/a/7/Moya/10.0.1/Moya.podspec.json#L26 that the library was pushed with, as that's also really tricky to verify in Podspec metadata.

KrauseFx:
Sounds like a great idea. From an outsiders perspective this seems like quite some work, and also isn't backwards compatible, meaning we only have those hashes on new releases, right?

I was thinking of updating the `pod trunk push` command to warn the library author that it's not encouraged to use `http` URLs, but still allowing them. At the same time, we can show a warning when the user runs `pod update` or `pod install` and one of the sources is unencrypted `http://`. 

This approach is also used by the new Google Chrome: By showing a `Not Secure` badge in the address bar, people **will** reach out to library authors and ask why this is not secure, and library authors will slowly move over to `https`.

Let me know if that's something that's interesting - I'm happy to jump on diving into the CocoaPods codebase (best forrest) and looking into how to best implement it 👍

orta:
Ah, so this is about just `http`? That's a bit different, I think passing a linter warning in `pod trunk push` which means you specifically have to declare you can ignore it is a good start. 

[...]

PR #7249, opened 2017-12-01
Show warning when Pod source uses unencrypted HTTP

KrauseFx:
Please let me know if this should be part of `cocoapods/cocoapods-downloader` instead. There are 2 reasons why I decided to add it to `pod_source_installer.rb` as part of `cocoapods/cocoapods` instead:

1) It seems like `cocoapods/cocoapods-downloader` was designed to not show any output, so it would add quite some extra complexity to show a warning as part of the check, without raising an exception
2) `cocoapods/cocoapods-downloader` doesn't have the knowledge of what Pod it's trying to install, just the necessary information to download, cache and store the Pod. I thought it makes sense to show the Pod name as part of the warning.

This PR addresses the first half of https://github.com/CocoaPods/CocoaPods/issues/7238, related to https://github.com/CocoaPods/CocoaPods/pull/7250

PR #7250, opened 2017-12-01
Show warning when SDK provider tries to push a version with an unencrypted HTTP source

KrauseFx:
Related to https://github.com/CocoaPods/CocoaPods/pull/7249, fixes https://github.com/CocoaPods/CocoaPods/issues/7238

I'll add tests, once I get the ok that this is the right location and the right approach. For now I used the `error` method to fail the validation. I believe that it's the right thing to do, there is no good excuse to transfer executable over non-encrypted protocols, however I do understand that this might throw off some SDK providers, and them being overwhelmed when they quickly need to push an update.

You all have more context around what the users are like, and if they have a way to work around this, just let me know what you're deciding on :+1:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Eric Amorde](https://github.com/amorde)
   [#7093](https://github.com/CocoaPods/CocoaPods/pull/7093)
 
+* Prohibit SDK providers to use non-encrypted HTTP links for SDKs  
+  [KrauseFx](https://github.com/KrauseFx)
+  [#7250](https://github.com/CocoaPods/CocoaPods/pull/7250)
+
 ##### Bug Fixes
 
 * Quote framework names in OTHER_LDFLAGS  
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Eric Amorde](https://github.com/amorde)
   [#7093](https://github.com/CocoaPods/CocoaPods/pull/7093)
 
-* Show warning when SDK provider tries to push a version with an unencrypted HTTP source
+* Show warning when SDK provider tries to push a version with an unencrypted HTTP source  
   [KrauseFx](https://github.com/KrauseFx)
   [#7250](https://github.com/CocoaPods/CocoaPods/pull/7250)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Eric Amorde](https://github.com/amorde)
   [#7093](https://github.com/CocoaPods/CocoaPods/pull/7093)
 
-* Prohibit SDK providers to use non-encrypted HTTP links for SDKs  
+* Show warning when SDK provider tries to push a version with an unencrypted HTTP source
   [KrauseFx](https://github.com/KrauseFx)
   [#7250](https://github.com/CocoaPods/CocoaPods/pull/7250)
 
--- a/lib/cocoapods/validator.rb
+++ b/lib/cocoapods/validator.rb
@@ -400,7 +400,7 @@ module Pod
       return if spec.source.nil? || spec.source[:http].nil?
       url = spec.source[:http]
       return if url.downcase.start_with?('https://')
-      warning('http', "The URL (#{url}) doesn't use the encrypted HTTPs protocol. " \
+      warning('http', "The URL (`#{url}`) doesn't use the encrypted HTTPs protocol. " \
               'It is crucial for Pods to be transferred over a secure protocol to protect your users from man-in-the-middle attacks. '\
               'This will be an error in future releases. Please update the URL to use https.')
     end
--- a/lib/cocoapods/validator.rb
+++ b/lib/cocoapods/validator.rb
@@ -402,7 +402,7 @@ module Pod
       return if url.downcase.start_with?('https://')
       warning('http', "The URL (#{url}) doesn't use the encrypted HTTPs protocol. " \
               'It is crucial for Pods to be transferred over a secure protocol to protect your users from man-in-the-middle attacks. '\
-              'Please update the URL to use https and try again.')
+              'This will be an error in future releases. Please update the URL to use https.')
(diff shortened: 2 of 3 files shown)


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

```markdown
* Prohibit SDK providers to use non-encrypted HTTP links for SDKs  
  [KrauseFx](https://github.com/KrauseFx)
```

</p>

dnkoutso:
LGTM.

KrauseFx:
Alright, thanks @dnkoutso, will update the Changelog and add a test 👍

dnkoutso:
Use `bundle exec rubocop -a` to ensure its using the rules of rubocop with the lib.

KrauseFx:
Ha, I did the rubocop command, it replaced the code style of 10,000 files (feels like). That's because the rubocop version isn't locked in the Gemfile, and I ran `bundle update`. What version do you have installed? Happy to submit a PR to the Gemfile 👍

KrauseFx:
Ready for review 👍

dnkoutso:
It looks like its locked in `Gemfile.lock`

```
rubocop (0.37.2)
```

KrauseFx:
Ah right, yah, that's why it didn't work for me, as I ran `bundle update` locally :)

orta:
OK, let's make this a warning in 1.4.0 and tell people it will be an error in 1.6.0. Warnings can be skipped, but errors cannot. So I'd rather give people some time to adjust to the change.

KrauseFx:
Makes sense, will update the PR and the tests 👍

KrauseFx:
Done :+1: Ready to be merged from my side whenever :)

dnkoutso:
@paulb777 @justinseanmartin @DanToml are you OK with this change?

lib/cocoapods/validator.rb
```diff
@@ -393,6 +394,17 @@ def validate_documentation_url(spec)
...
 
+    # Performs validations related to the `source` -> `http` attribute (if exists)
+    #
+    def validate_source_url(spec)
+      return if spec.source.nil? || spec.source[:http].nil?
+      url = spec.source[:http]
+      return if url.downcase.start_with?('https://')
+      warning('http', "The URL (#{url}) doesn't use the encrypted HTTPs protocol. " \
+              'It is crucial for Pods to be transferred over a secure protocol to protect your users from man-in-the-middle attacks. '\
+              'Please update the URL to use https and try again.')
```
paulb777: 'This will be an error in future releases. Please update the URL to use https.'

DanToml:
Yep - I have a couple of wording changes I want to make, WRT a timeline for requiring https, will review properly today

CHANGELOG.md
```diff
@@ -16,6 +16,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Eric Amorde](https://github.com/amorde)
   [#7093](https://github.com/CocoaPods/CocoaPods/pull/7093)
 
+* Prohibit SDK providers to use non-encrypted HTTP links for SDKs
```
paulb777: s/to use/from using/

DanToml:
That said, I'm ok with merging this is as is and following up in a second pr

CHANGELOG.md
justinseanmartin: Would also soften the wording of "Prohibit" to "Warn against" or something like that, but not a big deal.

justinseanmartin:
IIUC, this will generate a warning on `pod spec lint` / `pod lib lint`, but not prevent or warn someone from consuming the podspec on the other side. I assume the publisher can still bypass the warning with `--allow-warnings` as well. Is that right?

Assuming the above, this seems like a good first step. As a followup, on the consumption side we should require an explicit installation option of `install! 'cocoapods', allow_insecure_source_urls: true` to protect users from installing pods that have previously been published with `http` URLs.

KrauseFx:
So besides the text, is there anything else I should change?

KrauseFx:
Updated 👍  Thanks for the feedback everyone 👍

lib/cocoapods/validator.rb
```diff
@@ -393,6 +394,17 @@ def validate_documentation_url(spec)
       validate_url(spec.documentation_url) if spec.documentation_url
     end
 
+    # Performs validations related to the `source` -> `http` attribute (if exists)
+    #
+    def validate_source_url(spec)
+      return if spec.source.nil? || spec.source[:http].nil?
+      url = spec.source[:http]
+      return if url.downcase.start_with?('https://')
+      warning('http', "The URL (#{url}) doesn't use the encrypted HTTPs protocol. " \
```
dnkoutso: this is a nit but in most places around this file the strings are wrapped with ```, so `"The URL (`#{url}`)///`

CHANGELOG.md
```diff
@@ -16,6 +16,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Eric Amorde](https://github.com/amorde)
   [#7093](https://github.com/CocoaPods/CocoaPods/pull/7093)
 
+* Show warning when SDK provider tries to push a version with an unencrypted HTTP source
```
dnkoutso: nit, add two trailing spaces in the end of this line

dnkoutso:
Tiny nits.

CHANGELOG.md
dnkoutso: this line could also be updated to the commit message, as the word "prohibit" is not in effect here.

KrauseFx:
Best nits still :) Will address

orta:
> Assuming the above, this seems like a good first step. As a followup, on the consumption side we should require an explicit installation option of install! 'cocoapods', allow_insecure_source_urls: true to protect users from installing pods that have previously been published with http URLs.

I think this makes sense as a part of the "1.6" or whenever we declare switch it to an error. Worth making this a `TODO:` issue for it. 👍

KrauseFx:
Updated the PR, thanks @dnkoutso 👍

KrauseFx:
Mh, it doesn't seem like the Travis error was introduced with my PR?

dnkoutso:
Nope its not related to you. I need to fix this but I rekicked yours. Should go green.

PR #7289, opened 2017-12-17
Fix build

dnkoutso:


dnkoutso:
Merging since this is a test fix and has 0 changes in the actual logic of the library.

PR #7292, opened 2017-12-19
Restore `development_pod_targets` public method in installer

dnkoutso:
See comment https://github.com/CocoaPods/CocoaPods/pull/7110/files#r143264402 

Since `installer` is exposed in pre and post install hooks then we cannot easily remove public methods present as they can be used by external consumers.

This method is harmless and its convenient.

/cc @jmkk and @ob


lib/cocoapods/installer.rb
```diff
@@ -644,6 +644,19 @@ def run_podfile_post_install_hook
 
     #-------------------------------------------------------------------------#
 
+    public
+
+    # @return [Array<PodTargret>] The targets of the development pods generated by
```
paulb777: spelling - PodTarget

dnkoutso:
duh fixed typo, will merge on green.

PR #7293, opened 2017-12-19
Show warning when Pod source uses unencrypted HTTP

KrauseFx:
This PR addresses the first half of #7238, related to #7250 

Via https://github.com/CocoaPods/CocoaPods/pull/7249

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
-* None.  
+* Show warning when Pod source uses unencrypted HTTP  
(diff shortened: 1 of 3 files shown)


KrauseFx:
Let's see what the good Travis has to say

KrauseFx:
It's 💚

Coeur:
@KrauseFx is it possible to configure your Podfile or a commandline parameter (`pod update -something`) to automatically replace http sources with equivalent https sources?

KrauseFx:
@Coeur I don't think that would work well, as there is no guarantee that HTTPs is supported by the host. I think reaching out to the Pod's author is the way to go here, they should fix their issue ASAP.

PR #7318, opened 2018-01-08
Do not include test spec resource and framework paths of external targets into scripts

dnkoutso:


lib/cocoapods/target/pod_target.rb
```diff
@@ -468,10 +468,10 @@ def recursive_test_dependent_targets
       end
     end
 
-    # @return [Array<PodTarget>] the canonical list of test dependent targets this target has a dependency upon.
-    #         This includes the parent target as well as its transitive dependencies.
+    # @return [Array<PodTarget>] the canonical list of dependent targets this target has a dependency upon.
+    #         This list includes the target itself as well as its recursive dependent and test dependent targets.
     #
-    def all_test_dependent_targets
+    def all_dependent_targets
```
dnkoutso: method renamed for clarity.

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -401,9 +401,11 @@ def create_test_xcconfig_files
           #
           def create_test_target_copy_resources_script(test_type)
             path = target.copy_resources_script_path_for_test_type(test_type)
-            pod_targets = target.all_test_dependent_targets
+            pod_targets = target.all_dependent_targets
             resource_paths_by_config = target.user_build_configurations.keys.each_with_object({}) do |config, resources_by_config|
-              resources_by_config[config] = pod_targets.flat_map(&:resource_paths)
+              resources_by_config[config] = pod_targets.flat_map do |pod_target|
+                pod_target.resource_paths(pod_target == target)
```
dnkoutso: this one and below are the real fix. When we generate the `-Unit-Tests.resources.sh` script we no longer include paths of resources or vendored frameworks of dependent targets.

dnkoutso: We only care for test spec resources for our own target (self) not for others.

justinseanmartin: Might be worth storing `pod_target == target` in a variable named `include_test_dependencies` or something along those lines to make it clear as to what this represents from the callsite.

dnkoutso: done

PR #7326, opened 2018-01-12
 [Resolver] Fix multi-source resolution

### LINKED ISSUE #7031: Two different sources providing the same pod yields wrong resolved version
This is a link to https://github.com/CocoaPods/Molinillo/issues/88.

The summary is that the resolution of a pod that exists in at least two sources does not resolve properly.

It is unknown whether the bug is in the integration of CocoaPods and Molinillo 0.6.x or in Molinillo itself.

I am filing this issue in this project because an outside consumer of CocoaPods will not get the behavior expected when using CocoaPods master (or 1.4.x when it ships).

segiddins:
Closes #7031 by ensuring that specs are properly ordered by both version _and_ source preference.
Mostly took so long because I had to build out the test infrastructure to be able to reproduce the issue without live git repos.

cc @justinseanmartin

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,11 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
   [#7292](https://github.com/CocoaPods/CocoaPods/pull/7292)
 
+* Fix resolution when multiple sources provide the same pods, and there are 
+  (potential) dependencies between the sources.  
+  [Samuel Giddins](https://github.com/segiddins)
+  [#7031](https://github.com/CococaPods/CocoaPods/issues/7031)
+
 ## 1.4.0.rc.1 (2017-12-16)
 
 ##### Enhancements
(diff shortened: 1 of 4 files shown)


dnkoutso:
Let's ship this with 1.4.0.

valeriomazzeo:
Has this been fixed for `pod lib lint --sources` as well?

It doesn't seem to respect the order of sources, when `MyPod` exists in both a private repo and the master repo with the same name and similar versions.

`pod lib lint --allow-warnings --private --fail-fast --sources=git@github.com:myorg/cocoapods-specs.git,https://github.com/CocoaPods/Specs.git`

Podspec: `spec.dependency 'MyPod', '~> 1.0.0'` picks up the pod from master
Podspec: `spec.dependency 'MyPod', '1.0.0'` picks up the pod from the private repo

segiddins:
If the one in the master repo has a higher version, then that’s the intended behavior

valeriomazzeo:
I see.

What would be the use case scenario for that behaviour?

At the moment, I have a private pod which has a name collision with a public one, but I just want to use mine. I find difficult to find a case where I would want to override a public pod, in a private source, and then dynamically switch from one to the other based on the version.

In my opinion, when multiple reference are found, it would make more sense to stick to the pod available in the first source that contains a pod with that name, regardless of the version definition.

PR #7345, opened 2018-01-17
Simplify GitHub issue template

revolter:
- Replaced top quote with HTML code in case the user doesn't remove it
- Removed top instruction removal request as it's not needed anymore
- Replaced `this` with `these` so its clearer that the example on the second line must be removed too

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -25,22 +25,22 @@ Issues with `pod search`? Try deleting your cache `rm -rf ~/Library/Caches/Cocoa
 
 ## What did you do?
 
-ℹ Please replace these with what you did.  
+ℹ Please replace these two lines with what you did.  
 e.g. Run `pod install`
 
 ## What did you expect to happen?
 
-ℹ Please replace these with what you expected to happen.  
+ℹ Please replace these two lines with what you expected to happen.  
 e.g. Install all pod dependencies correctly.
 
 ## What happened instead?
 
-ℹ Please replace these with of what happened instead.  
+ℹ Please replace these two lines with of what happened instead.  
 e.g. Pod A is missing the subspec B for target C.
 
 ## CocoaPods Environment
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -1,45 +1,46 @@
...
 ## What happened instead?
 
-ℹ Please replace this with of what happened instead.  
+ℹ Please replace these with of what happened instead.  
 e.g. Pod A is missing the subspec B for target C.
 
 ## CocoaPods Environment
 
-ℹ Please replace this with the output of `pod env`.
+ℹ Please replace these with the output of `pod env`.
```
amorde: This doesn't seem to be more clear than using `this` here - any particular reason for this change?

amorde: (note: I did see your note in the PR description)

dnkoutso: +1 why change these?

dnkoutso:
Good with me if we revert back to using "this"

.github/ISSUE_TEMPLATE.md
paulb777: Perhaps "these two lines" to resolve the ambiguity of how much exactly should be replaced?

dnkoutso: yeap also "these two lines" is good also.

amorde: I agree either "these lines" or "these two lines" would be a worthwhile change

revolter: "This" is too ambiguous (this line? this section?), and I think that "these two lines" is the clearer version for anybody, even non-native english speakers.

revolter:
@dnkoutso, Let me know if the latest version is good 😄

revolter:
Thank you for the merge 😃

PR #7353, opened 2018-01-21
Unique all available pre-release versions when displaying

dnkoutso:


CHANGELOG.md
```diff
@@ -18,6 +18,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
+* Unique all available pre-release versions when displaying  
+  [Samuel Giddins](https://github.com/segiddins)
```
dnkoutso: @segiddins I added your name here since you did this change locally, I am only upstreaming it.

orta:
How do we end up with dupes in there in the first place? Different targets/platforms?

PR #7368, opened 2018-01-25
Always update input/output paths even if they are empty

dnkoutso:
Fixes a case in which a pod input/output path was added but then it does not get removed if the input/output paths are now empty.

This can also occur if someone uses `use_frameworks!` and switches out to static libs.

### DIFF
--- a/lib/cocoapods/installer/user_project_integrator/target_integrator.rb
+++ b/lib/cocoapods/installer/user_project_integrator/target_integrator.rb
@@ -71,15 +71,11 @@ module Pod
           #
           # @return [void]
           #
-          def add_embed_frameworks_script_phase_to_target(native_target, script_path, input_paths = [], output_paths = [])
+          def create_or_update_embed_frameworks_script_phase_to_target(native_target, script_path, input_paths = [], output_paths = [])
             phase = TargetIntegrator.create_or_update_build_phase(native_target, BUILD_PHASE_PREFIX + EMBED_FRAMEWORK_PHASE_NAME)
             phase.shell_script = %("#{script_path}"\n)
-            unless input_paths.empty?
-              phase.input_paths = input_paths
-            end
-            unless output_paths.empty?
-              phase.output_paths = output_paths
-            end
+            phase.input_paths = input_paths
+            phase.output_paths = output_paths
           end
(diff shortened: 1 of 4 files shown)


lib/cocoapods/installer/user_project_integrator/target_integrator.rb
```diff
@@ -71,15 +71,11 @@ class << self
           #
           # @return [void]
           #
-          def add_embed_frameworks_script_phase_to_target(native_target, script_path, input_paths = [], output_paths = [])
+          def create_or_update_embed_frameworks_script_phase_to_target(native_target, script_path, input_paths = [], output_paths = [])
             phase = TargetIntegrator.create_or_update_build_phase(native_target, BUILD_PHASE_PREFIX + EMBED_FRAMEWORK_PHASE_NAME)
             phase.shell_script = %("#{script_path}"\n)
-            unless input_paths.empty?
```
dnkoutso: just renamed method to match better what it actually does.

lib/cocoapods/installer/user_project_integrator/target_integrator.rb
```diff
@@ -111,16 +107,12 @@ def remove_embed_frameworks_script_phase_from_target(native_target)
           #
           # @return [void]
           #
-          def add_copy_resources_script_phase_to_target(native_target, script_path, input_paths = [], output_paths = [])
+          def create_or_update_copy_resources_script_phase_to_target(native_target, script_path, input_paths = [], output_paths = [])
```
dnkoutso: just renamed method to match better what it actually does.

lib/cocoapods/installer/user_project_integrator/target_integrator.rb
```diff
@@ -71,15 +71,11 @@ class << self
           #
           # @return [void]
           #
-          def add_embed_frameworks_script_phase_to_target(native_target, script_path, input_paths = [], output_paths = [])
+          def create_or_update_embed_frameworks_script_phase_to_target(native_target, script_path, input_paths = [], output_paths = [])
```
dnkoutso: just renamed method to match better what it actually does.

Issue #7362, opened 2018-01-24
Too many input, output paths cause a build failure


casatwy:
> the error says:

```
BUILD FAILED
Verify final result code for completed build operation
Build operation failed without specifying any errors. Individual build tasks may have failed for unknown reasons.
One possible cause is if there are too many (possibly zombie) processes; in this case, rebooting may fix the problem.
Some individual build task failures (up to 12) may be listed below.
```

> how to make this happen:

We put about 1700 images into this pod (in our project we do need that much...) at the path of `BLiOSAppImages/Images/*.png`, then publish this pod using `pod repo push XXXRepoHouse BLiOSAppImages.podspec --verbose --allow-warnings --use-libraries`

I am sure that this pod have nothing but images, no source code, no libs, only images. But it just can not build success to publish...

> what I have tried

1. I tried to create a new empty project, using `:path => '/pod/path'` in the podfile point to the 1700-image-pod, and run `pod install` then build this empty project, it can not work either, the build error remains the same.

2. If I just put one image into it, it works fine. Or if using cocoapods 1.2.1 it works fine, too.

3. reboot not works

4. using `kill $(ps -A -ostat,ppid | awk '/[zZ]/{print $2}')` to kill zombie proccess, and then build, not works.

> you can use this script to generate 1700 images:

```
#!/bin/bash
mkdir -p Images
for i in {1..1700}
do
    echo "cp temp.png ./Images/$i.png"
    cp temp.png ./Images/$i.png
done
```

> the pod spec is:

```
Pod::Spec.new do |s|
  s.name         = "BLiOSAppImages"
  s.version      = "535"
  s.summary      = "BLiOSAppImages."
  s.description  = 
<<-DESC
                    this is BLiOSAppImages
                   DESC
  s.homepage     = "https://xxx.xxx.com/iOS/BLiOSAppImages"
  s.license      = { :type => "MIT", :file => "FILE_LICENSE" }
  s.author             = { "xxx" => "xxx@xxx.com" }
  s.platform     = :ios, "8.0"
  s.source       = { :git => "https://xxx.xxx.com/iOS/BLiOSAppImages.git", :tag => s.version.to_s }
  s.ios.resources = ["BLiOSAppImages/Images/*.png", "BLiOSAppImages/Fonts/*.ttf"]
  s.requires_arc = true

end
```

> other information you may need

### Stack

```
   CocoaPods : 1.4.0
[...]

PR #7384, opened 2018-01-31
Clear input/output paths if they exceed an arbitrary limit

dnkoutso:
closes https://github.com/CocoaPods/CocoaPods/issues/7362

### DIFF
--- a/lib/cocoapods/installer/user_project_integrator/target_integrator.rb
+++ b/lib/cocoapods/installer/user_project_integrator/target_integrator.rb
@@ -40,6 +40,10 @@ module Pod
         #
         COPY_PODS_RESOURCES_PHASE_NAME = 'Copy Pods Resources'.freeze
 
+        # @return [Integer] the maximum number of input and output paths to use for a script phase
+        #
+        MAX_INPUT_OUTPUT_PATHS = 1000
+
         # @return [AggregateTarget] the target that should be integrated.
         #
         attr_reader :target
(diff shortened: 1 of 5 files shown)


lib/cocoapods/installer/user_project_integrator/target_integrator.rb
```diff
@@ -40,6 +40,10 @@ class TargetIntegrator
         #
         COPY_PODS_RESOURCES_PHASE_NAME = 'Copy Pods Resources'.freeze
 
+        # @return [Integer] the maximum number of input and output paths to use for a script phase
+        #
+        MAX_INPUT_OUTPUT_PATHS = 500
```
dnkoutso: magic numbers!

dnkoutso: Updated to 1000

dnkoutso:
Added couple of tests, all should be good now.

lib/cocoapods/installer/user_project_integrator/target_integrator.rb
```diff
@@ -183,6 +187,25 @@ def create_or_update_user_script_phases(script_phases, native_target)
             end
           end
 
+          # Script phases can have a limited number of input and output paths due to each one being exported to `env`.
+          # A large number can cause a build failure because of limitations in `env`. See issue
+          # https://github.com/CocoaPods/CocoaPods/issues/7362 and https://bugreport.apple.com/web/?problemID=36831832
```
amorde: Not sure if its a big deal, but that bug report link is only accessible to the account that created it so it isn't super helpful for others. Is there a radar filed that can be linked to on http://www.openradar.me ?

spec/unit/installer/user_project_integrator/target_integrator_spec.rb
```diff
@@ -299,6 +299,23 @@ module Pod
           phase.output_paths.should.be.empty
         end
 
+        it 'clears input and output paths from script phase if it exceeds limit' do
+          # The paths represented here will be 500 for input paths and 500 for output paths which will exceed the limit.
```
segiddins: it's actually 501 for each, which is why it's exceeding the threshold. If it was exactly 500 + 500, we wouldn't be clearing the list

dnkoutso: yes and to say that I had a comment because of the `.sh` script being part of these paths. Will amend comment.

dnkoutso: well I mean that its actually 502 because of the `.sh` file

dnkoutso: updated

lib/cocoapods/installer/user_project_integrator/target_integrator.rb
dnkoutso: good point

PR #7408, opened 2018-02-05
Preserve external sources on locked dependencies in the resolver

segiddins:
This PR preserves the `external_source` attribute on dependencies generated by the locking dependency analyzer, which ensures the resolver won't complain about pre-release versions for "locked" development pods.

cc @justinseanmartin


justinseanmartin:
We talked about adding specs for this, but it is not something that is easily testable and happens due to arbitrary ordering in the resolver.

PR #7409, opened 2018-02-05
Mixed language static frameworks

paulb777:
@segiddins - I started trying to make static frameworks work on top of the dani_static_swift branch.

Does this approach make sense? Or should I wait? Is there a good sample static library project I can look at as a reference?

The `copy generated compatibility header` Build phase is added as the last build phase of the static framework build, but for mixed languages I need the module maps and umbrella headers in place before the library build starts.

Even if I manually move the `copy generated compatibility header` Build phase to the top, the module map still does not get found.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -47,7 +47,7 @@ group :development do
 
   # Integration tests
   gem 'diffy'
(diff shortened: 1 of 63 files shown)


segiddins:
I guess this makes sense, if it works?

paulb777:
Replaced by #7519

Issue #7011, opened 2017-09-05
Option to set HEADER_SEARCH_PATHS to just "${PODS_ROOT}/Headers/Public"?


fabb:
This is more a question than a bug report.

Currently, CocoaPods generates a `HEADER_SEARCH_PATHS` that looks like this:
```
HEADER_SEARCH_PATHS = $(inherited) "${PODS_ROOT}/Headers/Public" "${PODS_ROOT}/Headers/Public/Pod1" "${PODS_ROOT}/Headers/Public/Pod2"
```

This allows to write imports in mixed style:
```
#import "Pod2.h"
#import "Pod2/Pod2.h"
#import <Pod2/Pod2.h>
```

The Swift compiler actually has a problem with that, as there currently is a bug when mixing imports of the same file:
https://miqu.me/blog/2017/06/16/fixing-autocompletion-on-mixed-objective-c-and-swift-projects/
TL;DR of that blog post: the Swift compiler does not notice that the same header was imported twice when importing it using mixed styles in the bridging header (or recursively in its imported headers), the result is that it sees duplicate symbols which breaks autocomplete.

I guess if CocoaPods would set `HEADER_SEARCH_PATHS = $(inherited) "${PODS_ROOT}/Headers/Public"`, the client project would be forced to use the folder name in the import, and the Swift bug could be avoided.

**So finally to the question: Is there an easy way to configure CocoPods via the Podfile to just set `HEADER_SEARCH_PATHS = $(inherited) "${PODS_ROOT}/Headers/Public"`?**

fabb:
https://bugs.swift.org/browse/SR-5842

dnkoutso:
See this PR https://github.com/CocoaPods/CocoaPods/pull/6743

Potentially planning to update the header search paths for 1.5.x maybe 1.4.x but its a big change that could break a lot of integrations today so it needs to be phased out.

dnkoutso:
Stay.

dnkoutso:
Fixed by https://github.com/CocoaPods/CocoaPods/pull/7412

PR #7412, opened 2018-02-06
Add support for modular header search paths, include "legacy" support

dnkoutso:
- Assuming `PodA`, `PodB` and `PodC` in a project.
- Assuming `PodA` depends only on `PodB` but **not** on `PodC`

Header search paths for `PodA`:

_before_ (<= 1.4, a.k.a today)
```
${PODS_ROOT}/Headers/Private
${PODS_ROOT}/Headers/Public
${PODS_ROOT}/Headers/Private/PodA
${PODS_ROOT}/Headers/Public/PodA
${PODS_ROOT}/Headers/Public/PodB
${PODS_ROOT}/Headers/Public/PodC
```

Notice that root `Public` folder but also that `PodC` is included in the pod's header search paths despite the fact that `PodA` does not depend on `PodC`.

_after_ (> 1.5)
```
${PODS_ROOT}/Headers/Private/PodA
${PODS_ROOT}/Headers/Public/PodA
${PODS_ROOT}/Headers/Public/PodA/PodA
${PODS_ROOT}/Headers/Public/PodB
${PODS_ROOT}/Headers/Public/PodB/PodB
```

Notice that `PodC` is correctly not included (due to https://github.com/CocoaPods/CocoaPods/pull/7116) but also that the root folder `Public` is no longer present which would have otherwise allowed consumers to include headers of `PodC` without directly depending on it in their podspec.

Instead, the root `Public` folder is "expanded" so users can continue to use `<>` and `""` header imports. Functionally, this is still the same and it shouldn't break anyone upgrading to 1.5.0 but it also makes it more clear by removing the root `Public` folder.

When using/enabling modular headers there is no "before" state because we never had modular header support in CocoaPods, but `PodA` header search paths look like this (using static libraries):

```
${PODS_ROOT}/Headers/Private/PodA
${PODS_ROOT}/Headers/Public/PodA
${PODS_ROOT}/Headers/Public/PodB
```

As for the file system all public headers are now nested an additional folder further so it looks like this:

Public header symlinks on the file system (notice the extra nested folder):
```
Pods/Headers/Public/PodA/PodA/PodA.h
Pods/Headers/Public/PodB/PodB/PodB.h
Pods/Headers/Public/PodC/PodC/PodC.h
```
Private header symlinks on the file system (notice no extra folder as all imports can work with `""`):
```
Pods/Headers/Private/PodA/PodA_Private.h
Pods/Headers/Private/PodB/PodB_Private.h
Pods/Headers/Private/PodC/PodC_Private.h
```

This means that `PodA` can now use `<>` imports of `PodB` but *not* use `""` imports for `PodB` private headers (unless maybe `USE_HEADERMAPS` is set to `YES`?).

The above only applies `HEADER_SEARCH_PATHS` which get populated when static libraries are used. When frameworks are used then `FRAMEWORK_SEARCH_PATHS` are used which continue to work as they are today.

Note that CocoaPods does not also change or interfere with the `USE_HEADERMAPS` flag, which can cause plenty of dark magic to find headers of targets that are otherwise not part of the header search paths.

closes https://github.com/CocoaPods/CocoaPods/issues/7011

### DIFF
--- a/lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
+++ b/lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
@@ -163,11 +163,11 @@ module Pod
 
         private
 
-        # Add build settings, which ensure that the pod targets can be imported
-        # from the integrating target by all sort of imports, which are:
-        #  - `#import <…>`
-        #  - `#import "…"`
-        #  - `@import …;` / `import …`
+        # Add build settings, which ensure that the pod targets can be imported from the integrating target.
+        # For >= 1.5.0 we use modular (stricter) header search paths this means that the integrated target will only be
+        # able to import public headers using `<>` or `@import` notation, but never import any private headers.
+        #
+        # For < 1.5.0 legacy header search paths the same rules apply: It's the wild west.
         #
         def generate_settings_to_import_pod_targets
           @xcconfig.merge! XCConfigHelper.search_paths_for_dependent_targets(target, pod_targets)
@@ -176,7 +176,6 @@ module Pod
             generator = AggregateXCConfig.new(search_paths_target, configuration_name)
             @xcconfig.merge! XCConfigHelper.search_paths_for_dependent_targets(nil, search_paths_target.pod_targets)
             @xcconfig.merge!(generator.settings_to_import_pod_targets)
-
             # Propagate any HEADER_SEARCH_PATHS settings from the search paths.
             XCConfigHelper.propagate_header_search_paths_from_search_paths(search_paths_target, @xcconfig)
           end
--- a/lib/cocoapods/installer/xcode/pods_project_generator/file_references_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/file_references_installer.rb
@@ -312,6 +312,7 @@ module Pod
           #
           def header_mappings(headers_sandbox, file_accessor, headers, visibility_scope = :private)
             consumer = file_accessor.spec_consumer
+            header_mappings_dir = consumer.header_mappings_dir
             dir = headers_sandbox
             dir += headers_sandbox if visibility_scope == :public
             dir += consumer.header_dir if consumer.header_dir
@@ -319,9 +320,8 @@ module Pod
             mappings = {}
             headers.each do |header|
               sub_dir = dir
-              if consumer.header_mappings_dir
-                header_mappings_dir = file_accessor.path_list.root + consumer.header_mappings_dir
-                relative_path = header.relative_path_from(header_mappings_dir)
+              if header_mappings_dir
+                relative_path = header.relative_path_from(file_accessor.path_list.root + header_mappings_dir)
                 sub_dir += relative_path.dirname
               end
               mappings[sub_dir] ||= []
--- a/lib/cocoapods/installer/xcode/pods_project_generator/file_references_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/file_references_installer.rb
@@ -147,11 +147,15 @@ module Pod
                     pod_target.build_headers.add_search_path(headers_sandbox, pod_target.platform)
                     sandbox.public_headers.add_search_path(headers_sandbox, pod_target.platform)
 
+                    # Private headers will always end up in Pods/Headers/Private/PodA/*.h
+                    # This will allow for `""` imports to work.
                     header_mappings(headers_sandbox, file_accessor, file_accessor.headers).each do |namespaced_path, files|
                       pod_target.build_headers.add_files(namespaced_path, files.reject { |f| f.to_path =~ framework_exp })
                     end
 
-                    header_mappings(headers_sandbox, file_accessor, file_accessor.public_headers).each do |namespaced_path, files|
+                    # Public headers on the other hand will be added in Pods/Headers/Public/PodA/PodA/*.h
+                    # The extra folder is intentional in order for `<>` imports to work.
+                    header_mappings(headers_sandbox, file_accessor, file_accessor.public_headers, :public).each do |namespaced_path, files|
                       sandbox.public_headers.add_files(namespaced_path, files.reject { |f| f.to_path =~ framework_exp })
                     end
                   end
@@ -295,6 +299,10 @@ module Pod
           #         The consumer file accessor for which the headers need to be
           #         linked.
           #
+          # @param  [Symbol] visibility_scope
+          #         The visibility scope to produce header mappings for. If set to :public then the headers
+          #         are nested an additional level deep. For example, 'Pods/Headers/Public/PodA/PodA'.
+          #
           # @param  [Array<Pathname>] headers
           #         The absolute paths of the headers which need to be mapped.
           #
@@ -302,9 +310,10 @@ module Pod
           #         headers folders as the keys and the absolute paths of the
           #         header files as the values.
           #
-          def header_mappings(headers_sandbox, file_accessor, headers)
+          def header_mappings(headers_sandbox, file_accessor, headers, visibility_scope = :private)
             consumer = file_accessor.spec_consumer
             dir = headers_sandbox
+            dir += headers_sandbox if visibility_scope == :public
             dir += consumer.header_dir if consumer.header_dir
 
             mappings = {}
--- a/lib/cocoapods/sandbox/headers_store.rb
+++ b/lib/cocoapods/sandbox/headers_store.rb
@@ -4,7 +4,7 @@ module Pod
     # the header search paths.
     #
     class HeadersStore
-      SEARCH_PATHS_KEY = Struct.new(:platform_name, :target_name)
+      SEARCH_PATHS_KEY = Struct.new(:platform_name, :target_name, :use_modular_headers)
 
       # @return [Pathname] the absolute path of this header directory.
       #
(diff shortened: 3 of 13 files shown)


paulb777:
This looks great! How will a podspec choose to "legacy" or "new" functionality?

dnkoutso:
@paulb777 see this comment https://github.com/CocoaPods/CocoaPods/pull/6966/files#r165800460

paulb777:
Isn't that comment only about Swift static libraries?  My understanding is that this PR making the imports more restrictive will break Pods that were implicitly depending on the incorrectly exposed headers - like using quotes instead of brackets or importing a private header.

lib/cocoapods/installer/xcode/pods_project_generator/file_references_installer.rb
```diff
@@ -295,16 +299,21 @@ def common_path(paths)
...
+          #
           # @param  [Array<Pathname>] headers
           #         The absolute paths of the headers which need to be mapped.
           #
           # @return [Hash{Pathname => Array<Pathname>}] A hash containing the
           #         headers folders as the keys and the absolute paths of the
           #         header files as the values.
           #
-          def header_mappings(headers_sandbox, file_accessor, headers)
+          def header_mappings(headers_sandbox, file_accessor, headers, visibility_scope = :private)
```
dnkoutso: I am thinking on removing `visibilty_scope` from this method and instead pass in the initializer. Some header stores are explicitly created as private stores but yet this method talks about a `public` header scope.

dnkoutso:
@paulb777 it is mostly tied to Swift static libs but not 100%, the `defines_module?` method has three rules:

1) all pods with Swift must use a module (static lib or not when #6966 lands)
2) If requires frameworks is on then true
3) If the pod author had specified `DEFINES_MODULE` => 'YES'` in their config.

I think "turning this on by default" might break the world. Any additional thoughts or ideas around this? 

We could potentially offer a `Podfile` flag to enable modules everywhere

paulb777:
I'm worried about (2). I suspect we have pure Objective C podspecs that are sloppy with `#import`'s  that would break. I would be fine to fix them, but can't speak for the world.

lib/cocoapods/installer/xcode/pods_project_generator/file_references_installer.rb
dnkoutso: done.

segiddins:
@paulb777 pods built as frameworks already dont get the header search paths, since they find headers via the `FRAMEWORK_SEARCH_PATHS`

lib/cocoapods/sandbox/headers_store.rb
```diff
@@ -22,36 +22,48 @@ def root
...
           matches_platform = entry[:platform] == platform.name
           matches_target = target_name.nil? || (entry[:path].basename.to_s == target_name)
           matches_platform && matches_target
         end
         headers_dir = root.relative_path_from(sandbox.root).dirname
-        @search_paths_cache[key] = ["${PODS_ROOT}/#{headers_dir}/#{@relative_path}"] + platform_search_paths.uniq.map { |entry| "${PODS_ROOT}/#{headers_dir}/#{entry[:path]}" }
+        @search_paths_cache[key] = search_paths.uniq.flat_map do |entry|
+          path = "${PODS_ROOT}/#{headers_dir}/#{entry[:path]}"
+          paths = [path]
+          paths.push("#{path}/#{entry[:path].basename}") if !use_modular_headers && @visibility_scope == :public
```
segiddins: we may also need to keep doing this for podspecs that define a header mappings dir, since they might be using something like j2objc that will expect things like `com/company/foo/bar/header.h` to be directly importable, without the module name prefix

lib/cocoapods/installer/xcode/pods_project_generator/file_references_installer.rb
dnkoutso: actually this comment is on the wrong file, I meant `header_store.rb` and yes I changed it there. This is fine to keep here.

segiddins:
Updated integration specs, also fixed compatibility with pods that define a `header_mappings_dir`

lib/cocoapods/target/pod_target.rb
```diff
@@ -195,6 +195,19 @@ def uses_swift?
...
+    #         (and thus will need a module map and umbrella header).
+    #
+    # @note   Static library targets can temporarily opt in to this behavior by setting
+    #         `DEFINES_MODULE = YES` in their specification's `pod_target_xcconfig`.
+    #
+    def defines_module?
+      return @defines_module if defined?(@defines_module)
+      return @defines_module = true if uses_swift? || requires_frameworks?
+
+      @defines_module = non_test_specs.any? { |s| s.consumer(platform).pod_target_xcconfig['DEFINES_MODULE'] == 'YES' }
```
segiddins: we need to decide if we want to hoist this setting up into the Podfile, instead of leaving it in the podspec

dnkoutso:
LGTM, thanks for the fix, do we want to merge this and continue discussion of `defines_module?` in https://github.com/CocoaPods/CocoaPods/pull/6966?

lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
```diff
@@ -163,11 +163,11 @@ def settings_to_import_pod_targets
...
-        # Add build settings, which ensure that the pod targets can be imported
-        # from the integrating target by all sort of imports, which are:
-        #  - `#import <…>`
-        #  - `#import "…"`
-        #  - `@import …;` / `import …`
+        # Add build settings, which ensure that the pod targets can be imported from the integrating target.
+        # For >= 1.5.0 we use modular (stricter) header search paths this means that the integrated target will only be
+        # able to import public headers using `<>` or `@import` notation, but never import any private headers.
+        #
+        # For < 1.5.0 legacy header search paths the same rules apply: Its the wild west.
```
paulb777: Its -> It's

spec/unit/generator/xcconfig/pod_xcconfig_spec.rb
```diff
@@ -133,13 +137,6 @@ module XCConfig
             @xcconfig.to_hash['PODS_TARGET_SRCROOT'].should == '${PODS_ROOT}/../../spec/fixtures/banana-lib'
           end
 
-          it 'adds the library build headers and public headers search paths to the xcconfig, with quotes' do
```
paulb777: Can/should this be converted to a negative test?

dnkoutso: I like it, I may have toyed with the idea but ultimately decided to remove. Will add back.

lib/cocoapods/target/pod_target.rb
```diff
@@ -195,6 +195,20 @@ def uses_swift?
...
+    #
+    # @note   Static library targets can temporarily opt in to this behavior by setting
+    #         `DEFINES_MODULE = YES` in their specification's `pod_target_xcconfig`.
+    #
+    def defines_module?
+      return @defines_module if defined?(@defines_module)
+      return @defines_module = true if uses_swift? || requires_frameworks?
+      return @defines_module = true if target_definitions.any? { |td| td.build_pod_as_module?(pod_name) }
+
+      @defines_module = non_test_specs.any? { |s| s.consumer(platform).pod_target_xcconfig['DEFINES_MODULE'] == 'YES' }
```
dnkoutso: we decided to keep this for now (and maybe this PR)

segiddins:
Updated, ready for final review!

dnkoutso:
👍 👍

efirestone:
A few possible gotchas I see:
1. By removing `${PODS_ROOT}/Headers/Public/` it means that `PodA` can no longer import `PodC`. This is correct, but has the potential to break existing pods, unless legacy support also brings back that import.
2. You make mention of using `""` to import files like `"PodA.h"`, but I don't believe this will work unless Pods is also populating the `USER_HEADER_SEARCH_PATHS` along with the standard `HEADER_SEARCH_PATHS`. By including the nested path (like `${PODS_ROOT}/Headers/Public/PodA/PodA`) you'll keep the ability to import things like `<PodA.h>`, but I don't believe `"PodA.h"` will work. This isn't a change from existing behavior, but I wanted to clarify based on your comments.
3. We will eventually want to remove the nested folders (like `${PODS_ROOT}/Headers/Public/PodA/PodA`) as they're a bad practice and can easily lead to conflicts. For example, if two dependencies have a `NSString+Additions.h` file, then which one gets imported with `<NSString+Additions.h>` is dependent on the order of the search paths. Instead we should require an explicit `<PodA/NSString+Additions.h>`.

segiddins:
> By removing ${PODS_ROOT}/Headers/Public/ it means that PodA can no longer import PodC. This is correct, but has the potential to break existing pods, unless legacy support also brings back that import.

Legacy support retains the ability to import everything

> You make mention of using "" to import files like "PodA.h", but I don't believe this will work unless Pods is also populating the USER_HEADER_SEARCH_PATHS along with the standard HEADER_SEARCH_PATHS. By including the nested path (like ${PODS_ROOT}/Headers/Public/PodA/PodA) you'll keep the ability to import things like <PodA.h>, but I don't believe "PodA.h" will work. This isn't a change from existing behavior, but I wanted to clarify based on your comments.

Correct

> We will eventually want to remove the nested folders (like ${PODS_ROOT}/Headers/Public/PodA/PodA) as they're a bad practice and can easily lead to conflicts. For example, if two dependencies have a NSString+Additions.h file, then which one gets imported with <NSString+Additions.h> is dependent on the order of the search paths. Instead we should require an explicit <PodA/NSString+Additions.h>.

The fully namespaced import is required when "modular_headers" are enabled for a pod

PR #7414, opened 2018-02-07
Store the Podfile checksum in a separate file in the sandbox

segiddins:
* The Podfile checksum is stored in a separate file in the Pods directory
  instead of in the Podfile.lock, allowing concurrent modification of the
  Podfile from multiple branches without causing merge conflicts.

The checksum is still verified in the “Check Manifest.lock” build script phase by running `shasum`.

### DIFF
--- a/lib/cocoapods/installer/user_project_integrator/target_integrator.rb
+++ b/lib/cocoapods/installer/user_project_integrator/target_integrator.rb
@@ -380,16 +380,29 @@ module Pod
             phase = TargetIntegrator.create_or_update_build_phase(native_target, BUILD_PHASE_PREFIX + phase_name)
             native_target.build_phases.unshift(phase).uniq! unless native_target.build_phases.first == phase
             phase.shell_script = <<-SH.strip_heredoc
-              diff "${PODS_PODFILE_DIR_PATH}/Podfile.lock" "${PODS_ROOT}/Manifest.lock" > /dev/null
-              if [ $? != 0 ] ; then
-                  # print error to STDERR
-                  echo "error: The sandbox is not in sync with the Podfile.lock. Run 'pod install' or update your CocoaPods installation." >&2
-                  exit 1
+              set -e
+              set -u
+              set -o pipefail
+
+              fail() {
+                # print error to STDERR
+                echo "error: The sandbox is not in sync with the Podfile.lock. Run 'pod install' or update your CocoaPods installation." $@ >&2
+                exit 1
+              }
+
+              diff -q "${PODS_PODFILE_DIR_PATH}/Podfile.lock" "${PODS_ROOT}/Manifest.lock" > /dev/null || fail "The manifest in the sandbox differs from your lockfile."
+
+              if [ -f "${PODS_ROOT}/Podfile.sha1" ]; then
+                (cd "${PODS_PODFILE_DIR_PATH}" && shasum --algorithm 1 --status --check "${PODS_ROOT}/Podfile.sha1") || fail "Your Podfile has been changed since the last time you ran 'pod install'."
               fi
+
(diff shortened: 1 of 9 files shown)


lib/cocoapods/installer/user_project_integrator/target_integrator.rb
```diff
@@ -380,16 +380,29 @@ def add_check_manifest_lock_script_phase
             phase = TargetIntegrator.create_or_update_build_phase(native_target, BUILD_PHASE_PREFIX + phase_name)
             native_target.build_phases.unshift(phase).uniq! unless native_target.build_phases.first == phase
             phase.shell_script = <<-SH.strip_heredoc
-              diff "${PODS_PODFILE_DIR_PATH}/Podfile.lock" "${PODS_ROOT}/Manifest.lock" > /dev/null
-              if [ $? != 0 ] ; then
-                  # print error to STDERR
-                  echo "error: The sandbox is not in sync with the Podfile.lock. Run 'pod install' or update your CocoaPods installation." >&2
-                  exit 1
+              set -e
```
DanToml: Please can you use long names for these?

DanToml:
this is an api breaking change for the lockfile... I don't think we should merge this without strong consideration.

segiddins:
> this is an api breaking change for the lockfile... I don't think we should merge this without strong consideration.

How is it API breaking? There are no public methods that currently expose that checksum, and it is _only_ used for the check manifest lock build phase

> There are many potential cases when a Pods/ dir is not checked in, that you would want to validate if the lockfile was correct for a given Podfile.

We still do that (when running `pod install`), but the checksum in the lockfile was only used in the diff of the Podfile.lock and Manifest.lock

lib/cocoapods/installer/user_project_integrator/target_integrator.rb
segiddins: I didn't know there were long names 😬
Do you also want me to change the other scripts that @keith recently updated?

justinseanmartin:
Rather than putting it out into the installer to create the two files, what about abstracting away the creation/reading of the separate file to be within the `Podfile.lock` code itself? Then it would be opaque to anyone using the lockfile parser APIs whether it is stored in the same file or not.

@DanToml - Have you seen tools that do direct Podfile.lock/Manifest.lock parsing?

lib/cocoapods/installer/user_project_integrator/target_integrator.rb
DanToml: Doesn't need to be in the same pr, but it would be nice to do a sweep of them. In scripts it's always preferrable to use the verbose names for everything imo.

segiddins:
> Rather than putting it out into the installer to create the two files, what about abstracting away the creation/reading of the separate file to be within the Podfile.lock code itself? Then it would be opaque to anyone using the lockfile parser APIs whether it is stored in the same file or not.

There isn't currently any `Lockfile` method to retrieve the podfile checksum, other than peeking inside the internal hash

segiddins:
If you don't think this will be mergable, I think I'll have to hack this change into a plugin -- that checksum is causing a lot of (avoidable) merge conflicts for us

dnkoutso:
I am 👍  to merge @DanToml has the final say here :D

DanToml:
@segiddins Honestly I think this would be best as a plugin.

segiddins:
😞

segiddins:
Closing in favor of https://github.com/segiddins/cocoapods-amicable.

keith:
Another benefit, besides the obvious ones, is that having this as a separate plugin requires that the build phases be re-written. For us this adds ~5 seconds on our no-op pod install time.

PR #7423, opened 2018-02-09
[PrivatePodXCConfig] Public Headers are not needed to build the targe…

aahil30060:
…t itself

The build headers are always given as a superset of the public headers. To build the target itself all headers are needed typically, not only the public headers. The public headers are needed for other targets (pod targets + user targets) depending on a target.


dnkoutso:
This is handled here anyways https://github.com/CocoaPods/CocoaPods/pull/7412

dnkoutso:
This is solved by https://github.com/CocoaPods/CocoaPods/pull/7631 and requires a rebase.  Going to close for now.

PR #7442, opened 2018-02-21
Use FRAMEWORK_SEARCH_PATHS to discover headers of vendored frameworks

segiddins:
This way, vendored frameworks integrated into static library aggregate targets can, e.g., use module maps.


dnkoutso:
Does this add `HEADER_SEARCH_PATHS` if the pod integrated also has sources?

segiddins:
> Does this add HEADER_SEARCH_PATHS if the pod integrated also has sources?

Yes, it adds header search paths for the headers not in the vendored framework

segiddins:
We might need to revert this, it's absolutely decimated our `pod install` times (because we're recursing into dependent targets a lot more, merging into xcconfigs more, which calls `Shellwords.shellsplit` more. Which is very slow)

PR #7451, opened 2018-02-23
Revert "Merge pull request #7442 from CocoaPods/seg-vendored-framewor…

segiddins:
…ks-no-sandbox-headers"

This reverts commit d16f648e38fe53b7910f6f00c55dd499a2d00999, reversing
changes made to d7c0b2b36e297df9799180fef277878bbe84ca44.

Reverts https://github.com/CocoaPods/CocoaPods/pull/7442

> We might need to revert this, it's absolutely decimated our pod install times (because we're recursing into dependent targets a lot more, merging into xcconfigs more, which calls Shellwords.shellsplit more. Which is very slow)


segiddins:
This makes `pod install` bearable for us again.

Issue #7460, opened 2018-02-26
`file:///` protocol considered insecure


herzi:
* [x] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

* We have a CI server that is supposed to test our binary framework.
* We started to run `pod spec lint --verbose --fail-fast` as part of our CI process (although using a local `file:///` URI).

## What did you expect to happen?

* Testing our pod this way would work and not introduce any new warnings.

## What happened instead?

We get this warning:

```
[!] 'AATKit' uses the unencrypted http protocol to transfer the Pod. Please be sure you're in a safe network with only trusted hosts in there. Please reach out to the library author to notify them of this security issue
```

While I totally agree to a whitelisting approach with regards to the protocol, I think the `file:///` protocol should be whitelisted as well (as I think, validating a pod this way is a useful way to find issues earlier in the development process).

### Stack

```
   CocoaPods : 1.4.0
        Ruby : ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-darwin16]
    RubyGems : 2.6.11
        Host : Mac OS X 10.12.6 (16G1212)
       Xcode : 9.2 (9C40b)
         Git : git version 2.14.3 (Apple Git-98)
Ruby lib dir : /Users/herzi/.brew/Cellar/ruby/2.4.1_1/lib
Repositories : master - https://github.com/CocoaPods/Specs.git @ 2acf562ab548acc6fad756f825586d26d5b2ea72
```

### Installation Source

```
Executable Path: /Users/herzi/.brew/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.2
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
cocoapods-trunk       : 1.3.0
cocoapods-try         : 1.1.0
```

### Podspec

```
[…]
  s.source           = { :http   => "file:///path/to/the/checkout/artifacts/aatkit-2.66.0.zip",
                         :sha256 => 'a7edb679591d9713ce516a19e6657e8e4a30142fc9dddf45c452b4f17aaa7c81'
[…]
```

orta:
Agree, happy to merge a PR adding this 👍

DanToml:
[...]

PR #7462, opened 2018-02-26
Do not warn when http source uses `file:///` URI scheme

dnkoutso:
closes https://github.com/CocoaPods/CocoaPods/issues/7460

### DIFF
--- a/lib/cocoapods/installer/pod_source_installer.rb
+++ b/lib/cocoapods/installer/pod_source_installer.rb
@@ -132,9 +132,8 @@ module Pod
       #
       def verify_source_is_secure(root_spec)
         return if root_spec.source.nil? || root_spec.source[:http].nil?
-        http_source = root_spec.source[:http]
-        return if http_source.downcase.start_with?('https://')
-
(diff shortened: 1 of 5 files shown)


lib/cocoapods/installer/pod_source_installer.rb
```diff
@@ -133,8 +133,7 @@ def download_source
       def verify_source_is_secure(root_spec)
         return if root_spec.source.nil? || root_spec.source[:http].nil?
         http_source = root_spec.source[:http]
-        return if http_source.downcase.start_with?('https://')
-
+        return if http_source.downcase.start_with?('https://', 'file:///')
```
segiddins: We should parse the URI and check the .scheme property, instead of just doing a string comparison

dnkoutso: will update

dnkoutso: updated to use `URI` and check the `scheme`

dnkoutso:
le green

PR #7473, opened 2018-03-01
Improve `pod install` performance for pods with exact file paths rather than glob patterns

manuyavuz:
This PR dramatically improves `pod install` performance for pod specs using exact file path lists for file attributes rather than glob patterns.

## Changes

- `FileReferencesInstaller` was reading file system from scratch using `read_file_system` multiple times for a `PathList` object if multiple file accessors are using that same `PathList` object. This is a common case for development pods having several subspecs which are independently added into a `Podfile`.
Now, file system will be read only once for each `PathList` object in `refresh_file_accessors` method.

- When `PathList.relative_glob` method runs, all string objects in `patterns` argument are regarded as glob patterns and `File.fnmatch` runs for each `(path_list_file, pattern)` pairs, which has complexity O(m*n). 
Now, we calculate the set intersection of list of all files exist under `PathList` and list of patterns, to find exact file path matches much faster. Then, we run `File.fnmatch` for remaining glob patterns.

Here is `time` output of my `pod install` run before and after changes:

*Before :* `bundle exec pod install  51.63s user 4.46s system 99% cpu 56.524 total`
*After :* `bundle exec pod install  7.08s user 1.97s system 98% cpu 9.176 total`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
+* Improve `pod install` performance for pods with exact file paths rather than glob patterns  
+  [Muhammed Yavuz Nuzumlalı](https://github.com/manuyavuz)
+  [#7473](https://github.com/CocoaPods/CocoaPods/pull/7473)
+
 * Add support for modular header search paths, include "legacy" support.  
   [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
   [#7412](https://github.com/CocoaPods/CocoaPods/pull/7412)
--- a/lib/cocoapods/sandbox/path_list.rb
+++ b/lib/cocoapods/sandbox/path_list.rb
@@ -128,19 +128,21 @@ module Pod
         else
           full_list = files
         end
-
-        list = Array(patterns).map do |pattern|
+        patterns_array = Array(patterns)
+        exact_matches = full_list & patterns_array
+        patterns_array -= exact_matches
+        all_patterns = patterns_array.map do |pattern|
           if directory?(pattern) && dir_pattern
             pattern += '/' unless pattern.end_with?('/')
             pattern += dir_pattern
           end
-          expanded_patterns = dir_glob_equivalent_patterns(pattern)
-          full_list.select do |path|
-            expanded_patterns.any? do |p|
-              File.fnmatch(p, path, File::FNM_CASEFOLD | File::FNM_PATHNAME)
-            end
-          end
+          dir_glob_equivalent_patterns(pattern)
         end.flatten
+        list = exact_matches + full_list.select do |path|
+          all_patterns.any? do |p|
+            File.fnmatch(p, path, File::FNM_CASEFOLD | File::FNM_PATHNAME)
+          end
+        end
 
         list = list.map { |path| Pathname.new(path) }
         if exclude_patterns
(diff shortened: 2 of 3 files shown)


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

```markdown
* Improve `pod install` performance for pods with exact file paths rather than glob patterns  
```

</p>

segiddins:
Wow, I can’t wait to test this out on our project!

dnkoutso:
omg.

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
+* Improve `pod install` performance for pods with exact file paths rather than glob patterns  
+  [Muhammed Yavuz Nuzumlalı](https://github.com/manuyavuz)
+  [#7473](https://github.com/CocoaPods/CocoaPods/issues/7473)
```
dnkoutso: this should be `pull/7473`

pointrlabs: Sorry, I just copied from danger template 🤦🏻‍♂️

lib/cocoapods/sandbox/path_list.rb
```diff
@@ -128,19 +128,21 @@ def relative_glob(patterns, options = {})
         else
           full_list = files
         end
-
-        list = Array(patterns).map do |pattern|
+        patterns_array = Array(patterns)
+        exact_matches = (full_list.to_set & patterns_array.to_set).to_a
```
segiddins: I don't think we need the `to_set` and `to_a` calls

manuyavuz: Oh yeah, I wouldn't know that array has set intersection operator 👍

manuyavuz:
@segiddins did you able to try on your project?

segiddins:
> @segiddins did you able to try on your project?

I've been traveling & out of office this week, I'll try it out first thing on Monday!

segiddins:
This doesn't seem to help our `pod install` times at all :(

lib/cocoapods/sandbox/path_list.rb
```diff
@@ -128,19 +128,21 @@ def relative_glob(patterns, options = {})
...
           end
-          expanded_patterns = dir_glob_equivalent_patterns(pattern)
-          full_list.select do |path|
-            expanded_patterns.any? do |p|
-              File.fnmatch(p, path, File::FNM_CASEFOLD | File::FNM_PATHNAME)
-            end
-          end
+          dir_glob_equivalent_patterns(pattern)
         end.flatten
+        list = exact_matches + full_list.select do |path|
```
segiddins: why can probably skip the `full_list.select` if `all_patterns.empty?`

manuyavuz:
Do you use array of exact file paths for attributes such as `source_files` or `public_header_files`?

In our project, we were having a PathList object for our development pod, and it had `40005` as number of files in it. And number of patterns given to `relative_glob` method was `446`, all of which were exact file paths. So we get rid of 40005x446 `File.fnmatch` calls.

If PathList root does not contain too many files or `podspec` file does not have many exact file paths, this change won't provide any improvements.

Maybe our case is a bit extreme, but changes does not have any measurable impact on projects using glob patterns, so it's good to have it :)

Merging.

lib/cocoapods/sandbox/path_list.rb
manuyavuz: Ow yeah, will do.

dnkoutso:
We mostly use a single glob as we've standardized our file structure. This is why we do not see as much gains in performance from this. We did once try on a pod to specify exactly the files and it made it super-slow. Your PR fixes this which is great.

PR #7488, opened 2018-03-08
Fix archiving when using Swift static libraries

segiddins:


dnkoutso:
@paulb777 I think its meant to be "fix archiving when using static libraries"?

PR #7501, opened 2018-03-13
Ensure `SWIFT_VERSION` is set for test only pod targets during validation

### LINKED ISSUE #7498: SWIFT_VERSION not being propagated down to test specs in pod spec
🌈

# Report

## What did you do?

1. Write a Podspec with a test spec that uses additional Swift dependencies (example below).
2. `pod lib lint --verbose`

The Podspec is tested, and the test dependencies are built as part of that.

## What happened instead?

The test dependencies are *not* built, with the following error being reported:

Testing failed:
	The “Swift Language Version” (SWIFT_VERSION) build setting must be set to a supported value for targets which use Swift. This setting can be set in the build settings editor.

If I copy the dependencies form the test spec into the root spec, they are built appropriately (probably as part of the normal build process) and tests run as expected.

### Stack

```
   CocoaPods : 1.4.0
        Ruby : ruby 2.3.3p222 (2016-11-21 revision 56859) [universal.x86_64-darwin17]
    RubyGems : 2.5.2
        Host : Mac OS X 10.13.3 (17D47)
       Xcode : 9.3 (9Q127n)
         Git : git version 2.16.1
Ruby lib dir : /System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib
Repositories : master - https://github.com/CocoaPods/Specs.git @ 140836b675a0d67b9273dce685eb15f735b7b02d
```

### Installation Source

```
Executable Path: /usr/local/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.2
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
cocoapods-trunk       : 1.3.0
cocoapods-try         : 1.1.0
```

I'm still getting the hang of CocoaPods, but I think I set this up right. The [spec in this repo](https://github.com/paxswill/SubSpecSwift) fails to build Quick and Nimble as part of `pod lib lint`. Unfortunately, the generated workspace doesn't seem to include an actual test target, but the verbose output includes the error. I also have another project (not currently uploaded anywhere) that exposes the error in the generated workspace, but I can upload it if you like.

dnkoutso:
closes #7498

### DIFF
--- a/lib/cocoapods/validator.rb
+++ b/lib/cocoapods/validator.rb
@@ -506,13 +506,10 @@ module Pod
     end
 
     def configure_pod_targets(targets, deployment_target)
+      test_only_pod_targets = @installer.pod_targets - targets.flat_map(&:pod_targets)
       targets.each do |target|
         target.pod_targets.each do |pod_target|
-          next unless (native_target = pod_target.native_target)
-          native_target.build_configuration_list.build_configurations.each do |build_configuration|
-            (build_configuration.build_settings['OTHER_CFLAGS'] ||= '$(inherited)') << ' -Wincomplete-umbrella'
-            build_configuration.build_settings['SWIFT_VERSION'] = (pod_target.swift_version || swift_version) if pod_target.uses_swift?
-          end
+          update_pod_target_build_settings(pod_target)
           if pod_target.uses_swift?
             pod_target.test_native_targets.each do |test_native_target|
               test_native_target.build_configuration_list.build_configurations.each do |build_configuration|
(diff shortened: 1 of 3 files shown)


spec/unit/validator_spec.rb
```diff
@@ -1068,6 +1068,26 @@ def test_swiftpod_with_dot_swift_version(version = '3.1.0')
...
+          validator = test_swiftpod
+          consumer = stub(:platform_name => 'iOS')
+          validator.instance_variable_set(:@consumer, consumer)
+          debug_configuration_one = stub(:build_settings => {})
+          debug_configuration_two = stub(:build_settings => {})
+          native_target_one = stub(:build_configuration_list => stub(:build_configurations => [debug_configuration_one]))
+          native_target_two = stub(:build_configuration_list => stub(:build_configurations => [debug_configuration_two]))
+          pod_target_one = stub(:uses_swift? => true, :swift_version => '4.0', :native_target => native_target_one, :test_native_targets => [])
+          pod_target_two = stub(:uses_swift? => true, :swift_version => '3.2', :native_target => native_target_two, :test_native_targets => [])
+          aggregate_target = stub(:pod_targets => [pod_target_one])
```
dnkoutso: aggregate target here is missing `pod_target_two` but we should still expect its version to be set to `3.2`

verified without my change this test goes red with `nil` set as the value but with my change it correctly gets set to `3.2`

paulz:
:thumbsup

lib/cocoapods/validator.rb
```diff
@@ -506,13 +506,11 @@ def install_pod
     end
 
     def configure_pod_targets(targets, deployment_target)
+      test_only_pod_targets = @installer.pod_targets.dup
```
segiddins: `test_only_pod_targets = @installer.pod_targets - targets.flat_map(&:pod_target)` ?

dnkoutso: sure

dnkoutso:
Will merge on green

Issue #7511, opened 2018-03-15
AppIcon cannot show in the Simulator Client


wangjindong:
![20180315093552040](https://user-images.githubusercontent.com/4929805/37439717-56d3feea-2834-11e8-90e0-e8e9317f6db4.png)

![20180315093616548](https://user-images.githubusercontent.com/4929805/37439724-60aa252a-2834-11e8-9172-12419692b331.png)

xcode：9.2
cocoapods：1.4.0

I have a resource project，it has only Images.xcassets。
But it causes appicon not to be displayed。
It's no problem in other images.

How to solve this problem

[demo](https://github.com/wangjindong/ResourcesDemo.git)

dnkoutso:
I do not think this is a cocoapods library issue. you should be setting the app icon to your project settings in Demo.xcodeproj.

CocoaPods does not mess with app icon integration.

wangjindong:
Please run the demo，app icon  really in Demo.xcodeproj。
delete #pod 'Resources', :path => '../'。App icon can be displayed normally!

dnkoutso:
ok I can investigate.

dnkoutso:
@wangjindong first of all confirmed. Second of all your project actually exposed a caching issue which I have fixed for faster re-builds!

Third, I apologize for closing quickly, let me see what I can do for your issue, I think one of your resources basically is stomping some files on your project.

dnkoutso:
@wangjindong quick work around is to use `s.resource_bundle` instead of `s.resources` in your podspec.

wangjindong:
Sorry,It need some time to be replaced for s.resource_bundle.

Is there no better solution?

dnkoutso:
This appears to be this issue https://github.com/CocoaPods/CocoaPods/pull/7020 I think

dnkoutso:
@wangjindong will close this one as dup of #7003

wangjindong:
ok,tks！

PR #7512, opened 2018-03-15
Fix script phase output path for `.xcasset` resources

dnkoutso:
When using `resources` DSL (note: *not* `resource_bundle`) then all `.xcassets` end up within `Assets.car` file.

This was causing re-builds of a project with many resources like the one in https://github.com/CocoaPods/CocoaPods/issues/7511 to re-compile very slow.

This makes re-compilation instant.

### DIFF
--- a/lib/cocoapods/installer/user_project_integrator/target_integrator.rb
+++ b/lib/cocoapods/installer/user_project_integrator/target_integrator.rb
@@ -230,6 +230,7 @@ module Pod
             when '.xcdatamodel'       then '.mom'
             when '.xcdatamodeld'      then '.momd'
             when '.xcmappingmodel'    then '.cdm'
+            when '.xcassets'          then '.car'
             else                      input_extension
             end
           end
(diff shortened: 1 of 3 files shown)


lib/cocoapods/installer/user_project_integrator/target_integrator.rb
```diff
@@ -311,8 +312,11 @@ def add_copy_resources_script_phase
               # convert input paths to output paths according to extensions
               output_paths = resource_paths_flattened.map do |input_path|
                 base_path = '${TARGET_BUILD_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}'
-                output_extension = TargetIntegrator.output_extension_for_resource(File.extname(input_path))
-                File.join(base_path, File.basename(input_path, File.extname(input_path)) + output_extension)
+                extname = File.extname(input_path)
+                basename = File.basename(input_path)
+                basename = 'Assets' if extname == '.xcassets' # All .xcassets end up within Assets.car by default.
```
amorde: `basename = 'Assets' if extname == '.xcassets' else File.basename(input_path)`

dnkoutso: sure

PR #7524, opened 2018-03-17
Allows to skip update operation for specific pods

### LINKED ISSUE #7334: How to ignore a single pod when using 'pod update' ?
When we use pod update, cocospod will update all pod of podfile.
Also we can use 'pod update < framework1 >' to update only framework1.

However, if I don't want to update framework1，then I have to do it like this:
'pod update framework2, framework3, framework4,......'
But it's really fool, isn't it?

So, any other smart way that I can ignore a single pod when using 'pod update' ?
Brainstorming, like 'pod update -ignore framework1' ?

I've asked [here](https://github.com/CocoaPods/CocoaPods/issues/7329),  but it's closed without solving my problem.

0mega:
This PR introduces `--exclude-pods` option to `pod update` command.
It allows to skip the update operation for specific pods.
E.g.: `pod update --exclude-pods=Alamofire` will update all installed pods from the lockfile except Alamofire.

When `--exclude-pods` option is used with non installed pods, an Informative error will be raised.

Resolves #7334

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
+
+* Add `--exclude-pods` option to `pod update` to allow excluding specific pods from update  
+  [Oleksandr Kruk](https://github.com/0mega)
+  [#7334](https://github.com/CocoaPods/CocoaPods/issues/7334)
+
 * Improve `pod install` performance for pods with exact file paths rather than glob patterns  
   [Muhammed Yavuz Nuzumlalı](https://github.com/manuyavuz)
   [#7473](https://github.com/CocoaPods/CocoaPods/pull/7473)
--- a/lib/cocoapods/command/update.rb
+++ b/lib/cocoapods/command/update.rb
@@ -23,6 +23,7 @@ module Pod
         [
           ['--sources=https://github.com/artsy/Specs,master', 'The sources from which to update dependent pods. ' \
            'Multiple sources must be comma-delimited. The master repo will not be included by default with this option.'],
+          ['--exclude-pods=podName', 'Pods to exclude during update. Multiple pods must be comma-delimited.'],
         ].concat(super)
       end
 
@@ -30,6 +31,7 @@ module Pod
         @pods = argv.arguments! unless argv.arguments.empty?
 
         source_urls = argv.option('sources', '').split(',')
+        excluded_pods = argv.option('exclude-pods', '').split(',')
         unless source_urls.empty?
           source_pods = source_urls.flat_map { |url| config.sources_manager.source_with_name_or_url(url).pods }
           unless source_pods.empty?
@@ -40,6 +42,20 @@ module Pod
               @pods = source_pods unless source_pods.empty?
             end
           end
+        end
+
+        unless excluded_pods.empty?
+          @pods ||= config.lockfile.pod_names.dup
+
+          non_installed_pods = (excluded_pods - @pods)
+          unless non_installed_pods.empty?
+            pluralized_words = non_installed_pods.length > 1 ? %w(Pods are) : %w(Pod is)
+            message = "Trying to skip `#{non_installed_pods.join('`, `')}` #{pluralized_words.first} " \
+                    "which #{pluralized_words.last} not installed"
+            raise Informative, message
+          end
+
+          @pods.delete_if { |pod| excluded_pods.include?(pod) }
         end
 
         super
--- a/spec/functional/command/update_spec.rb
+++ b/spec/functional/command/update_spec.rb
@@ -15,6 +15,8 @@ module Pod
         File.open(file, 'w') do |f|
           f.puts('platform :ios')
           f.puts('pod "BananaLib", "1.0"')
+          f.puts('pod "CoconutLib", "1.0"')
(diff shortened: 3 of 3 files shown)


lib/cocoapods/command/update.rb
```diff
@@ -23,13 +23,16 @@ def self.options
         [
           ['--sources=https://github.com/artsy/Specs,master', 'The sources from which to update dependent pods. ' \
            'Multiple sources must be comma-delimited. The master repo will not be included by default with this option.'],
+           ['--skip=podName', 'Pods to skip during update. ' \
+            'Multiple sources must be comma-delimited.'],
```
dnkoutso: wrong copy pasta documentation

dnkoutso: also would prefer a different name`--exclude-pods`

lib/cocoapods/command/update.rb
```diff
@@ -23,13 +23,15 @@ def self.options
         [
           ['--sources=https://github.com/artsy/Specs,master', 'The sources from which to update dependent pods. ' \
            'Multiple sources must be comma-delimited. The master repo will not be included by default with this option.'],
+          ['--skip=podName', 'Pods to skip during update. Multiple sources must be comma-delimited.'],
```
0mega: I experimented with:
```
pod update --exclude=Alamofire
pod update --ignore=Alamofire
pod update --skip=Alamofire
```
Skip sounded like the best to communicate the intent. But happy to accept suggestions.

lib/cocoapods/command/update.rb
```diff
@@ -42,6 +44,14 @@ def initialize(argv)
           end
         end
 
+        unless skipped_pods.empty?
+          if @pods
+            @pods.delete_if { |pod| skipped_pods.include?(pod) }
```
segiddins: I think this should be an error if there are overlapping pod names

lib/cocoapods/command/update.rb
```diff
@@ -42,6 +44,14 @@ def initialize(argv)
...
+          else
+            @pods = config.lockfile.pod_names.reject { |pod| skipped_pods.include?(pod) }
```
segiddins: should probably raise if you're skipping a pod that isn't in the lockfile

lib/cocoapods/command/update.rb
0mega: Thanks for reviewing @segiddins. My understanding was that the `@pods` from lockfile would be the complete list of the installed pods so I need to remove the ones that the user does not want to update. Similar logic would apply in case user specifies the source.
Is the `@pods` variable representing something different? Or maybe some other case I'm not considering?

lib/cocoapods/command/update.rb
dnkoutso: this still says "multiple sources" which is not relevant to this command

lib/cocoapods/command/update.rb
segiddins: Sorry, I thought it was the list of pods the user had explicitly passed. Eg

`pod update A —skip=A`

lib/cocoapods/command/update.rb
0mega: 🤦‍♂️

lib/cocoapods/command/update.rb
0mega: @segiddins added and informative error in case the user is trying to skip a pod that is not installed. Let me know what you think.

0mega:
@dnkoutso, @segiddins ♻️ all comments addressed

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
+* Add `--exclude-pods` option to `pod update` to allow excluding specific pods from update
```
dnkoutso: nit of nits, add two empty spaces after the first message

dnkoutso:
LGTM

lib/cocoapods/command/update.rb
```diff
@@ -42,6 +44,14 @@ def initialize(argv)
           end
         end
 
+        unless excluded_pods.empty?
+          @pods ||= config.lockfile.pod_names.dup
+          skipping_non_installed_pods = (excluded_pods - @pods).any?
+          raise Informative, 'Trying to skip pods that are not installed' if skipping_non_installed_pods
```
dnkoutso: this message could be a bit more descriptive and provide what actually went wrong. e.g. "Attempted to exclude pods `A,B,C` which they are not currently installed.`

CHANGELOG.md
0mega: Oh I see, for the line break, will do 👍

lib/cocoapods/command/update.rb
0mega: @dnkoutso also thought about it but at the time thought I'd have to test each skipped Pod for belonging to the installed set but turns out a simple subtract of two sets will do.
I did it slightly different from the way it's done for the [non installed Pods that are being updated ](https://github.com/CocoaPods/CocoaPods/blob/master/lib/cocoapods/command/update.rb#L57) because with the message I have it's easier to reuse the message and just pluralize two words.
If you value the consistency more though, I don't mind changing it to be like the other one.

PR #7544, opened 2018-03-25
Inhibit warnings for all dependencies during validation except for the one being validated  

### LINKED ISSUE #7434: Transitive dependencies output their warnings during `pod trunk push`
🌈

* [ x ] I've read and understood the [*CONTRIBUTING guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

    pod trunk push PromiseKit.podspec

## Description of Issue

The linting step output warnings for the transitive dependencies Bolts and OMGHTTPURLRQ. These are not warnings I can control, however any warnings cause the push to fail without specifying `--allow-warnings` and I don't want to get into the habit of always specifying this flag.
 
### Stack

```
   CocoaPods : 1.4.0
        Ruby : ruby 2.3.3p222 (2016-11-21 revision 56859) [universal.x86_64-darwin17]
    RubyGems : 2.5.2
        Host : Mac OS X 10.13.3 (17D47)
       Xcode : 9.2 (9C40b)
         Git : git version 2.16.1
Ruby lib dir : /System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib
Repositories : master - https://github.com/CocoaPods/Specs.git @ 728af8ad320b2b9e5c6fd50a685f492bd49b648c
```

### Installation Source

```
Executable Path: /usr/local/Cellar/cocoapods/1.4.0/libexec/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.2
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
cocoapods-trunk       : 1.3.0
cocoapods-try         : 1.1.0
```

https://github.com/mxcl/PromiseKit

dnkoutso:
closes #7434

### DIFF
--- a/lib/cocoapods/validator.rb
+++ b/lib/cocoapods/validator.rb
@@ -808,20 +808,22 @@ module Pod
       urls     = source_urls
       Pod::Podfile.new do
         install! 'cocoapods', :deterministic_uuids => false
+        # By default inhibit warnings for all pods, except the one being validated.
+        inhibit_all_warnings!
         urls.each { |u| source(u) }
         target 'App' do
           use_frameworks!(use_frameworks)
           platform(platform_name, deployment_target)
           if local
-            pod name, :path => podspec.dirname.to_s
+            pod name, :path => podspec.dirname.to_s, :inhibit_warnings => false
           else
-            pod name, :podspec => podspec.to_s
+            pod name, :podspec => podspec.to_s, :inhibit_warnings => false
           end
           test_spec_names.each do |test_spec_name|
             if local
-              pod test_spec_name, :path => podspec.dirname.to_s
+              pod test_spec_name, :path => podspec.dirname.to_s, :inhibit_warnings => false
             else
-              pod test_spec_name, :podspec => podspec.to_s
+              pod test_spec_name, :podspec => podspec.to_s, :inhibit_warnings => false
             end
           end
         end
--- a/spec/unit/validator_spec.rb
+++ b/spec/unit/validator_spec.rb
@@ -395,6 +395,13 @@ module Pod
           (!!target_definition.uses_frameworks?).should == false
           # rubocop:enable Style/DoubleNegation
         end
+
+        it 'inhibits warnings for all pods except the one being validated' do
+          podfile = @validator.send(:podfile_from_spec, :ios, '5.0')
+          target_definition = podfile.target_definitions['App']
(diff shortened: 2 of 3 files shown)


dnkoutso:
I think will require us to figure out where the warning is coming from from parsing the `xcodebuild` output.

Additionally, adding another option to lint would make things even more complicated. From a pod author perspective I do not think if Xcode 10 comes out I should be unable to publish because of new warnings introduced to pods I cannot control.

paulb777:
I agree that new warnings from Xcode 10 shouldn't stop publishing, but as a pod author, I want to know about the warnings caused by my dependencies. However, there are other ways to see them, so it it's not a simple implementation, probably not worth it.

lib/cocoapods/validator.rb
```diff
@@ -808,20 +808,22 @@ def podfile_from_spec(platform_name, deployment_target, use_frameworks = true, t
...
         install! 'cocoapods', :deterministic_uuids => false
+        # By default inhibit warnings for all pods, except the one being validated.
+        inhibit_all_warnings!
         urls.each { |u| source(u) }
         target 'App' do
           use_frameworks!(use_frameworks)
           platform(platform_name, deployment_target)
           if local
-            pod name, :path => podspec.dirname.to_s
+            pod name, :path => podspec.dirname.to_s, :inhibit_warnings => false
```
segiddins: such duplication 😭

spec/unit/validator_spec.rb
```diff
@@ -395,6 +395,12 @@ def podspec_path(name = 'JSONKit', version = '1.4')
           (!!target_definition.uses_frameworks?).should == false
           # rubocop:enable Style/DoubleNegation
         end
+
+        it 'inhibits warnings for all pods except the one being validated' do
+          podfile = @validator.send(:podfile_from_spec, :ios, '5.0')
+          target_definition = podfile.target_definitions['App']
+          target_definition.should.not.inhibits_warnings_for_pod?('JSONKit')
```
segiddins: should also test that other pods warnings are inhibited

dnkoutso: done

dnkoutso:
@paulb777 are you ultimately OK to land this this way?

paulb777:
Yes. I'm good with this change.

Issue #7558, opened 2018-03-29
Test target with inherit! :search_paths missing framework search paths


benasher44:
* [ ] I've read and understood the [*CONTRIBUTING guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

I updated our project to use proper target inheritance; in doing this, I used `inherit! :search_paths` for the test target.

## What did you expect to happen?

I expected the project to build successfully when building for testing.

## What happened instead?

The build fails when precompiling the bridging header for the test target. If I remove `inherit! :search_paths` from the test targets, this fixes the build.

### Stack

```
   CocoaPods : 1.4.0
        Ruby : ruby 2.4.0p0 (2016-12-24 revision 57164) [x86_64-darwin16]
    RubyGems : 2.6.10
        Host : Mac OS X 10.13.3 (17D102)
       Xcode : 9.3 (9Q127n)
         Git : git version 2.15.1 (Apple Git-101)
Ruby lib dir : /Users/basher/.rvm/rubies/ruby-2.4.0/lib
Repositories : master - https://github.com/CocoaPods/Specs.git @ 0864611ad37eb0f6134278ba484f35906c2eda0b
```

### Installation Source

```
Executable Path: /Users/basher/.rvm/gems/ruby-2.4.0/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.2
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
cocoapods-trunk       : 1.3.0
cocoapods-try         : 1.1.0
```

## Project that demonstrates the issue

[TestSearchPaths.zip](https://github.com/CocoaPods/CocoaPods/files/1861092/TestSearchPaths.zip)

benasher44:
I also noticed that when I remove `inherit! :search_paths`, the pods test target xcconfigs get properly scoped under the abstract_target (i.e. they have names like Pods-AbstractTarget-TestTarget). I would only expect the scoping to go away with `inherit! :none`, but maybe this could be a fundamental misunderstanding on my part. It's been awhile 😉

All of this behavior is demonstrated in the attached sample project by opening up its workspace and building for testing (command + shift + u). I've also tried with the beta release, and I got the same result.

dnkoutso:
[...]

PR #7559, opened 2018-03-29
Only include header search paths for dependencies of aggregate targets

dnkoutso:
Does not fix https://github.com/CocoaPods/CocoaPods/issues/7558 but the issue did surface this change right here.

Aggregate targets with static library mode are blindly including header search paths for everything that matches the aggregate targets platform.

With this change, only the settings for dependent targets will be included instead.

Given:

```ruby
platform :ios, '6.0'
pod 'Reachability', '3.1.0' # will be inherited by both concrete children

target 'SampleApp' do
  target 'SampleAppTests' do
    inherit! :search_paths
    pod 'JSONKit'
  end
end

target 'SampleApp_2' do
  pod 'JSONKit'
end
```

Then `Pods-SampleApp.debug.xcconfig` should *not* include header search paths for `JSONKit` because it does not link it or depend on it. Before this change this was wrong.

### DIFF
--- a/lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
+++ b/lib/cocoapods/generator/xcconfig/aggregate_xcconfig.rb
@@ -157,7 +157,9 @@ module Pod
             build_settings
           else
             # Make headers discoverable from $PODS_ROOT/Headers directory
-            header_search_paths = target.sandbox.public_headers.search_paths(target.platform)
+            header_search_paths = pod_targets.flat_map do |pod_target|
+              target.sandbox.public_headers.search_paths(target.platform, pod_target.pod_name, false)
+            end.uniq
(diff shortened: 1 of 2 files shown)


dnkoutso:
One more issue to resolve here around duplication that is annoying. I think this should go in 1.5.0.

dnkoutso:
Changing my mind this might be OK with 1.6.0 instead.

dnkoutso:
I think this will be fixed by the build settings revamp.

dnkoutso:
Will land this when https://github.com/CocoaPods/CocoaPods/pull/7631 lands

dnkoutso:
This is actually fixed and superseded by https://github.com/CocoaPods/CocoaPods/pull/7631!

PR #7541, opened 2018-03-24
Yes, improving performance again

segiddins:
- [Analyzer] Speed up #filter_pod_targets_for_target_definition

- [Resolver] Speed up #resolver_specs_by_target

This brings the time down from 90s to 4s for us, by reducing the amount of times we perform graph traversal from something like O(x^n^n) to O(log n)

- [Sandbox] Allow storing specs directly

This also allows up to only create the `Local Podspecs` dir once

- Clean up specs to make failures more obvious

- [Sandbox] Cache “Local Podspec” specs

- [SandboxAnalyzer] Don’t treat a pod as added if it’s local

- [PodVariant] Avoid allocating arrays in #hash

- [LockingDependencyAnalyzer] Only add a dependency string to the graph once

This saves almost 2 seconds on our project

- [Analyzer] Avoid checking checksums for the same pod multiple times

- [Analyzer] Don’t analyze multiple times

- Fix storing cached specs in the sandbox

- [Installer] raise when sandbox / podfile not given

- [Analyzer] Avoid loading specs when determining deps to fetch

- Avoid “creating” the same directory over & over

- Allow resetting Config.instance within a spec

- [HeadersStore] mkpath less

- [FileReferencesInstaller] Compute groups less

- [Analyzer] Only try to print pods state when verbose

- [Installer] Add an API that just returns AggregateTargets from a sandbox

- Fix validator specs

- Fix checksums for external source pods

PR #7582, opened 2018-04-04
[Resolver] Fix performance issue with exponential growth of dependency nodes cache

jmkk:
Calling "pod install" on our project (with over 100 podspec dependencies) was very slow - profiling showed dependency resolver to be the bottleneck, with the dependency cache returning hundreds of millions specs at the end.

The issue was that dependency nodes returned from `valid_dependencies_for_target_from_node` had thousands of duplicates, resulting in exponential dependency resolution times for each additional target.

This ensures we remove duplicates when returning target appropriate nodes from `valid_dependencies_for_target_from_node`, reducing our pod install times by a factor of 2x.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,8 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
-* None.  
+* Improve performance of the dependency resolver by removing duplicates for dependency nodes.
+  [Jacek Suliga](https://github.com/jmkk)
 
 ##### Bug Fixes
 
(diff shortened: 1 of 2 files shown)


bayareabear:
Super!

segiddins:
I believe this will be addressed in https://github.com/CocoaPods/CocoaPods/pull/7541, if you want to check that out (it needs tests fixed before it can be merged)

jmkk:
Thanks @segiddins! I checked #7541 and the podinstall times for us are as fast with your fix as with this PR. Looking forward to this and other perf improvements!

PR #7588, opened 2018-04-05
[Analyser] Inspect targets to integrate even if not integrating to user project

Ruenzuo:
I believe the `swift_version` support changes broke [CocoaPods/Rome](https://github.com/CocoaPods/Rome). The generated `Pods.xcodeproj` doesn't have a value for the `SWIFT_VERSION` build setting per Pod, so it can't generate the libraries to integrate. This is probably because the value is fetched either by inspecting the user's project target or using the specified value in the Podspec, if the Pod declares one.

Running `inspect_targets_to_integrate` even when not integrating fixes this, but I do not know if this has other side effects than the ones I'm describing here.

### DIFF
--- a/lib/cocoapods/installer/analyzer.rb
+++ b/lib/cocoapods/installer/analyzer.rb
@@ -74,9 +74,8 @@ module Pod
         validate_lockfile_version!
         @result = AnalysisResult.new
         @result.podfile_dependency_cache = @podfile_dependency_cache
-        if installation_options.integrate_targets?
-          @result.target_inspections = inspect_targets_to_integrate
-        else
+        @result.target_inspections = inspect_targets_to_integrate
+        unless installation_options.integrate_targets?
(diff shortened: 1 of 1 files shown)


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* [Analyser] Inspect targets to integrate even if not integrating to user project  
  [Ruenzuo](https://github.com/Ruenzuo)
```

</p>

segiddins:
This seems to break integrating when there's no user project present, which is the main purpose of disabling integration

Ruenzuo:
It all makes sense now 😂. I guess Rome can provide a Swift version as a parameter instead, thank you!

PR #7599, opened 2018-04-07
Make modular header private header access consistent

### LINKED ISSUE #7596: Access private headers with use_modular_headers
Private headers are importable from frameworks by explicitly naming them in imports and they're available in plain static libraries via ${PODS_ROOT}/Headers/Private. There should be a way to access them if modular_headers is on without making a custom module map.

For code that accesses private headers to work both as a static library and with modular headers, I need either a comparable path or an ifdef to flip between the explicit module import and the path in ${PODS_ROOT}/Headers/Private.

paulb777:
Make modular header private header access consistent with frameworks and static libraries

Closes #7596

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Paul Beusterien](https://github.com/paulb777)
   [#7592](https://github.com/CocoaPods/CocoaPods/issues/7592)
 
+* Make modular header private header access consistent with frameworks and static libraries.  
+  [Paul Beusterien](https://github.com/paulb777)
+  [#7596](https://github.com/CocoaPods/CocoaPods/issues/7596)
+
 * Inhibit warnings for all dependencies during validation except for the one being validated  
   [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
   [#7434](https://github.com/CocoaPods/CocoaPods/issues/7434)
--- a/lib/cocoapods/target/pod_target.rb
+++ b/lib/cocoapods/target/pod_target.rb
@@ -564,7 +564,7 @@ module Pod
     #
     def header_search_paths(include_test_dependent_targets = false)
       header_search_paths = []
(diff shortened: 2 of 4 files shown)


paulb777:
Right, otherwise pods aren't portable between the app developer using `use_frameworks!`, `use_modular_headers!`, and neither.

I'll rebase and make an integration test PR on approval.

dnkoutso:
I was under the impression with modular headers we are more strict and we do not want to allow pods to import private headers of other pods?

paulb777:
I think the compatibility argument is compelling on its own.  pod suppliers shouldn't have to write different code based on what their users put in the Podfile.

Beyond that, I would argue that it is overly prescriptive for CocoaPods to mandate programming practices so strongly. A private header is already harder to access than a public header since it is not part of the module and needs to be explicitly imported.  And harder, but not impossible, seems like the right balance to me.

dnkoutso:
@paulb777 sounds good, please rebase and we can land this.

paulb777:
Rebased and integration-spec pr at https://github.com/CocoaPods/cocoapods-integration-specs/pull/158

dnkoutso:
@paulb777 merged integration PR, feel free to rebase to get this green and merged.

paulb777:
Thanks @dnkoutso!  It's green :)

dnkoutso:
@paulb777 in an internal version of me using master i am receiving errors of "Reference to "X" is ambiguous" 

This is actually specific to Protobuf pod (and a bunch of protos we use). It only works after I remove the private symlinked headers of Protobuf...

Trying to investigate further.

This is with static libs

dnkoutso:
Yes it works if I remove the `${PODS_ROOT}/Headers/Private'`entry from `HEADER_SEARCH_PATHS`

dnkoutso:
@paulb777 I believe I can, let me see what I can do and get back to you.

dnkoutso:
@paulb777 not yet ready for the sample but very quickly we also do have `GPB_USE_PROTOBUF_FRAMEWORK_IMPORTS` as turned on

PR #7602, opened 2018-04-07
Cleanup old TODOs

amorde:


lib/cocoapods/installer.rb
```diff
@@ -275,8 +275,6 @@ def validate_build_configurations
     #         regenerated from scratch and any file which might not be
     #         overwritten.
     #
-    # @todo   [#247] Clean the headers of only the pods to install.
```
segiddins: I think this TODO is still relevant?

lib/cocoapods/installer/analyzer/sandbox_analyzer.rb
```diff
@@ -43,11 +43,7 @@ class SandboxAnalyzer
         alias_method :update_mode?, :update_mode
 
         # @return [Lockfile] The lockfile of the installation as a fall-back if
-        #         there is no sandbox manifest. This is indented as a temporary
-        #         solution to prevent the full re-installation from users which
-        #         are upgrading from CP < 0.17.
-        #
-        # @todo   Remove for CP 0.18.
```
segiddins: maybe we can remove this now ;)

lib/cocoapods/installer.rb
dnkoutso: yeap still relevant. we only had a change in which we do not delete the target support files directory for pod targets because then we could keep the `xcconfig` contents and help xcode not to rebuild stuff after each `pod install` invocation.

lib/cocoapods/installer/analyzer/sandbox_analyzer.rb
amorde: Haha good point. I can do that in another PR?

lib/cocoapods/installer.rb
amorde: Ah okay, I tried to follow the issues to see if it was still relevant but they were all closed. Will revert

lib/cocoapods/installer/analyzer/sandbox_analyzer.rb
segiddins: 🎊

PR #7603, opened 2018-04-08
[Pods Project Generator] Improve performance by removing redundant processing for native target dependencies

jmkk:
When profiling slow `podinstall` times for our project, almost a third of the time was being spent in `set_target_dependencies`, with hundreds of calls to the expensive `add_dependent_targets_to_native_target` function. Debugging showed that function gets called multiple times for the same native target, which in the end results in no new dependencies getting added - but slow product name resolution meant slow execution of this task (note over 5M calls made c99ext_identifier - for a project with about 100 pods and 60 targets).

profile_before

This fix, or rather a work around, adds a cache to short-circuit and return if we've already handled adding dependencies for a given target. This speeds up `podinstall` for us by 25% (same gain observed when applied on top of #7541).

profile_after

Now, it's likely this is not the right fix. Please treat this PR more like a flag and discussion started around performance work in the Pods Project Generator. I feel `set_target_dependencies` needs some work to remove dups in the outer arrays, as even with the fix we're still iterating over 1.6M items for our project. In the meantime, we're monkey-patching with this change to get the performance boost.

### DIFF
--- a/lib/cocoapods/installer/xcode/pods_project_generator.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator.rb
@@ -312,6 +312,10 @@ module Pod
         end
 
         def add_dependent_targets_to_native_target(dependent_targets, native_target, is_app_extension, requires_frameworks, frameworks_group)
+          @processed_native_targets ||= Set.new
+          return if @processed_native_targets.include? native_target
+          @processed_native_targets << native_target
+
           dependent_targets.each do |pod_dependency_target|
             next unless pod_dependency_target.should_build?
             native_target.add_dependency(pod_dependency_target.native_target)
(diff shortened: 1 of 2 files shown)


segiddins:
I feel like the correct way to fix is this probably to figure out a way to just call the method once per native target anyways?

segiddins:
Sweet! Can’t wait to check this out on Monday!

lib/cocoapods/installer/xcode/pods_project_generator.rb
```diff
@@ -239,28 +241,39 @@ def set_target_dependencies
               aggregate_target.native_target.add_dependency(pod_target.native_target)
               configure_app_extension_api_only_for_target(pod_target) if is_app_extension
 
-              add_dependent_targets_to_native_target(pod_target.dependent_targets,
-                                                     pod_target.native_target, is_app_extension,
-                                                     pod_target.requires_frameworks? && !pod_target.static_framework?,
-                                                     frameworks_group)
+              if is_app_extension
+                pod_extension_targets << pod_target
```
dnkoutso: kind of surprised this works, I was under the impression `pod_target` does not implement equality

segiddins: That means it gets the default implementation, which is identity for equality and object id for hash value

PR #7611, opened 2018-04-10
Properly namespace Info.plist names during target installation

dnkoutso:
The generated name before was ambiguous, although it works, this just makes it easier when browsing files.

screen_shot_2018-04-09_at_10_01_51_pm


dnkoutso:
Integration specs https://github.com/CocoaPods/cocoapods-integration-specs/pull/152

PR #7646, opened 2018-04-16
Track test dependent targets per test spec instead of flat list

dnkoutso:
This is in preparation of getting each test spec to actually create a test bundle target (one-to-one map) instead of grouping them together by supported test types.

This PR can go by itself since it only changes the way we store dependencies but has no effect on how the integration works today.

It will help with figuring out which pods to to use for each test bundle that gets created instead of linking them all blindly.

### DIFF
--- a/lib/cocoapods/installer.rb
+++ b/lib/cocoapods/installer.rb
@@ -224,7 +224,9 @@ module Pod
     #
     def pod_targets
       aggregate_target_pod_targets = aggregate_targets.flat_map(&:pod_targets)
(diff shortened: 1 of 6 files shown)


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Track test dependent targets per test spec instead of flat list  
```

</p>

PR #7688, opened 2018-04-26
Do not include test dependencies input and output paths

dnkoutso:
The actual `.sh` script generated does the right thing but the input and output paths for a test target will include input and output paths from other test targets. 

This can cause cases of recompilation for test targets that have dependencies to other pods that also have test specs and resources or frameworks.


spec/unit/installer/xcode/pods_project_generator/pod_target_integrator_spec.rb
```diff
@@ -71,8 +82,31 @@ class PodTargetIntegrator
                 ]
               end
 
+              it 'excludes test framework and resource paths from dependent targets' do
```
dnkoutso: verified this goes red without my change

PR #7720, opened 2018-05-09
Allow EXPANDED_CODE_SIGN_IDENTITY to be unset

keith:
This is unset for some test target configurations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Samuel Giddins](https://github.com/segiddins)
   [#7570](https://github.com/CocoaPods/CocoaPods/pull/7570)
 
+* Allow `EXPANDED_CODE_SIGN_IDENTITY` to be unset.  
+  [Keith Smiley](https://github.com/keith)
+  [#7708](https://github.com/CocoaPods/CocoaPods/issues/7708)
 
 ## 1.5.2 (2018-05-09)
 
--- a/lib/cocoapods/generator/embed_frameworks_script.rb
+++ b/lib/cocoapods/generator/embed_frameworks_script.rb
@@ -138,7 +138,7 @@ module Pod
 
           # Signs a framework with the provided identity
           code_sign_if_enabled() {
-            if [ -n "${EXPANDED_CODE_SIGN_IDENTITY}" -a "${CODE_SIGNING_REQUIRED:-}" != "NO" -a "${CODE_SIGNING_ALLOWED}" != "NO" ]; then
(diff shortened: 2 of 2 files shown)


keith:
Hmm, this doesn't appear related:

```
Bacon::Error: ["UIKit.framework"].==(["Foundation.framework", "UIKit.framework"]) failed
	spec/unit/installer/xcode/pods_project_generator_spec.rb:184:in `block (3 levels) in <class:Xcode>': Generating Pods Project - adds system frameworks to targets```

**Comment by segiddins** [comment]
Nope, that's not related. im about to fix that

File: CHANGELOG.md [hunk 1]
```diff
@@ -40,6 +40,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 * Improve performance of grouping pods by configuration.  
   [Samuel Giddins](https://github.com/segiddins)
 
+* Allow `EXPANDED_CODE_SIGN_IDENTITY` to be unset
```
segiddins: `.  `

segiddins:
Thanks! I can deal with bumping the integration specs

keith:
I feel like you've been doing that too much for me 😛 I can update them if you'd like (although it looks like they're out of date related to that other issue). Assuming `bundle exec rake spec:rebuild_integration_fixtures` is enough?

keith:
Thanks!

lib/cocoapods/generator/embed_frameworks_script.rb
```diff
@@ -138,7 +138,7 @@ def script
 
           # Signs a framework with the provided identity
           code_sign_if_enabled() {
-            if [ -n "${EXPANDED_CODE_SIGN_IDENTITY}" -a "${CODE_SIGNING_REQUIRED:-}" != "NO" -a "${CODE_SIGNING_ALLOWED}" != "NO" ]; then
+            if [ -n "${EXPANDED_CODE_SIGN_IDENTITY:-}" -a "${CODE_SIGNING_REQUIRED:-}" != "NO" -a "${CODE_SIGNING_ALLOWED}" != "NO" ]; then
```
saagarjha: Have you considered using the one-character-shorter `"${EXPANDED_CODE_SIGN_IDENTITY-}", which only performs the expansion to the empty string if `EXPANDED_CODE_SIGN_IDENTITY` is unset (rather than the empty string or unset, which is what we have here)?

Buju77:
Is it possible to back port this simple fix (adding 2 chars `:-`) to 1.5.x and release it as 1.5.4?

Because we are having issues building our app on our CI with Xcode 10 GM + CP 1.5.3.

We just tried adding those 2 chars into the script and then our CI is working fine again. But always having to edit the script is obviously not scaleable.

We are not confident and ready to use 1.6.0 beta right now. Or somebody else provide a suitable workaround?

PR #7724, opened 2018-05-10
Fix a bug that umbrella  header import fails in a specific case

chuganzy:
In case that:

1. static library is enabled
2. modular header is enabled
3. header_dir is specified in PodSpec

the project fails to build since the umbrella header's import paths are wrong - it does not take header_dir into account so far which is still fine with Embedded Frameworks since it force flatten the header directories but not with Static libraries

Example projects which causes this issue: [ReactNative](https://github.com/facebook/react-native/blob/master/React.podspec) / [Lottie](https://github.com/airbnb/lottie-ios/blob/master/lottie-ios.podspec)

Example Podfile:

```ruby
...

use_modular_headers!
pod 'React', :path => '../node_modules/react-native', :subspecs => [
  'Core',
  'CxxBridge',
  'DevSupport',
  'RCTText',
  'RCTNetwork',
  'RCTWebSocket',
  'RCTAnimation',
]

...
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
+* Umbrella header import path gets wrong when `header_dir` is specified in PodSpec + try to build statically + Modular header is enabled  
+  [chuganzy](https://github.com/chuganzy)
+  [#7724](https://github.com/CocoaPods/CocoaPods/pull/7724)
+
 * Do not build pod target if it only contains script phases  
   [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
   [#7746](https://github.com/CocoaPods/CocoaPods/issues/7746)
--- a/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
@@ -58,13 +58,23 @@ module Pod
                   generator.headers.concat module_map_additional_headers
                 end
                 create_umbrella_header(native_target) do |generator|
-                  generator.imports += if header_mappings_dir
-                                         file_accessors.flat_map(&:public_headers).map do |pathname|
-                                           pathname.relative_path_from(header_mappings_dir)
-                                         end
-                                       else
-                                         file_accessors.flat_map(&:public_headers).map(&:basename)
+                  generator.imports += file_accessors.flat_map do |file_accessor|
+                    header_dir = if !target.requires_frameworks? && dir = file_accessor.spec_consumer.header_dir
+                                   Pathname.new(dir)
+                    end
+
+                    file_accessor.public_headers.map do |public_header|
+                      public_header = if header_mappings_dir
+                                        public_header.relative_path_from(header_mappings_dir)
+                                      else
+                                        public_header.basename
                                       end
+                      if header_dir
+                        public_header = header_dir.join(public_header)
+                      end
+                      public_header
+                    end
+                  end
                 end
               end
 
--- a/spec/unit/installer/xcode/pods_project_generator/pod_target_installer_spec.rb
+++ b/spec/unit/installer/xcode/pods_project_generator/pod_target_installer_spec.rb
@@ -261,6 +261,35 @@ module Pod
                 content.should.not =~ /"CoconutTestHeader.h"/
               end
 
+              it 'uses header_dir to umbrella header imports' do
+                @coconut_pod_target.file_accessors.first.spec_consumer.stubs(:header_dir).returns('Coconut')
+                @coconut_pod_target.stubs(:requires_frameworks?).returns(false)
+                @coconut_pod_target.stubs(:defines_module?).returns(true)
+                @installer.install!
+                content = @coconut_pod_target.umbrella_header_path.read
+                content.should =~ %r{"Coconut/Coconut.h"}
+              end
+
+              it 'uses header_dir and header_mappings_dir to umbrella header imports' do
+                @coconut_pod_target.file_accessors.first.spec_consumer.stubs(:header_dir).returns('Coconut2')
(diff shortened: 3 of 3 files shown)


dnkoutso:
Please update the PR with the following:

* Need a CHANGELOG entry and depict which issue its actually fixing.
* Add tests for your change
* Update your commit message to be more descriptive.

I am uncertain if this change is correct but the above must be corrected before and if we merge this.

chuganzy:
I couldn’t find the issue. Should I open a new one?

dnkoutso:
Nah, I thought there was one already, my bad. Would be interested to see a sample that demonstrates the problem though. Helps understand the change itself as well.

alexcurylo:
@chuganzy This sounds like it's probably the same as a problem I'm having with turning [the Lottie pod](https://github.com/airbnb/lottie-ios) into a static library with 1.5.2, "lottie-ios" vs "Lottie". Could you check if this fixes that pod as well? TIA ❤️

chuganzy:
Looks so, let me try and will share with you. Stay tuned.

dnkoutso:
This LGTM. @segiddins when you get the time can you also review?

chuganzy:
@alexcurylo I've tried and this PR solves the issue with Lottie!

CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

chuganzy:
Got a conflict to rebased.

CHANGELOG.md
```diff
@@ -20,6 +20,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
+* Umbrella header import path gets wrong when `header_dir` is specified in PodSpec + try to build statically + Modular header is enabled
```
dnkoutso: nit: 2 empty spaces at the end of this line

chuganzy: OK, I'll add it with squash!

chuganzy: Done!

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -58,13 +58,23 @@ def install!
...
                 end
                 create_umbrella_header(native_target) do |generator|
-                  generator.imports += if header_mappings_dir
-                                         file_accessors.flat_map(&:public_headers).map do |pathname|
-                                           pathname.relative_path_from(header_mappings_dir)
-                                         end
-                                       else
-                                         file_accessors.flat_map(&:public_headers).map(&:basename)
+                  generator.imports += file_accessors.flat_map do |file_accessor|
+                    header_dir = if !target.requires_frameworks? && dir = file_accessor.spec_consumer.header_dir
```
segiddins: I dont this this is correct -- can different subspecs really have a different `header_dir`?

chuganzy: I guess so, as React Native has one.
https://github.com/facebook/react-native/blob/master/React.podspec#L244

dnkoutso:
@chuganzy going to merge this, which issues does it fix that we can close? I believe there is 1 or 2.

alexcurylo:
@dnkoutso The Lottie guys opened #7751 there but saw this PR so promptly closed it!

chuganzy:
@dnkoutso Thanks for merging. I looked into again but could not find any open issue related..

JoeSzymanski:
Any chance this can be pushed up to the 1.5.3 release instead of the 1.6.0? This is the last piece I'm aware of that's blocking us from updating.

chuganzy:
Actually I’m also in the same situation and hoping so..

JoeSzymanski:
@dnkoutso Any update on when this will get released? It's been over a month now and 1.6.0 hasn't even hit beta yet. This issue is blocking my team from a few projects and we don't have a clean way to use bundler at all.

amorde:
> and we don't have a clean way to use bundler at all.

Could you describe what you mean by this? Bundler would let you point to whichever git SHA works best for your project

JoeSzymanski:
I mostly mean that, at the moment, our process doesn't allow for a clean change to start using bundler without a significant amount of work. I'm looking to find out when the next release will be going out, since it's weeks after I was told the first betas were likely to show up.

dnkoutso:
@JoeSzymanski apologies for leading you on a specific range. However, historically it appears that there is no schedule of performing releases. You can check the progress of 1.6.0 here https://github.com/CocoaPods/CocoaPods/milestone/34 which the only thing blocking a beta is wrapping up an enhancement feature (as we can fix the rest of the issues in subsequent betas).

Having said that if you cannot use Bundler easily you could also monkey patch the method in your podfile that includes a fix. Its not ideal but it would unblock your team until 1.6.0 is out.

Sorry for getting your hopes up on providing a range for 1.6.0 I should probably stop doing that.

JoeSzymanski:
Do you have any references regarding steps that would need to be taken to monkey patch via the podfile? I'm not sure where to even start with regards to getting this change in place.

dnkoutso:
`Podfile` is all Ruby so anything standard Ruby would work. Here's an example that redefines the method `dependencies` within the `Podfile` class (alter to fit your needs):
```ruby
# Until CocoaPods 1.6.0 is released we need this monkey patch in order for the project to work.
# See https://github.com/CocoaPods/CocoaPods/pull/7724
podfile_class = ::Pod::Podfile
podfile_class.send(:define_method, :dependencies) do
  # do your custom logic here
end
```

Generally ugly but given you cannot easily add Bundler it would do.

dnkoutso:
You can put that at the top of your `Podfile` or any place for that matter.

JoeSzymanski:
Thanks. I'll see if that's something I can get working.

Issue #7746, opened 2018-05-15
Adding script_phase to vendored_frameworks in Podspec causes `dyld: Symbol not found:`


vkedwardli:
* [x] I've read and understood the [*CONTRIBUTING guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

Adding a `script_phase` to a `vendored_frameworks` podspec

## What did you expect to happen?

Only the `script_phase` should be added.

## What happened instead?

1. There is a new target inside the Pods.xcodproj, with the `script_phase` added to `Build Phases` (Not harmful to me actually, just stating the difference, without `script_phase` I can't see the target for vendored framework)

2. What actual breaks:
A redundant `install_framework` function is added to `Pods-$(Target)-frameworks.sh`
```diff
   install_framework "${PODS_ROOT}/../Foo/Foo.framework"
   install_dsym "${PODS_ROOT}/../Foo/Foo.framework.dSYM"
+  install_framework "${BUILT_PRODUCTS_DIR}/Foo/Foo.framework"
```
(First 2 line is the original lines with only `vendored_frameworks` in podspec, the newline appears right after adding a `script_phase`)

Causing the contents of the vendored framework is stripped like other compiled frameworks, resulting `dyld: Symbol not found:`

### Stack

```
   CocoaPods : 1.5.2
        Ruby : ruby 2.4.2p198 (2017-09-14 revision 59899) [x86_64-darwin16]
    RubyGems : 2.6.13
        Host : Mac OS X 10.13.4 (17E199)
       Xcode : 9.3.1 (9E501)
         Git : git version 2.13.0
Ruby lib dir : /usr/local/Cellar/ruby/2.4.2/lib
Repositories : master - https://github.com/CocoaPods/Specs.git @ a6d38046fc6442739144e0ff2c0d0ab434a4ee57
```

### Installation Source

```
Executable Path: /usr/local/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.2
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
cocoapods-trunk       : 1.3.0
cocoapods-try         : 1.1.0
```

dnkoutso:
Yeap most likely an issue, do you have a sample app to reproduce ?

vkedwardli:
@dnkoutso thanks for verifying

[...]

PR #7749, opened 2018-05-15
Do not build pod target if it contains script phases

dnkoutso:
In 1.6.0 we now add a `PBXAggregateTarget` for pod targets that should not be built. This means we can integrate script phases to them instead of forcing a pod target to be actually built (`should_build?` to return `true`).

closes https://github.com/CocoaPods/CocoaPods/issues/7746

This is how it looks after this change:

screen_shot_2018-05-15_at_10_42_49_am

### DIFF
--- a/lib/cocoapods/target/pod_target.rb
+++ b/lib/cocoapods/target/pod_target.rb
@@ -146,7 +146,6 @@ module Pod
     #
     def should_build?
       return @should_build if defined? @should_build
-      return @should_build = true if contains_script_phases?
       accessors = file_accessors.reject { |fa| fa.spec.test_specification? }
       source_files = accessors.flat_map(&:source_files)
       source_files -= accessors.flat_map(&:headers)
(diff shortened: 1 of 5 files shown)


lib/cocoapods/target/pod_target.rb
```diff
@@ -146,7 +146,6 @@ def product_module_name
     #
     def should_build?
       return @should_build if defined? @should_build
-      return @should_build = true if contains_script_phases?
```
dnkoutso: this is the fix.

spec/unit/target/build_settings/aggregate_target_settings_spec.rb
```diff
@@ -587,7 +587,8 @@ def specs
 
             it 'adds values from all subspecs' do
               @consumer_b.stubs(:user_target_xcconfig).returns('OTHER_CPLUSPLUSFLAGS' => '-std=c++1y')
-              consumer_c = mock('consumer_c', :user_target_xcconfig => { 'OTHER_CPLUSPLUSFLAGS' => '-stdlib=libc++' }, :script_phases => [], :spec => mock(:test_specification? => false), :frameworks => [], :libraries => [])
+              consumer_c = mock('consumer_c', :user_target_xcconfig => { 'OTHER_CPLUSPLUSFLAGS' => '-stdlib=libc++' },
```
dnkoutso: `:script_phases => [],` was removed from the mock

Issue #7705, opened 2018-05-02
git:// sources are not encrypted


KrauseFx:
After working on https://github.com/CocoaPods/CocoaPods/pull/7250 and https://github.com/CocoaPods/CocoaPods/pull/7293 to add better protection for the user against human-in-the-middle attacks for un-encrypted network traffic, I just learned that `git://` seems to also be unencrypted by default.

Thanks to @larsxschneider for bringing this up with https://github.com/KrauseFx/krausefx.com/commit/0628e01f84ff993c578c127a350a37c8e97d8fe8#r28804197

I didn't verify that this is actually the case, but according to the docs, this is accurate.

Maybe someone here knows more about this, but if it's true, we should update CocoaPods to also warn when specifying a `git://` URL instead of `ssh://` 👍

larsxschneider:
That sounds great! However, I would recommend to use `https://` instead of `git://`. The reason: `ssh://` requires configured SSH credentials even for public repos. `https://` should work on almost any platform for public repos with less trouble.

larsxschneider:
Plus, at the last Git contributor summit @peff asked if it would make sense to deprecate the anonymous Git protocol (search "git anonymous protocol" in the [notes](https://public-inbox.org/git/alpine.DEB.2.20.1803091557510.23109@alexmv-linux/)). AFAIK there is no specifc plan to do this any time soon but the topic seems to be on some peoples radar.

KrauseFx:
Agreed, I personally always use `https://`, however I do know of developers who don't have `https` auth set up

larsxschneider:
@KrauseFx AFAIK you don't need `https` auth setup for publicly accessible repos.

PR #7752, opened 2018-05-16
Warn if 'git://' protocol is used as the source of a pod

dnkoutso:
closes https://github.com/CocoaPods/CocoaPods/issues/7705

/cc @KrauseFx @larsxschneider

### DIFF
--- a/lib/cocoapods/installer/pod_source_installer.rb
+++ b/lib/cocoapods/installer/pod_source_installer.rb
@@ -8,6 +8,8 @@ module Pod
     # @note This class needs to consider all the activated specs of a Pod.
     #
     class PodSourceInstaller
+      UNENCRYPTED_PROTOCOLS = %w(http git).freeze
+
       # @return [Sandbox] The installation target.
       #
       attr_reader :sandbox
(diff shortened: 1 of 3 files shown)


lib/cocoapods/installer/pod_source_installer.rb
```diff
@@ -110,18 +112,24 @@ def download_source
...
+      #
+      # @return [void]
       #
       def verify_source_is_secure(root_spec)
-        return if root_spec.source.nil? || root_spec.source[:http].nil?
-        http_source = URI(root_spec.source[:http])
-        return if http_source.scheme == 'https' || http_source.scheme == 'file'
-        UI.warn "'#{root_spec.name}' uses the unencrypted http protocol to transfer the Pod. " \
+        return if root_spec.source.nil? || (root_spec.source[:http].nil? && root_spec.source[:git].nil?)
+        source = if !root_spec.source[:http].nil?
```
dnkoutso: anyone with more idiomatic ruby? suggestions welcomed

lib/cocoapods/installer/pod_source_installer.rb
```diff
@@ -110,18 +112,24 @@ def download_source
...
+                   URI(root_spec.source[:http])
+                 elsif !root_spec.source[:git].nil?
+                   URI(root_spec.source[:git].to_s)
```
amorde: Why does `:git` need `to_s` but `:http` does not?

dnkoutso: @amorde good Q, for some of the tests and podspecs the `git` value resolves to a `Pathname`... I could also make the `:http` one to use `to_s` it should be harmless.

lib/cocoapods/installer/pod_source_installer.rb
```diff
@@ -110,18 +112,24 @@ def download_source
...
+        return if root_spec.source.nil? || (root_spec.source[:http].nil? && root_spec.source[:git].nil?)
+        source = if !root_spec.source[:http].nil?
+                   URI(root_spec.source[:http].to_s)
+                 elsif !root_spec.source[:git].nil?
+                   URI(root_spec.source[:git].to_s)
+                 end
+        if UNENCRYPTED_PROTOCOLS.include?(source.scheme)
+          UI.warn "'#{root_spec.name}' uses the unencrypted '#{source.scheme}' protocol to transfer the Pod. " \
                 'Please be sure you\'re in a safe network with only trusted hosts in there. ' \
                 'Please reach out to the library author to notify them of this security issue.'
```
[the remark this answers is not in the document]
dnkoutso: done

Issue #7745, opened 2018-05-15
Assets NOT recompiled because Xcode skip run `Copy Pod Resources`


Whirlwind:
CP add some `Input Files` and `Output Files` to `Copy Pod Resources` in `Build Phases`. And Xcode will not run again if the `Input Files` and `Output Files` not be changed, include their contents are not changed. It is good job.

But, there are a logic which search assets will not be run again in the `Copy Pod Resources` script:
```
  OTHER_XCASSETS=$(find "$PWD" -iname "*.xcassets" -type d)
  while read line; do
    if [[ $line != "${PODS_ROOT}*" ]]; then
      XCASSET_FILES+=("$line")
    fi
```

There is a strange logic in the CP: 
CP find all assets in my project directory, then run `actool` to compile my project's assets and Pod's assets together. `actool` will generate a `Assets.car` in `xxx.app`, and instead of origin `Assets.car` which be generated by Xcode.

eg:
```
My Projects: Assets.xcassets and IM.xcassets
Pods: ShareSDK.xcassets
```
Xcode will compile `Assets.xcassets` and `IM.xcassets` to `Assets.car` first. but it is a small file.
The Pod will recompile `Assets.xcassets`, `IM.xcassets` and `ShareSDK.xcassets` to a bigger `Assets.car` to instead of the small one.

It is OK, if there is NOT the `Input Files` and `Output Files`....

When the `Input Files` and `Output Files` not be changed, Xcode will skip the script.
If I update the `Assets.xcassets`, the Xcode will recompile the `Assets.xcassets` and `IM.xcassets` to a small `Assets.car` to overwrite the bigger which generated by Pods. But the Pod will not run the script again. So I will lose the `ShareSDK.xcassets` in my `Assets.car`.

I think the assets logic should be a Independent script.

Could you understand what I said?

Whirlwind:
The Demo is very easy: https://github.com/Whirlwind/CocoaPodsTestAssets
First, you build and run the `MyApp`, you will see a imageview in simulator. Then, please add any image to the `Assets.xcassets` in the `MyApp`. Run again, you will lost the image in the imageView!

Whirlwind:
Test by Cocoapods v1.5.2

dnkoutso:
definitely a bug. investigated will be a bit tough to fix. See more here https://github.com/CocoaPods/CocoaPods/pull/7779#discussion_r190765971

Fab1n:
[...]

Issue #6159, opened 2016-11-11
Pod install includes all xcassets from all targets


mfclarke:
> Before you start, are you using the CocoaPods beta release?
> a lot has changed with Xcode 8 that is not backwards compatible.
>
> Xcode 8 Status: Only works with CocoaPods 1.1.0
> Swift 3: You need to add a `.swift-version` file, see the CHANGELOG
> China: Use https://gems.ruby-china.org/ instead of https://gems.ruby-china.org/ for the RC see #5910

* [x] I've read and understood the [*CONTRIBUTING guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

Run pod install using CocoaPods 1.1.1. Then build a multi target project.

## What did you expect to happen?

The target being built would contain only the xcassets that are associated with the target. 

## What happened instead?

xcassets were bundled in from all targets in the project

### Stack

```
   CocoaPods : 1.1.1
        Ruby : ruby 2.2.3p173 (2015-08-18 revision 51636) [x86_64-darwin15]
    RubyGems : 2.6.1
        Host : Mac OS X 10.11.6 (15G31)
       Xcode : 7.3 (7D175)
         Git : git version 2.6.4 (Apple Git-63)
Ruby lib dir : /Users/clarkema/.rbenv/versions/2.2.3/lib
Repositories : master - https://github.com/CocoaPods/Specs.git @ 18f0f12182e1056fd893af80c31fecc4c711fd34
               master-1 - git@github.com:CocoaPods/Specs.git @ 27c05d97d019671c4bcde9f3f2f4a4731d8c0036
               redacted - internal company repos
```

### Installation Source

```
Executable Path: /Users/clarkema/.rbenv/versions/2.2.3/bin/pod
```

### Plugins

```
cocoapods-deintegrate  : 1.0.1
cocoapods-dependencies : 0.5.2
cocoapods-plugins      : 1.0.0
cocoapods-search       : 1.0.0
cocoapods-stats        : 1.0.0
cocoapods-trunk        : 1.1.1
cocoapods-try          : 1.1.0
```

### Podfile

```ruby
source 'git@github.com:CocoaPods/Specs.git'
source 'redacted - internal company repo'

platform :ios, "8.0"
inhibit_all_warnings!

def common_pods
  pod 'Reachability', '~> 3.2'
  pod 'SDWebImage', '~> 3.7'
[...]

PR #7779, opened 2018-05-24
🌈Using xcodeproj to generate .xcassets list

i-dama:
Related open issues:
https://github.com/CocoaPods/CocoaPods/issues/7745
https://github.com/CocoaPods/CocoaPods/issues/6159

The PR is not yet ready, I'd just like to get some feedback to see if the direction is good.
**The problem:**
As stated in the generated `copy_pods_resuources` script
```
# Find all other xcassets (this unfortunately includes those of path pods and other targets).
OTHER_XCASSETS=$(find "$PWD" -iname "*.xcassets" -type d)
```
The script searches for all `.xcassets` in the project folder recursively, excluding `Pods` folder afterwards. This might pick up some `.xcassets` we don't want in a certain target.

**Proposed solution**
Use the `xcodeproj` gem to extract the list of `.xcassets` to be copied for a certain target and only use those

Code that prints out paths to `.xcassets` for a given target and project:
```
require 'xcodeproj';
project_path = 'Project.xcodeproj'
target_name = 'MyTarget'

target = project.targets.find { |t| t.name == target_name }
files = target.resources_build_phase.files
xcassets = files.select { | file | file.display_name.end_with?('.xcassets') }

xcassets.each { | file | puts file.file_ref.real_path }
```

In order to do that, `copy_resources_script.rb` needs to know the location of the user project and the target we're adding the pods to.

Stuff I'd like to get feedback on:
1. Is this an acceptable way of solving this problem?
2. Is there a better way of getting a reference to the `user_project` (Currently works for `AggregatedTargetInstaller`, not sure about `PodTargetInstaller`)
3. Is there a better way of getting the user project target name?
4. The script inserts ruby code in the `.sh` script, is that acceptable? (`.xcassets` files can be added after `pod install` is ran, so the `xcassets` resolution needs to be done in the build script)

I ran the modified code on an example project, and no unwanted `.xcassets` file were included.

### DIFF
--- a/lib/cocoapods/generator/copy_resources_script.rb
+++ b/lib/cocoapods/generator/copy_resources_script.rb
@@ -6,20 +6,31 @@ module Pod
       #
       attr_reader :resources_by_config
 
-      # @return [Platform] The platform of the library for which the copy
-      #         resources script is needed.
+      # @return [Xcodeproj::Project] the user project that this target will
+      #         integrate as identified by the analyzer.
       #
+      attr_reader :user_project_path
+
+      # @return [Target] target
+      #         The library whose target needs to be generated.
+      #
+      attr_reader :target_name
       attr_reader :platform
 
       # Initialize a new instance
       #
+      # @param  [Pod::Project] project
+      #         @see project
+      #
+      # @param  [Target] target
+      #         @see target
+      #
       # @param  [Hash<String, Array<String>>] resources_by_config
       #         @see resources_by_config
       #
-      # @param  [Platform] platform
-      #         @see platform
-      #
-      def initialize(resources_by_config, platform)
+      def initialize(resources_by_config, platform, user_project_path, target_name)
+        @user_project_path = user_project_path
+        @target_name = target_name
         @resources_by_config = resources_by_config
         @platform = platform
       end
@@ -71,6 +82,28 @@ module Pod
         end
       end
 
+      # @return [String] The contents of the target xcassets script.
+      #
+      def target_xcassets
+
+        project_path = user_project_path.to_path
+        user_target_name = target_name.sub("Pods-", "")
+
+        script = <<-EOS 
+        require 'xcodeproj';
+        project_path = '#{project_path}';
+        target_name = '#{user_target_name}';
+        project = Xcodeproj::Project.open(project_path);
+
+        target = project.targets.find { |t| t.name == target_name };
+        files = target.resources_build_phase.files;
+        xcassets = files.select { | file | file.display_name.end_with?('.xcassets') };
+
+        xcassets.each { | file | puts file.file_ref.real_path };
+        EOS
+        script.delete!("\n")
+      end
+
       # @return [String] The contents of the copy resources script.
       #
       def script
@@ -89,7 +122,7 @@ module Pod
         end
 
         script += RSYNC_CALL
-        script += XCASSETS_COMPILE
+        script += xcassets_compile
         script
       end
 
@@ -199,17 +232,15 @@ if [[ "${ACTION}" == "install" ]] && [[ "${SKIP_INSTALL}" == "NO" ]]; then
 fi
 rm -f "$RESOURCES_TO_COPY"
 EOS
-
-      XCASSETS_COMPILE = <<EOS
+      
+      def xcassets_compile 
+        <<-EOS
 
 if [[ -n "${WRAPPER_EXTENSION}" ]] && [ "`xcrun --find actool`" ] && [ -n "${XCASSET_FILES:-}" ]
 then
-  # Find all other xcassets (this unfortunately includes those of path pods and other targets).
-  OTHER_XCASSETS=$(find "$PWD" -iname "*.xcassets" -type d)
+  OTHER_XCASSETS=$(ruby -e "#{target_xcassets}")
   while read line; do
-    if [[ $line != "${PODS_ROOT}*" ]]; then
-      XCASSET_FILES+=("$line")
-    fi
+    XCASSET_FILES+=("$line")
   done <<<"$OTHER_XCASSETS"
 
   if [ -z ${ASSETCATALOG_COMPILER_APPICON_NAME+x} ]; then
@@ -219,6 +250,8 @@ then
   fi
 fi
 EOS
+      end
+
     end
   end
 end
--- a/lib/cocoapods/installer/xcode/pods_project_generator/aggregate_target_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/aggregate_target_installer.rb
@@ -136,7 +136,7 @@ module Pod
           #
           def create_copy_resources_script
             path = target.copy_resources_script_path
-            generator = Generator::CopyResourcesScript.new(target.resource_paths_by_config, target.platform)
+            generator = Generator::CopyResourcesScript.new(target.resource_paths_by_config, target.platform, target.user_project.path, target.name)
             update_changed_file(generator, path)
             add_file_to_support_group(path)
           end
--- a/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
@@ -469,7 +469,7 @@ module Pod
                 pod_target.resource_paths(include_test_spec_paths)
               end
             end
-            generator = Generator::CopyResourcesScript.new(resource_paths_by_config, target.platform)
+            generator = Generator::CopyResourcesScript.new(resource_paths_by_config, target.platform, target.user_project.path, target.name)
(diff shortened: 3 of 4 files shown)


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* 🌈Using xcodeproj to generate .xcassets list  
  [i-dama](https://github.com/i-dama)
```

</p>

lib/cocoapods/generator/copy_resources_script.rb
```diff
@@ -71,6 +82,28 @@ def install_resources_function
...
 
+      # @return [String] The contents of the target xcassets script.
+      #
+      def target_xcassets
+
+        project_path = user_project_path.to_path
+        user_target_name = target_name.sub("Pods-", "")
+
+        script = <<-EOS 
+        require 'xcodeproj';
```
segiddins: I dont think this change is going to work -- cocoapods needs to generate a workspace that works even on machines that don't have cocoapods installed

i-dama: I see. Didn't take that into consideration. 

Could I check for that and use this method only on machines that have it installed, and fall back to the shell command otherwise?
Or would you prefer not to have different behaviors on different machines?

lib/cocoapods/generator/copy_resources_script.rb
```diff
@@ -71,6 +82,28 @@ def install_resources_function
...
+        project_path = '#{project_path}';
+        target_name = '#{user_target_name}';
+        project = Xcodeproj::Project.open(project_path);
```
dnkoutso: this also has slowdown implications for large projects. Each target script phase will actually re-open the xcodeproj to calculate these which can be fairly slow.

We have to come up with a different strategy to figure this out. I'd like to understand what was the reason the initial script phase was so naive.

dnkoutso: For example this is very old code (back in 2013 and 2015) as this commit https://github.com/CocoaPods/CocoaPods/commit/e06296847aa99d0e31cee1451dab9995200ec4b5

i-dama: Another approach could be adding the pod ‘.xcassets’ in the user project during ‘pod install’. That way the assets catalogue doesn’t need to be recompiled.

However, this could complicate ‘pod deintegrate’

dnkoutso: There is also `${PODS_TARGET_SRCROOT}` now set which is always relative to where the pod is (local dev pod or remote). 

I am playing around with a few samples to try to understand existing logic.

i-dama: If I understood it correctly, this is what’s happening:
1. The user project compiles ‘.xcassets’ into Assets.car
2. There is a pod shipped as a static library (scenario I’m experiencing)
3. The pod expects the assets to be in the the main bundle
4. Copy resources copies all resources into the main bundle
5. Since ‘.xcassets’ need to be compiled, and adding to Assets.car is impossible, it is regenerated
6. The script picks up all ‘.xcassets’ to compile them, accidentally getting some that are not intended to go in

dnkoutso: @i-dama you are correct. Still trying to understand what was the reasoning of adding 
```sh
  # Find all other xcassets (this unfortunately includes those of path pods and other targets).
  OTHER_XCASSETS=$(find "$PWD" -iname "*.xcassets" -type d)
```

dnkoutso: @i-dama I am understanding this more. It seems its trying to pick up the xcassets used by the app so it can compile all of them together.

i-dama: What about the approach suggested above? Adding the pod .xcassets in the app’s “copy bundle resources” phase during ‘pod install’ and avoiding recompiling the Assets.car altogether

dnkoutso: it would probably work, but then de integrate will become a problem.

dnkoutso: It would also pollute the users project further.

dnkoutso: @i-dama the only way I can think of doing this, is what you mentioned with polluting the users app with these references.

dnkoutso: @i-dama admittedly another crappy way to resolve this provide an ENV variable that is set by the user that specifies the xcconfigs that part of their app. Its pretty crappy but I cant think of anything better without literally re-designing how all of the resource copying works in CocoaPods.

i-dama: Agree. I think any solution that requires the user to have to manually do something whenever they add an `.xcassets` would be requiring too much from them.

I'm going to do a bit more digging, and try to come up with another proposal. I think people might be affected by this without realizing it, since you wouldn't notice the extra images in the `.ipa` file unless you manually inspect it

Anyway, thanks for taking the time to investigate this 💯

dnkoutso: Given where we are this might work. So you will be able to find which `.xcassets` folder the `Name` exists into and use that?

Yes there could be some unwanted ones but much much smaller chance

dnkoutso: @i-dama it might be a bit scary too because if there are collisions we must ensure we dont pick the wrong one as it could be a different image

i-dama: @dnkoutso I'd propose this logic:
An `.xcassets` is included if *all* of the `xcimageset`s are present in the compiled `Assets.car` file.

From what I know, there is no way of instructing Xcode to compile only certain `.xcimageset`s in a given `.xcassets` catalogue, so by checking that *all* are present, we're most likely not going to exclude any `.xcassets` that should be included.

However, the problem would still persist if:
* Target 1 - Assets1: `1`, `2`, `3`
* Target 2 - Assets2: `1`, `2`

If compiling for `Target 1`, Assets2 would be included.

I don't see a way of knowing the distinction without tapping into the information from `.xcodeproj`.

However, the current state has the same problem, and doesn't handle collisions, so doing this would (hopefully) solve it for a subset of affected configurations.

dnkoutso: OK. want to try to make a PR for it? We *might* want to consider hiding it behind some sort of flag or ENV variable presence to ensure we slowly migrate to that.

dnkoutso: Seems like it will be a lot of bash scripting....

i-dama: Another approach would be to create a temporary `Pods.xcassets` catalogue, and copy all matching `.xcimageset`s into it.
After `Assets.car` is compiled, it can be deleted.

However, the collision problem still persists.

dnkoutso: > However, the collision problem still persists. you mean your initial (1,2,3) one but not the wrong image right? 

Could work as well. We need to ensure also the script phase does not run again (input and output paths) on subsequent builds.

i-dama: I don't see the distinction between the (1,2,3) and the wrong image problem.
In the 1,2,3 problem, images 1 and 2 get copied twice, and it's not known which one is going to get used.

I'd be glad to try and create a PR with it. Any preference on the approach? (I prefer the `Pods.xcassets` approach, seems safer)

Regarding the input/output, I don't think anything needs to be changed. The only thing changing is the logic of compiling `Assets.car` (Include all vs filter out some), the inputs/outputs stay the same.

dnkoutso: We could output an Xcode warning in that case.

You can go with `Pods.xcassets` since its all internal implementation detail of pods but I still believe we should hide this behind an ENV flag just to be sure we do not "break the world" (although admittedly is already broken ;))

i-dama: Actually, an input would be the `Assets.car` file. Probably not there atm.

i-dama: Alright 👍 
I'm going to try and make a PR this weekend. Should I just update this one or create a new one?
Agree on an ENV flag too 👍 

Funny thing is, `actool` (I believe) emits a warning for this, and it's also present in the current state. But it doesn't propagate to Xcode runtime warnings:
image

dnkoutso: Let's make a new PR! We've had plenty of technical conversation here :)

PR #7781, opened 2018-05-25
Add support if project path is symbol link

### LINKED ISSUE #7761: project symlink not work
CocoaPods v1.5.0 and v1.5.2:

I have a project, a development pods and other development pods in the project directory, but all projects are under the symlink directory. eg:
```
- MyApp -> .mbox/repos/MyApp
                      |---MySDK2
                             |---MySDK2.podspec
- MySDK -> .mbox/repos/MySDK
                      |---MySDK.podspec
- Podfile
```

In the `Podfile`:
```
project 'MyApp/MyApp.xcodeproj'

target 'MyApp' do
    pod 'MySDK', :path=>"MySDK"
    pod 'MySDK2', :path=>"MyApp/MySDK2"
end
```

I found the `Development Pods` in `Pods` project group will not work:

image

Demo here: https://github.com/Whirlwind/CocoaPodsTestSymlink

Whirlwind:
To fix #7761 .

I change some codes:

1. set the `Pods` group location to `Pods/` directory path in user project.
- set the `Pods-xxx.xcconfig` location to relative from parent path (change `Pods/Target Support Files/Pods-AFNetworking Example/Pods-AFNetworking Example.debug.xcconfig` to `Target Support Files/Pods-AFNetworking Example/Pods-AFNetworking Example.debug.xcconfig`, remove the `Pods`)
- all shell script paths changed to `${PODS_ROOT}` from `${SRCROOT}/Pods`
- if the dependency source is local path, create a symlink in `Pods/`, eg: `Pods/XXX` -> `../XXX`.
- if the project is symlink, create a Pods symlink in the project directory.
- update read_file_system logic, it will list all files in symlink directory if the directory links to another directory without dependency's directory.
- update some path, use `cleanpath` instead of `realpath`, will not resolve symlink.

If we change so many codes to support symlink, I think it is too dangerous and impossible.
So I use the symlink to fix this symlink bug....

```
$ cat Podfile
pod 'AFNetworking', :path => './AFNetworking'
$ tree .
|--.mbox
|       |--repos
|              |--MyApp
|                        |--Pods -> ../../../Pods
|--Podfile
|--Podfile.lock
|--MyApp -> .mbox/repos/MyApp
|--AFNetworking
|--Pods
       |--AFNetworking -> ../AFNetworking
```

First, I keep some symlink paths, not resolve to realpath. Just remove `.././` with `cleanpath`.

Then, If the user project is symlink, I create the Pods symlink to the `Pods` realpath, so user project will keep `PODS_ROOT=${SRCROOT}/Pods`. If developer makes multiple symlinks to a project, CP will just update the symlink target instead of change all settings (eg: environment paths, script paths in user project)

If the dependency is local, I create the dependency symlink in Pods directory : `AFNetworking -> ../AFNetworking`,
So the user project will access the dependency directory by `${PODS_ROOT}/AFNetworking` instead of `${SRCROOT}/../../../AFNetworking`.

### DIFF
--- a/spec/unit/target/aggregate_target_spec.rb
+++ b/spec/unit/target/aggregate_target_spec.rb
@@ -201,7 +201,7 @@ module Pod
           )
           @target.framework_paths_by_config['Debug'].should == [
             { :name => 'FrameworkA.framework',
-              :input_path => '${PODS_ROOT}/../../../../../../../../some/absolute/path/to/FrameworkA.framework',
+              :input_path => '${PODS_ROOT}/../../../../../../../some/absolute/path/to/FrameworkA.framework',
               :output_path => '${TARGET_BUILD_DIR}/${FRAMEWORKS_FOLDER_PATH}/FrameworkA.framework' },
           ]
         end
@@ -230,7 +230,7 @@ module Pod
           )
           @target.framework_paths_by_config['Debug'].should == [
             { :name => 'FrameworkA.framework',
-              :input_path => '${PODS_ROOT}/../../../../../../../../absolute/path/to/FrameworkA.framework',
+              :input_path => '${PODS_ROOT}/../../../../../../../absolute/path/to/FrameworkA.framework',
               :output_path => '${TARGET_BUILD_DIR}/${FRAMEWORKS_FOLDER_PATH}/FrameworkA.framework' },
           ]
         end


dnkoutso:
@amorde can you take an initial first look since you've done some prior work for this?

amorde:
I'm pretty swamped today but will take a look this weekend

amorde:
@Whirlwind Overall looks pretty good!

In general I like the idea of linking the pod source into `Pods/` as it makes the paths match and also makes relative globbing work. I'm wondering about some edge cases, like multiple local pods coming from the same directory. This also reduces the performance of `read_file_system`, but I don't have a large project available to test on to see how much it would impact the runtime of `pod install`. I also feel like some of the logic is in the wrong place but I don't have better suggestions. In particular I don't think the `analyzer` should have to handle creating the symlinks.

I've made some edits to help with some of the documentation and code style, slightly improve performance, and avoid running shell commands using the `cd ..` syntax. Please take a look:

```diff
diff --git a/lib/cocoapods/installer/analyzer.rb b/lib/cocoapods/installer/analyzer.rb
index 18f22761bdf..2404c174f61 100644
--- a/lib/cocoapods/installer/analyzer.rb
+++ b/lib/cocoapods/installer/analyzer.rb
@@ -650,14 +650,17 @@ module Pod
                       target_definitions, file_accessors, scope_suffix).tap do |target|
           name = target.root_spec.name
           if sandbox.local?(name)
-            # we will create a symlink `Pods/#{name}` to the realpath of the pod,
-            # then we could access the Pod's files by the `${PODS_ROOT}/#{name}` instead of the realpath.
+            # Create a symlink at `$PODS_ROOT/#{name}` to the path of the local pod,
+            # making the local pod's files available at `${PODS_ROOT}/#{name}`
             poddir = sandbox.pod_dir(target.root_spec.name)
             realdir = sandbox.pod_realdir(name)
             unless sandbox.local_path_was_absolute?(name)
               realdir = realdir.realpath.relative_path_from(poddir.dirname.realpath)
             end
-            `cd "#{poddir.dirname}" && rm -rf "#{poddir.basename}" && ln -fs "#{realdir}" "#{poddir.basename}"`
+            if File.symlink?(poddir)
+              File.unlink(poddir)
+            end
+            File.symlink(realdir, poddir)
           end
         end
       end
diff --git a/lib/cocoapods/installer/user_project_integrator/target_integrator.rb b/lib/cocoapods/installer/user_project_integrator/target_integrator.rb
index c2d489d7bc8..5da4352c0ba 100644
--- a/lib/cocoapods/installer/user_project_integrator/target_integrator.rb
+++ b/lib/cocoapods/installer/user_project_integrator/target_integrator.rb
@@ -272,7 +272,7 @@ module Pod
             add_copy_resources_script_phase
             add_check_manifest_lock_script_phase
             add_user_script_phases
-            link_pods_directory
+            add_sandbox_symlink
           end
         end
 
@@ -411,17 +411,21 @@ module Pod
           end
         end
 
-        # Adds a fake Pods directory that link to real Pods directory,
-        # if the user project is a symlink
+        # Adds a symlink at $PROJECT_DIR/Pods pointing to $PODS_ROOT if the user project is behind a symlink
         #
-        # @note   will remove origin Pods directory
+        # @note Replaces any existing symlinks at $PROJECT_DIR/Pods
         #
         # @return [void]
         #
-        def link_pods_directory
+        def add_sandbox_symlink
           if target.user_project_is_symlink
             project_dir = target.user_project_path.dirname.realpath
-            `cd "#{project_dir}" && rm -rf Pods && ln -s "#{target.sandbox.root.realpath.relative_path_from(project_dir)}" Pods`
+            link_source = target.sandbox.root.realpath.relative_path_from(project_dir)
+            link_path = project_dir + target.sandbox.root.basename
+            if File.symlink?(link_path)
+              File.unlink(link_path)
+            end
+            File.symlink(link_source, link_path)
           end
         end
 
diff --git a/lib/cocoapods/installer/user_project_integrator/target_integrator/xcconfig_integrator.rb b/lib/cocoapods/installer/user_project_integrator/target_integrator/xcconfig_integrator.rb
index 34fc14630ed..39f86ff7425 100644
--- a/lib/cocoapods/installer/user_project_integrator/target_integrator/xcconfig_integrator.rb
+++ b/lib/cocoapods/installer/user_project_integrator/target_integrator/xcconfig_integrator.rb
@@ -42,7 +42,7 @@ module Pod
           #         The build configuration.
           #
           def self.set_target_xcconfig(pod_bundle, target, config)
-            # Xcode root group's path is a realpath, we must get the relative path of sandbox to user project's realpath
+            # Xcode root group's path is absolute, we must get the relative path of the sandbox to the user project
             group_path = pod_bundle.relative_pods_root_path
             path = pod_bundle.xcconfig_relative_path(config.name)
             group = config.project['Pods'] || config.project.new_group('Pods', group_path)
diff --git a/lib/cocoapods/sandbox.rb b/lib/cocoapods/sandbox.rb
index d57a32aaf14..737b4c401ed 100644
--- a/lib/cocoapods/sandbox.rb
+++ b/lib/cocoapods/sandbox.rb
@@ -142,8 +142,8 @@ module Pod
       target_support_files_root + name
     end
 
-    # Returns the path where the Pod with the given name is stored, taking into
-    # account whether the Pod is locally sourced.
+    # Returns the path where the Pod with the given name is stored
+    # within the Sandbox. Does not account for where local pods are stored
     #
     # @param  [String] name
     #         The name of the Pod.
@@ -155,6 +155,14 @@ module Pod
       sources_root + root_name
     end
 
+    # Returns the path where the Pod with the given name is stored, taking into
+    # account whether the Pod is locally sourced.
+    #
+    # @param  [String] name
+    #         The name of the Pod.
+    #
+    # @return [Pathname] the path of the Pod.
+    #
     def pod_realdir(name)
       root_name = Specification.root_name(name)
       if local?(root_name)
diff --git a/lib/cocoapods/sandbox/path_list.rb b/lib/cocoapods/sandbox/path_list.rb
index ad99bb548f8..0fb73c1cbcb 100644
--- a/lib/cocoapods/sandbox/path_list.rb
+++ b/lib/cocoapods/sandbox/path_list.rb
@@ -46,19 +46,20 @@ module Pod
 
       # List all files and directories under {dir}. This method could be executed recursively.
       #
-      # @param  [Pathname] dir
+      # @param  [Pathname] dir directory to traverse
       #
       # @return [Array<Array<String>>] The array of directories and
-      #         files, the pathes are relative from current `pwd`
+      #         files in {dir}. The paths are relative from 
       #
-      def list_files_under(dir)
+      def list_files_under(dir, _basepath=nil)
         dirs = []
         files = []
+        _basepath = dir.realpath unless _basepath
         dir.children.each do |path|
           if path.directory?
             dirs << path.to_s
-            if !path.symlink? || path.realpath.relative_path_from(Pathname.new('.').realpath).to_s.start_with?('..')
-              tmp_dirs, tmp_files = list_files_under(path)
+            if !path.symlink? || path.realpath.relative_path_from(_basepath).to_s.start_with?('..')
+              tmp_dirs, tmp_files = list_files_under(path, _basepath)
               dirs.concat(tmp_dirs)
               files.concat(tmp_files)
             end
@@ -79,9 +80,7 @@ module Pod
 
         dirs = []
         files = []
-        Dir.chdir(root.cleanpath.to_s) do
-          dirs, files = list_files_under(Pathname.new('.'))
-        end
+        dirs, files = list_files_under(root.cleanpath)
 
         dirs.sort_by!(&:upcase)
         files.sort_by!(&:upcase)
```

spec/unit/sandbox/path_list_spec.rb
```diff
@@ -336,12 +336,20 @@ module Pod
...
       end
 
-      it 'doesn\'t include file in symlinked dir' do
+      it 'includes file in symlinked dir if link to other directory which not in root' do
         @path_list.instance_variable_set(:@files, nil)
         @path_list.instance_variable_set(:@dirs, nil)
         File.symlink(@tmpdir, @symlink_dir)
 
-        @path_list.files.map(&:to_s).should.not.include?('Classes/symlinked/someheader.h')
+        @path_list.files.map(&:to_s).should.include?('Classes/symlinked/someheader.h')
```
amorde: Will this affect headers in vendored frameworks?

spec/unit/sandbox/path_list_spec.rb
```diff
@@ -336,12 +336,20 @@ module Pod
...
+      end
+
+      it 'doesn\'t include file in symlinked dir if link to other directory in root' do
+        @path_list.instance_variable_set(:@files, nil)
+        @path_list.instance_variable_set(:@dirs, nil)
+        File.symlink(@path_list.root + 'framework/Source', @symlink_dir)
+
+        @path_list.files.map(&:to_s).should.not.include?('Classes/symlinked/MoreBanana.h')
```
amorde: I'm not really sure what this is testing

spec/unit/sandbox/path_list_spec.rb
Whirlwind: It is work for local development pods: `pod 'xx', :path=>'../xxx'`

Whirlwind:
How to make CI pass the test before cocoapods-integration-specs merged?

Whirlwind:
Hi @amorde , Thanks for your review.

I update my code as you said, but there is a error:
```

diff --git a/lib/cocoapods/sandbox/path_list.rb b/lib/cocoapods/sandbox/path_list.rb
index ad99bb548f8..0fb73c1cbcb 100644
--- a/lib/cocoapods/sandbox/path_list.rb
+++ b/lib/cocoapods/sandbox/path_list.rb
@@ -46,19 +46,20 @@ module Pod
 
       # List all files and directories under {dir}. This method could be executed recursively.
       #
-      # @param  [Pathname] dir
+      # @param  [Pathname] dir directory to traverse
       #
       # @return [Array<Array<String>>] The array of directories and
-      #         files, the pathes are relative from current `pwd`
+      #         files in {dir}. The paths are relative from 
       #
-      def list_files_under(dir)
+      def list_files_under(dir, _basepath=nil)
         dirs = []
         files = []
+        _basepath = dir.realpath unless _basepath
         dir.children.each do |path|
           if path.directory?
             dirs << path.to_s
-            if !path.symlink? || path.realpath.relative_path_from(Pathname.new('.').realpath).to_s.start_with?('..')
-              tmp_dirs, tmp_files = list_files_under(path)
+            if !path.symlink? || path.realpath.relative_path_from(_basepath).to_s.start_with?('..')
+              tmp_dirs, tmp_files = list_files_under(path, _basepath)
               dirs.concat(tmp_dirs)
               files.concat(tmp_files)
             end
@@ -79,9 +80,7 @@ module Pod
 
         dirs = []
         files = []
-        Dir.chdir(root.cleanpath.to_s) do
-          dirs, files = list_files_under(Pathname.new('.'))
-        end
+        dirs, files = list_files_under(root.cleanpath)
 
         dirs.sort_by!(&:upcase)
         files.sort_by!(&:upcase)
```

We need file list with relative path:
```
-        Dir.chdir(root.cleanpath.to_s) do
-          dirs, files = list_files_under(Pathname.new('.'))
-        end
+        dirs, files = list_files_under(root.cleanpath)
```
So I use `Dir.chdir` and `Pathname.new('.')` to get the relative path from `Pathname.children`.
We will get the absolute paths without the `Dir.chdir`.

Whirlwind:
```
            if File.symlink?(poddir)
              File.unlink(poddir)
            end
            File.symlink(realdir, poddir)
```
If the project is not under symlink directory, but now I use it as a symlink. The `Pods` exists and not be a symlink, it will fail.

Whirlwind:
```
In particular I don't think the analyzer should have to handle creating the symlinks.
```
I create the symlinks at the Installer, but the unit test not call the `Installer.install!` at the time when test analayzer. It will cause the symlink not exists and can not find the pods.

segiddins:
@dnkoutso want to check what performance implications this has on a project with many many development pods? And see if it works if those development pods share directories?

amorde:
>             if File.symlink?(poddir)
>               File.unlink(poddir)
>             end
>             File.symlink(realdir, poddir)
> If the project is not under symlink directory, but now I use it as a symlink. The Pods exists and not be a symlink, it will fail.

In this scenario the project was originally in the same directory, and then moved somewhere else and a symlink was added? I guess we could handle that but we are really getting into edge cases here. Mostly wanted to avoid running `rm -rf` unless there are no other options

> We need file list with relative path:
> 
> -        Dir.chdir(root.cleanpath.to_s) do
> -          dirs, files = list_files_under(Pathname.new('.'))
> -        end
> +        dirs, files = list_files_under(root.cleanpath)
> So I use Dir.chdir and Pathname.new('.') to get the relative path from Pathname.children.
> We will get the absolute paths without the Dir.chdir.

Ah okay, but you can use `relative_path_from` for that if that is the desired behavior

amorde:
I think rather than performance, just removing the dependency on what `pwd` is at any given time is a plus. The code should work even if somehow `pwd` is not what we expect, it's just calculations on paths

lib/cocoapods/installer/user_project_integrator/target_integrator.rb
```diff
@@ -410,6 +411,25 @@ def add_check_manifest_lock_script_phase
           end
         end
 
+        # Adds a symlink at $PROJECT_DIR/Pods pointing to $PODS_ROOT if the user project is behind a symlink
+        # if the user project is a symlink
```
amorde: The second line `if the user project is a symlink` is unneccessary

lib/cocoapods/installer/user_project_integrator/target_integrator/xcconfig_integrator.rb
```diff
@@ -42,9 +42,13 @@ def self.integrate(pod_bundle, targets)
           #         The build configuration.
           #
           def self.set_target_xcconfig(pod_bundle, target, config)
-            path = pod_bundle.xcconfig_relative_path(config.name)
-            group = config.project['Pods'] || config.project.new_group('Pods')
-            file_ref = group.files.find { |f| f.path == path }
+            # Xcode root group's path is absolute, we must get the relative path of the sandbox to the user project
+            group_path = pod_bundle.relative_pods_root_path
+            group = config.project['Pods'] || config.project.new_group('Pods', group_path)
+            group_path = Pathname.new(group.path || "") # supoort user custom path of Pods group
```
amorde: `supoort` -> `support` in comment

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -224,8 +224,8 @@ def add_files_to_build_phases(native_target, test_native_targets)
                               end
 
               headers = file_accessor.headers
-              public_headers = file_accessor.public_headers.map(&:realpath)
-              private_headers = file_accessor.private_headers.map(&:realpath)
+              public_headers = file_accessor.public_headers
+              private_headers = file_accessor.private_headers
```
amorde: Mentioned this elsewhere but wondering if this would affect vendored frameworks which often use symlinks to point to different versions vs. "current"

amorde:
I think using symlinks for development pods could work out really well, but I'm unsure about the other changes. It does feel like a whole lot of changes to handle an uncommon use case.

@dnkoutso any thoughts?

lib/cocoapods/installer/user_project_integrator/target_integrator/xcconfig_integrator.rb
Whirlwind: done

lib/cocoapods/installer/user_project_integrator/target_integrator.rb
Whirlwind: done

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
Whirlwind: I change the get file list logic and it will list all files in symlink directory.

segiddins:
> I think rather than performance, just removing the dependency on what pwd is at any given time is a plus. The code should work even if somehow pwd is not what we expect, it's just calculations on paths

CocoaPods performance is a major concern for us at the moment. `pod install` takes anywhere from 80-120 seconds, and further regressions from that would be incredibly painful

amorde:
Definitely, I meant even if the performance impact was minimal I'd still avoid `Dir.chdir` anyway.

Whirlwind:
I agree with @amorde and @segiddins . The `Dir.chdir` is unsafe in multithread, and I want to void to use it, too. But I didn't find a better way.

Whirlwind:
Is there any idea about this PR?

amorde:
Sorry for the delay, there's so much changing in this PR it can be a bit hard to review.

Would you be open to moving parts of this into separate PRs? These changes would be good in their own PRs

> * set the Pods group location to Pods/ directory path in user project.
> * all shell script paths changed to ${PODS_ROOT} from ${SRCROOT}/Pods

Other than that I don't have any additional comments on the code

Whirlwind:
> all shell script paths changed to ${PODS_ROOT} from ${SRCROOT}/Pods

The changed must be done for this PR. All the shell script will must use the `${PODS_ROOT}`, because I will create a symlink `Pods/` to the path of Pods folder.

Issue #7486, opened 2018-03-07
Cannot set the CFBundleDevelopmentRegion in Info.plist of resource bundles


xingheng:
* [*] I've read and understood the [*CONTRIBUTING guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

Clone the lastest CocoaPods project to review the resource bundle's Info.plist's generate rules.
Check the issue list.

## What did you expect to happen?

Apply the app target's `development region` to all the pods' resource bundle.

## What happened instead?

The `CFBundleDevelopmentRegion` key in Info.plist is set to `en` with a default key.

### Stack

```
   CocoaPods : 1.3.1
        Ruby : ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-darwin16]
    RubyGems : 2.6.13
        Host : Mac OS X 10.12.6 (16G29)
       Xcode : 9.2 (9C40b)
         Git : git version 2.14.3 (Apple Git-98)
Ruby lib dir : /Users/hanwei/.rvm/rubies/ruby-2.4.1/lib
Repositories : bitbucket - https://bitbucket.org/UItachi/specs.git @ 6aa9c96ed1c82b32f906285dc004c0a50bc16666
               master - https://github.com/CocoaPods/Specs.git @ bec882057e88704a796126484d0e627b654b43eb
```

### Installation Source

```
Executable Path: /Users/hanwei/.rvm/gems/ruby-2.4.1/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.1
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
cocoapods-trunk       : 1.3.0
cocoapods-try         : 1.1.0
```
...

## Project that demonstrates the issue

I found the [onepassword-app-extension](https://github.com/agilebits/onepassword-app-extension) has the same issue there.

---

I found [this issue](https://github.com/CocoaPods/Xcodeproj/issues/259) is the starter, resource bundle needs an Info.plist as a standard. For the `CFBundleDevelopmentRegion` key,  the region will be used in some system API to localize some identifiers like `UIBarButtonSystemItemDone` and `UIBarButtonSystemItemCancel` in pod projects.

[...]

PR #7784, opened 2018-05-26
Fix `INFOPLIST_FILE` being overridden when set in `pod_target_xcconfig`

### LINKED ISSUE #7530: Question: How to set CFBundleVersion in Pod Framework?
I apologize for the clutter, but I've googled around for an answer to this and haven't found an answer. I'm developing an SDK in a Pod as a framework that wraps a REST api. I don't care what the users of this pod set their `CFBundleVersion` value to. That number has no meaning to our system. Instead, I need to declare the `CFBundleVersion` for each release of the pod so that we can know what versions are out in the wild. And it seems like bad form to require each user of the framework to put the correct `post_install` script in their Podfile.

I'm sure there must be a way to do this and that I just haven't found the right documentation. Can someone enlighten me.

🌈

amorde:
Fixes #7530

@dnkoutso We may also be able to mark #3032 and #7486 as done, unless they should stay open for future DSL to be added

Podspecs utilizing this may need to set `s.resources = 'PATH_TO_INFO_PLIST.plist'` so that the file is not discarded during install

### DIFF
--- a/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
@@ -79,7 +79,7 @@ module Pod
               end
 
               if target.requires_frameworks?
-                unless target.static_framework?
+                unless skip_info_plist?(native_target)
                   create_info_plist_file(target.info_plist_path, native_target, target.version, target.platform)
                 end
                 create_build_phase_to_symlink_header_folders(native_target)
@@ -112,6 +112,18 @@ module Pod
           #
           def skip_pch?(specs)
             specs.any? { |spec| spec.prefix_header_file.is_a?(FalseClass) }
+          end
+
+          # True if info.plist generation should be skipped
+          #
+          # @param [PXNativeTarget] native_target
+          #
+          # @return [Boolean] Whether the target should build an Info.plist file
+          #
+          def skip_info_plist?(native_target)
+            return true if target.static_framework?
+            existing_setting = native_target.resolved_build_setting('INFOPLIST_FILE', true).values.compact
+            !existing_setting.empty?
           end
 
           # Remove the default headers folder path settings for static library pod
--- a/spec/unit/installer/xcode/pods_project_generator/pod_target_installer_spec.rb
+++ b/spec/unit/installer/xcode/pods_project_generator/pod_target_installer_spec.rb
@@ -90,6 +90,16 @@ module Pod
               end
             end
 
+            it 'respects INFOPLIST_FILE of pod_target_xcconfig' do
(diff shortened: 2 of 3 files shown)


amorde:
After a second look I don't think this fixes #7486

lib/cocoapods/installer/xcode/pods_project_generator/target_installer.rb
```diff
@@ -170,6 +170,7 @@ def create_info_plist_file(path, native_target, version, platform, bundle_packag
               add_file_to_support_group(path) if add_to_support_group
 
               native_target.build_configurations.each do |c|
+                next unless c.resolve_build_setting('INFOPLIST_FILE').nil?
```
dnkoutso: this might not be the right spot to debounce it. This means the file was actually written to disk but not "linked" to the project.

Could we check on this a bit earlier within `pod_target_installer` instead?

dnkoutso:
Yeah I am not sure about #3032 and #7486. We can probably comment and announce this is a relatively easily work around but there might still be a case to include it in the CocoaPods library.

amorde:
Will squash once green

spec/unit/installer/xcode/pods_project_generator/pod_target_installer_spec.rb
```diff
@@ -757,6 +767,34 @@ class PodsProjectGenerator
...
+              @pod_target.stubs(:requires_frameworks?).returns(true)
+              @installer.install!
+              group = @project['Pods/BananaLib/Support Files']
+              group.children.map(&:display_name).sort.should == [
+                'BananaLib-Info.plist',
+                'BananaLib-dummy.m',
+                'BananaLib-prefix.pch',
+                'BananaLib.modulemap',
+                'BananaLib.xcconfig',
+              ]
```
amorde: I couldn't find an existing unit test that covered this - if there is one I'll remove this

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -114,6 +114,18 @@ def skip_pch?(specs)
...
 
+          # True if info.plist generation should be skipped
+          #
+          # @param [PXNativeTarget] native_target
+          #
+          # @return [Boolean] Whether the target should build an Info.plist file
+          #
+          def skip_info_plist?(native_target)
+            return true if target.static_framework?
+            existing_setting = native_target.resolved_build_setting('INFOPLIST_FILE', true).values.compact
```
dnkoutso: I guess passing `true` here makes sense since we know the `.xcconfig` has been created and assigned there and its within our control.

Issue #7617, opened 2018-04-11
Pod [V1.5.0] lib lint fail with error "AppIcon" in case of its dependency or subspec contains a xcassets type resources


seuzxh:
* [√] I've read and understood the [*CONTRIBUTING guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

- I create a test Pod with command "pod lib create"
- I add a subspec named "TestSub1" in the podspec file
- Add resources of "TestSub1" with the xcassets file
- run "pod lib lint  --allow-warnings --verbose --fail-fast --no-clean"

## What did you expect to happen?

- lint success

## What happened instead?

- lint fail with message 

```
/* com.apple.actool.errors */
/Users/zhaoxianghui/Documents/XCProject/TestPod1.5.0/TestSub1/Assets/testpodAset.xcassets: error: None of the input catalogs contained a matching stickers icon set or app icon set named  "AppIcon".
/* com.apple.actool.compilation-results */
/Users/zhaoxianghui/Library/Developer/Xcode/DerivedData/App-ephhbjaiwsmhsbexfmoptnatintl/Build/Products/Release-iphonesimulator/App.app/Assets.car
/Users/zhaoxianghui/Library/Developer/Xcode/DerivedData/App-ephhbjaiwsmhsbexfmoptnatintl/Build/Products/Release-iphonesimulator/assetcatalog_generated_info.plist
Command /bin/sh failed with exit code 1
** BUILD FAILED **
The following build commands failed:
PhaseScriptExecution [CP]\ Copy\ Pods\ Resources /Users/zhaoxianghui/Library/Developer/Xcode/DerivedData/App-ephhbjaiwsmhsbexfmoptnatintl/Build/Intermediates.noindex/App.build/Release-iphonesimulator/App.build/Script-2991DDF45C0A25EC16418712.sh
```

### Stack

```
   CocoaPods : 1.5.0
        Ruby : ruby 2.3.3p222 (2016-11-21 revision 56859) [x86_64-darwin16]
    RubyGems : 2.6.8
        Host : Mac OS X 10.13.4 (17E199)
       Xcode : 9.3 (9E145)
         Git : git version 2.9.2
Ruby lib dir : /usr/local/Cellar/ruby/2.3.3/lib
[...]

PR #7785, opened 2018-05-26
[WIP] Selecting_xcassets

i-dama:
🌈
Continuation of: 
https://github.com/CocoaPods/CocoaPods/pull/7779

Related issues 
https://github.com/CocoaPods/CocoaPods/issues/7745
https://github.com/CocoaPods/CocoaPods/issues/6159
https://github.com/CocoaPods/CocoaPods/issues/7617

Still WIP, but I'd like to get some feedback early on.
Mostly:
* Is using bash 4 ok? Having to iterate over the `INCLUDED_IMAGESETS` array for each `.imageset` makes it very slow. Bash 4 has associative array, those should be faster (I think)
* Is using a magical regex for parsing the `Name` out of the json ok? I couldn't find any better solution for bash without requiring additional tools.

Managed to get a sample project to run successfully with this tho 🎉

### DIFF
--- a/lib/cocoapods/generator/copy_resources_script.rb
+++ b/lib/cocoapods/generator/copy_resources_script.rb
@@ -204,13 +204,53 @@ EOS
 
 if [[ -n "${WRAPPER_EXTENSION}" ]] && [ "`xcrun --find actool`" ] && [ -n "${XCASSET_FILES:-}" ]
 then
-  # Find all other xcassets (this unfortunately includes those of path pods and other targets).
+
+  ASSETS_CAR_PATH="${BUILT_PRODUCTS_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}/Assets.car"
+  TMP_DIR="$TARGET_TEMP_DIR"
+
+  # TODO: Does PLATFORM_NAME match the input?
+  # TODO: Need to figure out an alternative to parse out `Name` in bash
+  # TODO: Is bash 4 an option? The matching is very inefficient
+  INCLUDED_IMAGESETS=$(xcrun --sdk "$PLATFORM_NAME" assetutil --info "$ASSETS_CAR_PATH" | ruby -e "require 'json'; JSON.parse(STDIN.read).each { |o| puts o['Name'] };")
+  echo "$INCLUDED_IMAGESETS"
+  PODS_XCASSETS="$TMP_DIR/Pods.xcassets"
+  rm -rf "$PODS_XCASSETS"
+  mkdir "$PODS_XCASSETS"
+
   OTHER_XCASSETS=$(find "$PWD" -iname "*.xcassets" -type d)
-  while read line; do
-    if [[ $line != "${PODS_ROOT}*" ]]; then
-      XCASSET_FILES+=("$line")
-    fi
+  while read xcassets; do
+
+    # TODO: Can files other than .imageset and .appiconset be in a xcassets file
+    ASSETS_PATHS_IN_CATALOGUE=$(find "$xcassets" -iname "*.imageset" -or -iname "*.appiconset" -type d)
+    while read path; do
+      folder_name_with_extension=${path##*/}
+      folder_name=${folder_name_with_extension%%.*}
+
+      ASSET_INCLUDED=0
+      while read included_asset; do
+        if [ "$included_asset" == "$folder_name" ]; then
+          echo "including $included_asset"
+          ASSET_INCLUDED=1
+        fi
+      done <<<"$INCLUDED_IMAGESETS"
+      
+      if [ "$ASSET_INCLUDED" == "1" ]; then
+        cp -r "$path" "$PODS_XCASSETS"
+      fi
+      
+    done <<<"$ASSETS_PATHS_IN_CATALOGUE"
   done <<<"$OTHER_XCASSETS"
+
+  XCASSET_FILES+=("$PODS_XCASSETS")
+
+  # TODO: Add environment switch
+  # Find all other xcassets (this unfortunately includes those of path pods and other targets).
+  # OTHER_XCASSETS=$(find "$PWD" -iname "*.xcassets" -type d)
+  # while read line; do
+    # if [[ $line != "${PODS_ROOT}*" ]]; then
+      # XCASSET_FILES+=("$line")
+    # fi
+  # done <<<"$OTHER_XCASSETS"
 
   if [ -z ${ASSETCATALOG_COMPILER_APPICON_NAME+x} ]; then
     printf "%s\\0" "${XCASSET_FILES[@]}" | xargs -0 xcrun actool --output-format human-readable-text --notices --warnings --platform "${PLATFORM_NAME}" --minimum-deployment-target "${!DEPLOYMENT_TARGET_SETTING_NAME}" ${TARGET_DEVICE_ARGS} --compress-pngs --compile "${BUILT_PRODUCTS_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}"
--- a/lib/cocoapods/generator/copy_resources_script.rb
+++ b/lib/cocoapods/generator/copy_resources_script.rb
@@ -89,15 +89,7 @@ module Pod
         end
 
         script += RSYNC_CALL
-        script += XCASSETS_COMPILE_BEGIN
-        # TODO: Proper way of checking the env?
-        if ENV['XCASSETS_SELECTIVE_COMPILE'].is_set?
-          script += XCASSETS_COMPILE_FIND_SELECTIVE
-        else
-          script += XCASSETS_COMPILE_FIND_ALL  
-        end
-        
-        script += XCASSETS_COMPILE_END
+        script += XCASSETS_COMPILE
         script
       end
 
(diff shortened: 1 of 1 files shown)


i-dama:
@dnkoutso 😄

CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* [WIP] Selecting_xcassets  
  [i-dama](https://github.com/i-dama)
```

</p>

lib/cocoapods/generator/copy_resources_script.rb
```diff
@@ -204,14 +204,54 @@ def script
...
+
+  ASSETS_CAR_PATH="${BUILT_PRODUCTS_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}/Assets.car"
+  TMP_DIR="$TARGET_TEMP_DIR"
+
+  # TODO: Does PLATFORM_NAME match the input?
+  # TODO: Need to figure out an alternative to parse out `Name` in bash
+  # TODO: Is bash 4 an option? The matching is very inefficient
+  INCLUDED_IMAGESETS=$(xcrun --sdk "$PLATFORM_NAME" assetutil --info "$ASSETS_CAR_PATH" | ruby -e "require 'json'; JSON.parse(STDIN.read).each { |o| puts o['Name'] };")
+  echo "$INCLUDED_IMAGESETS"
+  PODS_XCASSETS="$TMP_DIR/Pods.xcassets"
```
dnkoutso: where does `$TMP_DIR` match to? Can we use a temp dir provided by Xcode intermediates folder maybe?

lib/cocoapods/generator/copy_resources_script.rb
```diff
@@ -204,14 +204,54 @@ def script
...
+  mkdir "$PODS_XCASSETS"
+
   OTHER_XCASSETS=$(find "$PWD" -iname "*.xcassets" -type d)
-  while read line; do
-    if [[ $line != "${PODS_ROOT}*" ]]; then
-      XCASSET_FILES+=("$line")
-    fi
+  while read xcassets; do
+
+    # TODO: Can files other than .imageset and .appiconset be in a xcassets file
```
dnkoutso: can check here but we will never be guaranteed if apple adds something else :) https://developer.apple.com/library/content/documentation/Xcode/Reference/xcode_ref-Asset_Catalog_Format/FolderStructure.html#//apple_ref/doc/uid/TP40015170-CH33-SW1

lib/cocoapods/generator/copy_resources_script.rb
dnkoutso: ah nevermind I see above.

lib/cocoapods/generator/copy_resources_script.rb
```diff
@@ -204,14 +204,54 @@ def script
...
 if [[ -n "${WRAPPER_EXTENSION}" ]] && [ "`xcrun --find actool`" ] && [ -n "${XCASSET_FILES:-}" ]
 then
-  # Find all other xcassets (this unfortunately includes those of path pods and other targets).
+
+  ASSETS_CAR_PATH="${BUILT_PRODUCTS_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}/Assets.car"
+  TMP_DIR="$TARGET_TEMP_DIR"
+
+  # TODO: Does PLATFORM_NAME match the input?
+  # TODO: Need to figure out an alternative to parse out `Name` in bash
+  # TODO: Is bash 4 an option? The matching is very inefficient
```
dnkoutso: I think Bash 4 would not be an option. I am on Bash 3 right now I'd imagine it would cause issues.

i-dama:
@dnkoutso Resolved the `// TODO:`'s

Need to take care of 2 more things:
1. Is what I did a proper way of checking for ENV here?
2. How can I specify the input/output you mentioned?

lib/cocoapods/generator/copy_resources_script.rb
```diff
@@ -89,7 +89,15 @@ def script
         end
 
         script += RSYNC_CALL
-        script += XCASSETS_COMPILE
+        script += XCASSETS_COMPILE_BEGIN
+        # TODO: Proper way of checking the env?
+        if ENV['XCASSETS_SELECTIVE_COMPILE'].is_set?
```
dnkoutso: I was suggesting to make this a test inside the bash script itself. Here it will only be checked when `pod install` runs versus the bash script in which it will run each time the script runs. The user can go add the "Build Setting" in the `.xcodeproj` and then re-run a clean build.

dnkoutso:
This is looking better and better! I can also try it on my sample project to verify it works

i-dama:
@dnkoutso Moved the env check. 
Tried the fix out on a huge project, got the same results as with a hotfix I did initially for that project.

Do we need to specify the Xcode inputs/outputs for this?
I guess the only input that is not the same as before is the compiled `Assets.car` file.

dnkoutso:
@i-dama have you tried re-running the build of a large project? Do it and then select 'Recent' tab and see if there are any messages at all. There should be none since everything should be skipped.

Need to think a bit more how this can go out of sync.

dnkoutso:
@i-dama its a bit tricky because there are two paths in the script with different input/output paths as you said `Assets.car`. The list of I/O is only a single flat list so there is a chance we will cause the original script to be executed over and over again for everyone. Need to think about it more.

dnkoutso:
@i-dama could we make `$TMP/Pods.xcassets` an input path? If we do we might need to make the "old" case to `mkdir` this (and leave it as empty)

i-dama:
@dnkoutso if I got it correctly, the input/output paths are used by Xcode to determine if a scripts needs to be ran or not.

In that case, I think that `Assets.car` should be the input in both cases. Consider this scenario using the `find all` method:
1. The user builds the project
2. The scripts picks up all `.xcassets` and regenerates `Assets.car`
3. The user adds another asset to a `xcassets` folder they manage
4. The user re-builds the project

Since `Assets.car` isn't in the input, the script isn't ran, and the newly added image isn't in the final `Assets.car`. (Or the `Assets.car` file got rewritten by the `Copy bundle resources` phase)

I think the script gets ran each time in the current state, otherwise I don't see how it would work correctly. I'll try what you suggested on my big project

i-dama:
Ran my big project twice. The second run re-executed both `Embed Pods Frameworks` and `Copy Pods Resources` scripts.
CocoaPods 1.5.2

dnkoutso:
It shouldnt run everytime. I have tested it before and if you expand the script phase in the Xcode logs it will tell you why.

The reason probably is that your `Assets.car` file that we produce since its an input path it has changed because of our own script causing the script to re-run.

dnkoutso:
@i-dama want to purse this any further?

i-dama:
I do! I'll try to find some time in the next couple of days.

I'm traveling atm - that's why I haven't replied in 2 days :)

dnkoutso:
nice!

i-dama:
@dnkoutso How can I tell if the script got ran or not? In the Recent tab I can see the phase and the logs it prints (using the current state of the script)

Also, what are your thoughts on my reasoning above about adding `Assets.car` for all configurations.

dnkoutso:
@i-dama I havent forgotten about this, will need this week for WWDC and come back to it next week.

dnkoutso:
@i-dama nothing with Xcode 10 that has improved upon how we can complete this right?

i-dama:
@dnkoutso Didn't try with Xcode 10 so far.

If I got it correctly, we need to figure out the input/output of the build script.

I'm failing to see how the current state of the script can correctly work without being ran every time.

robinkunde:
I hope this is the right place to bring this up. Happy to move this elsewhere if not. You probably know most of this, but I want to make sure we're on the same page in case I missing something.

I've been looking into this issue with Xcode 10 this past week. Xcode 10 issues a warning when both the app target's "Copy Bundle Resource" step and CP's "Copy Pods Resources" step produce an Assets.car file.

Example:
```
ignoring duplicated output file: '/Users/username/Library/Developer/Xcode/DerivedData/AppName-gxjlfmqfsldftcfezkhwcevxbwtn/Build/Products/Debug-iphonesimulator/AppName.app/Assets.car' (in target 'AppName')
```

Based on my research, the kind of solution CP is currently using is is not supported well enough by actool and Xcode's build pipeline to make work without major downsides (happy to be wrong here). What I found is this (1.53, 1.6.0 beta1 behaves identical IMO):

1st run/build
1. App target Copy Bundle Resource step generates Assets.car
2. Copy Pods Resources step generates completely new Assets.car to include pods catalogs

2nd+ run/build
1. App target Copy Bundle Resource step notices that Assets.car has changed, and generates Assets.car again from app resources only
2. Copy Pods Resources step is not re-run because none of its input files have changed, so Assets.car will be missing pods catalogs now (Xcode 10 will put the output from the last time the step was run (1.2) in the build log though. I verified this by having Pods-FDConsumer-resources.sh output timestamps. Idk if that is a bug or intentional)


Fixes I've considered:
1. Because we don't have control over how actool is invoked by Xcode's Copy Bundle Resource step and actool can't be invoked incrementally, we could regenerate the Assets.car file every time. 
Downsides: 
a. Increasing build times because Assets.car will be generated twice each time and also has to be copied during installation. 
b. The invocation parameters CP uses on actool are not optimal. Xcode knows the target device type and can skip generating assets that will not be needed, producing a smaller catalog in a shorter amount of time. CP has to generate a full catalog without that information. 
c. It doesn't solve the potential namespace collision issue.
d. Doesn't fix warning.
2. Add pods' catalogs to the app target's Copy Bundle Resource step
Upsides:
a. Eliminates duplicated work, fixing build time issues.
b. Can take advantage of target device info.
c. Fixes warning.
Downsides:
a. Messy, breaks encapsulation
b. Doesn't address namespace collision.
3. Deprecate this solution and encourage pod authors to move to using resource bundles. We could start by issuing a warning during pod validation and directing them to a page that helps with migration.
Upsides:
a. Eliminates duplicated work, fixing build time issues.
b. Can take advantage of target device info.
c. Addresses namespace collision issue.
d. Fixes warning.
Downsides:
a. Some amount of work for pod authors to migrate. I did this for PKHUD and it was straightforward because loading resources from bundles was already abstracted away.

The last solution seems like the best long term fix to me. It produces some work for pod authors but the improvements in correctness and determinism are worth it in my opinion. The first solution can be used in the short term if needed. Let me know what you think.

robinkunde:
@i-dama @dnkoutso Sorry, forgot to tag you last week. Please let me know what you think.

i-dama:
@robinkunde I'm experiencing a different issue (Path pods resource bundles being included in the Assets.car file), but it's the same root cause.
I think I mentioned the same issue with the script not being ran in one of the comments above. I don't see how the current approach can work without re-running the script every time.

Regarding your 3rd proposal, I'm not sure how this would play with static libraries. Do you know if they still use bundles, or rely on the Assets.car file?

I think it might be best to open up another issue, not to mix things up. I see it as 2 separate issues:
1. Selection being too naive and selecting *all* assets
2. The script not being ran every time

@dnkoutso Thoughts?

robinkunde:
@i-dama When using resource bundles, it doesn't matter whether it's a static or dynamic library, AFAIK. It generates a bundle file through a separate build target, which then contains a separate Assets.car file, along with whatever other assets (nibs, storyboards, strings files, etc). Either way, the library author will have add some code to locate the resource bundle and pass it to all methods that take a bundle parameter.

Example: https://github.com/pkluz/PKHUD/pull/235/files

Fab1n:
Regarding this issue I have also filed one pull request #8136 and contributed in another issue (#7745).

I have setup a sample project which seem to fix these 2 bugs regarding xcassets compilation.
You can check it out here:
https://github.com/Fab1n/PodResourcesBug

The sample project actually shows 2 recent issues regarding asset catalog compilation.

The test procedure is written in detail in the repos README.md

Whirlwind:
I made a cocoapods plugin to fix the assets lost:

1. Split the assets script into a single script from `Copy Pods Resources` ( I call it `Compile Assets Resources`)
```
XCASSET_FILES=()

# Input Files
COUNTER=0
while [ $COUNTER -lt ${SCRIPT_INPUT_FILE_COUNT} ]; do
  tmp="SCRIPT_INPUT_FILE_$COUNTER"
  line=${!tmp}

  XCASSET_FILES+=("$line")
  echo "$line"
  let COUNTER=COUNTER+1
done

# Input File Lists
COUNTER=0
while [ $COUNTER -lt ${SCRIPT_INPUT_FILE_LIST_COUNT} ]; do
  tmp="SCRIPT_INPUT_FILE_LIST_$COUNTER"
  file=${!tmp}
  while read line; do
    XCASSET_FILES+=("$line")
    echo "$line"
  done <"$file"
  let COUNTER=COUNTER+1
done

if [[ -n "${WRAPPER_EXTENSION}" ]] && [ "`xcrun --find actool`" ] && [ -n "${XCASSET_FILES:-}" ]
then

  if [ -z ${ASSETCATALOG_COMPILER_APPICON_NAME+x} ]; then
    printf "%s\0" "${XCASSET_FILES[@]}" | xargs -0 xcrun actool --output-format human-readable-text --notices --warnings --platform "${PLATFORM_NAME}" --minimum-deployment-target "${!DEPLOYMENT_TARGET_SETTING_NAME}" ${TARGET_DEVICE_ARGS} --compress-pngs --compile "${BUILT_PRODUCTS_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}"
  else
    printf "%s\0" "${XCASSET_FILES[@]}" | xargs -0 xcrun actool --output-format human-readable-text --notices --warnings --platform "${PLATFORM_NAME}" --minimum-deployment-target "${!DEPLOYMENT_TARGET_SETTING_NAME}" ${TARGET_DEVICE_ARGS} --compress-pngs --compile "${BUILT_PRODUCTS_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}" --app-icon "${ASSETCATALOG_COMPILER_APPICON_NAME}" --output-partial-info-plist "${TARGET_TEMP_DIR}/assetcatalog_generated_info_cocoapods.plist"
  fi
fi
```
2. Move all assets in pods resources into the input file list of `Compile Assets Resources` (`input.xcfilelist`)
3. Remove all user assets from `Copy Bundle Resources` in user target, and add all them into the `Input Files` of `Compile Assets Resources`.
4. Set the output files: `${BUILT_PRODUCTS_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}/Assets.car` and `${TARGET_TEMP_DIR}/assetcatalog_generated_info_cocoapods.plist`

Now, the script of `Compile Assets Resources` will have `Input Files`, `Input File Lists` and `Output Files`, and the `Input Files` will be set by user.

Upsides:
1. No duplicate compile, the `Copy Bundles Resources` will have not any assets.
2. User modify the Assets will make the xcode rerun the `Compile Assets Resources`, but not to rerun `Copy Pods Resources`

But there is a bug: the app icon in the `Info.plist` lost, there is not any icons in the `Copy Bundles Resources`, lose the `-additionalcontentfile`.

So I add other script `Fill App Icon into Info.plist` to add the appicon info into the `Info.plist`:
```
/usr/libexec/PlistBuddy -c "Merge \"${TARGET_TEMP_DIR}/assetcatalog_generated_info_cocoapods.plist\"" "${BUILT_PRODUCTS_DIR}/${INFOPLIST_PATH}"
```

But the target will lose the icon in the `Scheme` and `Targets`. It not effects.

dnkoutso:
Closing for inactivity. We will re-think this again for 1.9.0

Issue #7790, opened 2018-05-29
URI::InvalidURIError - bad URI(is not URI?)


Whirlwind:
```
URI::InvalidURIError - bad URI(is not URI?): git@code.byted.org:ugc/OpenCV.git
/Users/whirlwind/.rvm/rubies/ruby-2.4.1/lib/ruby/2.4.0/uri/rfc3986_parser.rb:67:in `split'
/Users/whirlwind/.rvm/rubies/ruby-2.4.1/lib/ruby/2.4.0/uri/rfc3986_parser.rb:73:in `parse'
/Users/whirlwind/.rvm/rubies/ruby-2.4.1/lib/ruby/2.4.0/uri/common.rb:231:in `parse'
/Users/whirlwind/.rvm/rubies/ruby-2.4.1/lib/ruby/2.4.0/uri/common.rb:737:in `URI'
/Users/whirlwind/Documents/Developer/CocoaPods/CocoaPods/lib/cocoapods/installer/pod_source_installer.rb:126:in `verify_source_is_secure'
/Users/whirlwind/Documents/Developer/CocoaPods/CocoaPods/lib/cocoapods/installer/pod_source_installer.rb:107:in `download_source'
/Users/whirlwind/Documents/Developer/CocoaPods/CocoaPods/lib/cocoapods/installer/pod_source_installer.rb:62:in `install!'
/Users/whirlwind/Documents/Developer/CocoaPods/CocoaPods/lib/cocoapods/installer.rb:367:in `install_source_of_pod'
/Users/whirlwind/Documents/Developer/CocoaPods/CocoaPods/lib/cocoapods/installer.rb:332:in `block (2 levels) in install_pod_sources'
/Users/whirlwind/Documents/Developer/CocoaPods/CocoaPods/lib/cocoapods/user_interface.rb:86:in `titled_section'
/Users/whirlwind/Documents/Developer/CocoaPods/CocoaPods/lib/cocoapods/installer.rb:331:in `block in install_pod_sources'
/Users/whirlwind/Documents/Developer/CocoaPods/CocoaPods/lib/cocoapods/installer.rb:314:in `each'
/Users/whirlwind/Documents/Developer/CocoaPods/CocoaPods/lib/cocoapods/installer.rb:314:in `install_pod_sources'
/Users/whirlwind/Documents/Developer/CocoaPods/CocoaPods/lib/cocoapods/installer.rb:163:in `block in download_dependencies'
[...]

PR #7794, opened 2018-05-29
Fix source validation for local URLs and SSH repos

dnkoutso:
closes https://github.com/CocoaPods/CocoaPods/issues/7790

Tried `git@code.byted.org:ugc/OpenCV.git` from the issue and it worked.

/cc @Whirlwind

### DIFF
--- a/spec/unit/installer/pod_source_installer_spec.rb
+++ b/spec/unit/installer/pod_source_installer_spec.rb
@@ -62,6 +62,30 @@ module Pod
         UI.warnings.should.include 'uses the unencrypted \'git\' protocol'
       end
 
+      it 'does not warn for local repositories with spaces' do
+        @spec.source = { :git => '/Users/kylef/Projects X', :tag => '1.0' }
+        dummy_response = Pod::Downloader::Response.new
+        Downloader.stubs(:download).returns(dummy_response)
+        @installer.install!
+        UI.warnings.length.should.equal(0)
+      end
+
+      it 'does not warn for SSH repositories' do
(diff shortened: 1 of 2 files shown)


spec/unit/installer/pod_source_installer_spec.rb
```diff
@@ -62,6 +62,30 @@ module Pod
...
+        @installer.install!
+        UI.warnings.length.should.equal(0)
+      end
+
+      it 'does not warn for SSH repositories on Github' do
+        @spec.source = { :git => 'git@github.com:kylef/test.git', :tag => '1.0' }
+        dummy_response = Pod::Downloader::Response.new
+        Downloader.stubs(:download).returns(dummy_response)
+        @installer.install!
+        UI.warnings.length.should.equal(0)
```
amorde: This seems identical to the test above - is there expected to be different behavior if the URL host is `github.com`?

amorde: How would `git@github.com:kylef/test.git` and `git@bitbucket.org:kylef/test.git` differ?

dnkoutso: Well not so much actually...I used the same tests that were applied here https://github.com/CocoaPods/Core/pull/84/files

amorde: Huh, that's interesting..

dnkoutso: @amorde overall harmless I say :) but you were right to comment on it.

Issue #7793, opened 2018-05-29
[Master] OTHER_LDFLAGS should not be present for static library integrated targets


Whirlwind:
I use master branch of Cocoapods, but always get very many fails. I compare the xcconfigs between generated by Master and by v1.5.3, I found many differences:

1. master add `USE_HEADERMAP = NO`, it cause all import failed.
I found that when I import other files in my project, I always get the error which can not found the header files. But all files are in same project. I think the import behavior should could import all header files in same project without any changes.

2. When my target is a static library, The xcconfig which generated by v1.5.3 has not `OTHER_LDFLAGS`. But the master will link all dependencies. I think it should not link the dependency if target is library/framework.

Is it right?

Whirlwind:
The `USE_HEADEMAP` maybe is #1216 , the master branch lost this fix.

Whirlwind:
Hello, somebody could help me?

Whirlwind:
I know that.
But I must make a sure that it is a bug, then I could make the MR.

Whirlwind:
There are many `OTHER_LDFLAGS ` in the xcconfig when user target is a static library. And if the user target is static framework, the `OTHER_LDFLAGS` is empty.

dnkoutso:
re-opened, @Whirlwind care to make a PR? You did it already for one fo the two issues here. Try to compare 1.5.3 code and 1.6.0 (master). It changed rather significantly for performance reasons but there are obviously a few regressions.

Whirlwind:
Sorry, I try to do it, but I still can not find the reason.... :(

dnkoutso:
@Whirlwind I will check it out.

dnkoutso:
@Whirlwind This is what I see for 1.5.3 vs 1.6.0 (master)

1.5.3 Pods-LibraryDemo.debug
```
GCC_PREPROCESSOR_DEFINITIONS = $(inherited) COCOAPODS=1
HEADER_SEARCH_PATHS = $(inherited) "${PODS_ROOT}/Headers/Public" "${PODS_ROOT}/Headers/Public/AFNetworking" "${PODS_ROOT}/Headers/Public/SDWebImage"
LIBRARY_SEARCH_PATHS = $(inherited) "${PODS_CONFIGURATION_BUILD_DIR}/AFNetworking"
[...]

Issue #1216, opened 2013-07-22
A setting like `USE_HEADERMAPS` should be kept private.


alloy:
Should this be handled in a configurable way, i.e. specify it in the spec, or should this simply be added to the default settings to ignore?

For more info see: #1195.

/cc @irrationalfab, @dhinviethoa

Whirlwind:
I found this logic had been remove in the master branch, is it a bug?

PR #7808, opened 2018-06-03
Exclude `USE_HEADERMAP`

Whirlwind:
#1216 #7793

It seem that the code is missing.


dnkoutso:
Seems like a good regression fix. At least in terms of maintaining the same logic as we used to have before the revamped build settings.

segiddins:
Thanks for the PR!

PR #7812, opened 2018-06-05
Fix markdown in GitHub issue template

revolter:


revolter:
Wooow, this has been like this for the past 2 years 😲

orta:
Hah, nice catch

PR #7815, opened 2018-06-06
[Gemspec] Bump ruby-macho to '~> 1.2'

woodruffw:
macOS 10.14 introduces (or at least begins to use) two new load commands, so ruby-macho has been updated accordingly.

The 1.3 release will contain an option to ignore new load commands in the future, but 1.2 is just a stopgap.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -120,7 +120,7 @@ PATH
       gh_inspector (~> 1.0)
       molinillo (~> 0.6.5)
       nap (~> 1.0)
-      ruby-macho (~> 1.1)
+      ruby-macho (~> 1.2)
       xcodeproj (>= 1.5.8, < 2.0)
 
 GEM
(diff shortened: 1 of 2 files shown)


dnkoutso:
@woodruffw needs a `bundle install` I believe to pass CI.

dnkoutso:
@woodruffw is https://github.com/Homebrew/ruby-macho/commit/378aff7a09108e1f355e565f0d882cfe073d15b5 going to be a problem? I believe CocoaPods is meant to work with Ruby 2.0.0

woodruffw:
> is Homebrew/ruby-macho@378aff7 going to be a problem? I believe CocoaPods is meant to work with Ruby 2.0.0

Hmm, it could be. ruby-macho itself is still completely compatible with Ruby 2.0.0, but its development dependencies (namely RuboCop) aren't. I don't plan on making any changes in 1.3.0 that break that compatibility, but I don't actually have a clean way to test it in my current CI setup.

1.2.0 will continue to support Ruby 2.0.0 (I made that change after the release), while 1.3.0 and later will probably have Ruby 2.1.0 as a hard requirement (even if the library itself is technically 2.0.0 compatible).

dnkoutso:
Ah yes you are right it was after the release 1.2.0.

LGTM.

segiddins:
Development dependencies aren’t an issue

PR #7817, opened 2018-06-07
Do not add copy resources script phase for static library user targets

dnkoutso:
Given:
```ruby
target 'StaticLibrary'
  pod 'Something' # which includes resource bundles
end
```

CocoaPods integrates the `Copy Pods Resources` script phase  into user static library targets unnecessarily, including the input and output paths which could cause unnecessary and confusing diffs.

This change mirrors what we do for `add_embed_frameworks_script_phase` which is only integrated and added for targets that make sense.

### DIFF
--- a/lib/cocoapods/installer/user_project_integrator/target_integrator.rb
+++ b/lib/cocoapods/installer/user_project_integrator/target_integrator.rb
@@ -322,6 +322,8 @@ module Pod
         #
         def add_copy_resources_script_phase
           native_targets.each do |native_target|
+            # Static library targets cannot include resources. Skip this phase from being added instead.
+            next if native_target.symbol_type == :static_library
             script_path = target.copy_resources_script_relative_path
             resource_paths_by_config = target.resource_paths_by_config
             if resource_paths_by_config.values.all?(&:empty?)


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Do not add copy resources phase script for static library user targets  
```

</p>

lib/cocoapods/installer/user_project_integrator/target_integrator.rb
```diff
@@ -321,7 +321,7 @@ def add_pods_library
         # @return [void]
         #
         def add_copy_resources_script_phase
-          native_targets.each do |native_target|
+          native_targets_to_embed_in.each do |native_target|
```
dnkoutso: This mirrors https://github.com/CocoaPods/CocoaPods/blob/master/lib/cocoapods/installer/user_project_integrator/target_integrator.rb#L359

dnkoutso:
This actually might need a bit more. Right now the list used to ignore it does not include dynamic frameworks so basically it will break it for user dynamic framework targets.

Will follow up.

dnkoutso:
Ok updated to be a bit more scoped and easy. 

I know there is also "debouncing" mechanism on the script phase itself in case an env variable is not set but that does not prevent CocoaPods from "polluting" the users project with unnecessary input/output paths and a script phase addition.

Whirlwind:
@dnkoutso Could you remove exist build phase from previous version?

Issue #7814, opened 2018-06-05
Resolve symlink before call to `lipo` to compile in Xcode 10


DarkDust:
* [X] I've read and understood the [*CONTRIBUTING guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md). 🌈

# Report

## What did you do?

Tried to build my project in Xcode 10 beta 1 which includes a pod that wraps an existing framework. This framework contains i386 code which is now getting stripped.

## What did you expect to happen?

Correctly run the embed pods phase.

## What happened instead?

The "[CP] Embed Pods Frameworks" phase now fails on a pod that merely wraps a framework with the following error message:

    <path to myframework>: bundle format is ambiguous (could be app or framework)

### Stack

```
   CocoaPods : 1.5.3
        Ruby : ruby 2.2.2p95 (2015-04-13 revision 50295) [x86_64-darwin14]
    RubyGems : 2.4.8
        Host : Mac OS X 10.13.5 (17F70a)
       Xcode : 9.3.1 (9E501)
         Git : git version 2.15.1 (Apple Git-101)
Ruby lib dir : /Users/haisenko/.rvm/rubies/ruby-2.2.2/lib
Repositories : eqpodspecs - git@vcs.muc.equinux.net:equinux/eqPodSpecs.git @ e011b702a3bea20d71c595d0cef20b77061f7db0
               equinux-eqpodspecs - ssh://git@vcs.muc.equinux.net/equinux/eqPodSpecs.git @ e011b702a3bea20d71c595d0cef20b77061f7db0
               master - https://github.com/CocoaPods/Specs.git @ 8ef5b361deaaa25607575fa29d9a4a1b41078896
```

**NOTE**: Xcode is actually version 10.0 beta (10L176w)

### Installation Source

```
Executable Path: /Users/haisenko/.rvm/gems/ruby-2.2.2/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.2
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
cocoapods-trunk       : 1.3.0
cocoapods-try         : 1.1.0
```

# Analysis

The problem is that `lipo` replaces the symlink to the framework binary with the stripped binary. So with a structure like:

    My.framework/My -> Versions/A/My
    My.framework/Versions/A/My

[...]

Issue #7587, opened 2018-04-05
strip_invalid_archs() breaks vendored frameworks


CodingMarkus:
A framework looks like that:
```
FW.framework
FW.framework/FW -> Versions/Current/FW
FW.framework/Headers -> Versions/Current/Headers
FW.framework/Resources -> Versions/Current/Resources
FW.framework/Versions
FW.framework/Versions/A
FW.framework/Versions/Current -> A
FW.framework/Versions/A/FW
FW.framework/Versions/A/Headers
FW.framework/Versions/A/Resources
```
Note that the 2nd line above is a **symlink** to the real binary which resists in the `A` folder. However, if that binary is a fat binary and you call `strip_invalid_archs()` to thin it, the resulting framework looks like that:
```
FW.framework
FW.framework/FW
FW.framework/Headers -> Versions/Current/Headers
FW.framework/Resources -> Versions/Current/Resources
FW.framework/Versions
FW.framework/Versions/A
FW.framework/Versions/Current -> A
FW.framework/Versions/A/FW
FW.framework/Versions/A/Headers
FW.framework/Versions/A/Resources
```
Note that the 2nd line is **no symlink** anymore. Now the framework contains two binaries. A thin one ```FW.framework/FW``` and the original, unmodified fat one ```FW.framework/Versions/A/FW```.

This results in `codesign` to fail with the following error:

```bundle format is ambiguous (could be app or framework)```

**Comment by dnkoutso** [comment]
I assume this happens pre-1.5.0 too correct?

**Comment by dnkoutso** [comment]
@CodingMarkus can you also upload a quick sample app?

**Comment by CodingMarkus** [comment]
It happens because of this code line
```
lipo -remove "$arch" -output "$binary" "$binary" || exit 1
```
If you need a test universal binary, just download a freeware, like Bonjour Browser:
http://www.tildesoft.com/#RendezvousBrowser

And here is the problem:
```
# lipo -info "Bonjour Browser.app/Contents/MacOS/Bonjour Browser" 
Architectures in the fat file: Bonjour Browser.app/Contents/MacOS/Bonjour Browser are: ppc i386

# ln -s "Bonjour Browser.app/Contents/MacOS/Bonjour Browser" test

# ls -l test
...  test -> Bonjour Browser.app/Contents/MacOS/Bonjour Browser

# lipo -info test
Architectures in the fat file: test are: ppc i386 

[...]

PR #7819, opened 2018-06-07
Properly follow symlinks within macOS universal frameworks

dnkoutso:
closes https://github.com/CocoaPods/CocoaPods/issues/7814
closes https://github.com/CocoaPods/CocoaPods/issues/7587

@CodingMarkus I spent a bit more time on this and figured out why I could not repro. I did have a universal framework but:
```bash
  # Strip invalid architectures so "fat" simulator / device frameworks work on device
  if [[ "$(file "$binary")" == *"dynamically linked shared library"* ]]; then
    strip_invalid_archs "$binary"
  fi
```
This check was failing and so no stripping was happening at all, even though I had applied your fix within `strip_invalid_archs`.

That is because the result of `"$(file "$binary")"` returned `"/path/to/framework/binary : symbolic link to Versions/Current/framework"` and so it was never executing.

The fix was to properly figure out the final location of the binary was and propagate it through like you recommended but at an earlier stage.

@CodingMarkus, I'd appreciate it if you can manually apply the fix to your own `Pods-<yourapp>-frameworks.sh` file and verify that it works.

Also verified the same fix is not required if a `dSYM` is present. Architectures were properly stripped on the the `.framework` and the `dSYM`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
   [#7825](https://github.com/CocoaPods/CocoaPods/pull/7825)
 
+* Properly follow symlinks within macOS universal frameworks  
+  [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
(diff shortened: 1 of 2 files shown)


dnkoutso:
Integration specs https://github.com/CocoaPods/cocoapods-integration-specs/pull/167

Issue #7738, opened 2018-05-12
[1.5 swift] pod lib lint failed with '--use-libraries' param


0x5e:
## Step to reproduce

1. Run `pod lib create TYTestLib`
2. Add a swift pod depedency in podspec: `s.dependency 'Alamofire'`
3. Run `pod lib lint --allow-warnings --use-libraries --verbose`

```
.....

    ** BUILD FAILED **


    The following build commands failed:
    	Ld /Users/admin/Library/Developer/Xcode/DerivedData/App-cbhtfphhmqekmebvaiewmnbdhvxq/Build/Intermediates.noindex/App.build/Release-iphonesimulator/App.build/Objects-normal/x86_64/App normal x86_64
    	Ld /Users/admin/Library/Developer/Xcode/DerivedData/App-cbhtfphhmqekmebvaiewmnbdhvxq/Build/Intermediates.noindex/App.build/Release-iphonesimulator/App.build/Objects-normal/i386/App normal i386
    	CreateUniversalBinary /Users/admin/Library/Developer/Xcode/DerivedData/App-cbhtfphhmqekmebvaiewmnbdhvxq/Build/Products/Release-iphonesimulator/App.app/App normal i386\ x86_64
    	GenerateDSYMFile /Users/admin/Library/Developer/Xcode/DerivedData/App-cbhtfphhmqekmebvaiewmnbdhvxq/Build/Products/Release-iphonesimulator/App.app.dSYM /Users/admin/Library/Developer/Xcode/DerivedData/App-cbhtfphhmqekmebvaiewmnbdhvxq/Build/Products/Release-iphonesimulator/App.app/App
    	CodeSign /Users/admin/Library/Developer/Xcode/DerivedData/App-cbhtfphhmqekmebvaiewmnbdhvxq/Build/Products/Release-iphonesimulator/App.app
    (5 failures)
   Testing with `xcodebuild`.
 -> TYTestLib (0.1.0)
    - WARN  | summary: The summary is not meaningful.
    - WARN  | url: The URL (https://github.com/0x5e/TYTestLib) is not reachable.
    - WARN  | [iOS] swift: The validator used Swift 3.2 by default because no Swift version was specified. To specify a Swift version during validation, add the `swift_version` attribute in your podspec. Note that usage of the `--swift-version` parameter or a `.swift-version` file is now deprecated.
    - ERROR | [iOS] xcodebuild: Returned an unsuccessful exit code.
[...]

PR #7821, opened 2018-06-08
Validator adds a Swift file if any of the pod targets use Swift

dnkoutso:
closes https://github.com/CocoaPods/CocoaPods/issues/7738

If the pod being linted does not use Swift but any of its dependencies do, then a dummy Swift file is added to the 'App' in order to ensure Swift will be linked.

### DIFF
--- a/lib/cocoapods/generator/app_target_helper.rb
+++ b/lib/cocoapods/generator/app_target_helper.rb
@@ -38,18 +38,38 @@ module Pod
       # @param  [Symbol] platform
       #         the platform of the target. Can be `:ios` or `:osx`, etc.
       #
-      # @param  [Boolean] use_frameworks
-      #         whether to use frameworks or not when generating the contents of the import file.
+      # @param  [String] name
+      #         The name to use for the target, defaults to 'App'.
+      #
+      # @return [Array<PBXBuildFile>] the created build file references.
+      #
+      def self.add_app_project_import(project, target, pod_target, platform, name = 'App')
+        source_file = AppTargetHelper.create_app_import_source_file(project, pod_target, platform, name)
+        group = project[name] || project.new_group(name, name)
+        source_file_ref = group.new_file(source_file)
+        target.add_file_references([source_file_ref])
+      end
+
+      # Creates and links an empty Swift file for the given target.
+      #
+      # @param  [Project] project
+      #         the Xcodeproj to generate the target into.
+      #
+      # @param  [PBXNativeTarget] target
+      #         the native target to link the generated import file into.
       #
       # @param  [String] name
       #         The name to use for the target, defaults to 'App'.
       #
       # @return [Array<PBXBuildFile>] the created build file references.
       #
-      def self.add_app_project_import(project, target, pod_target, platform, use_frameworks, name = 'App')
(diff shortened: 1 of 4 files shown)


lib/cocoapods/generator/app_target_helper.rb
```diff
@@ -116,26 +113,23 @@ def self.add_swift_version(target, swift_version)
...
-      #
       # @param  [String] name
       #         The name of the folder to use and save the generated main file.
       #
       # @return [Pathname] the new source file that was generated.
       #
-      def self.create_app_import_source_file(project, pod_target, platform, use_frameworks, name = 'App')
-        language = pod_target.uses_swift? ? :swift : :objc
+      def self.create_app_import_source_file(project, pod_target, platform, name = 'App')
+        language = [pod_target, *pod_target.recursive_dependent_targets].any?(&:uses_swift?) ? :swift : :objc
```
dnkoutso: We now check if the target itself or any of dependents use Swift.

lib/cocoapods/generator/app_target_helper.rb
```diff
@@ -116,26 +113,23 @@ def self.add_swift_version(target, swift_version)
...
 
         if language == :swift
           source_file = project.path.dirname.+("#{name}/main.swift")
           source_file.parent.mkpath
-          import_statement = use_frameworks && pod_target.should_build? ? "import #{pod_target.product_module_name}\n" : ''
+          import_statement = pod_target.should_build? && pod_target.defines_module? ? "import #{pod_target.product_module_name}\n" : ''
```
dnkoutso: With Swift static libraries I believe this was outdated. We should attempt to import the module if the target supports it. No longer scoped to frameworks.

Is this correct?

lib/cocoapods/generator/app_target_helper.rb
dnkoutso: actually this can be a bit wrong as an ObjC pod will not be imported by the App unless it uses modules. 

The correct fix here is to just add a dummy swift file if any of the dependencies uses Swift but *not* the pod target itself.

lib/cocoapods/validator.rb
```diff
@@ -528,8 +528,9 @@ def add_app_project_import
       app_project = Xcodeproj::Project.open(validation_dir + 'App.xcodeproj')
       app_target = app_project.targets.first
       pod_target = @installer.pod_targets.find { |pt| pt.pod_name == spec.root.name }
-      Pod::Generator::AppTargetHelper.add_app_project_import(app_project, app_target, pod_target, consumer.platform_name, use_frameworks)
+      Pod::Generator::AppTargetHelper.add_app_project_import(app_project, app_target, pod_target, consumer.platform_name)
       Pod::Generator::AppTargetHelper.add_xctest_search_paths(app_target) if @installer.pod_targets.any? { |pt| pt.spec_consumers.any? { |c| c.frameworks.include?('XCTest') } }
+      Pod::Generator::AppTargetHelper.add_empty_swift_file(app_project, app_target) if @installer.pod_targets.any?(&:uses_swift?)
```
dnkoutso: this is the fix. If any of the pod targets use swift then we add a dumb file just to link Swift to the target app.

PR #7831, opened 2018-06-13
:sparkles: :tada: [fish] tab-completion support 

archmagees:
Add the tab-completion support for fish shell. 🎉

### DIFF
--- a/pod.fish
+++ b/pod.fish
@@ -1,0 +1,663 @@
+# Author: Arthur
+# Email:  archmagees+cocoapods@gmail.com
+# GitHub: https://github.com/archmagees
+function get_commands
+  for c in (commandline -opc)
+    if not string match -q -- '-*' $c
+      echo $c
+    end
+  end
+end
+
+function get_options
+  for c in (commandline -opc)
+    if string match -q -- '-*' $c
+      echo $c
+    end
+  end
+end
+
+function commands_syntax_is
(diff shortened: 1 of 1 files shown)


orta:
This is great, how do people use it?

dnkoutso:
I think you have to use http://fishshell.com/ ?

orta:
I'm a fish user 👍, but I don't see anything that would get shipped with the gem for example - so I'm interested in how anyone wanting to use this could get it all hooked up

archmagees:
@orta 
You just need to link it to the following directories:
```shell
cat $fish_complete_path
```
Just pick a directory, and put `pod.fish` in it.
Then you can enjoy the tab-completion for CocoaPods in fish shell.

dnkoutso:
LGTM. Don't know much but it seems like it works :)

segiddins:
This should probably be integrated into https://github.com/CocoaPods/claide-completion, since we try to avoid bundling auto-complete definitions directly into CocoaPods, since they require maintenance

amorde:
@archmagees interested in submitting this to https://github.com/CocoaPods/claide-completion? since these files won't ship with the gem, it's not easy for end users to install, and as mentioned above we'd prefer to keep shell completions out of this repo for easier maintenance

dnkoutso:
Closing for inactivity.

Issue #6074, opened 2016-10-24
The famous third-party framework does not include i386, how can I pass the "pod lib lint"?


mdsb100:
* [x] I've read and understood the [*CONTRIBUTING guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

I create a pod repo include the famous third-party framework. The framework does not include i386.

The framework is [TencentOpenAPI](http://qzonestyle.gtimg.cn/qzone/vas/opensns/res/doc/iOS_SDK_V3.1.0.zip). 

You must know Tencent in China! I have to use it in my pod repo!

## What did you expect to happen?

How can I pass the ```pod lib lint```? May I ignore the building of i386?

## What happened instead?

When I run ```pod lib lint```, it failed by missing i386 symbols.

## CocoaPods Environment

```
### Stack

```
   CocoaPods : 1.1.1
        Ruby : ruby 2.3.0p0 (2015-12-25 revision 53290) [x86_64-darwin15]
    RubyGems : 2.5.1
        Host : Mac OS X 10.11.6 (15G1004)
       Xcode : 8.0 (8A218a)
         Git : git version 2.8.4 (Apple Git-73)
Ruby lib dir : /Users/apple/.rvm/rubies/ruby-2.3.0/lib
Repositories : baidao-ios-ytx-pod-specs - http://gitlab.baidao.com/ios/ytx-pod-specs.git @ ccb4920607fd89f6ac85adb958e06571bfb9d9a6
               master - https://github.com/CocoaPods/Specs.git @ 58cb4955df65e47d5a5405c9e139b81d3a8b472c
```

### Installation Source

```
Executable Path: /Users/apple/.rvm/gems/ruby-2.3.0/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.1
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
cocoapods-trunk       : 1.1.1
cocoapods-try         : 1.1.0
```
```

## Project that demonstrates the issue
[TencentOpenAPI](http://qzonestyle.gtimg.cn/qzone/vas/opensns/res/doc/iOS_SDK_V3.1.0.zip)

orta:
This is realistically a dupe of https://github.com/CocoaPods/CocoaPods/issues/5854

benasher44:
[...]

Issue #5854, opened 2016-09-13
Pod lint fails when containing dynamic-frameworks without simulator architectures


eldewy:
Currently `pod trunk` is not usable for us as `pod lint` fails because simulator architectures (i386 and x86_64) are missing. BUT Apple rejects apps containing dynamic-frameworks with simulator architectures.

We deploy two variants of dynamic-framworks
- one with all device and simulator architectures (for debugging)
- one with device-only architectures (for app-submission).

`pod lint` fails with the following errors:

> x86_64 in file SampleSDK/iOS/Sample.framework/Sample (3 slices)
> - NOTE  | [SampleSDK/SampleSDK and more...] xcodebuild:  fatal error: lipo: -remove's specified would result in an empty fat file
> - NOTE  | [iOS] [SampleSDK/SampleSDK] xcodebuild:  ld: warning: ld: warning: ignoring file SampleSDK/iOS/Sample.framework/Sample, missing required architecture x86_64 in file 

We were forced to provide device-only architectures as Apple rejects app-submissions containing any simulator-architectures with

> ERROR: ERROR ITMS-90087: "Unsupported Architectures. The executable for Sample.app/Frameworks/Sample.framework contains unsupported architectures '[x86_64, i386]'."

Similar issues: 
https://github.com/CocoaPods/CocoaPods/issues/5275
http://stackoverflow.com/questions/36618252/cocoapods-podspec-push-without-build-simulator-architecture

Philipp372:
I am also affected by this problem, that I aim to distribute a dynamic framework pod without the simulator architectures as they are not allowed when submitting an app to the AppStore.
Cocoapods team, pls advise.

LeT0C:
Thank you benasher44, it indeed is the same problem for me :
https://github.com/CocoaPods/CocoaPods/issues/5877

We haven't submit to Apple yet, so I can't confirm eldewy's problem, but if it's the same here, it's a huge issue for my team.

Cocoapods team, could you please do something or advise us a working solution ?

Thanks

orta:
I'm not really sure of the answer here, 1.0 now has more strict linting rules because successfully working with other people's projects is becoming more complicated due to things like code-signing, frameworks and code-signing.

Perhaps a PR improving the linter specifically towards binary/dynamic framework pods that can take into account it's constraints better could be a way to move forwards?

eldewy:
@orta
This is a problem only with dynamic.frameworks...
[...]

Issue #5664, opened 2016-07-23
Undefined symbols for architecture i386 even with --use-libraries


igagis:
* [x ] I've read and understood the [*CONTRIBUTING guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

```
pod spec lint --use-libraries cocoapods/libmorda.podspec --sources=https://github.com/igagis/cocoapods-repo.git,https://github.com/CocoaPods/Specs.git --verbose
```

## What did you expect to happen?

Spec validates

## What happened instead?

Spec did not validate. There were linker errors like "Undefined symbols for architecture i386" and for x86_64 too. It was complaining about missing symbols from all dependency pods.

piece of build log:

```
Undefined symbols for architecture x86_64:
  "_OBJC_CLASS_$_GLKViewController", referenced from:
      _OBJC_CLASS_$_ViewController in liblibmorda.a(main-2DAD2A9D83562E83.o)
  "_OBJC_METACLASS_$_GLKViewController", referenced from:
      _OBJC_METACLASS_$_ViewController in liblibmorda.a(main-2DAD2A9D83562E83.o)
  "morda::createApp(int, char const**, utki::Buf<unsigned char>)", referenced from:
      -[AppDelegate application:didFinishLaunchingWithOptions:] in liblibmorda.a(main-2DAD2A9D83562E83.o)
  "std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >::find(char, unsigned long) const", referenced from:
      (anonymous namespace)::Parser::getNamespace(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) in liblibsvgdom.a(dom.o)
  "std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >::rfind(char, unsigned long) const", referenced from:
      papki::File::ext() const in liblibpapki.a(File.o)
      papki::File::dir() const in liblibpapki.a(File.o)
      papki::File::notDir() const in liblibpapki.a(File.o)
  "std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >::compare(char const*) const", referenced from:
      papki::FSFile::listDirContents(unsigned long) const in liblibpapki.a(FSFile.o)
[...]

PR #7836, opened 2018-06-14
Add a way to still build\lint iOS frameworks that may still include invalid archs

fischerdan:
There's a handful of issues stemming from xcodebuild taking the default route of building all architectures. This adds a very basic way to allow the build process to continue instead of forcing through via skipping the build step on lint or push.

https://github.com/CocoaPods/CocoaPods/issues/5854
https://github.com/CocoaPods/CocoaPods/issues/6074
https://github.com/CocoaPods/CocoaPods/issues/5664

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
+* Add `--build-active-ios-arch-only` parameter to `pod spec lint` and `pod lib lint` and `pod repo push` to allow linting and pushing of frameworks that may still have i386 support.  
+  [Dan Fischer](https://github.com/fischerdan)
+  [#5854](https://github.com/CocoaPods/CocoaPods/issues/5854)
+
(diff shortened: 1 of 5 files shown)


segiddins:
I'm not entirely sold on why this is necessary -- the active arch is arbitrary, based upon whichever simulator cocoapods has decided to build for. A better way to do this might be to check and see if the minimum deployment target is higher than one that supports 32bit ? We dont want to encourage pods shipping without support for all the platforms the SDK supports

fischerdan:
The gap seems to be caused by the gray area of support i386 has currently. If you want to support iOS 9, you tangentially have to support i386, even though an end user wouldn't be affected by 32 bit simulator support.

One of the issues this would resolve:
https://github.com/TextureGroup/Texture/issues/873#issuecomment-386036467

I suppose it could be argued that if you want to support iOS9 you have to support iOS9 simulators.

segiddins:
> I suppose it could be argued that if you want to support iOS9 you have to support iOS9 simulators.

That's my personal opinion

dnkoutso:
Closing stale PRs.

PR #7842, opened 2018-06-18
[PodTargetInstaller] Symlink static lib custom modulemaps into the Headers directory

segiddins:
This means we wont be overwriting the underlying file, so `pod install` wont cause full recompiles

### DIFF
--- a/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
@@ -645,7 +645,7 @@ module Pod
           def create_module_map(native_target)
             return super(native_target) unless custom_module_map
 
-            path = target.module_map_path
+            path = target.module_map_path_to_write
             UI.message "- Copying module map file to #{UI.path(path)}" do
               contents = custom_module_map.read
               unless target.requires_frameworks?
@@ -655,8 +655,14 @@ module Pod
               update_changed_file(generator, path)
               add_file_to_support_group(path)
 
+              linked_path = target.module_map_path
+              if path != linked_path
+                linked_path.dirname.mkpath
+                FileUtils.ln_sf(path, linked_path)
+              end
+
+              relative_path = target.module_map_path.relative_path_from(sandbox.root).to_s
               native_target.build_configurations.each do |c|
-                relative_path = path.relative_path_from(sandbox.root)
                 c.build_settings['MODULEMAP_FILE'] = relative_path.to_s
               end
             end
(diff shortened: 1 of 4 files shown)


lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -644,7 +644,7 @@ def apply_xcconfig_file_ref_to_resource_bundle_targets(resource_bundle_targets,
           def create_module_map(native_target)
             return super(native_target) unless custom_module_map
 
-            path = target.module_map_path
+            path = Target.instance_method(:module_map_path).bind(target).call
```
dnkoutso: can you explain why this change?

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -654,6 +654,11 @@ def create_module_map(native_target)
               update_changed_file(generator, path)
               add_file_to_support_group(path)
 
+              if path != (linked_path = target.module_map_path)
```
dnkoutso: can this be tested?

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -776,6 +781,11 @@ def add_placeholder_target
           # @return [Void]
           #
           def add_swift_static_library_compatibility_header_phase(native_target)
+            if custom_module_map
+              raise Informative, 'Using swift static libraries with custom module maps does not yet work.' \
```
dnkoutso: nit: "Swift"

dnkoutso:
How were we overwriting the file? I thought `update_changed_file` was not changing the contents of files written.

dnkoutso:
Overall LGTM!

segiddins:
> How were we overwriting the file? I thought update_changed_file was not changing the contents of files written.

Because it was in `Pods/Headers`, it was always getting deleted in `Installer#clean_sandbox`

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
segiddins: sure, will add this in a comment or factor this to make it less hideous:

We want always to store the actual file in the `Target Support Files` directory in the sandbox, but the `PodTarget#module_map_path` method refers to the path used to _import_ the module map in build settings, which needs to be next to the headers in question

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
segiddins: integration specs cover this pretty thoroughly, I can look into adding a unit test

segiddins:
If this looks good, I'll go ahead and update the submodule and write some unit tests

dnkoutso:
👍 verified in a sample project that rebuilds do not happen.

Issue #7573, opened 2018-04-02
Do not group test specs by type and instead create a test bundle for each test spec


dnkoutso:
Given two test specs within a podspec, CocoaPods should create a separate test bundle for each. Right now (<= 1.5.0) CocoaPods will group everything by test type.

Today this causes problems where a test spec might need an app host while another does not. Furthermore, library authors would like to group different tests together like ViewSnapshot tests and Unit tests.

dnkoutso:
Basically revive this https://github.com/CocoaPods/CocoaPods/pull/6981

PR #7843, opened 2018-06-18
Add new spec fixture for multiple test spec support

dnkoutso:
This makes it easier to write test cases for https://github.com/CocoaPods/CocoaPods/issues/7573


spec/unit/installer/xcode/pods_project_generator/pod_target_integrator_spec.rb
```diff
@@ -85,13 +86,7 @@ class PodTargetIntegrator
               end
 
               it 'excludes test framework and resource paths from dependent targets' do
-                test_framework_paths = [{ :name => 'TestVendored.framework',
```
dnkoutso: this test was a false positive. Now its more real by ensuring that a resource bundle of another test spec is not part of the input/output paths of another test native target.

dnkoutso: It was a false positive because the `banana_pod_target` never had test specs so it was never loading or returning paths for a test spec.

```ruby
-                @banana_pod_target.stubs(:resource_paths).returns('BananaLib' => [], 'BananaLib/Tests' => test_resource_paths)
```

this was stupid.

dnkoutso:
All icons used here are also free to use.

PR #7845, opened 2018-06-19
Performance improvements & refactoring

amorde:
Used the built in support for ruby-prof to try and profile some of the heavy areas. I plan to do some more but need to get a better test project to measure the performance changes with.

@dnkoutso 
@segiddins can you try this with one of your projects?

PR #7851, opened 2018-06-19
Fix PBXAggregateTargets to include an xcconfig

dnkoutso:
This is a master only issue.

In 1.6.0 we changed it so targets that `should_build?` return `false` will now integrate a PBXAggregateTarget and return early.

By returning early we no longer create an xcconfig for these targets (which 1.5.3 used to do) which means none of the other variables such as `PODS_TARGET_SRCROOT` are available. If this target includes a script phase then it will not be able to reference this env variable or other `PODS_*` variables

This change fixes this regression and ensures an `xcconfig` is created for these targets.

### DIFF
--- a/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
@@ -29,6 +29,7 @@ module Pod
           #
           def install!
             UI.message "- Installing target `#{target.name}` #{target.platform}" do
+              create_support_files_dir
               test_file_accessors, file_accessors = target.file_accessors.partition { |fa| fa.spec.test_specification? }
 
               unless target.should_build?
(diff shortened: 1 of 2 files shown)


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Fix PBXAggregateTargets to include an xcconfig  
```

</p>

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -29,18 +29,18 @@ def initialize(sandbox, project, target, umbrella_headers_by_dir = nil)
           #
           def install!
             UI.message "- Installing target `#{target.name}` #{target.platform}" do
+              create_support_files_dir
```
amorde: Reminder to myself to update #7845 to include this scenario if necessary - there I changed it so that we clean at the end of install, but I might need to add a call to clean within the `unless target.should_build?` section

amorde: Whoops, totally misread this, thought it was manually creating the temporary directory.

PR #7862, opened 2018-06-20
[Installer] Avoid locking&unlocking sources if there are no plugin post-install hooks

segiddins:
This should provide a nice little performance boost in that case

cc @amorde

### DIFF
--- a/lib/cocoapods/hooks_manager.rb
+++ b/lib/cocoapods/hooks_manager.rb
@@ -76,6 +76,20 @@ module Pod
         @registrations[hook_name] << Hook.new(hook_name, plugin_name, block)
       end
 
+      # Returns all the hooks to run for the given event name
+      # and set of whitelisted plugins
+      #
+      # @see run
+      #
+      # @return [Array<Hook>] the hooks to run
+      #
+      def hooks_to_run(name, whitelisted_plugins = nil)
+        return [] unless registrations
+        hooks = registrations.fetch(name, [])
+        return hooks unless whitelisted_plugins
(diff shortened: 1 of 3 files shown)


lib/cocoapods/installer.rb
```diff
@@ -422,11 +422,9 @@ def run_plugins_pre_install_hooks
     # @return [void]
     #
     def perform_post_install_actions
-      unlock_pod_sources
       run_plugins_post_install_hooks
       warn_for_deprecations
       warn_for_installed_script_phases
-      lock_pod_sources
```
amorde: Don't we still want the sources to end up locked after installation? Or is there another mechanism for making that happen?

segiddins: oof... good catch. at the very least, we can skip unlocking ? So 1/2 as much benefit?

lib/cocoapods/hooks_manager.rb
```diff
@@ -94,23 +108,21 @@ def run(name, context, whitelisted_plugins = nil)
...
-                  else
-                    block.call(context)
-                  end
-                end
+        hooks = hooks_to_run(name, whitelisted_plugins)
+        return if hooks.empty?
+
+        UI.message "- Running #{name.to_s.tr('_', ' ')} hooks" do
+          hooks.each do |hook|
+            next if whitelisted_plugins && !whitelisted_plugins.key?(hook.plugin_name)
```
amorde: Shouldn't need to check this anymore since `hooks_to_run` handles the whitelist

PR #7868, opened 2018-06-22
Reduce memory allocation in TargetInstaller#update_changed_file

amorde:
Currently `update_changed_file` writes the new file to disk, checks if they are the same with `FileUtils#identical?` (which then loads both files into memory) and then moves the new file over if the contents have changed.

By avoiding reloading the file that we just wrote back into memory, we can save a large amount of memory allocation. Since garbage collection accounts for a major portion of `pod install` times, this should hopefully help.

### Memory Allocation during `pod install`
Tested using the `memory_profiler` gem

(note: this was tested on a test project meant to stress CocoaPods, and may not be representative of the average project that utilizes CocoaPods)

#### Before

Total allocated: 12547967800 bytes (**11.69 GB**)
Total retained:  14801813 bytes  (**14.8 MB**)

#### After

Total allocated: 6329712267 bytes (12363884 objects) (**5.9 GB**)
Total retained:  14801970 bytes (171337 objects) (**14.8 MB**)

##

Although total retained remains the same, total allocated was cut almost in half

cc @segiddins

### DIFF
--- a/lib/cocoapods/generator/dummy_source.rb
+++ b/lib/cocoapods/generator/dummy_source.rb
@@ -11,12 +11,13 @@ module Pod
       # @return [String] the string contents of the dummy source file.
       #
       def generate
-        result = ''
-        result << "#import <Foundation/Foundation.h>\n"
-        result << "@interface #{class_name} : NSObject\n"
-        result << "@end\n"
-        result << "@implementation #{class_name}\n"
-        result << "@end\n"
+        result = <<-source.strip_heredoc
+          #import <Foundation/Foundation.h>
+          @interface #{class_name} : NSObject
+          @end
+          @implementation #{class_name}
+          @end
+        source
         result
       end
 
--- a/lib/cocoapods/generator/dummy_source.rb
+++ b/lib/cocoapods/generator/dummy_source.rb
@@ -8,13 +8,21 @@ module Pod
         @class_name = "PodsDummy_#{validated_class_name_identifier}"
       end
 
+      # @return [String] the string contents of the dummy source file.
+      #
+      def generate
+        result = ''
+        result << "#import <Foundation/Foundation.h>\n"
+        result << "@interface #{class_name} : NSObject\n"
+        result << "@end\n"
+        result << "@implementation #{class_name}\n"
+        result << "@end\n"
+        result
+      end
+
       def save_as(pathname)
         pathname.open('w') do |source|
-          source.puts '#import <Foundation/Foundation.h>'
-          source.puts "@interface #{class_name} : NSObject"
-          source.puts '@end'
-          source.puts "@implementation #{class_name}"
-          source.puts '@end'
+          source.write(generate)
         end
       end
     end
--- a/lib/cocoapods/installer/xcode/pods_project_generator/target_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/target_installer.rb
@@ -110,21 +110,12 @@ module Pod
           #
           def update_changed_file(generator, path)
             if path.exist?
-              if !generator.respond_to?(:generate)
-                # the generator doesn't support generating string contents, so write it to a tmp file and compare
-                # it with the existing file
-                generator.save_as(support_files_temp_dir)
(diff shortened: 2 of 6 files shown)


lib/cocoapods/installer/xcode/pods_project_generator/target_installer.rb
```diff
@@ -106,19 +108,27 @@ def custom_build_settings
           #
           # Saves the content the provided path unless the path exists and the contents are exactly the same.
           #
-          # @return [Object] the result of the generator.
```
amorde: The return value from here was never actually used anywhere

lib/cocoapods/installer/xcode/pods_project_generator/target_installer.rb
```diff
@@ -106,19 +108,27 @@ def custom_build_settings
...
-          #
           def update_changed_file(generator, path)
             if path.exist?
-              result = generator.save_as(support_files_temp_dir)
-              unless FileUtils.identical?(support_files_temp_dir, path)
-                FileUtils.mv(support_files_temp_dir, path)
+              if !generator.respond_to?(:generate)
+                # the generator doesn't support generating string contents, so write it to a tmp file and compare
```
amorde: The one generator that broke this is the `plist` generator - it uses a function in `Xcodeproj` that only accepts a path. I could work on getting that one but I think getting all the other generators to support this is still worth it

CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Reduce memory allocation in TargetInstaller#update_changed_file  
```

</p>

lib/cocoapods/installer/xcode/pods_project_generator/target_installer.rb
dnkoutso: it was being used but I think it no longer did due to the build settings re-write.

dnkoutso:
Is there a way to define an interface in Ruby? Something that all of this classes can conform to that provides the `generate` method?

dnkoutso:
This is awesome

lib/cocoapods/generator/dummy_source.rb
```diff
@@ -8,13 +8,21 @@ def initialize(class_name_identifier)
         @class_name = "PodsDummy_#{validated_class_name_identifier}"
       end
 
+      # @return [String] the string contents of the dummy source file.
+      #
+      def generate
+        result = ''
```
segiddins: this could use a HEREDOC to avoid needing to allocate 5 intermediary strings

lib/cocoapods/installer/xcode/pods_project_generator/target_installer.rb
segiddins: could use the underlying serializer from nanaimo if you wanted:

```
plist = Nanaimo::Plist.new(hash, :xml)
Nanaimo::Writer::XMLWriter.new(plist, :pretty => true, :output => f, :strict => false).write
```

amorde: Yup, I thought of that too. Would need to use another StringIO here so it can "write" to it then pull the string out.

I'll go ahead and do that 👍

segiddins: 👍

amorde:
@dnkoutso I don't think there's anything like interfaces in Ruby, but I'm still a novice in Ruby

dnkoutso:
I believe you do that via `Modules` see https://metabates.com/2011/02/07/building-interfaces-and-abstract-classes-in-ruby/.

It has already been done in other places within CocoaPods code.

dnkoutso:
Don't let me derail you with this module stuff by the way.

amorde:
Yeah, I think I'm good with this for now since it is covered by integration tests, unless you guys disagree

Issue #7546, opened 2018-03-26
Can not strip the frameworks with single quote in the name


hydro0:
# Report

## What did you do?

Build project that contains single quotes `'` in its name and uses `CocoaPods`.

## What did you expect to happen?

Successful build. 

## What happened instead?

The build was failed in the `Run Script` phase, script name `$PROJECT_NAME-frameworks.sh`. 

### Stack

```
   CocoaPods : 1.4.0
        Ruby : ruby 2.2.5p319 (2016-04-26 revision 54774) [x86_64-darwin15]
    RubyGems : 2.4.8
        Host : Mac OS X 10.13.3 (17D102)
       Xcode : 9.2 (9C40b)
         Git : git version 2.14.3 (Apple Git-98)
Ruby lib dir : /Users/orestsavchak/.rvm/rubies/ruby-2.2.5/lib
Repositories : master - https://github.com/CocoaPods/Specs.git @ e51b40cfb010e9383e38c543fba86750fc6f2030
```

### Installation Source

```
Executable Path: /Users/orestsavchak/.rvm/gems/ruby-2.2.5/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.2
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
cocoapods-trunk       : 1.3.0
cocoapods-try         : 1.1.0
```

### Podfile

```ruby
# Uncomment the next line to define a global platform for your project
platform :ios, '9.0'

target 'Twist \'n\' Slide' do
  # Uncomment the next line if you're using Swift or would like to use dynamic frameworks
   use_frameworks!

  # Pods for Twist 'n' Slide
  pod "GetSocial/Core"
  pod "GetSocial/UI"
end
```

## Project that demonstrates the issue

[GitHub example](https://github.com/hydro0/Twist-n-Slide)

## Workaround

I've fixed that by replacing single quotes with double quotes on the line `98` in the `$PROJECT_NAME-frameworks.sh`.

segiddins:
- Seems like we may want to `shellescape` the strings we interpolate into build script phases

pcarollo84:
Hi, I think I found a solution. I have create this [PR](https://github.com/CocoaPods/CocoaPods/pull/7870). 
I couldn't find any similar unit test that cover this part. But if there are let me know that I will happily write them.

Issue #6121, opened 2016-11-02
Embed Pod Frameworks fails if the target has spaces in the name


harlanhaskins:
* [x] I've read and understood the [*CONTRIBUTING guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).
🌈

# Report

## What did you do?

Attempt to build my project after running `pod install`

## What did you expect to happen?

A successful build

## What happened instead?

A cryptic error message, because the frameworks path is not quoted:
```
/usr/bin/codesign --force --sign 3529B9D2157072B8C666FCDCB5561B2FB83CD9F2  --preserve-metadata=identifier,entitlements /Users/harlan/Library/Developer/Xcode/DerivedData/Bryx_911-etlctfgizhnmsigivwwqqolxshkr/Build/Products/Debug-iphoneos/Bryx 911.app/Frameworks/Alamofire.framework
/Users/harlan/Library/Developer/Xcode/DerivedData/Bryx_911-etlctfgizhnmsigivwwqqolxshkr/Build/Products/Debug-iphoneos/Bryx: No such file or directory
```

### Stack

```
   CocoaPods : 1.2.0.beta.1
        Ruby : ruby 2.0.0p648 (2015-12-16 revision 53162) [universal.x86_64-darwin16]
    RubyGems : 2.0.14.1
        Host : Mac OS X 10.12.2 (16C32e)
       Xcode : 8.1 (8B62)
         Git : git version 2.7.0
Ruby lib dir : /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib
Repositories : master - https://github.com/CocoaPods/Specs.git @ b558170a7d2e713784f599de9e9c799368474e3a
```

### Installation Source

```
Executable Path: /usr/local/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.1
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
cocoapods-trunk       : 1.1.1
cocoapods-try         : 1.1.0
```

### Podfile

```ruby
source 'https://github.com/CocoaPods/Specs.git'
platform :ios, "8.0"
use_frameworks!
inhibit_all_warnings!

def shared_pods
    pod 'FormatterKit'
    pod 'FXKeychain'
    pod 'RMPhoneFormat', :git => 'https://github.com/davbeck/RMPhoneFormat.git'
[...]

PR #7870, opened 2018-06-22
Fix for #7546 Can not strip the frameworks in project with single quote in the name

pcarollo84:
🌈 
When code signing there wasn't any escaping for the framework path. If the path contained a single quote it fails with the following error
``` bash
.../Pods/Target Support Files/Pods-campusM/Pods-campusM-frameworks.sh: eval: line 113: unexpected EOF while looking for matching `''
```
as mentioned on issue #7546 

I fixed replacing the single quote `'` with `'\''` as mentioned on this answer on [stack overflow](https://stackoverflow.com/questions/32122586/curl-escape-single-quote) directly in the shell script. 
I didn't find any similar unit test available for this PR. 
But if there are please let me know which and I will happily write relevant unit tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
   [#7883](https://github.com/CocoaPods/CocoaPods/pull/7883)
   
+* Can not strip the frameworks in project with single quote in the name   
+  [Paolo Carollo](https://github.com/pcarollo84)
+  [#7546](https://github.com/CocoaPods/CocoaPods/issues/7546)
+
 * Set `CURRENT_PROJECT_VERSION` for generated app host targets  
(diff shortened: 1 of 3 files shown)


dnkoutso:
@pcarollo84 thanks! One of a things I remember is that various changes like these have caused other issues in the past.

See this issue: https://github.com/CocoaPods/CocoaPods/issues/6121 we need to ensure it does not regress.

There are only integration specs that test this bash logic (https://github.com/CocoaPods/cocoapods-integration-specs) would be AMAZING to add a new integration spec with various different cases of framework names, one with spaces and one with a single quote for example.

Really thanks for this!

dnkoutso:
@pcarollo84 any chance of pushing this forward?

pcarollo84:
Hi @dnkoutso! Yes, I will! Sorry I couldn't work on it in the past days. I will add some new integration tests and update the pull request soon.

dnkoutso:
@pcarollo84 ping on this.

dnkoutso:
Sorry going to close this one as it has been idle for too long. Feel free to re-open with a test if you want to land it.

PR #7876, opened 2018-06-24
Remove handling for CocoaPods 0.35.0 backwards compatibility in resolver

amorde:


dnkoutso:
@amorde still want a changelog?

amorde:
I think it’s probably fine without one honestly

PR #7878, opened 2018-06-25
change shell script relative path to ${PODS_ROOT} instead of ${SRCROOT}/Pods

Whirlwind:
Keep all script path is relative the Pods folder, rather than `SRCROOT`. So we could use the symlink of `Pods` folder.

This is a separate PR for #7781.

PR #7879, opened 2018-06-25
Add support if project path is symbol link

Whirlwind:
Re-create a separate PR from https://github.com/CocoaPods/CocoaPods/pull/7781, close #7781 .

First, I keep some symlink paths, not resolve to `realpath`. Just remove `.././` with `cleanpath`.

Then, If the user project is symlink, I create the `Pods` symlink to the `Pods` `realpath`, so user project will keep `PODS_ROOT=${SRCROOT}/Pods`. If developer makes multiple symlinks to a project, CP will just update the symlink target instead of change all settings (eg: environment paths, script paths in user project)

If the dependency is local, I create the dependency symlink in Pods directory : `AFNetworking -> ../AFNetworking`,
So the user project will access the dependency directory by `${PODS_ROOT}/AFNetworking` instead of `${SRCROOT}/../../../AFNetworking`.

Please review #7878 first.


lib/cocoapods/installer/analyzer.rb
```diff
@@ -658,7 +658,22 @@ def generate_pod_target(target_definitions, target_inspections, specs, scope_suf
         platform = determine_platform(specs, target_definitions, host_requires_frameworks)
         file_accessors = create_file_accessors(specs, platform)
         PodTarget.new(sandbox, host_requires_frameworks, user_build_configurations, archs, platform, specs,
-                      target_definitions, file_accessors, scope_suffix)
+                      target_definitions, file_accessors, scope_suffix).tap do |target|
```
segiddins: this looks like the wrong place to be doing this, it should be done in an installer

lib/cocoapods/installer/user_project_integrator/target_integrator.rb
```diff
@@ -412,6 +413,24 @@ def add_check_manifest_lock_script_phase
           end
         end
 
+        # Adds a symlink at $PROJECT_DIR/Pods pointing to $PODS_ROOT if the user project is behind a symlink
+        #
+        # @note Replaces any existing symlinks at $PROJECT_DIR/Pods
+        #
+        # @return [void]
+        #
+        def add_sandbox_symlink
```
segiddins: since this has nothing to do with the particular target, it probably belongs in the `UserProjectIntegrator`

lib/cocoapods/sandbox.rb
```diff
@@ -150,7 +163,7 @@ def target_support_files_dir(name)
     #
     # @return [Pathname] the path of the Pod.
     #
-    def pod_dir(name)
+    def pod_realdir(name)
```
segiddins: this seems weird to me, why do we need two methods for the `pod_dir` ?

lib/cocoapods/sandbox/path_list.rb
```diff
@@ -44,6 +44,32 @@ def dirs
...
+      #         files in {dir}. The paths are relative from
+      #
+      def list_files_under(dir, basepath = nil)
+        dirs = []
+        files = []
+        basepath = dir.realpath unless basepath
+        dir.children.each do |path|
+          if path.directory?
+            dirs << path.to_s
+            if !path.symlink? || path.realpath.relative_path_from(basepath).to_s.start_with?('..')
```
segiddins: using `relative_path_from` instead of the string slicing here will be _incredibly_ expensive

spec/unit/installer/post_install_hooks_context_spec.rb
```diff
@@ -3,8 +3,8 @@
 module Pod
   describe Installer::PostInstallHooksContext do
     it 'offers a convenience method to be generated' do
-      pods_project = Project.new('/path/Pods.xcodeproj')
-      sandbox = stub(:root => '/path', :project => pods_project)
+      pods_project = Project.new('/tmppath/Pods.xcodeproj')
```
segiddins: why have these specs changed?

lib/cocoapods/installer/user_project_integrator/target_integrator.rb
Whirlwind: If the Pod is a local development pod, I will create the symlink, too. So it is in `TargetIntegrator`.

CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Add support if project path is symbol link  
```

</p>

lib/cocoapods/installer/analyzer.rb
segiddins: I don't understand

lib/cocoapods/installer/user_project_integrator/target_integrator.rb
segiddins: but this still has nothing to do with an individual _target_, only an individual _project_

lib/cocoapods/sandbox.rb
amorde: would calling `realpath` on the result of `pod_dir` be different from this implementation?

Whirlwind: I create a new `pod_dir` method, and is always `${PODS_ROOT}/PODNAME`. It is symbol link path when the pod is local pod ( with `:path` ).
So this `pod_realdir` will return a realpath of the pod.

amorde:
@Whirlwind thinking about this some more, I feel like this is a lot of changes for something that isn't used that often. I'm concerned about our ability to maintain this going forward. It also has a significant performance impact.

This might be something that would be better suited as a plugin. Plugins have access to all of Ruby's monkey patching, it's not ideal but this is a very specific use case

Whirlwind:
I am sorry to hear that. I have many plugins to hook the CocoaPods. But this PR changes too many logic and it is hard to hook. eg: changed `realpath` to `cleanpath`.

If the CocoaPods methods are changed, the plugin will not work. So I hope that CocoaPods support the symlink directly, and use the so many test to keep them work.

amorde:
I definitely understand, I think specifically the use case of having the user project be a symlink is too much of an edge case to maintain support for, but symlinks in local/development pods is definitely something that has been discussed before.

The ideal scenario from my perspective:

* Symlinks are supported for development pods
* The performance penalty of resolving symlinks only impacts development pods, since externally sourced pods will have their sources copied into the sandbox
* Supporting user projects that are symlinks is kept outside of CocoaPods in its own plugin

amorde:
As this has become out of date and is currently not a change that is likely to be accepted, I'm going to close this. Comments [here](https://github.com/CocoaPods/CocoaPods/pull/7879#issuecomment-414787792) still stand though 👍

PR #7882, opened 2018-06-25
Extract app host out of pod target installation

dnkoutso:
`pod_target_installer.rb` is becoming too big and since we are using only one app host per test type it does not make sense to keep it within `pod_target_installer`. We also had to debounce to prevent adding an app host many times.

### DIFF
--- a/lib/cocoapods/installer/xcode/pods_project_generator.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator.rb
@@ -4,12 +4,14 @@ module Pod
       # The {PodsProjectGenerator} handles generation of the 'Pods/Pods.xcodeproj'
       #
       class PodsProjectGenerator
+        require 'cocoapods/installer/xcode/pods_project_generator/target_installer_helper'
         require 'cocoapods/installer/xcode/pods_project_generator/pod_target_integrator'
         require 'cocoapods/installer/xcode/pods_project_generator/target_installer'
         require 'cocoapods/installer/xcode/pods_project_generator/target_installation_result'
         require 'cocoapods/installer/xcode/pods_project_generator/pod_target_installer'
         require 'cocoapods/installer/xcode/pods_project_generator/file_references_installer'
         require 'cocoapods/installer/xcode/pods_project_generator/aggregate_target_installer'
+        require 'cocoapods/installer/xcode/pods_project_generator/app_host_installer'
 
         # @return [Sandbox] The sandbox where the Pods should be installed.
         #
@@ -63,9 +65,10 @@ module Pod
         def generate!
           prepare
           install_file_references
-          @target_installation_results = install_libraries
+          @target_installation_results = install_targets
           integrate_targets(@target_installation_results.pod_target_installation_results)
-          wire_target_dependencies(@target_installation_results)
+          app_hosts_by_host_key = install_app_hosts
+          wire_target_dependencies(@target_installation_results, app_hosts_by_host_key)
           @target_installation_results
         end
 
@@ -114,6 +117,7 @@ module Pod
 
         private
 
+        AppHostKey = Struct.new(:test_type, :platform)
         InstallationResults = Struct.new(:pod_target_installation_results, :aggregate_target_installation_results)
 
         def create_project
@@ -175,7 +179,7 @@ module Pod
           installer.install!
         end
 
-        def install_libraries
+        def install_targets
           UI.message '- Installing targets' do
             umbrella_headers_by_dir = pod_targets.map do |pod_target|
               next unless pod_target.should_build? && pod_target.defines_module?
@@ -200,6 +204,36 @@ module Pod
             end]
 
             InstallationResults.new(pod_target_installation_results, aggregate_target_installation_results)
+          end
+        end
+
+        def install_app_hosts
+          pod_target_with_test_specs = pod_targets.reject do |pod_target|
+            pod_target.test_specs.empty? || pod_target.test_spec_consumers.none?(&:requires_app_host?)
+          end
+
+          return if pod_target_with_test_specs.empty?
+
+          UI.message '- Installing app hosts' do
+            app_host_keys = pod_target_with_test_specs.flat_map do |pod_target|
+              pod_target.supported_test_types.flat_map do |test_type|
+                AppHostKey.new(test_type, pod_target.platform)
+              end.uniq
+            end
+
+            app_host_keys_by_test_type = app_host_keys.group_by do |app_host_key|
+              [app_host_key.test_type, app_host_key.platform.symbolic_name]
+            end
+
+            app_host_keys_by_test_type.map do |(test_type, platform_symbol), keys|
+              deployment_target = keys.map { |k| k.platform.deployment_target }.max
+              platform = Platform.new(platform_symbol, deployment_target)
+              AppHostKey.new(test_type, platform)
+            end
+
+            Hash[app_host_keys.map do |app_host_key|
+              [app_host_key, AppHostInstaller.new(sandbox, project, app_host_key.platform, app_host_key.test_type).install!]
+            end]
           end
         end
 
@@ -239,9 +273,12 @@ module Pod
         #         the installation results that were produced when all targets were installed. This includes
         #         pod target installation results and aggregate target installation results.
         #
+        # @param  [Hash{AppHostKey=>Array<PBXNativeTarget>}] app_hosts_by_host_key
+        #         the app hosts by test type that were installed in #install_app_hosts
+        #
         # @return [void]
         #
-        def wire_target_dependencies(target_installation_results)
+        def wire_target_dependencies(target_installation_results, app_hosts_by_host_key)
           frameworks_group = project.frameworks_group
           pod_target_installation_results_hash = target_installation_results.pod_target_installation_results
           aggregate_target_installation_results_hash = target_installation_results.aggregate_target_installation_results
@@ -294,10 +331,13 @@ module Pod
                   end
                   test_native_target.add_dependency(dependency_installation_result.native_target)
                   add_framework_file_reference_to_native_target(test_native_target, pod_target, test_dependent_target, frameworks_group)
-                  test_spec_consumers = test_specs.map { |test_spec| test_spec.consumer(pod_target.platform) }
-                  if test_spec_consumers.any?(&:requires_app_host?)
-                    app_host_target = project.targets.find { |t| t.name == pod_target.app_host_label(test_specs.first.test_type) }
-                    test_native_target.add_dependency(app_host_target)
+                  # Wire app host dependencies to test native target
+                  if pod_target.test_spec_consumers.any?(&:requires_app_host?)
+                    pod_target.supported_test_types.each do |test_type|
+                      app_host_target = app_hosts_by_host_key[AppHostKey.new(test_type, pod_target.platform)]
+                      test_native_target.add_dependency(app_host_target)
+                      configure_app_host_to_native_target(app_host_target, test_native_target)
+                    end
                   end
                 end
               end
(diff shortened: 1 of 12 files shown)


dnkoutso:
Build will fail

CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* [WIP] extract app host out of pod target installation  
```

</p>

lib/cocoapods/installer/xcode/pods_project_generator.rb
```diff
@@ -294,10 +329,15 @@ def wire_target_dependencies(target_installation_results)
...
                   test_native_target.add_dependency(dependency_installation_result.native_target)
                   add_framework_file_reference_to_native_target(test_native_target, pod_target, test_dependent_target, frameworks_group)
-                  test_spec_consumers = test_specs.map { |test_spec| test_spec.consumer(pod_target.platform) }
-                  if test_spec_consumers.any?(&:requires_app_host?)
-                    app_host_target = project.targets.find { |t| t.name == pod_target.app_host_label(test_specs.first.test_type) }
-                    test_native_target.add_dependency(app_host_target)
+                  # Wire app host dependencies to test native target
+                  if pod_target.test_spec_consumers.any?(&:requires_app_host?)
+                    pod_target.supported_test_types.each do |test_type|
+                      app_host_target = app_hosts_by_host_key.select do |app_host_key|
```
dnkoutso: I think you can use `find` here instead of `select`

lib/cocoapods/installer/xcode/pods_project_generator/app_host_installer.rb
```diff
@@ -0,0 +1,134 @@
...
+          #
+          attr_reader :test_type
+
+          # Initialize a new instance
+          #
+          # @param [Pod::Project] project @see #project
+          # @param [Platform] platform @see #platform
+          # @param [Symbol] test_type @see #test_type
+          #
+          def initialize(sandbox, project, platform, test_type)
```
dnkoutso: nit: add `sandbox` param documentation to the initializer

lib/cocoapods/installer/xcode/pods_project_generator/app_host_installer.rb
```diff
@@ -0,0 +1,134 @@
...
+          #
+          # @param  [Platform] platform
+          #         the platform to use for when generating this Info.plist file.
+          #
+          # @param  [Symbol] bundle_package_type
+          #         the CFBundlePackageType of the target this Info.plist file is for.
+          #
+          # @return [void]
+          #
+          def create_info_plist_file(path, native_target, version, platform, bundle_package_type = :fmwk)
```
dnkoutso: @segiddins this method is copied pasted from `pod_target_installer.rb`, should we extract it out or make it a class method? The same question for the `update_changed_file` method below

spec/unit/installer/xcode/pods_project_generator_spec.rb
```diff
@@ -267,6 +288,46 @@ class Xcode
             @generator.project.targets.find { |t| t.name == 'CoconutLib-iOS' }.isa.should == 'PBXAggregateTarget'
           end
 
+          it 'creates and links app host with an iOS test native target' do
+            @generator.generate!
+            @generator.project.targets.count.should == 19
+            app_host_target = @generator.project.targets[17]
+            app_host_target.name.should == 'AppHost-iOS-Unit-Tests'
+            app_host_target.symbol_type.should == :application
+            app_host_target.build_configurations.each do |bc|
```
dnkoutso: nit you can remove these assertions since they are already in `app_host_installer_spec.rb`. You only care to assert only about connection between the test native target and the app host

spec/unit/installer/xcode/pods_project_generator_spec.rb
```diff
@@ -267,6 +288,46 @@ class Xcode
             @generator.project.targets.find { |t| t.name == 'CoconutLib-iOS' }.isa.should == 'PBXAggregateTarget'
           end
 
+          it 'creates and links app host with an iOS test native target' do
+            @generator.generate!
+            @generator.project.targets.count.should == 19
```
dnkoutso: nit: can remove this assertion. Its superseded by prior tests and also does not help that much. (I know it was a left over from the transferred test)

spec/unit/installer/xcode/pods_project_generator_spec.rb
```diff
@@ -267,6 +288,46 @@ class Xcode
...
+            app_host_target = @generator.project.targets[17]
```
dnkoutso: can also prefer to do `@generator.project.targets.find { |t| t.name == 'AppHost-iOS-Unit-Tests` instead of the index based finding.

Then just assert its not `nil` (via `should.not.be.nil`)

spec/unit/installer/xcode/pods_project_generator_spec.rb
```diff
@@ -267,6 +288,46 @@ class Xcode
...
+            test_native_target = @generator.project.targets[8]
+            test_native_target.build_configurations.each do |bc|
+              bc.build_settings['TEST_HOST'].should == '$(BUILT_PRODUCTS_DIR)/AppHost-iOS-Unit-Tests.app/AppHost-iOS-Unit-Tests'
+            end
+            @generator.project.root_object.attributes['TargetAttributes'][test_native_target.uuid.to_s].should == {
+              'TestTargetID' => app_host_target.uuid.to_s,
+            }
+          end
+
+          it 'creates and links app host with an OSX test native target' do
```
dnkoutso: same comments for this test as the ones above

dnkoutso:
Thank you for pushing this through! Since we added new text in the log we need to update the integration tests. You can do that via `bundle exec rake:update-integration-fixtures` (if I remember correctly) and make a PR to https://github.com/CocoaPods/cocoapods-integration-specs (this will get CI green in this PR).

lib/cocoapods/installer/xcode/pods_project_generator.rb
```diff
@@ -63,9 +64,10 @@ def initialize(sandbox, aggregate_targets, pod_targets, analysis_result, install
         def generate!
           prepare
           install_file_references
-          @target_installation_results = install_libraries
+          @target_installation_results = install_targets
           integrate_targets(@target_installation_results.pod_target_installation_results)
-          wire_target_dependencies(@target_installation_results)
+          app_hosts = install_app_hosts
```
dnkoutso: nit: that was me, but rename to `app_hosts_by_host_key`

lib/cocoapods/installer/xcode/pods_project_generator/target_installer.rb
```diff
@@ -157,17 +157,13 @@ def support_files_temp_dir
           # @param  [Symbol] bundle_package_type
           #         the CFBundlePackageType of the target this Info.plist file is for.
           #
-          # @param  [Boolean] add_to_support_group
-          #         Whether to add the generated file into the support files group by default.
-          #
           # @return [void]
           #
-          def create_info_plist_file(path, native_target, version, platform, bundle_package_type = :fmwk,
```
dnkoutso: so glad this "hack" parameter is removed.

spec/fixtures/watermelon-lib/WatermelonLib.podspec
```diff
@@ -12,6 +12,7 @@ Pod::Spec.new do |s|
   }
   
   s.ios.deployment_target = '9.0'
+  s.osx.deployment_target = '10.11'
```
dnkoutso: <3

spec/fixtures/watermelon-lib/WatermelonLib.podspec
```diff
@@ -22,6 +23,7 @@ Pod::Spec.new do |s|
   end
 
   s.test_spec 'SnapshotTests' do |test_spec|
+    test_spec.requires_app_host = true
```
dnkoutso: 👍

lib/cocoapods/installer/xcode/pods_project_generator.rb
jkap: they behave slightly differently but `find` won't actually do what we need. it's weird to explain so here's a sample:

```ruby
a = {:foo => "bar", :a => "b"}

a.find {|a| true}
# [:foo, "bar"]

a.select {|a| true}
# {:foo => "bar", :a => "b}
```

lib/cocoapods/installer/xcode/pods_project_generator/app_host_installer.rb
```diff
@@ -0,0 +1,134 @@
+module Pod
+  class Installer
+    class Xcode
+      class PodsProjectGenerator
+        # Installs an app host target to a given project.
+        #
+        class AppHostInstaller
+          # @return [Sandbox] sandbox
+          #         The sandbox where the support files should be generated.
```
dnkoutso: sorry this was my bad to tell you to copy paste it but you can make it `The sandbox used for this installation.`

lib/cocoapods/installer/xcode/pods_project_generator.rb
dnkoutso: ah I was wrong

jkap: it's all good! this behavior is extremely unintuitive and not even documented! 🙃

jkap:
test fixtures updated in CocoaPods/cocoapods-integration-specs#172

dnkoutso:
💯 
only question lingering is the `create_info_plist` and `update_changed_file` methods and whether they should be duplicated or not. I do not think they should be and most likely need to become class methods for now. ¯\_(ツ)_///¯

spec/unit/installer/xcode/pods_project_generator_spec.rb
```diff
@@ -267,6 +288,36 @@ class Xcode
             @generator.project.targets.find { |t| t.name == 'CoconutLib-iOS' }.isa.should == 'PBXAggregateTarget'
           end
 
+          it 'creates and links app host with an iOS test native target' do
+            @generator.generate!
+            app_host_target = @generator.project.targets.find { |t| t.name == 'AppHost-iOS-Unit-Tests' }
+            app_host_target.name.should.not.be.nil?
```
dnkoutso: nit: `app_host_target.should.not.be.nil?`?

spec/unit/installer/xcode/pods_project_generator_spec.rb
```diff
@@ -267,6 +288,36 @@ class Xcode
...
+            end
+            @generator.project.root_object.attributes['TargetAttributes'][test_native_target.uuid.to_s].should == {
+              'TestTargetID' => app_host_target.uuid.to_s,
+            }
+          end
+
+          it 'creates and links app host with an OSX test native target' do
+            @generator.generate!
+            app_host_target = @generator.project.targets.find { |t| t.name == 'AppHost-macOS-Unit-Tests' }
+            app_host_target.name.should.not.be.nil?
```
dnkoutso: same question as above

jkap:
does it make sense to make those class methods on a `TargetInstallerHelper` or something? since they're being used in two places now

dnkoutso:
I am OK with that

lib/cocoapods/installer/xcode/pods_project_generator/target_installer.rb
```diff
@@ -160,16 +138,8 @@ def support_files_temp_dir
...
-              update_changed_file(generator, path)
-              add_file_to_support_group(path)
-
-              relative_path_string = path.relative_path_from(sandbox.root).to_s
-              native_target.build_configurations.each do |c|
-                c.build_settings['INFOPLIST_FILE'] = relative_path_string
-              end
-            end
+            TargetInstallerHelper.create_info_plist_file(@sandbox, path, native_target, version, platform, bundle_package_type)
+            add_file_to_support_group(path)
```
jkap: note here: `add_file_to_support_group` is only needed when `create_info_plist_file` is called by subclasses of `PodTargetInstaller` and extracting that function out to `TargetInstallerHelper` would require more changes that feel excessive, so this is my workaround there. is this an acceptable compromise

lib/cocoapods/installer/xcode/pods_project_generator/app_host_installer.rb
segiddins: should probably stick both of them in a module and include that module in both places

lib/cocoapods/installer/xcode/pods_project_generator.rb
jkap: didn't know that two structs with identical values were equivalent in ruby. learned something new today

jkap:
if we're good now, i'm gonna squash this down

lib/cocoapods/installer/xcode/pods_project_generator/target_installer_helper.rb
```diff
@@ -0,0 +1,65 @@
...
+        class TargetInstallerHelper
+          # @param [Generator] generator
+          #        the generator to use for generating the content.
+          #
+          # @param [Pathname] path
+          #        the pathname to save the content into.
+          #
+          # Saves the content the provided path unless the path exists and the contents are exactly the same.
+          #
+          def self.update_changed_file(generator, path)
```
segiddins: we can use `module_function` instead of explicitly declaring these as `self.`, and then include the module in the 2 classes that use it

jkap:
squashed, ready to land @segiddins

PR #7914, opened 2018-07-09
Only build pod as module when all target definitions say to do so

segiddins:


justinseanmartin:
Rather than silently dropping the `modular_headers` request, maybe this should generate an error or warning about the mismatch?

segiddins:
> Rather than silently dropping the modular_headers request, maybe this should generate an error or warning about the mismatch?

the problem there is that you can have a target definition that doesn’t explicitly have anything for that pod, but have it do `use_modular_headers!`

PR #7917, opened 2018-07-10
Use CP_HOME_DIR as the base for all default directories

mcfedr:
Currently if you want to change the folders used by Cocoapods you have to set 3 separate envvars.

This is completely non-obvious and not really very helpful. It wasn't until I search though the code that I worked out why setting `CP_HOME_DIR` apparently did nothing.

I have also updated the help messages to use these global variables, as before they would be misleading in an environment where the dirs have been changed.

This has the potential for being a breaking changing for some people, but given that `CP_HOME_DIR` isn't documented anywhere, its probably not very widely used. I would expect that 99% of people using `CP_HOME_DIR` are also setting `CP_REPOS_DIR` and `CP_TEMPLATES_DIR`, and so they would be unaffected by this change.

Also, 🌈

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -131,6 +131,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   as a module when all of the targets it is linked into have opted into it.  
   [Samuel Giddins](https://github.com/segiddins)
 
+* Use `CP_HOME_DIR` as the base for all default directories  
+  [mcfedr](https://github.com/mcfedr)
+  [#7917](https://github.com/CocoaPods/CocoaPods/pull/7917)
 
 ## 1.5.3 (2018-05-25)
 
--- a/lib/cocoapods/config.rb
+++ b/lib/cocoapods/config.rb
@@ -100,10 +100,18 @@ module Pod
     def initialize(use_user_settings = true)
       configure_with(DEFAULTS)
 
+      unless ENV['CP_HOME_DIR'].nil?
+        @cache_root = home_dir + 'cache'
+      end
+
       if use_user_settings && user_settings_file.exist?
         require 'yaml'
         user_settings = YAML.load_file(user_settings_file)
         configure_with(user_settings)
+      end
+
+      unless ENV['CP_CACHE_DIR'].nil?
+        @cache_root = Pathname.new(ENV['CP_CACHE_DIR']).expand_path
       end
     end
 
--- a/lib/cocoapods/config.rb
+++ b/lib/cocoapods/config.rb
@@ -127,7 +127,7 @@ module Pod
     # @return [Pathname] the directory where the CocoaPods sources are stored.
     #
     def repos_dir
-      @repos_dir ||= Pathname.new(ENV['CP_REPOS_DIR'] || '~/.cocoapods/repos').expand_path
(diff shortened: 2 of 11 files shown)


dnkoutso:
Thanks! Could you also add an entry to `CHANGELOG`? Additionally if possible is there a test that can be added here?

CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Use CP_HOME_DIR as the base for all default directories  
  [mcfedr](https://github.com/mcfedr)
```

</p>

mcfedr:
@dnkoutso Of course! Added both of these things

CHANGELOG.md
```diff
@@ -45,6 +45,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Whirlwind](https://github.com/Whirlwind)
   [#7878](https://github.com/CocoaPods/CocoaPods/pull/7878)
 
+* Use CP_HOME_DIR as the base for all default directories
```
dnkoutso: nit: I think this is a bug fix, not enhancement. Also can wrap in CP_HOME_DIR in back ticks.

mcfedr: I agree, but wasnt sure the maintainers would see it in that way :)

dnkoutso:
@mcfedr looks good. I'd prefer if one other maintainer also takes a look before we merge.

mcfedr:
So now a question, I'd like to change the default `cache_root`, in the case that `CP_HOME_DIR` is set, to be `$CP_HOME_DIR/cache` - does that seem like a reasonable change?

I feel like, again, in spirit of one var to change the pods folder, this would make sense, but maybe 'the project' would see this differently?

I'm mostly interested in changing/fixing this for a shared CI server, were I want to minimise the risk of one project affecting another, and also the case where private repos are used by some projects. I'm realise that being on the same machine they could still see each others caches, but moving the complete `.cocoapods` dir into the project does reduce the chances for accidentally using another projects cache

mcfedr:
@dnkoutso I've added a commit that does the change to the cache_root. I feel like this is a good additional change

lib/cocoapods/config.rb
```diff
@@ -100,11 +100,19 @@ def cache_root
...
+
       if use_user_settings && user_settings_file.exist?
         require 'yaml'
         user_settings = YAML.load_file(user_settings_file)
         configure_with(user_settings)
       end
+
+      unless ENV['CP_CACHE_DIR'].nil?
+        @cache_root = Pathname.new(ENV['CP_CACHE_DIR']).expand_path
+      end
```
amorde: Any reason for splitting this into two sections?

This will also override any values set by the yaml config file if it exists. If that's the desired behavior, we should have a test for that and document it as well.

mcfedr: The reason is to allow overriding the config.

The first case could actually go into the `DEFAULTS` but then it becomes untestable because its static.

Then we load your config file, but `CP_CACHE_DIR` can still override the conf, that seemed like the most idiomatic way to combine config file and env vars.

Ill add a test to show that this is the case.

dnkoutso:
@mcfedr looks good, just needs a rebase.

amorde:
@mcfedr Thanks for the contribution!

PR #7933, opened 2018-07-16
Add a deployment mode for installs

segiddins:
It errors if there are any changes to the podfile or the lockfile, similiar to `bundle install --deployment`.
It would be useful to have for CI.

- [x] Specs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Whirlwind](https://github.com/Whirlwind)
   [#7878](https://github.com/CocoaPods/CocoaPods/pull/7878)
 
+* Add a `--deployment` flag to `pod install` that errors if there are any
+  changes to the Podfile or Lockfile.  
+  [Samuel Giddins](https://github.com/segiddins)
+
 ##### Bug Fixes
 
 * Do not display a source was changed if it uses different casing  
--- a/lib/cocoapods/command/install.rb
+++ b/lib/cocoapods/command/install.rb
@@ -30,7 +30,13 @@ module Pod
       def self.options
         [
           ['--repo-update', 'Force running `pod repo update` before install'],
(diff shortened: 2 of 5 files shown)


lib/cocoapods/command/install.rb
```diff
@@ -30,14 +30,21 @@ class Install < Command
       def self.options
         [
           ['--repo-update', 'Force running `pod repo update` before install'],
+          ['--deployment', 'Dissallow any changes to the Podfile or the Podfile.lock during installation'],
```
cysp: `Disallow`? :-)

dnkoutso:
👍 with specs from me.

CHANGELOG.md
```diff
@@ -45,6 +45,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Whirlwind](https://github.com/Whirlwind)
   [#7878](https://github.com/CocoaPods/CocoaPods/pull/7878)
 
+* Add a `--deployment` flag to `pod install` that errors if there are any
+  changes to the Podfile or Lockfile.  
+  [Samuel Giddins](https://github.com/segiddins)
```
amorde: The description sounds like its saying that it will error if the command itself changes the Podfile or Lockfile.

Maybe "... that errors if the Podfile and Lockfile are not in sync" or "... if `pod install` would trigger an update to the Lockfile"

segiddins:
> This is great. I sort of wish it was a default if ENV['CI'] lol

That's a whole  other can of worms I'd like not to open :P

CHANGELOG.md
dnkoutso: I dont think it's about "sync" though, its about them being dirty unless thats what you meant with "sync" on an individual file basis

amorde: hmm yeah. words are hard

dnkoutso:
This does not have the best test coverage admittedly but the code itself is structured in a bad way to actually write the spec.

PR #7946, opened 2018-07-21
Generate -ObjC in OTHER_LDFLAGS for apps with static frameworks

paulb777:
-ObjC was not getting generated for apps linking static frameworks that were accessing podspec's other than via spec repo's. 

This PR makes the static framework linking behavior consistent with static libraries.

More details and an example at https://github.com/firebase/firebase-ios-sdk/issues/1341


paulb777:
There is still an open question about why this bug is dependent on the location of the podspec, but, nonetheless, this PR is an incremental improvement.

PR #7181, opened 2017-10-29
Only add resources and frameworks build phase when needed

keith:
This changes the project integrator to only add build phases to the
user's project if there are resources of frameworks that will be copied
in that phase.

PR #7957, opened 2018-07-27
Don't generate scripts if they will not be added to the user project

amorde:
Follow up to #7181 - if we aren't going to add the scripts to the user project, don't generate the files at all and don't write them to disk.

cc @keith

### DIFF
--- a/lib/cocoapods/installer/xcode/pods_project_generator/aggregate_target_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/aggregate_target_installer.rb
@@ -30,9 +30,9 @@ module Pod
               # cause an App Store rejection because frameworks cannot be
               # embedded in embedded targets.
               #
-              create_embed_frameworks_script unless target.requires_host_target?
+              create_embed_frameworks_script if target.includes_frameworks? && !target.requires_host_target?
               create_bridge_support_file(native_target)
-              create_copy_resources_script
+              create_copy_resources_script if target.includes_resources?
               create_acknowledgements
               create_dummy_source(native_target)
               clean_support_files_temp_dir
--- a/lib/cocoapods/target/aggregate_target.rb
+++ b/lib/cocoapods/target/aggregate_target.rb
@@ -191,6 +191,19 @@ module Pod
       pod_targets.any?(&:uses_swift?)
     end
 
+    # @return [Boolean] Whether the target contains any resources
+    #
+    def includes_resources?
+      !resource_paths_by_config.values.all?(&:empty?)
+    end
+
+    # @return [Boolean] Whether the target contains framework to be embedded into
+    #         the user target
+    #
+    def includes_frameworks?
+      !framework_paths_by_config.values.all?(&:empty?)
+    end
+
     # @return [Hash{String => Array<Hash{Symbol => [String]}>}] The vendored dynamic artifacts and framework target
     #         input and output paths grouped by config
     #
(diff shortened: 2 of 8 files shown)


lib/cocoapods/installer/xcode/pods_project_generator/aggregate_target_installer.rb
```diff
@@ -6,6 +6,11 @@ class PodsProjectGenerator
         # project and the relative support files.
         #
         class AggregateTargetInstaller < TargetInstaller
+          # @return [AggregateTarget] target
+          #         the target to install
+          #
+          attr_reader :target
```
dnkoutso: why is this needed? Isnt it available from the base class?

amorde: was trying to update the type annotation to be `AggregateTarget` for this subclass - is there an easier way to do that?

lib/cocoapods/target/aggregate_target.rb
```diff
@@ -191,6 +191,19 @@ def uses_swift?
...
+    # @return [Boolean] Whether the target contains any resources
+    #
+    def has_resources?
+      !resource_paths_by_config.values.all?(&:empty?)
+    end
+
+    # @return [Boolean] Whether the target contains framework to be embedded into
+    #         the user target
+    #
+    def has_frameworks?
```
amorde: rubocop doesn't like the name of these - any suggestions

spec/unit/installer/user_project_integrator/target_integrator_spec.rb
```diff
@@ -49,16 +49,6 @@ module Pod
...
 
-        it 'fixes the copy resource scripts of legacy installations' do
-          @target_integrator.integrate!
-          target = @target_integrator.send(:native_targets).first
-          phase_name = @copy_pods_resources_phase_name
-          phase = target.shell_script_build_phases.find { |bp| bp.name == phase_name }
-          phase.shell_script = %("${PODS_ROOT}/Pods-resources.sh"\n)
-          @target_integrator.integrate!
-          phase.shell_script.strip.should == '"${PODS_ROOT}/Target Support Files/Pods/Pods-resources.sh"'
-        end
```
amorde: The "legacy installations" here refers to pre-0.34, I think we can drop any special handling for that by now

lib/cocoapods/installer/xcode/pods_project_generator/aggregate_target_installer.rb
segiddins: You could use yard's `@!parse` macro: https://www.rubydoc.info/gems/yard/file/docs/Tags.md#parse

lib/cocoapods/target/aggregate_target.rb
segiddins: `includes_...?`

dnkoutso:
Does this change mean we actually *do* create the files on disk (the `.sh` scripts) but we just don't add the script phase if there is nothing to copy or embed?

amorde:
> Does this change mean we actually do create the files on disk (the .sh scripts) but we just don't add the script phase if there is nothing to copy or embed?

yep exactly that

dnkoutso:
OK. Was wondering if we wanted to take it further and not even write the `.sh` files! but this change is great as it is.

amorde:
Err sorry, this PR is actually doing that @dnkoutso - now we won't make the files at all if they aren't going to be added to the build phases :)

amorde:
what I meant is right now we don't add them to the build phase if there aren't resources / frameworks, but on master we still make the actual files.

Issue #7683, opened 2018-04-25
Unable to lint a pod that requires modular headers


JDongKhan:
# Report

## What did you do?

`pod lib lint SportsKit.podspec  --allow-warnings --verbose --use-libraries `

## What did you expect to happen?

lint success！！！

## What happened instead?

The Swift pod `SportsKit` depends upon `YYImage`, which do not define modules. To opt into those targets generating module maps (which is necessary to import them from Swift when building as static libraries), you may set `use_modular_headers!` globally in your Podfile, or specify `:modular_headers => true` for particular dependencies.

## CocoaPods Environment

ℹ Please replace these two lines with the output of `pod env`.

### Stack

```
   CocoaPods : 1.5.0
        Ruby : ruby 2.3.3p222 (2016-11-21 revision 56859) [universal.x86_64-darwin17]
    RubyGems : 2.7.6
        Host : Mac OS X 10.13.4 (17E199)
       Xcode : 9.3 (9E145)
         Git : git version 2.15.1 (Apple Git-101)
Ruby lib dir : /System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib
Repositories :master - https://github.com/CocoaPods/Specs.git @ 0f66338609cb62a08c18919bbce219fc799dde18
```

### Installation Source

```
Executable Path: /usr/local/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.2
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
cocoapods-trunk       : 1.3.0
cocoapods-try         : 1.1.0
```

### Podfile

```ruby
platform:ios,'8.0'

inhibit_all_warnings!
use_frameworks!

target "SwiftAndOc" do

pre_install do |installer|
    # workaround for https://github.com/CocoaPods/CocoaPods/issues/3289
    Pod::Installer::Xcode::TargetValidator.send(:define_method, :verify_no_static_framework_transitive_dependencies) {}
end

pod 'SportsKit', :path => 'SportsKit.podspec'

end

```

## Project that demonstrates the issue
[Demo](https://github.com/wangjindong/SwiftAndOc.git)

amorde:
Think this is related: https://github.com/CocoaPods/CocoaPods/issues/7642

wangjindong:
@amorde

feifeiwuxian:
[...]

PR #7959, opened 2018-07-28
Add --use-modular-headers to lint and push commands

amorde:
Progress towards #7683 

Still need to add the flag to `pod trunk`

### DIFF
--- a/lib/cocoapods/validator.rb
+++ b/lib/cocoapods/validator.rb
@@ -243,6 +243,10 @@ module Pod
     #
     attr_accessor :use_frameworks
 
+    # @return [Boolean] Whether modular headers should be used for the installation.
+    #
+    attr_accessor :use_modular_headers
+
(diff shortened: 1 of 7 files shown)


lib/cocoapods/validator.rb
```diff
@@ -847,7 +851,7 @@ def initialize(type, attribute_name, message, public_only = false)
     # @note   The generated podfile takes into account whether the linter is
     #         in local mode.
     #
-    def podfile_from_spec(platform_name, deployment_target, use_frameworks = true, test_spec_names = [])
+    def podfile_from_spec(platform_name, deployment_target, use_frameworks = false, test_spec_names = [], use_modular_headers = false)
```
segiddins: why is this default changing?

amorde: wasn't intentional, will revert

Issue #7867, opened 2018-06-21
LoadError - cannot load such file -- macho


chshapiro:
I tried with versions 1.5.0 and 1.5.3 and got this error for both versions.

### Command

```
/Users/xxxx/xxxx/.gems/bin/pod install
```

### Report

* What did you do?

tried to run `pod install`

* What did you expect to happen?

Pods get installed

* What happened instead?

I got the following error

### Stack

```
   CocoaPods : 1.5.3
        Ruby : ruby 2.2.4p230 (2015-12-16 revision 53155) [x86_64-darwin17]
    RubyGems : 2.4.5.1
        Host : Mac OS X 10.13.4 (17E199)
       Xcode : 9.4 (9F1027a)
         Git : git version 2.17.1
Ruby lib dir : /Users/xxxx/.rbenv/versions/2.2.4/lib
Repositories : master - https://github.com/CocoaPods/Specs.git @ 44b861022b9f0c4ba03a664bbfa04ed4628bf2b7
```

### Plugins

```
cocoapods-deintegrate : 1.0.2
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
cocoapods-trunk       : 1.3.0
cocoapods-try         : 1.1.0
```

### Error

```
LoadError - cannot load such file -- macho
/Users/xxxx/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/cocoapods-1.5.3/lib/cocoapods/sandbox/file_accessor.rb:432:in `rescue in dynamic_binary?'
/Users/xxx/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/cocoapods-1.5.3/lib/cocoapods/sandbox/file_accessor.rb:427:in `dynamic_binary?'
/Users/xxx/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/cocoapods-1.5.3/lib/cocoapods/sandbox/file_accessor.rb:164:in `block in vendored_dynamic_frameworks'
/Users/xxx/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/cocoapods-1.5.3/lib/cocoapods/sandbox/file_accessor.rb:163:in `select'
[...]

PR #7980, opened 2018-08-06
Use `require` for MachO instead of `autoload`

amorde:
Closes https://github.com/CocoaPods/CocoaPods/issues/7867

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -175,6 +175,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 * Don't generate framework or resource scripts if they will not be used  
   [Eric Amorde](https://github.com/amorde)
 
+* Fix a crash when loading Ruby MachO library in certain environments  
+  [Eric Amorde](https://github.com/amorde)
+  [#7867](https://github.com/CocoaPods/CocoaPods/issues/7867)
+
 ## 1.5.3 (2018-05-25)
 
 ##### Enhancements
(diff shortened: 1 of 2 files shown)


segiddins:
This shouldn't be necessary.... `autoload` calls `require` under the hood when `const_missing` is called with the given constant

amorde:
@segiddins any insights on why this change fixes the issue for multiple people then? I agree it _shouldn't_ be necessary but quite a few people are having this crash

segiddins:
It’s possible their ruby environment is in a weird state

amorde:
I do think it's worth it if it can fix the issue for a lot of people but if you disagree I'll close this.

segiddins:
Up to you

amorde:
Let's get it into the 1.6.0 beta and we can revert if it impacts start up time or causes other issues

amorde:
I also don't know enough about autoload vs require (I did do a bit of research into it) to know how this could be breaking for anyone

icecrystal23:
I manually replaced autoload with require and that worked for me locally, too.

amorde:
@icecrystal23 this was included in the 1.6.0.beta.1 release

icecrystal23:
Was it? Because I had to fix it manually after installing 1.6.0.beta.1
I'm not 100% sure I'm running the beta even though I installed it, though.  Do I have to call it differently or anything? pod --version is still telling me 1.5.3

PR #7998, opened 2018-08-10
fix a same name issue under different ruby

Whirlwind:
First, I don't know why it cause a issue. CocoaPods could not call the `plist` method and cause the plist file to not be generated. I try to put some logs in the `plist` method and output nothing. And the log in `generate` method is normal.

I check everything, but could not find the reason. I try to rename the `plist` name to other name, everything is OK. It maybe the method name conflicts with local var name ( I am not sure.)

### DIFF
--- a/lib/cocoapods/generator/acknowledgements/plist.rb
+++ b/lib/cocoapods/generator/acknowledgements/plist.rb
@@ -8,19 +8,19 @@ module Pod
       end
 
       def save_as(path)
-        Xcodeproj::Plist.write_to_path(plist, path)
+        Xcodeproj::Plist.write_to_path(plist_hash, path)
       end
 
       # @return [String] The contents of the plist
       #
       def generate
-        plist = Nanaimo::Plist.new(plist, :xml)
+        plist = Nanaimo::Plist.new(plist_hash, :xml)
         contents = StringIO.new
         Nanaimo::Writer::XMLWriter.new(plist, :pretty => true, :output => contents, :strict => false).write
         contents.string
       end
 
-      def plist
+      def plist_hash
         {
           :Title => plist_title,
           :StringsTable => plist_title,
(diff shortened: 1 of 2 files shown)


Whirlwind:
There is a strange issue in the CI.

segiddins:
Could you add a test for the bug this is fixing?

Whirlwind:
It is hard to write a unit test. It just a problem about a method name.

Whirlwind:
It seams that it cause by ruby or other environment. It crash today. But It is OK before.

segiddins:
Instead of changing the method name, why not change the local variable name that it conflicts with?

dnkoutso:
Agreed, lets change the local variable name instead of the method name:
```diff
diff --git a/lib/cocoapods/generator/acknowledgements/plist.rb b/lib/cocoapods/generator/acknowledgements/plist.rb
index 162e11ccff9..6ca8b638277 100644
--- a/lib/cocoapods/generator/acknowledgements/plist.rb
+++ b/lib/cocoapods/generator/acknowledgements/plist.rb
@@ -14,9 +14,9 @@ module Pod
       # @return [String] The contents of the plist
       #
       def generate
-        plist = Nanaimo::Plist.new(plist, :xml)
+        new_plist = Nanaimo::Plist.new(plist, :xml)
         contents = StringIO.new
-        Nanaimo::Writer::XMLWriter.new(plist, :pretty => true, :output => contents, :strict => false).write
+        Nanaimo::Writer::XMLWriter.new(new_plist, :pretty => true, :output => contents, :strict => false).write
         contents.string
       end
```

Whirlwind:
We have `header_hash` and `footer_hash` method name, I want to keep same style, so I change the method `plist` to `plist_hash`.

dnkoutso:
Guess that is a true statement! Can you rebase to latest master?

Whirlwind:
@dnkoutso  Sorry, I could not understand what you said..... My English is not good.

dnkoutso:
@Whirlwind I am saying you are right that other methods in `plist.rb` use `def method_hash` so we should proceed with your change...

Whirlwind:
done

dnkoutso:
we are about to launch 1.6.0 this week so I will wait to merge this end of this week!

Whirlwind:
Could you merge it to master branch ? I have some other branches, and they could not pass the unit test.

PR #7999, opened 2018-08-10
Allow duplicate dependencies when they have no conflict

Whirlwind:
I have some cases. 

## case 1:
Podfile:
```
target :A do
    pod 'SDK'
end
target :B do
    pod 'SDK', '> 1.0'
end
target :C do
    pod 'SDK', '2.0'
end
```
I think the Podfile is allowed and the `SDK` should be `2.0`. But the CocoaPods will have some warning and some error.

## case 2
A `SDK/SDK.podspec` :
```
s.name = 'SDK'
s.version = '2.0'
s.dependency 'AFNetworking', '> 1.0'
```
A `AFNetworking/AFNetworking.podspec`:
```
s.name = 'AFNetworking'
s.version = '2.0'
```

The `Podfile`:
```
target :A do
    pod 'SDK', :path=>'../SDK'
end
target :B do
    pod 'AFNetworking', :path=>'../AFNetworking'
end
target :C do
    pod 'AFNetworking', '2.0'
end
```
I think it is allowed, too. But the CocoaPods will failed.

## case 3
`SDK.podspec`:
```
s.name = 'SDK'
s.version = '2.0'
s.dependency 'AFNetworking', '> 1.0'
```
`Podfile`:
```
target :SDK do
    podspec
end
target :B do
    pod 'AFNetworking', :path=>'../AFNetworking'
end
```
It likes the case 2. It should have nothing warning and error.

## Idea
I want to allow the duplicate dependencies if they have no conflict.
In case 1, the `SDK` will use the `2.0`. Becasue the `2.0` satisfies `>1.0` and `=2.0`.
In case 2, the  `AFNetworking` will use the `:path`. Because the external source will be used first. And the `AFNetworking` in the `../AFNetworking` is `2.0`, it satisfies `> 1.0` and `=2.0`.

So, If the pod has external source, I will use it and use the version.
If the pod has many version conditions, I will use them to make a version arbitration.

### DIFF
--- a/lib/cocoapods/installer/analyzer.rb
+++ b/lib/cocoapods/installer/analyzer.rb
@@ -794,7 +794,7 @@ module Pod
 
       def verify_no_pods_with_different_sources!
         deps_with_different_sources = podfile_dependencies.group_by(&:root_name).
-          select { |_root_name, dependencies| dependencies.map(&:external_source).uniq.count > 1 }
+          select { |_root_name, dependencies| dependencies.map(&:external_source).compact.uniq.count > 1 }
         deps_with_different_sources.each do |root_name, dependencies|
           raise Informative, 'There are multiple dependencies with different ' \
           "sources for `#{root_name}` in #{UI.path podfile.defined_in_file}:" \
@@ -884,7 +884,7 @@ module Pod
       #
       def resolve_dependencies(locked_dependencies)
         duplicate_dependencies = podfile_dependencies.group_by(&:name).
-          select { |_name, dependencies| dependencies.count > 1 }
+          select { |_name, dependencies| dependencies.count(&:external_source) > 1 }
         duplicate_dependencies.each do |name, dependencies|
           UI.warn "There are duplicate dependencies on `#{name}` in #{UI.path podfile.defined_in_file}:\n\n" \
            "- #{dependencies.map(&:to_s).join("\n- ")}"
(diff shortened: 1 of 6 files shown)


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Allow duplicate dependencies when they have no conflict  
```

</p>

dnkoutso:
I am not sure this 100% complete. What sources does `SDK` ends up having in the `Pods.xcodeproj`? I dont think CocoaPods currently supports version variants for targets.

Whirlwind:
I don't change the `Molinillo`, and the version which be used will be calculated by `molinillo`.
For the external source, I change the delegete `search_for` of `molinillo`. It will only return a spec from the external source if the spec'version is compatibled. If not, will return anything and get a fail.

dnkoutso:
You might fix the resolver but what version of `SDK` appears in Pods.xcodeproj for the user to use? What sources are added (does it include both sources from SDK 1 and SDK 2?)

dnkoutso:
I would expect CocoaPods to create a `SDK-1.0` and an `SDK-2.0` target.

Whirlwind:
If the `SDK` `1.1.0` have different dependencies with `2.0`, it will maybe cause all dependencies have two versions.

But user maybe just want to a version for many targets, the targets are in a app or for `free` or `paid` version. So the `SDK` in different targets should be same version.

dnkoutso:
I believe this would introduce a ton of edge cases. This feature needs to be thought more carefully before we make changes to support it imo.

Whirlwind:
Under the current version, CocoaPods still only supports a version for a Pod and many variants for a version. I keep this logic and CP still only use a version.

If CP support many version for a Pod, it will be very hard. And user maybe not want it.

Whirlwind:
The Case 3 should be a correct way which use the `podspec`, but the CP will raise a error with multi source. I often use the `podspec` DSL, and it cause that I could not use the `podspec`.

Whirlwind:
I still want you to consider this issue. :)

dnkoutso:
Closing for now as I do not think we will implement this.

PR #8004, opened 2018-08-12
Pull out target generation from the analyzer

amorde:
## Changes

* Pull out the handling of target generation into its own class
* Fix / add lots of type annotations

This exposed all of the variables needed to generate the targets, and will make it easier to decompose it further down the road. Currently the target generation doesn't have too much direct test coverage, and instead is tested at a higher level. I would write more direct unit tests if I even knew what the expected behavior was based on all the inputs :p


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Pull out target generation from the analyzer  
```

</p>

lib/cocoapods/installer/analyzer/target_generator.rb
```diff
@@ -0,0 +1,397 @@
...
+            dedupe_cache = {}
+            pod_targets = resolver_specs_by_target.flat_map do |target_definition, specs|
+              grouped_specs = specs.group_by(&:root).values.uniq
+              grouped_specs.flat_map do |pod_specs|
+                generate_pod_target([target_definition], target_inspections, pod_specs.map(&:spec), sandbox, installation_options).scoped(dedupe_cache)
+              end
+            end
+          end
+
+          all_resolver_specs = resolver_specs_by_target.values.flatten.map(&:spec).uniq
```
amorde: The code starting from here was duplicated within each branch of the `if`, with the only difference being:

```
if installation_options.integrate_targets?
  ...
  # within target.test_dependent_targets_by_spec_name =
  hash[test_spec.name] = filter_dependencies(test_dependencies, pod_targets_by_name, target
else
  ...
 # within target.test_dependent_targets_by_spec_name =
  hash[test_spec.name] = pod_targets.reject { |t| test_dependencies[t.root_spec].nil? }
end
```

now both branches are using `filter_dependencies`, but is that valid? tests were passing but not sure if there's an edge case not covered by the tests

lib/cocoapods/installer/analyzer/target_generator.rb
```diff
@@ -0,0 +1,397 @@
...
+            target_inspection = target_inspections[target_definition]
+            raise "missing inspection: #{target_definition.name}" unless target_inspection
+            user_project = target_inspection.project
+            client_root = user_project.path.dirname.realpath
+            user_target_uuids = target_inspection.project_target_uuids
+            user_build_configurations = target_inspection.build_configurations
+            archs = target_requires_64_bit ? ['$(ARCHS_STANDARD_64_BIT)'] : target_inspection.archs
+          else
+            user_project = nil
+            client_root = Config.instance.installation_root
```
amorde: not happy about this part, but trying to keep the number of arguments from growing to a paragraph

lib/cocoapods/installer/analyzer/target_inspector.rb
```diff
@@ -95,13 +97,12 @@ def compute_project_path
...
-        #         the user project
+        # @param  [Array<PBXNativeTarget>] native_targets
+        #         all native targets in the user project
         #
-        # @return [Array<PBXNativeTarget>]
+        # @return [Array<PBXNativeTarget>] the targets to be integrated
         #
-        def compute_targets(user_project)
-          native_targets = user_project.native_targets
+        def compute_targets(native_targets)
```
amorde: changed to accept just the array of native targets, since that's all it was using the project for

lib/cocoapods/resolver.rb
```diff
@@ -133,12 +133,12 @@ def resolve
     #
     def resolver_specs_by_target
       @resolver_specs_by_target ||= {}.tap do |resolver_specs_by_target|
-        @podfile_dependency_cache.target_definition_list.each do |target|
+        @podfile_dependency_cache.target_definition_list.each do |target_definition|
```
amorde: renamed for clarity, no other logic changes here

amorde:
not sure what's wrong with the build, I don't think the failures are related to this

segiddins:
Restarted build

lib/cocoapods/installer/analyzer.rb
```diff
@@ -196,6 +193,8 @@ def update_mode
         end
       end
 
+      # @return [Array<Pod::Dependency>]
```
dnkoutso: nit add more context here?

lib/cocoapods/installer/analyzer/target_generator.rb
```diff
@@ -0,0 +1,397 @@
+module Pod
+  class Installer
+    class Analyzer
+      class TargetGenerator
+        # @return [String] The version of iOS which requires binaries with only 64-bit architectures
+        #
+        IOS_64_BIT_ONLY_VERSION = Version.new('11.0')
```
dnkoutso: `.freeze`?

amorde: ah, now I remember why I didn't do this - comparing versions with `>=` triggers a lazy initializer on `_segments` inside `Version`, which is a modification. so basically comparing one version with another modifies the version instance :/

amorde:
looks like something broke after a rebase. will look into it at some point

lib/cocoapods/installer/analyzer/target_generator.rb
```diff
@@ -0,0 +1,397 @@
...
+              test_dependencies = dependencies_for_specs([test_spec], target.platform, all_specs).group_by(&:root)
+              test_dependencies.delete_if { |k| dependencies.key? k }
+              hash[test_spec.name] = filter_dependencies(test_dependencies, pod_targets_by_name, target)
+            end
+          end
+        end
+
+        #-----------------------------------------------------------------------#
+
+        class << self
```
segiddins: please be consistent, and either use `class << self` or `def self.`, but not both

lib/cocoapods/installer/analyzer/target_generator.rb
amorde: @segiddins this is the only thing that could potential affect functionality as far as I can tell

lib/cocoapods/installer/analyzer/target_generator.rb
amorde: I was trying to make this section private class methods, but it seems that private class methods aren't really a thing unless you use `class << self` is that correct? should I use `class << self` for the whole class then?

segiddins: `private_class_ method :foo` is also a thing. I think we should just be consistent within a single class

dnkoutso:
For 1.7.0

amorde:
Going to close this and maybe revisit it in another PR.

PR #8005, opened 2018-08-12
Generate a separate app host per pod

segiddins:
For Podfiles that include _many_ test specs that require app hosts, this will help keep the size of those app hosts down, by only installing one or two xctest bundles into each one. This greatly speeds up launching tests.


dnkoutso:
Integration specs LGTM too.

segiddins:
It can’t go back into the pod target installer as implemented, because we use the pod name. We’d have to switch to using the pod target label to be able to do that (though we probably should)

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -308,6 +311,37 @@ def add_test_targets
...
+          #
+          # @param  [Array<PBXNativeTarget>] test_native_targets
+          #         the test native targets that have been created to use as a lookup when linking the app host to.
+          #
+          # @return [Array<PBXNativeTarget>] the app host targets created.
+          #
+          def add_test_app_host_targets(test_native_targets)
+            target.test_spec_consumers.select(&:requires_app_host?).group_by(&:test_type).map do |test_type, test_spec_consumers|
+              platform = target.platform
+              name = "AppHost-#{target.label}-#{test_type.capitalize}-Tests"
```
dnkoutso: @segiddins I went with `target.label` which is already scoped by the platform/suffix.

This is now much better and much more symmetric to other names we use.

dnkoutso:
All 👍

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
segiddins: 👍

lib/cocoapods/installer/xcode/pods_project_generator/target_installation_result.rb
```diff
@@ -33,21 +33,28 @@ class TargetInstallationResult
           #
           attr_reader :test_resource_bundle_targets
 
+          # @return [Array<PBXNativeTarget>] test_app_host_targets
+          #         The test app host native targets that were produced for this target. Can be empty.
+          #
+          attr_reader :test_app_host_targets
```
segiddins: this seems to be unused?

dnkoutso: it is but its correct. This is the entire result of `pod_target_installer.rb` it might be useful to carry through this information. I know...YAGNI. but whatevez

spec/unit/installer/xcode/pods_project_generator/app_host_installer_spec.rb
```diff
@@ -11,35 +11,35 @@ class PodsProjectGenerator
           end
 
           it 'correctly installs an iOS app host target to the project' do
-            installer = AppHostInstaller.new(config.sandbox, @project, Platform.ios, :unit)
+            installer = AppHostInstaller.new(config.sandbox, @project, Platform.ios, 'AppHost-PodName-iOS-Unit-Tests')
             installer.install!
-            @project.targets.map(&:name).sort.should == ['AppHost-iOS-Unit-Tests']
+            @project.targets.map(&:name).find { |name| name == 'AppHost-PodName-iOS-Unit-Tests' }.should.not.be.nil
```
dnkoutso: actually changing this assertion is stupid

Issue #7872, opened 2018-06-24
[1.6.0.beta.1] weak_frameworks lost


Whirlwind:
In CP v1.3.1, the `weak_frameworks` from podspec works, but in CP master branch, all `weak_frameworks` lost. I will check the code later.
The Podspec;
```
  s.frameworks = 'OpenAL','VideoToolbox','CoreMedia', 'AudioToolbox','MediaPlayer','CoreMotion','AVFoundation','SystemConfiguration'
  s.weak_frameworks = 'MetalKit'
```

dnkoutso:
very possible due to build_settings re-write.

Whirlwind:
the function `weak_frameworks_to_import` in class `PodTargetSettings` in the `build_settings.rb`:

```
        # @return [Array<String>]
        define_build_settings_method :weak_frameworks_to_import, :memoized => true, :sorted => true, :uniqued => true do
-          []
+          spec_consumers.flat_map(&:weak_frameworks)
        end
```

Whirlwind:
The code is very nice and I afraid that I do a error modify, so I could not make a PR.

dnkoutso:
we need to find out if there was a test in 1.5.3 that covers this case.

Whirlwind:
It seams that there is not any test for the `weak_frameworks` in the `AggregateTarget`, only for `PodTarget`. So the user target will not link weak framework.

Whirlwind:
Could somebody help me about this issue? I could not fix it by myself.

dnkoutso:
@Whirlwind  1.6.0.beta.1 is out is this still an issue? Can you upload a sample app?

Whirlwind:
I test my app, this still an issue.

Whirlwind:
```
[...]

PR #8026, opened 2018-08-17
Fix `weak_frameworks` missing regression.

dnkoutso:
closes https://github.com/CocoaPods/CocoaPods/issues/7872

/cc @Whirlwind

### DIFF
--- a/lib/cocoapods/target/build_settings.rb
+++ b/lib/cocoapods/target/build_settings.rb
@@ -552,7 +552,7 @@ module Pod
 
         # @return [Array<String>]
         define_build_settings_method :weak_frameworks_to_import, :memoized => true, :sorted => true, :uniqued => true do
-          []
+          spec_consumers.flat_map(&:weak_frameworks)
         end
 
         # @return [Array<String>]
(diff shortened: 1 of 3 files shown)


spec/unit/target/build_settings/aggregate_target_settings_spec.rb
```diff
@@ -465,6 +465,48 @@ def pod_target(spec, target_definition)
             @generator.other_ldflags.should == %w(-ObjC -l"StaticLibrary" -l"VendoredDyld" -l"xml2" -framework "PodTarget" -framework "VendoredFramework" -framework "XCTest")
```
dnkoutso: @segiddins I don't understand why `-l"xml2"` or `-framework "XCTest"` are included in the settings of this aggregate target since we are using frameworks...

dnkoutso: wait I think I do...we always propagate these to the aggregate target even if we use frameworks...

dnkoutso:
Wait a minute this might be a bigger regression than I thought.

dnkoutso:
Well nevermind the difference is now in 1.6.0.beta.1 we link all system frameworks (including weak with this change) into the aggregate target...no regression there but this PR still stands and fixes this issue.

lib/cocoapods/target/build_settings.rb
```diff
@@ -552,7 +552,7 @@ def initialize(target, test_spec = nil)
 
         # @return [Array<String>]
         define_build_settings_method :weak_frameworks_to_import, :memoized => true, :sorted => true, :uniqued => true do
-          []
+          spec_consumers.flat_map(&:weak_frameworks)
```
dnkoutso: I _think_ this is the right change? I am having a hard time to follow it here.

Issue #8031, opened 2018-08-21
Error running unit tests for a static framework


felipe-valio-movile:
* [X] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

1. Create a pod with `pod lib create TestExample`
2. Turn the pod in a static framework by adding `s.static_framework = true`
3. Create a unit test class in `TestExample/Tests/FrameworkTests.swift`
4. Create a test spec in `TestExample.podspec` like this:

```
s.test_spec 'Tests' do |test_spec|
  test_spec.source_files = 'TestExample/Tests/**/*'
end
```

5. Add `TestExample-Unit-Tests` to tests action of the project scheme
6. Run the tests

## What did you expect to happen?

The pod tests should run correctly.

## What happened instead?

The app tests runs fine, but the pod tests fails with this console message:

```
2018-08-21 13:59:45.787495-0300 xctest[40101:8820276] The bundle “TestExample-Unit-Tests” couldn’t be loaded because it is damaged or missing necessary resources. Try reinstalling the bundle.
2018-08-21 13:59:45.788556-0300 xctest[40101:8820276] (dlopen_preflight(/Users/felipe.valio/Library/Developer/Xcode/DerivedData/TestExample-gyfgqphahycqtigtvqsrufgjzjlq/Build/Products/Debug-iphonesimulator/TestExample-Unit-Tests.xctest/TestExample-Unit-Tests): Library not loaded: @rpath/TestExample.framework/TestExample
  Referenced from: /Users/felipe.valio/Library/Developer/Xcode/DerivedData/TestExample-gyfgqphahycqtigtvqsrufgjzjlq/Build/Products/Debug-iphonesimulator/TestExample-Unit-Tests.xctest/TestExample-Unit-Tests
  Reason: image not found)
Program ended with exit code: 82
```

### Stack

```
   CocoaPods : 1.5.3
        Ruby : ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-darwin16]
    RubyGems : 2.6.11
        Host : Mac OS X 10.13.6 (17G65)
       Xcode : 9.4.1 (9F2000)
         Git : git version 2.8.1
Ruby lib dir : /Users/felipe.valio/.rvm/rubies/ruby-2.4.1/lib
[...]

PR #8032, opened 2018-08-21
Clear out `MACH_O_TYPE` for unit test bundles that use static frameworks.

dnkoutso:
closes https://github.com/CocoaPods/CocoaPods/issues/8031

### DIFF
--- a/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
@@ -285,6 +285,9 @@ module Pod
                 # requires frameworks. For tests we always use the test target name as the product name
                 # irrelevant to whether we use frameworks or not.
                 configuration.build_settings['PRODUCT_NAME'] = name
+                # target_installer sets 'MACH_O_TYPE' for static frameworks ensure this does not propagate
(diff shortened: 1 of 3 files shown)


lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -285,6 +285,9 @@ def add_test_targets
                 # requires frameworks. For tests we always use the test target name as the product name
                 # irrelevant to whether we use frameworks or not.
                 configuration.build_settings['PRODUCT_NAME'] = name
+                # target_installer sets 'MACH_O_TYPE' for static frameworks ensure this does not propagate
+                # to test target.
+                configuration.build_settings.delete('MACH_O_TYPE')
```
dnkoutso: I really dislike that super class sets a bunch of these and we have to delete them but I won't refactor it for 1.6.0.

amorde: agree this isn't ideal

dnkoutso:
This bug actually exists in 1.5.3 too.

amorde:
Yeah looks like the bug was reported while using 1.5.3

PR #8059, opened 2018-08-29
Validate all generated native targets have source files

segiddins:
Validate that all generated `PBXNativeTarget`s contain source files to build,
  so specs (including test specs) with no source files won't fail at runtime
  due to the lack of a generated executable.

### DIFF
--- a/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
@@ -50,6 +50,7 @@ module Pod
               test_resource_bundle_targets = add_resources_bundle_targets(test_file_accessors)
 
               add_files_to_build_phases(native_target, test_native_targets)
+              validate_targets_contain_sources(test_native_targets + [native_target])
 
               create_xcconfig_file(native_target, resource_bundle_targets)
               create_test_xcconfig_files(test_native_targets, test_resource_bundle_targets)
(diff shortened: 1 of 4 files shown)


lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -821,6 +822,14 @@ def add_swift_static_library_compatibility_header_phase(native_target)
             )
           end
 
+          def validate_targets_contain_sources(native_targets)
+            native_targets.each do |native_target|
+              next unless native_target.source_build_phase.files.empty?
+
+              raise Informative, "Unable to install the `#{target.label}` pod, because the #{native_target} target in Xcode would have no sources to compile"
```
dnkoutso: nit: please add a period at the end of the sentence

dnkoutso: wrap `#{native_target}` backticks?

spec/unit/installer/xcode/pods_project_generator/pod_target_installer_spec.rb
```diff
@@ -251,6 +251,13 @@ class PodsProjectGenerator
                 installation_result.test_native_targets.count.should == 2
               end
 
+              it 'raises when a test spec has no source files' do
+                @watermelon_pod_target.test_spec_consumers.first.stubs(:source_files).returns([])
+                e = ->() { @installer.install! }.should.raise Informative
+                e.message.should.
+                  include 'Unable to install the `WatermelonLib` pod, because the `WatermelonLib-Unit-Tests` target in Xcode would have no sources to compile.'
+              end
```
amorde: could you add a test for the non-test target case?

dnkoutso: sure!

dnkoutso: Actually this cannot be tested. The reason being is that `pod_target_installer.rb` checks for `unless target.should_build?` and therefore a target without any sources would be integrated via a `PBXAggregateTarget` instead.

Are you OK with that?

amorde: Yeah makes sense. could / should `should_build?` be used here as well?

dnkoutso: I think its more of a `should_build_test_spec?` but I think its becoming convoluted. We really ought to split a separate pod target instance even for test specs maybe...`pod_target` is a god object.

PR #8076, opened 2018-09-05
Refactor framework path hash map into its own model

amorde:


lib/cocoapods/target/framework.rb
```diff
@@ -0,0 +1,30 @@
+module Pod
+  class Framework
```
amorde: I'm terrible at naming things - let me know if you have a better name idea

lib/cocoapods.rb
```diff
@@ -43,6 +43,7 @@ def message
   autoload :Deintegrator,              'cocoapods_deintegrate'
   autoload :Executable,                'cocoapods/executable'
   autoload :ExternalSources,           'cocoapods/external_sources'
+  autoload :Framework,                 'cocoapods/target/framework'
```
amorde: is this a decent place to put this?

CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Refactor framework path hash map into its own model  
```

</p>

lib/cocoapods/target/pod_target.rb
```diff
@@ -244,34 +244,27 @@ def contains_test_specifications?
...
-                          :output_path => "${TARGET_BUILD_DIR}/${FRAMEWORKS_FOLDER_PATH}/#{framework_path.basename}" }
+            framework_source = "${PODS_ROOT}/#{relative_path_to_sandbox}"
             # Until this can be configured, assume the dSYM file uses the file name as the framework.
             # See https://github.com/CocoaPods/CocoaPods/issues/1698
             dsym_name = "#{framework_path.basename}.dSYM"
             dsym_path = Pathname.new("#{framework_path.dirname}/#{dsym_name}")
-            if dsym_path.exist?
-              framework[:dsym_name] = dsym_name
-              framework[:dsym_input_path] = "${PODS_ROOT}/#{relative_path_to_sandbox}.dSYM"
-              framework[:dsym_output_path] = "${DWARF_DSYM_FOLDER_PATH}/#{dsym_name}"
```
amorde: turns out `:dsym_name` & `:name` were unused, and `:output_path` and `:dsym_output_path` were only used to build the script output path list. so I pulled the logic for the output paths out of the target since it seems like it really shouldn't be there anyway

spec/unit/installer/user_project_integrator/target_integrator_spec.rb
```diff
@@ -637,6 +612,47 @@ module Pod
...
+          ]
+          TargetIntegrator.framework_output_paths(paths).sort.should == %w(
+            ${DWARF_DSYM_FOLDER_PATH}/SomeFramework.framework.dSYM
+            ${TARGET_BUILD_DIR}/${FRAMEWORKS_FOLDER_PATH}/CompiledFramework.framework
+            ${TARGET_BUILD_DIR}/${FRAMEWORKS_FOLDER_PATH}/SomeFramework.framework
+          )
+        end
+
+        it 'calculates the output paths of the copy resources script' do
+          resource_paths = %w(
```
amorde: I didn't change this logic but it didn't have a dedicated unit test so I added one

lib/cocoapods/target/framework.rb
```diff
@@ -0,0 +1,30 @@
...
+    # @return [String] the path to the .framework
+    #
+    attr_reader :source_path
```
amorde: also not sure about the name on this one - maybe `:root_path`? 🤷‍♂️

lib/cocoapods/target/framework.rb
dnkoutso: `FrameworkPath`?

spec/unit/installer/user_project_integrator/target_integrator_spec.rb
dnkoutso: could have sworn this existed was it only framework paths?

lib/cocoapods/target/framework.rb
```diff
@@ -0,0 +1,30 @@
...
+        other.source_path == @source_path && other.dsym_path == @dsym_path
+      else
+        false
+      end
+    end
+
+    alias eql? ==
+
+    def hash
+      source_path.hash
```
dnkoutso: was this generated?

dnkoutso:
I kinda wanted to do the same. Was thinking if a `Struct` could also do the job but structs are not immutable versus this will be.

spec/unit/installer/user_project_integrator/target_integrator_spec.rb
amorde: there are tests but not for this individual function, they are more like integration tests

https://github.com/CocoaPods/CocoaPods/blob/ab5bd128bbd4a78328619993c3bbb76ef9304078/spec/unit/installer/user_project_integrator/target_integrator_spec.rb#L323-L363

lib/cocoapods/target/framework.rb
amorde: nope - did I do this incorrectly? I'm mostly used to having to implement equality in hashing for custom models but I'm still a noob at Ruby

amorde:
yeah I thought of a `Struct` as well, but I thought a class would be more flexible going forward. could change it to a `Struct` if that's preferred though

lib/cocoapods/target/framework_paths.rb
```diff
@@ -0,0 +1,30 @@
...
+  class FrameworkPaths
+    # @return [String] the path to the .framework
+    #
+    attr_reader :source_path
+
+    # @return [String, Nil] the dSYM path, if one exists
+    #
+    attr_reader :dsym_path
+
+    def initialize(source_path, dsym_path)
```
dnkoutso: nit can make `dsym_path = nil` as default

spec/unit/installer/user_project_integrator/target_integrator_spec.rb
dnkoutso: cool!

lib/cocoapods/target/framework_paths.rb
dnkoutso: this way you can get rid of the `, nil` everywhere else

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_integrator.rb
```diff
@@ -78,14 +78,15 @@ def add_embed_frameworks_script_phase(native_target, test_spec)
             framework_paths = target.dependent_targets_for_test_spec(test_spec).flat_map do |dependent_target|
               spec_paths_to_include = dependent_target.non_test_specs.map(&:name)
               spec_paths_to_include << test_spec.name if dependent_target == target
-              dependent_target.framework_paths.values_at(*spec_paths_to_include).flatten.compact.uniq
-            end
+              dependent_target.framework_paths.values_at(*spec_paths_to_include).flatten.compact
+            end.uniq
```
dnkoutso: any reason this moved here?

lib/cocoapods/target/framework_paths.rb
```diff
@@ -0,0 +1,30 @@
...
+        other.source_path == @source_path && other.dsym_path == @dsym_path
+      else
+        false
+      end
+    end
+
+    alias eql? ==
+
+    def hash
+      source_path.hash
```
dnkoutso: I still think the hash should include the `dsym_path` if its not `nil`

lib/cocoapods/target/framework_paths.rb
amorde: yeah makes sense

lib/cocoapods/installer/user_project_integrator/target_integrator.rb
```diff
@@ -254,6 +255,23 @@ def resource_output_paths(resource_input_paths)
...
+          # @return [Array<String>] The framework output paths
+          #
+          def framework_output_paths(framework_input_paths)
+            framework_input_paths.flat_map do |framework_path|
+              framework_output_path = "${TARGET_BUILD_DIR}/${FRAMEWORKS_FOLDER_PATH}/#{File.basename(framework_path.source_path)}"
+              dsym_path = if (dsym_input_path = framework_path.dsym_path)
+                            "${DWARF_DSYM_FOLDER_PATH}/#{File.basename(dsym_input_path)}"
+                          end
+              [framework_output_path, dsym_path]
+            end.flatten.compact.uniq
```
segiddins: given this is already `flat_map`ped, the `flatten` probably isn't needed

lib/cocoapods/target/framework_paths.rb
```diff
@@ -0,0 +1,34 @@
+module Pod
+  class FrameworkPaths
```
segiddins: this constant should be nested in the `Target` namespace

amorde:
Moved the class into the `Target` namespace, let me know if you have any other suggestions

lib/cocoapods/target/framework_paths.rb
```diff
@@ -0,0 +1,36 @@
...
+        if other.class == self.class
+          other.source_path == @source_path && other.dsym_path == @dsym_path
+        else
+          false
+        end
+      end
+
+      alias eql? ==
+
+      def hash
```
dnkoutso: do you use rubymine? I am curious if you were to generate this from the IDE what it would output?

lib/cocoapods/installer/user_project_integrator/target_integrator.rb
```diff
@@ -254,6 +255,23 @@ def resource_output_paths(resource_input_paths)
...
           end
+
+          # Returns the framework output paths for the given input paths
+          #
+          # @param  [Array<Target::FrameworkPaths>] framework_input_paths
+          #         The framework input paths to map to output paths.
+          #
+          # @return [Array<String>] The framework output paths
+          #
+          def framework_output_paths(framework_input_paths)
```
dnkoutso: alternatively we can have each instance of `FrameworkPaths` declare a `def output_path` and then `map(&:output_path)`?

spec/unit/installer/user_project_integrator/target_integrator_spec.rb
```diff
@@ -417,24 +415,13 @@ module Pod
         end
 
         it 'adds embed frameworks build phase input and output paths for vendored and non vendored frameworks' do
-          debug_vendored_framework = { :name => 'DebugVendoredFramework.framework',
```
dnkoutso: much cleaner

lib/cocoapods/target/framework_paths.rb
amorde: I do use rubymine but I'm not sure how to generate `hash` implementations

are you worried about the correctness of the implementation? for hashing the main requirement is if something is `==` to something else, their hashes must also be equal, but not the other way around. collisions are OK as long as `==` is correct

also since we're piggy backing on `Array`'s implementation we get it for free, if `Array`'s `hash` doesn't work then there are bigger problems :)

lib/cocoapods/installer/user_project_integrator/target_integrator.rb
amorde: I think the logic for where the frameworks are placed in the built product shouldn't live within that model. it's a bit weird though because the embed frameworks script actually duplicates the logic of both the input and the output paths

https://github.com/CocoaPods/CocoaPods/blob/92385cabea6b572867ab39ec742c862cd942fc07/lib/cocoapods/generator/embed_frameworks_script.rb#L76-L84

so really, this is only controlling what the script's input and output paths are set in Xcode, and it doesn't affect the actual script contents at all. should probably unify that logic too

amorde: this also matches nicely with the `resource_output_paths` function right above this one, so at least its consistent? 🤷‍♂️

dnkoutso:
@amorde just needs a rebase and we merge.

PR #8097, opened 2018-09-17
Provide an installation option to preserve folder structure

dacaiguoguogmail:
Make user can keep folder structure by add "keep_folder_structure" into Podfile

I know this is not a good practice.
What is the correct way?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 * Remove contraction from docs to fix rendering on the website.  
   [stevemoser](https://github.com/stevemoser)
   [#8131](https://github.com/CocoaPods/CocoaPods/pull/8131)
+* Provide an installation option to preserve folder structure  
+  [dacaiguoguogmail](https://github.com/dacaiguoguogmail)
+  [#8097](https://github.com/CocoaPods/CocoaPods/pull/8097)
 
 * mark_ruby_file_ref add indent width and tab width config  
   [dacaiguoguogmail](https://github.com/dacaiguoguogmail)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 * Remove contraction from docs to fix rendering on the website.  
   [stevemoser](https://github.com/stevemoser)
   [#8131](https://github.com/CocoaPods/CocoaPods/pull/8131)
+
 * Provide an installation option to preserve folder structure  
   [dacaiguoguogmail](https://github.com/dacaiguoguogmail)
   [#8097](https://github.com/CocoaPods/CocoaPods/pull/8097)
--- a/lib/cocoapods/installer/installation_options.rb
+++ b/lib/cocoapods/installer/installation_options.rb
@@ -109,7 +109,7 @@ module Pod
       option :warn_for_multiple_pod_sources, true
       option :share_schemes_for_development_pods, false
       option :disable_input_output_paths, false
-      option :preserve_file_structure, false
+      option :preserve_pod_file_structure, false
 
       module Mixin
         module ClassMethods
--- a/lib/cocoapods/installer/installation_options.rb
+++ b/lib/cocoapods/installer/installation_options.rb
@@ -109,6 +109,7 @@ module Pod
       option :warn_for_multiple_pod_sources, true
       option :share_schemes_for_development_pods, false
       option :disable_input_output_paths, false
+      option :preserve_file_structure, false
 
       module Mixin
         module ClassMethods
--- a/lib/cocoapods/installer/xcode/pods_project_generator/file_references_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/file_references_installer.rb
@@ -18,36 +18,21 @@ module Pod
           #
           attr_reader :pods_project
 
-          # @return [bool] Whether you want the folder structure.
+          # @return [Bool] preserve file structure when value is true.
           #
-          attr_reader :keep_folder_structure
-
-          PODFILE_NAMES = [
-            'CocoaPods.podfile.yaml',
-            'CocoaPods.podfile',
-            'Podfile',
-          ].freeze
-
+          attr_reader :preserve_file_structure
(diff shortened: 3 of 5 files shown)


lib/cocoapods/installer/xcode/pods_project_generator/file_references_installer.rb
```diff
@@ -28,6 +38,16 @@ def initialize(sandbox, pod_targets, pods_project)
             @sandbox = sandbox
             @pod_targets = pod_targets
             @pods_project = pods_project
+            @keep_folder_structure = false
+            
+            PODFILE_NAMES.each do |filename|
+              candidate = (Pathname sandbox.root).parent + filename
+              if candidate.exist?
+                if candidate.read.include? 'keep_folder_structure'
```
amorde: this definitely not the way functionality would be added to the Podfile - if you want to introduce new features there then most likely the Podfile DSL is where to do it. See here: https://github.com/CocoaPods/Core/blob/master/lib/cocoapods-core/podfile/dsl.rb

lib/cocoapods/installer/xcode/pods_project_generator/file_references_installer.rb
```diff
@@ -204,6 +224,11 @@ def add_file_accessors_paths_to_pods_group(file_accessor_key, group_key = nil, r
 
               pod_name = file_accessor.spec.name
               local = sandbox.local?(pod_name)
+              if keep_folder_structure
+                local = true
```
amorde: maybe rename the variable to `preserve_file_structure` rather than conflating the idea of what is a local / development pod

dacaiguoguogmail:
Thank you for your suggestion, I will finish it in the next few days. @amorde

dacaiguoguogmail:
@amorde new commit has been pushed

dnkoutso:
Should the flag be more explicit that it pertains to local pods only?

dacaiguoguogmail:
Add more explicit comment for preserve_file_structure @dnkoutso

amorde:
this change makes it apply to all pods, correct? apologies haven't had time to try it myself

dacaiguoguogmail:
Yes, this change makes it apply to all pods

dacaiguoguogmail:
Is there any other place to modify? @dnkoutso @amorde

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 * Remove contraction from docs to fix rendering on the website.  
   [stevemoser](https://github.com/stevemoser)
   [#8131](https://github.com/CocoaPods/CocoaPods/pull/8131)
+* Provide an installation option to preserve folder structure
```
dnkoutso: press enter here

lib/cocoapods/installer/xcode/pods_project_generator/file_references_installer.rb
```diff
@@ -18,16 +18,21 @@ class FileReferencesInstaller
           #
           attr_reader :pods_project
 
+          # @return [Bool] add support for preserving the file structure of externally sourced pods, in addition to local pods.
+          #
+          attr_reader :preserve_file_structure
```
dnkoutso: also press enter here

amorde:
have been thinking about this for a bit - is this really something we want to support? what was the motivation for adding this feature? 

wanting to avoid feature creep wherever it makes sense. thanks for your patience :)

dnkoutso:
I agree. I think it adds yet another option. What are we trying to solve for?

dacaiguoguogmail:
There are a lot of pods in my project.
Each pod has different folders divided by function.
Most can't be treated as a sub pod.
Only the development pod or sub pod can be displayed by group in the pod definition.
Many sub pods will cause `pod install` to be slow.
So I want to add an option to preserve file structure instead of adding a lot of sub pods for this purpose.

amorde:
Makes sense, but are you frequently reading through the source of external Pods? It seems like if its a big part of your workflow then they should be development pods instead, but maybe I am missing something

dacaiguoguogmail:
I use our private CocoaPods repository to Modularize a Big iOS App, There have been many pods that iterate along with the App version

amorde:
I see - so the primary use case sounds like private pods that aren't development pods, but are still internal source code that would be easier to read if it included the file structure. Make sense to me

dacaiguoguogmail:
Yes,  thanks for your patience :) .

lib/cocoapods/installer/installation_options.rb
```diff
@@ -109,6 +109,7 @@ def hash
       option :warn_for_multiple_pod_sources, true
       option :share_schemes_for_development_pods, false
       option :disable_input_output_paths, false
+      option :preserve_file_structure, false
```
amorde: could you rename this to something that makes it clear it's meant for pod sources? maybe `preserve_pod_file_structure`

lib/cocoapods/installer/xcode/pods_project_generator/file_references_installer.rb
```diff
@@ -203,11 +209,11 @@ def add_file_accessors_paths_to_pods_group(file_accessor_key, group_key = nil, r
...
 
               pod_name = file_accessor.spec.name
-              local = sandbox.local?(pod_name)
-              base_path = local ? common_path(paths) : nil
+              preserve_file_structure_flag = (sandbox.local?(pod_name) || preserve_file_structure)
+              base_path = preserve_file_structure_flag ? common_path(paths) : nil
               group = pods_project.group_for_spec(pod_name, group_key)
               paths.each do |path|
-                pods_project.add_file_reference(path, group, local && reflect_file_system_structure_for_development, base_path)
+                pods_project.add_file_reference(path, group, preserve_file_structure_flag && reflect_file_system_structure_for_development, base_path)
```
amorde: the `reflect_file_system_structure_for_development` function argument could be repurposed here to be `reflect_file_system_structure`, and remove the check for `local`. then the places that call `add_file_accessors_paths_to_pods_group` could reference the `preserve_file_structure` setting. that keeps this function flexible and it would rely less on internal state

dacaiguoguogmail:
I tried to remove the check of the pod, but it caused the unit test to fail.

Issue #8073, opened 2018-09-03
XCode10 (new build system) - if incremental build, Embed script doesn't run


mkubista:
* [x] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

ℹ I made changes in some code files in local development pod 

## What did you expect to happen?

ℹ After build and run in xCode, the changes are incorporated into the resulting package.

## What happened instead?

ℹ The old binary is in package. The `[CP] Embed Pods Frameworks` script doesnt run. Its because of build optimalisation in new xCode - when it detects no change in input or output files specified for script, the script is not started.
When we did enter the binary file from the content of the `.framework` folder into the input files list, everything is OK. Clean build is also OK.
So the solution would be to have no input and out files at all, or during `pod install` add also content files of frameworks into the input files list.

Strange thing - in the build log in Report navigator, the output of the `[CP] Embed Pods Frameworks` appeared even when it didn't actually run (I modified the script to echo current time, so I can see the same time stamps)

## CocoaPods Environment

ℹ 
```
   CocoaPods : 1.6.0.beta.1
        Ruby : ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-darwin16]
    RubyGems : 2.6.11
        Host : Mac OS X 10.14 (18A371a)
       Xcode : 10.0 (10L232m)
         Git : git version 2.17.1 (Apple Git-112)
Ruby lib dir : /usr/local/Cellar/ruby/2.4.1_1/lib
Repositories : artsy - https://github.com/Artsy/Specs.git @ 8c00e26ca1bd8ba163587f0eb592e9cd5a241d0e
               master - https://github.com/CocoaPods/Specs.git @ 058895fdadffd84d53c3ddc27a31591ea2f89ce7
               SSCPWrapper - https://tfs.kswr.cz/tfs/DefaultCollection/Jablotron/_git/MyCompany-iOS @ 518f383f1524b77aad7ddaa1d50eb72e4b824594
```

## Project that demonstrates the issue

ℹ Unfortunately, project cannot be shown in public.

Legoless:
I have the same issue. This causes me having to rebuild the entire project, which takes about 15 minutes. But I only changed a single line in code in one of the development pods.

mkubista:
I have a possible temporary solution
[...]

PR #8105, opened 2018-09-18
Provide an installation option to disable usage of input/output paths.

dnkoutso:
closes https://github.com/CocoaPods/CocoaPods/issues/8073

We cannot do anything about this until Apple resolves the issue with Xcode 10, dynamic frameworks and script phases not being re-run due to a change in the folder. 

In the meantime we provide users the option to disable input/output paths so the script phase will always run. 

For those who opt in to use this option, it mean slower but correct incremental builds.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,11 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [#8064](https://github.com/CocoaPods/CocoaPods/issues/8064)
 
 ##### Bug Fixes
-  
+
+* Provide an installation option to disable usage of input/output paths.  
+  [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
+  [#8073](https://github.com/CocoaPods/CocoaPods/issues/8073)
+
 * Scope prefix header setting to each test spec.  
   [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
   [#8046](https://github.com/CocoaPods/CocoaPods/pull/8046)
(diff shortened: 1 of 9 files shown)


orta:
Is this something that's really useful in the long run? It _feels_ like an underlaying tooling abstraction that we're exposing as an option to users.

If it's temporary, it might be worth prefixing it with `tmp` or something, any as any DSL additions will stick around basically forever at this point. At least we can set expectations then.

dnkoutso:
Not that useful to be honest...but somewhat harmless to have? I just do not know if and when Apple will fix that issue and it seems to be a major blocker for folks to update to Xcode 10 and new build system...I am out of ideas :|

orta:
A fair argument, can you think of a way so that it could be easily exposed to the `post_install` hook maybe? I'm mainly trying to set expectations that the DSL addition should be temporary

amorde:
yeah good points on the DSL - maybe just a boolean on `Installer` that could be overridden inside a `pre_install`?

dnkoutso:
The only reason I would push back on this is that I do not know of any prior precedent of configuration installation via `pre_install` hooks. In other words are we unnecessarily just introducing a new way for folks to interface with installation ?

orta:
It's been done before but only in a larger refactoring [in 0.38](http://blog.cocoapods.org/CocoaPods-0.38/) we revised the API to consolidate the sort of random tweaks people were doing (build settings, swift versions (maybe, might have predated that))

But I do get what you're saying, maybe it's going off too far into the woods. Just trying to fight the good fight for minimal config.

amorde:
@zeusent in your Podfile, you can use the `install!` directive to specify installation options

```ruby
install! 'cocoapods', :disable_input_output_paths => true
```

See: https://guides.cocoapods.org/syntax/podfile.html#install_bang

dnkoutso:
Also see https://github.com/CocoaPods/CocoaPods/pull/9125

PR #8134, opened 2018-09-26
[Gemspec] Bump ruby-macho to '~> 2.1'

woodruffw:
See https://github.com/Homebrew/ruby-macho/pull/113 and https://github.com/Homebrew/ruby-macho/pull/114.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -120,7 +120,7 @@ PATH
       gh_inspector (~> 1.0)
       molinillo (~> 0.6.6)
       nap (~> 1.0)
-      ruby-macho (~> 1.2)
+      ruby-macho (~> 1.3, >= 1.3.1)
       xcodeproj (>= 1.6.0, < 2.0)
 
 GEM
@@ -236,7 +236,7 @@ GEM
       ruby-progressbar (~> 1.7)
       unicode-display_width (~> 0.3)
     ruby-graphviz (1.2.2)
-    ruby-macho (1.2.0)
+    ruby-macho (1.3.1)
     ruby-prof (0.15.2)
     ruby-progressbar (1.7.5)
     safe_yaml (1.0.4)
@@ -297,4 +297,4 @@ DEPENDENCIES
   xcodeproj!
 
 BUNDLED WITH
-   1.16.3
+   1.16.4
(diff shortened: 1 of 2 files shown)


dnkoutso:
Seems like this does not work in CI:
```
ruby-macho-2.1.0 requires ruby version >= 2.2, which is incompatible with the
current version, ruby 2.0.0p647
```

@segiddins @DanToml is 2.0.0 of Ruby required to be tested?

segiddins:
Yes, CocoaPods still supports ruby 2.0.0

dnkoutso:
@woodruffw is it easy to cherry-pick this your fix in the 1.x release?

shekhar-suman:
@woodruffw Thanks for working on this issue. Can we please merge this soon. Xcode 10 is already out and this is breaking our build.

amorde:
@shekhar-suman you can use [Bundler](https://bundler.io/) to pull in the version of `ruby-macho` you need if this is a blocker for you

shekhar-suman:
@amorde I do have latest ruby-macho and it works fine if used standalone. My problem is with Cocoapod as it is still dependent and using older version of ruby-macho. Hence my 'pod install' is failing. It can not use a vendored_framework for a framework which has arm64_32 arch.

woodruffw:
@shekhar-suman For a temporary fix, you can always modify your local copy of CocoaPods (I don't know anything about how CocoaPods is installed, but if it's anything like Homebrew you can just check this branch out and run `bundle install`). Similarly, unless you're actually using that `arm64_32` slice, you can always temporarily disable it in your build.

@dnkoutso Probably not, since there were additional changes in the 1.x tree that are marked as non-2.0 compatible. However, as I mentioned in https://github.com/CocoaPods/CocoaPods/pull/7815, the library *probably* still works on 2.0, I just don't have a CI setup to test it on 2.0 anymore. I'll look at tweaking the Ci to remedy that.

woodruffw:
Tried applying some hacks to bring things back down to 2.0 in https://github.com/Homebrew/ruby-macho/pull/115, but the library's changed a bit more than I was aware (since I usually develop on the latest ruby and let the CI sort things out). In particular, ruby-macho currently makes use the `String#+@` operator for unfreezing literals and uses `Integer#negative?` in a few places, both of which didn't appear until Ruby 2.3.

macOS 10.13 and higher supply at least Ruby 2.3, which would work just fine; would it be possible for CocoaPods to do something similar to Homebrew and use a vendored Ruby version on older system versions?

segiddins:
CocoaPods always supports back 2 OS X / Xcode versions and is installed as a gem, so vendoring an older ruby version isn’t possible

woodruffw:
I've made one last Ruby 2.0/2.1/2.2 compatible release, containing all of the feature additions and bug fixes from the latest (incompatible) release.

segiddins:
Thanks!

PR #8136, opened 2018-09-26
Fixes the double .xcasset folder compilation issue

Fab1n:
The old code compares a **lowercase** part of a string ( of`$line`) with an uppercase part of a string (of `${PODS_ROOT}`)?
My dev directory lives in `/Users/<username>/Documents/...`.

The `find` command seems to ignore the **uppercase** `Documents` and prints it **lowercase**.
The problem is `$POD_ROOT` contains `Documents` as **uppercase**.

Thus `[[ $line != "${PODS_ROOT}*" ]]` can never be true in my case, am I right?

These changes fix the problem. Any suggestions?

Seems to fix:
#6159

Related issues:
#7745
#7779
#7617
#7785

### DIFF
--- a/lib/cocoapods/generator/copy_resources_script.rb
+++ b/lib/cocoapods/generator/copy_resources_script.rb
@@ -218,9 +218,7 @@ then
   # Find all other xcassets (this unfortunately includes those of path pods and other targets).
   OTHER_XCASSETS=$(find "$PWD" -iname "*.xcassets" -type d)
   while read line; do
-    LOWER_LINE="$(echo $line | tr '[:upper:]' '[:lower:]')"
-    LOWER_PODS_ROOT="$(echo $PODS_ROOT | tr '[:upper:]' '[:lower:]')"
-    if [[ $LOWER_LINE != ${LOWER_PODS_ROOT}* ]]; then
+    if [[ $line != ${PODS_ROOT}* ]]; then
       XCASSET_FILES+=("$line")
     fi
   done <<<"$OTHER_XCASSETS"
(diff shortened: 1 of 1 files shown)


dnkoutso:
How does it resolve https://github.com/CocoaPods/CocoaPods/issues/6159? The problem here happens with multiple pods containing `.xcasset` files right?

Fab1n:
I'll check your feedback and come back to all that

Fab1n:
Ok, now I have updated the fix to the bare minimum, I also edit my initial explanation.
With the newest updates the bug is fixed and I think #6159 should then also be fixed (if I understood the problem right).

For everyone who is curious of how to reproduce the issue, I have setup a detailed sample project here:
https://github.com/Fab1n/PodResourcesBug

The sample project actually shows 2 recent issues regarding asset catalog compilation.
For this issue here you will need to checkout the second bug.

The test procedure is written in detail in the repos [README.md](https://github.com/Fab1n/PodResourcesBug/blob/master/README.md)

Fab1n:
The test expectations have to be updated also accordingly, but only if that fix is accpeted

JimmyYezeguelian:
Excuse me, but have you find a solution/fix for this issue ? 
Thanks a lot guys

Fab1n:
> Excuse me, but have you find a solution/fix for this issue ?
> Thanks a lot guys

I did not get any response.
But what I did is update the pod that was using  an `xcasset` folder in the `resources` attribute of the spec.
For details, see an the real fix here:
https://github.com/pkluz/PKHUD/commit/f8110f601d2a6ac73fa260c7744b93c2802ad7a2

Then the `xcassets` folder will not be referenced twice because it is in a separate bundle.

dnkoutso:
Will circle back to this for, sorry havent had time lately.

dnkoutso:
Closing as outdated. We will be working xcassets at some point in the future.

PR #6401, opened 2017-01-11
[Generator] Replace -iquote by -isystem for frameworks in xcconfig

guillaumealgis:
🌈

- Replace usage of the `-iquote` compiler flag by `-isystem` when searching for framework headers. This will silence compilation warning in the Pods' headers files.
> "Header files found in directories added to the search path with the `-isystem` and `-idirafter` command-line options are treated as system headers for the purposes of diagnostics"
> "All warnings [...] are suppressed while GCC is processing a system header"
[https://gcc.gnu.org/onlinedocs/cpp/System-Headers.html](https://gcc.gnu.org/onlinedocs/cpp/System-Headers.html)

- Update unit tests to reflect the above change.

Integrations tests are updated in https://github.com/CocoaPods/cocoapods-integration-specs/pull/94.

This fixes #5589.

When I tested locally with `bundle exec rake spec:all`, it failed on the Alamofire example, but I *think* this is normal and the example was not updated for Xcode 8 (please correct me if I'm wrong). When running the same tests as Travis thought, everything passed.

```
[...]

PR #8111, opened 2018-09-19
Nests test specs host apps inside that Pod's directory for cleaner project navigators.

dostrander:
@segiddins @dnkoutso

PR #8153, opened 2018-10-02
Revert to CocoaPods integration specs

amorde:
Accidentally switched to a fork in #6401


amorde:
Never mind, will be done in https://github.com/CocoaPods/CocoaPods/pull/8111

Issue #8139, opened 2018-09-27
Bug with Unit Tests on Mojave: "The bundle couldn’t be loaded because it is damaged or missing necessary resources"


jegnux:
* [x] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report
Since I upgraded to macOS Mojave, I can't run unit test of modules including cocoapods. Please check the sample app to understand/reproduce.

## What did you do?

- Download sample project at https://cl.ly/990d99adec49
- Unzip
- In CLI, go to the project directory and run `pod install`
- Open the `UnitTestIssue.xcworkspace` file with **Xcode 10** on **macOS Mojave**
- Run unit tests on *App* scheme

ℹ The sample project is a workspace composed by an iOS application target (`App`) and a framework (`AppModule`) imported by the app. Both target includes `RxSwift` (as example) through CocoaPods.

## What did you expect to happen?

✅ `AppTests` is executed and all its tests correctly pass
✅ `AppModuleTests` is executed and all its tests correctly pass

## What happened instead?

✅ `AppTests` is executed and all its tests correctly pass
❌ `AppModuleTests` is **not** executed and the console displays the following error:

```
2018-09-27 16:49:34.580550+0200 xctest[3033:761430] The bundle “AppModuleTests” couldn’t be loaded because it is damaged or missing necessary resources. Try reinstalling the bundle.
2018-09-27 16:49:34.580849+0200 xctest[3033:761430] (dlopen_preflight(/Users/jeromealves/Library/Developer/Xcode/DerivedData/UnitTestIssue-akzscdcjjahzpmdxkzgmyhdxntrl/Build/Products/Debug-iphonesimulator/AppModuleTests.xctest/AppModuleTests): Library not loaded: @rpath/RxSwift.framework/RxSwift
  Referenced from: /Users/jeromealves/Library/Developer/Xcode/DerivedData/UnitTestIssue-akzscdcjjahzpmdxkzgmyhdxntrl/Build/Products/Debug-iphonesimulator/AppModuleTests.xctest/AppModuleTests
  Reason: image not found)
Program ended with exit code: 82
```

### Stack

```
   CocoaPods : 1.6.0.beta.1
        Ruby : ruby 2.3.7p456 (2018-03-28 revision 63024) [universal.x86_64-darwin18]
    RubyGems : 2.5.2.3
[...]

PR #8172, opened 2018-10-10
Set 'ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES' to YES in Debug mode

icecrystal23:
This works around a bug in Xcode 10 where unit test targets don't correctly embed the necessary Swift libraries into unit test targets.

### DIFF
--- a/spec/unit/installer/xcode/pods_project_generator/target_installer_spec.rb
+++ b/spec/unit/installer/xcode/pods_project_generator/target_installer_spec.rb
@@ -62,6 +62,16 @@ module Pod
               'AppStore' => 'staticlib',
             }
           end
+
+          it 'embeds swift libraries in Debug mode' do
+            @target.stubs(:product_type).returns(:framework)
+            @installer.send(:add_target).resolved_build_setting('ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES').should == {
+              'Release' => nil,
+              'Debug' => 'YES',
+              'Test' => nil,
+              'AppStore' => nil,
+            }
+          end
         end
       end
     end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   with the same dependencies.  
   [Samuel Giddins]
 
+* Set 'ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES' to YES in Debug mode  
+  [icecrystal23](https://github.com/icecrystal23)
+
 ##### Bug Fixes
 
 * Don't generate unencrypted source warnings for localhost.  
--- a/lib/cocoapods/installer/xcode/pods_project_generator/target_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/target_installer.rb
@@ -68,6 +68,13 @@ module Pod
 
             native_target.build_configurations.each do |configuration|
               configuration.build_settings.merge!(custom_build_settings)
+
+              # This works around a unit test issue introduced in Xcode 10 where swift libraries
+              # required by embedded frameworks are not always added to unit test targets
+              require_xcode10_hack = configuration.debug? && target.product_type == :framework
+              if require_xcode10_hack
+                configuration.build_settings['ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES'] = 'YES'
+              end
             end
 
             native_target


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Set 'ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES' to YES in Debug mode  
  [icecrystal23](https://github.com/icecrystal23)
```

</p>

dnkoutso:
@amorde isn't this maybe fixed with updated xcodeproj gem settings?

icecrystal23:
@dnkoutso Does the xcodeproj gem generate some of the build settings for CocoaPods?

icecrystal23:
I tried updating the spec submodule to fix the tests, but I think that caused other failures

amorde:
hmm, the only thing that changed with my fix to Xcodeproj were Swift _optimization_ settings, I didn't see any changes related to this

amorde:
but yes @icecrystal23 Xcodeproj defines the default build settings for the different targets (ex. App, Framework, static library, watch extension, etc.)

icecrystal23:
Should I make this fix in there, then?

amorde:
haven't had a chance to take a deeper look at this - what does Xcode do if you were to mimic the setup using Xcode's UI? As in creating the targets manually within Xcode

icecrystal23:
The Xcode UI does not set this setting.  This is actually a workaround for a bug in Xcode 10 (I had a bug filed with Apple, and they told me to use this workaround... I'm not sure if I convinced them that it is their bug or not).  It only seems to affect unit test targets.  For some reason, Xcode 10 fails to recognize which Swift standard libraries are required for frameworks embedded in unit test targets.

icecrystal23:
We had been working around this with a post_install hook for our projects, but then I started seeming the same unit test failure when I tried to push one of our internal CocoaPod libraries to our internal spec repo.

amorde:
I see, thanks for the background info. Do you have an openradar link?

If this is a workaround for an Xcode bug I feel like it should be in CocoaPods instead of Xcodeproj but don't feel strongly either way

spec/unit/installer/xcode/pods_project_generator/target_installer_spec.rb
```diff
@@ -62,6 +62,16 @@ class PodsProjectGenerator
               'AppStore' => 'staticlib',
             }
           end
+
+          it 'embeds swift libraries in Debug mode' do
+            @target.stubs(:product_type).returns(:framework)
+            @installer.send(:add_target).resolved_build_setting('ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES').should == {
+              'Release' => nil,
+              'Debug' => 'YES',
+              'Test' => nil,
```
amorde: it seems like `Test` would need the fix as well, no? is the fix only applied to `Debug` configurations so that it only applies to unit tests?

icecrystal23: It was only applied to Debug mostly so it doesn't affect Release mode and app size.  I don't actually know what the Test configuration is typically used for; we don't use that one ourselves.

amorde: CocoaPods has the concept of build configurations which behave similar to the standard `Debug` & `Release` (can't seem to find any docs on it atm) so I think this would need to be applied to any Debug-like configuration

icecrystal23:
@amorde http://www.openradar.me/radar?id=4970867087900672
Apple claims it is related to mismatched Swift versions, but my sample project to them had all Swift versions set to 4.  They say it works if I set the Swift versions to all 4.2; I didn't validate that case since the Pod I was testing with doesn't work with 4.2 and I didn't have time to update it to do so.

amorde:
Thanks for the radar link, sounds like this is something specific about mismatched Swift versions and test bundles.

I think the change might be too broad here since it will apply to any target (not just targets which have associated test bundles). I'd like another dev to weigh in on this as I feel like I'm missing part of the picture here

amorde:
Also realized this seems related: https://github.com/CocoaPods/CocoaPods/issues/8139

spec/unit/installer/xcode/pods_project_generator/target_installer_spec.rb
icecrystal23: Is there some way to tell which configurations are supposed to behave like Debug?

segiddins: the `debug?` method on `Xcodeproj::Project::Object::XCBuildConfiguration`

mgrebenets:
We've got this problem and our Swift version is currently set to 4.0 for all targets.
Moving to 4.2 will take some minor effort due to all renamings in Swift, so we can't quickly verify how it works with 4.2.

spec/unit/installer/xcode/pods_project_generator/target_installer_spec.rb
icecrystal23: I updated the implementation to use the `debug?` check, but it still results in `Test` being `nil` for this test

amorde:
Haven't forgotten about this - if anyone needs a workaround while this is in review:

```ruby

post_install do |installer|
        config.build_settings['ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES'] = 'YES'
    end
  end
end
```

lukecharman:
Has it still being forgotten about? :)

amorde:
I'm hesitant to put specific workarounds like this in if it's something that Xcode will fix, especially given it's specific to certain Swift versions and its something we might have to keep around for a while or immediately remove on the next Xcode patch

amorde:
Looks like this is fixed in the latest Xcode beta, so I'm going to close this. Still, thanks for the PR!

icecrystal23:
We're seeing similar issues again in Xcode 11 betas (as noted in #8139 ).  The workaround is simple and only affects DEBUG builds.  Can you reconsider accepting this PR?

PR #8178, opened 2018-10-11
Always set the JSON checksum on first install

### LINKED ISSUE #8021: [1.6.0.beta.1] podspec checksum for `:podspec` is inconsistent across installs
Podfile:
```
pod 'A', :path=> 'xx/A/'
pod 'B', :podspec=> 'xx/B.podspec'
```
There are two pods in the Podfile. I found that CP get the checksum of podspec will have different logic:

If the pod use the `path`, CP get the checksum of `xx/A/A.podspec`.
If the pod use the `podspec`, CP get the checksum of `Pods/Local Podspecs/B.podspec.json`, instead of `xx/B.podspec`.

The logic for the `path` seams that be changed from the v1.5. I remember that its logic is same with `podspec` at past. I think the logic which get the checksum of origin podspec file is correct way. So, should we also change the logic of `podspec`?

segiddins:
Fixes #8021

### DIFF
--- a/lib/cocoapods/sandbox.rb
+++ b/lib/cocoapods/sandbox.rb
@@ -254,25 +254,29 @@ module Pod
       file_name = json ? "#{name}.podspec.json" : "#{name}.podspec"
       output_path = specifications_root + file_name
 
-      case podspec
-      when String
-        output_path.open('w') { |f| f.puts(podspec) }
-      when Pathname
-        unless podspec.exist?
-          raise Informative, "No podspec found for `#{name}` in #{podspec}"
+      spec =
+        case podspec
+        when String
+          output_path.open('w') { |f| f.puts(podspec) }
+          Specification.from_file(output_path)
+        when Pathname
+          unless podspec.exist?
+            raise Informative, "No podspec found for `#{name}` in #{podspec}"
+          end
+          FileUtils.copy(podspec, output_path)
+          Specification.from_file(podspec)
+        when Specification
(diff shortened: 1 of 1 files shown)


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Always set the JSON checksum on first install  
  [segiddins](https://github.com/segiddins)
```

</p>

lib/cocoapods/sandbox.rb
```diff
@@ -271,8 +271,8 @@ def store_podspec(name, podspec, _external_source = false, json = false)
         raise ArgumentError, "Unknown type for podspec: #{podspec.inspect}"
       end
 
-      spec ||= Specification.from_file(output_path)
-      spec.defined_in_file ||= output_path
+      spec = spec ? spec.dup : Specification.from_file(output_path)
```
amorde: should this go into the `when` cases that require it? I noticed that the `when Specification` case already calls `dup`, so this would dup it twice. is this only needed in `when Pathname`?

amorde: also probably worth a comment explaining why its necessary

dnkoutso: or remove the `.dup` above?

dnkoutso: agree on comment

Issue #7046, opened 2017-09-18
Size of ~/.cocoapods/repos exceeds 1gb


chrismanderson:
The size of my `~/.cocoapods/repos` folder is over 1gb (1.3gb to be specific). This seems much too large. Can I remove this folder? Or how can I shrink it down? Trying to maximize every iota of space on a MacBook with minimal storage.

dnkoutso:
@chrismanderson it's because of the specs master repo most likely (https://github.com/CocoaPods/Specs). I do not think anything can be done at this moment to reduce its size but I do understand it is becoming a problem.

chrismanderson:
Yeah, I know folder came about from http://blog.cocoapods.org/Master-Spec-Repo-Rate-Limiting-Post-Mortem/, but still, I hope there is something that can be done about it in the future. Just too big a folder to require people to keep.

orta:
We've taken a [few attacks](https://github.com/orta/repo_commit_squasher) that this, but realistically it's a hard problem that no-one has volunteered time to fix and maintain. If this is something you're interested at fixing, there's a few ideas:

* Create a smaller binary version of the specs repo that stays local
* Move trunk specs resolving to a server, similar to how bunder works
* Regularly squash the specs repo

amorde:
What if two separate repositories existed, one which `pod trunk` etc. interacted with and retained the full history, and one that is kept in sync but routinely squashed? 

It seems like the main issue is that the history of who/when adds specs is useful and worth keeping, but every user of CocoaPods does not need to be downloading the entire history and keeping a copy on their machine.

orta:
That's what my repo tried to do, but it turns out you don't really gain much from the process alas

amorde:
Just tried checking out the Specs repo and removing the `.git` folder, and it while there is a significant reduction it's most likely not worth the effort.

`192,884,971 bytes (773.4 MB on disk) for 404,973 items`

Attempting to even create a fresh repo takes quite a long time and increases the size dramatically.

davidair:
What if (and I know this is probably a crazy idea) the CocoaPods repo was split into multiple repositories based on some spec partition that would optimize for most common cases? Then the pod installation/update could conceivably function like that:

- Sync the smaller "current" repo first, attempt to resolve all dependencies. If it worked, done.
- Keep syncing additional "archival" repos until the dependencies are fully resolved.

[...]

PR #8197, opened 2018-10-17
[WIP] Do not setup the master spec repo automatically

amorde:
## Background

Currently, CocoaPods will automatically setup the master repo if it has not been setup already. Given that the repo has grown to be quite large (see #7046), this can be a problem. Going forward, we should only initialize the master repo if explicitly requested to do so.

## Changes

With this PR, CocoaPods will no longer automatically checkout the master repo, and instead will emit an error when there are no spec repos available.

The current implementation in this PR will emit this error on `pod spec lint` and `pod lib lint` if
there are no source repos available:

```
 There are no source repositories available. Either:

    a.) Setup the master repo by running `pod init`
    b.) Specify a list of private source repos with the `--sources` option
    c.) If the master repo exists but is incompatibile with this version of CocoaPods, run `pod repo update master`
```

When running `pod install`, the error will be as follows:

```
There are no sources specified in your Podfile. Either: 
              
    a.) Add the master repo to your Podfile, ex. `source "master"`
    b.) Add a private private repo at the source of your choosing, ex. `source "https://github.com/artsy/Specs"`

Note: The master repo is no longer included by default.
```

## Notes

I'm looking for a review on my basic approach, and since this is a (very) breaking change this won't go out with 1.6.0. Maybe it will land in 1.7.0, or 2.0.0, or if we decide this isn't something we want to do then we can just scrap it.

## Todo

- [ ] Integration specs
- [ ] Finalize error text
- [ ] Update cocoapods-core
- [ ] Update integration specs
- [ ] Update `pod init` template to include an explicit source
- [ ] Update example projects to include explicit sources

### DIFF
--- a/lib/cocoapods/command.rb
+++ b/lib/cocoapods/command.rb
@@ -92,13 +92,19 @@ module Pod
       end
     end
 
-    # Ensure that the master spec repo exists
+    # Ensure that at least one repo exists
     #
     # @return [void]
     #
-    def ensure_master_spec_repo_exists!
-      unless config.sources_manager.master_repo_functional?
-        Setup.new(CLAide::ARGV.new([])).run
+    def ensure_source_repo_available(source_urls)
+      if source_urls.empty? && config.sources_manager.all.empty?
+        raise Informative, <<-EOS
+          There are no source repositories available. Either:
+
+            a.) Setup the master repo by running `pod init`
+            b.) Specify a list of private source repos with the `--sources` option
+            c.) If the master repo exists but is incompatibile with this version of CocoaPods, run `pod repo update master`
+        EOS
       end
     end
 
--- a/lib/cocoapods/installer/analyzer.rb
+++ b/lib/cocoapods/installer/analyzer.rb
@@ -155,7 +155,15 @@ module Pod
           if all_dependencies_have_sources
             sources = dependency_sources
           elsif has_dependencies? && sources.empty? && plugin_sources.empty?
-            sources = ['https://github.com/CocoaPods/Specs.git']
+            raise Informative, <<-EOS.strip_heredoc
+            There are no sources specified in your Podfile. Either: 
+              
+              a.) Add the master repo to your Podfile, ex. `source "master"`
+              b.) Add a private private repo at the source of your choosing, ex. `source "https://github.com/artsy/Specs"`
+
+            Note: The master repo is no longer included by default.
+
+            EOS
           else
             sources += dependency_sources
           end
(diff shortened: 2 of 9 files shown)


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* [WIP] Do not setup the master spec repo automatically  
```

</p>

paulb777:
Clarification questions: This is a speed-up solution only for those that never need to use the master repo? - only for their initial installation?

lib/cocoapods/installer/analyzer.rb
```diff
@@ -155,7 +155,15 @@ def sources
...
-            sources = ['https://github.com/CocoaPods/Specs.git']
+            raise Informative, <<-EOS.strip_heredoc
+            There are no sources specified in your Podfile. Either: 
+              
+              a.) Add the master repo to your Podfile, ex. `source "master"`
+              b.) Add a private private repo at the source of your choosing, ex. `source "https://github.com/artsy/Specs"`
+
+            Note: The master repo is no longer included by default.
+
+            EOS
```
orta: We shouldn't do this if it breaks every existing Podfile relying on the default sources

segiddins:
This needs to wait until 2.0, it's a very big breaking change.

lib/cocoapods/command.rb
```diff
@@ -92,13 +92,19 @@ def initialize(argv)
       end
     end
 
-    # Ensure that the master spec repo exists
+    # Ensure that at least one repo exists
     #
     # @return [void]
     #
-    def ensure_master_spec_repo_exists!
```
segiddins: This only gets called on `try` and `search`, it seems to me

amorde:
@paulb777 yes my thinking here was to not start a 1GB+ download unless that is what was intended.

Totally understand how breaking this change is - are there parts of this that could be released gradually? For example, maybe `pod install` stays the same, but `pod spec lint` and `pod lib lint` don't automatically set up the master repo unless requested to do so with `--sources`

amorde:
Sounding like this is a bit premature - I'll close this for now and rethink how / when / where this should change, if at all

dnkoutso:
For posterity, I love the spirit of this and I cant wait for it to be true but we do need a detailed plan on how we will do it.

amorde:
👍  wanted to get the chance to hear everyones thought on it. maybe I'll write up an RFC so we can have a public discussion

orta:
I think this is feasible without making it a breaking change: the rules as they are today:

- 0 sources, only get trunk
- 1+ source, you have to define trunk ([example](https://github.com/artsy/eigen/blob/master/Podfile#L8-L9))

So it should be totally fine to check if master isn't in the list of explicit sources, and that there *is* a list of explicit sources

PR #8221, opened 2018-10-24
Small refactoring of pod project generator.

sebastianv1:
Functions in this class now rely less on the assumption that a project object will be created by the time the function is ran and requires a `project` parameter to be passed into itself. Since the `generate!` function is responsible for actually creating the Pods project file, we return it wrapped inside a `PodsProjectGeneratorResult` object alongside the `target_installation_results` that is also returned from `generate!`.

### DIFF
--- a/lib/cocoapods/installer.rb
+++ b/lib/cocoapods/installer.rb
@@ -206,15 +206,15 @@ module Pod
     #
     def generate_pods_project(generator = create_generator)
       UI.section 'Generating Pods project' do
-        @target_installation_results = generator.generate!
-        @pods_project = generator.project
+        pod_project_generation_result = generator.generate!
+        @target_installation_results = pod_project_generation_result.target_installation_results
+        @pods_project = pod_project_generation_result.project
         run_podfile_post_install_hooks
-        generator.write
-        generator.share_development_pod_schemes
+        generator.write(@pods_project, @target_installation_results)
+        generator.share_development_pod_schemes(@pods_project)
         write_lockfiles
       end
     end
-
     #-------------------------------------------------------------------------#
 
     public
--- a/lib/cocoapods/installer/xcode/pods_project_generator.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator.rb
@@ -11,14 +11,11 @@ module Pod
         require 'cocoapods/installer/xcode/pods_project_generator/pod_target_installer'
         require 'cocoapods/installer/xcode/pods_project_generator/file_references_installer'
         require 'cocoapods/installer/xcode/pods_project_generator/aggregate_target_installer'
+        require 'cocoapods/installer/xcode/pods_project_generator_result'
 
         # @return [Sandbox] The sandbox where the Pods should be installed.
         #
         attr_reader :sandbox
-
-        # @return [Pod::Project] the `Pods/Pods.xcodeproj` project.
-        #
-        attr_reader :project
 
         # @return [Array<AggregateTarget>] The model representations of an
         #         aggregation of pod targets generated for a target definition
@@ -62,15 +59,15 @@ module Pod
         end
 
         def generate!
-          prepare
-          install_file_references
-          @target_installation_results = install_targets
-          integrate_targets(@target_installation_results.pod_target_installation_results)
-          wire_target_dependencies(@target_installation_results)
-          @target_installation_results
+          project = prepare
+          install_file_references(project)
+          target_installation_results = install_targets(project)
+          integrate_targets(target_installation_results.pod_target_installation_results)
+          wire_target_dependencies(project, target_installation_results)
+          PodsProjectGeneratorResult.new(project, target_installation_results)
         end
 
-        def write
+        def write(project, target_installation_results)
           UI.message "- Writing Xcode project file to #{UI.path sandbox.project_path}" do
             project.pods.remove_from_project if project.pods.empty?
             project.development_pods.remove_from_project if project.development_pods.empty?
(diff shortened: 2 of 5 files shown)


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Small refactoring of pod project generator.  
  [sebastianv1](https://github.com/sebastianv1)
```

</p>

lib/cocoapods/installer/xcode/pods_project_generator_result.rb
```diff
@@ -0,0 +1,18 @@
+module Pod
+  class Installer
+    class Xcode
+      class PodsProjectGenerator
+        class PodsProjectGeneratorResult
+
+          attr_reader :project
```
dnkoutso: nit docs are usually added on most classes.

lib/cocoapods/installer/xcode/pods_project_generator.rb
```diff
@@ -62,15 +59,15 @@ def initialize(sandbox, aggregate_targets, pod_targets, analysis_result, install
...
         def generate!
-          prepare
-          install_file_references
-          @target_installation_results = install_targets
-          integrate_targets(@target_installation_results.pod_target_installation_results)
-          wire_target_dependencies(@target_installation_results)
-          @target_installation_results
+          project = prepare
+          install_file_references(project)
+          @target_installation_results = install_targets(project)
```
dnkoutso: does this need to be an instance variable anymore?

lib/cocoapods/installer.rb
```diff
@@ -206,11 +206,12 @@ def create_generator
...
-        @target_installation_results = generator.generate!
-        @pods_project = generator.project
+        pod_project_generation_result = generator.generate!
+        @target_installation_results = pod_project_generation_result.target_installation_results
+        @pods_project = pod_project_generation_result.project
         run_podfile_post_install_hooks
-        generator.write
-        generator.share_development_pod_schemes
+        generator.write(pods_project)
+        generator.share_development_pod_schemes(project)
```
dnkoutso: not sure if you want in this PR but I think those should become class methods. They contain no state

dnkoutso: actually they do. bleh

lib/cocoapods/installer/xcode/pods_project_generator_result.rb
sebastianv1: Does this look okay? 😬

dnkoutso: you can add "the project that was generated as part of the generator"

also add the ones in the `def initialize` method. See `pod_target.rb` for example

spec/unit/installer_spec.rb
```diff
@@ -92,7 +92,8 @@ def @installer.clean_pod_sources
         @installer.unstub(:generate_pods_project)
         generator = @installer.send(:create_generator)
         @installer.stubs(:create_generator).returns(generator)
-        generator.stubs(:generate!)
+        generator_result = Installer::Xcode::PodsProjectGenerator::PodsProjectGeneratorResult.new(nil, [])
+        generator.stubs(:generate!).returns(generator_result)
```
dnkoutso: nice

lib/cocoapods/installer.rb
```diff
@@ -206,15 +206,15 @@ def create_generator
     #
     def generate_pods_project(generator = create_generator)
       UI.section 'Generating Pods project' do
-        @target_installation_results = generator.generate!
-        @pods_project = generator.project
+        pod_project_generation_result = generator.generate!
+        @target_installation_results = pod_project_generation_result.target_installation_results
+        @pods_project = pod_project_generation_result.project
```
segiddins: could this not be an ivar?

sebastianv1: `pod_project_generation_result`? Don't see why it should be if it's not used outside this scope and only passes them off to the existing ivars.

dnkoutso: I think he means `@pods_project` but for the scope of this PR maybe not change `installer.rb` at all, since it also exposes a public API in `post_install`.

lib/cocoapods/installer/xcode/pods_project_generator.rb
```diff
@@ -62,15 +59,15 @@ def initialize(sandbox, aggregate_targets, pod_targets, analysis_result, install
...
+          project = prepare
+          install_file_references(project)
+          target_installation_results = install_targets(project)
+          integrate_targets(target_installation_results.pod_target_installation_results)
+          wire_target_dependencies(project, target_installation_results)
+          PodsProjectGeneratorResult.new(project, target_installation_results)
         end
 
-        def write
+        def write(project, target_installation_results)
```
dnkoutso: should these now become class methods?

lib/cocoapods/installer/xcode/pods_project_generator.rb
```diff
@@ -127,54 +124,55 @@ def create_project
 
         # Creates the Pods project from scratch if it doesn't exists.
```
dnkoutso: typo nit (not yours): 'if it doesn't exist'

lib/cocoapods/installer/xcode/pods_project_generator.rb
sebastianv1: Looks like generator is still using the `installation_options` ivar in this function. We could pass it in from the installer and I think that would allow us to make it a class method. What do you think?

amorde: I'm actually working on a refactor that will remove the need for `installation_options` in this location - if this PR lands then I can rework this a bit in my refactor

dnkoutso: @amorde done!

PR #8158, opened 2018-10-04
App spec integration into cocoapods

dostrander:
@dnkoutso @segiddins

PR #8244, opened 2018-11-02
Don't have libraries add app targets to their scheme for building.

dostrander:
Introduced with the introduction of app specs:
https://github.com/CocoaPods/CocoaPods/pull/8158


dostrander:
@dnkoutso @segiddins

PR #8287, opened 2018-11-17
Generate Info.plist files for static frameworks

paulb777:
The new Xcode 10 build system spits out a warning about missing Info.plist files for static frameworks like:

`    - NOTE  | [iOS] xcodebuild:  warning: Skipping code signing because the target does not have an Info.plist file. (in target 'FirebaseCore')`

Even though Info.plist files should not be needed since static frameworks are fully linked into apps before bundling, they should be harmless to include.

Reported at https://github.com/firebase/firebase-ios-sdk/issues/1846

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
+* Generate Info.plist files for static frameworks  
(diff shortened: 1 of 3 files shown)


segiddins:
Thanks!

oleghnidets:
I have such issue on CocoaPods 1.6.1. It doesn't generates and links Info.plist

PR #8339, opened 2018-12-07
Creates a stage sandbox and sandbox dir cleaner step.

sebastianv1:
Inside stage sandbox we now add all the respective header search paths to the sandbox `HeaderStore` and each pod target's `HeaderStore`. This sets up the work for incremental pod installation because we would like to analyze each pod target's build settings to determine if it has changed; however, we previously only added the header search paths during installation which is too late in the pipline.

Instead of removing the entire Headers and Target Support Files directory in the Sandbox for every installation, we now only remove the headers and target support files that are stale.

integration specs: https://github.com/CocoaPods/cocoapods-integration-specs/pull/197

### DIFF
--- a/lib/cocoapods/installer.rb
+++ b/lib/cocoapods/installer.rb
@@ -30,16 +30,18 @@ module Pod
   # source control.
   #
   class Installer
-    autoload :Analyzer,                   'cocoapods/installer/analyzer'
-    autoload :InstallationOptions,        'cocoapods/installer/installation_options'
-    autoload :PostInstallHooksContext,    'cocoapods/installer/post_install_hooks_context'
-    autoload :PreInstallHooksContext,     'cocoapods/installer/pre_install_hooks_context'
-    autoload :SourceProviderHooksContext, 'cocoapods/installer/source_provider_hooks_context'
-    autoload :PodfileValidator,           'cocoapods/installer/podfile_validator'
-    autoload :PodSourceInstaller,         'cocoapods/installer/pod_source_installer'
-    autoload :PodSourcePreparer,          'cocoapods/installer/pod_source_preparer'
-    autoload :UserProjectIntegrator,      'cocoapods/installer/user_project_integrator'
-    autoload :Xcode,                      'cocoapods/installer/xcode'
+    autoload :Analyzer,                     'cocoapods/installer/analyzer'
+    autoload :InstallationOptions,          'cocoapods/installer/installation_options'
+    autoload :PostInstallHooksContext,      'cocoapods/installer/post_install_hooks_context'
+    autoload :PreInstallHooksContext,       'cocoapods/installer/pre_install_hooks_context'
+    autoload :SourceProviderHooksContext,   'cocoapods/installer/source_provider_hooks_context'
+    autoload :PodfileValidator,             'cocoapods/installer/podfile_validator'
+    autoload :PodSourceInstaller,           'cocoapods/installer/pod_source_installer'
+    autoload :PodSourcePreparer,            'cocoapods/installer/pod_source_preparer'
+    autoload :UserProjectIntegrator,        'cocoapods/installer/user_project_integrator'
+    autoload :Xcode,                        'cocoapods/installer/xcode'
+    autoload :SandboxHeaderPathsInstaller,  'cocoapods/installer/sandbox_header_paths_installer'
+    autoload :SandboxDirCleaner,            'cocoapods/installer/sandbox_dir_cleaner'
 
     include Config::Mixin
 
@@ -171,7 +173,6 @@ module Pod
       UI.section 'Analyzing dependencies' do
         analyze(analyzer)
         validate_build_configurations
-        clean_sandbox
       end
 
       UI.section 'Verifying no changes' do
@@ -190,6 +191,20 @@ module Pod
       end
     end
 
+    # Stages the sandbox after analysis.
+    #
+    # @param [Sandbox] sandbox
+    #        The sandbox to stage.
+    #
+    # @param [Array<PodTarget>] pod_targets
+    #        The list of all pod targets.
+    #
+    # @return [void]
+    #
+    def stage_sandbox(sandbox, pod_targets)
+      SandboxHeaderPathsInstaller.new(sandbox, pod_targets).install!
+    end
+
     #-------------------------------------------------------------------------#
 
     # @!group Pods Project Generation
@@ -206,7 +221,15 @@ module Pod
 
     # Generates the Xcode project(s) that go inside the `Pods/` directory.
     #
-    def generate_pods_project(generator = create_generator(installation_options.generate_multiple_pod_projects))
+    def generate_pods_project
+      stage_sandbox(sandbox, pod_targets)
+      clean_sandbox(pod_targets)
+      create_and_save_projects
+      write_lockfiles
+      SandboxDirCleaner.new(sandbox, pod_targets, aggregate_targets).clean!
+    end
+
+    def create_and_save_projects(generator = create_generator(installation_options.generate_multiple_pod_projects))
       UI.section 'Generating Pods project' do
         pod_project_generation_result = generator.generate!
         @target_installation_results = pod_project_generation_result.target_installation_results
@@ -227,8 +250,6 @@ module Pod
         all_projects_by_pod_targets.each do |project, pod_targets|
           generator.share_development_pod_schemes(project, development_pod_targets(pod_targets))
         end
-
-        write_lockfiles
       end
     end
 
@@ -314,25 +335,14 @@ module Pod
       end
     end
 
-    # @return [void] In this step we clean all the folders that will be
-    #         regenerated from scratch and any file which might not be
-    #         overwritten.
+    # @return [void] In this step we clean all the header folders for pod targets that will be
+    #         regenerated from scratch and cleanup any pods that have been removed.
     #
-    # @todo   [#247] Clean the headers of only the pods to install.
-    #
-    def clean_sandbox
-      sandbox.public_headers.implode!
-      target_support_dirs = sandbox.target_support_files_root.children.select(&:directory?)
+    def clean_sandbox(pod_targets)
       pod_targets.each do |pod_target|
-        pod_target.build_headers.implode!
-        target_support_dirs.delete(pod_target.support_files_dir)
+        pod_target.build_headers.implode_path!(pod_target.headers_sandbox)
+        sandbox.public_headers.implode_path!(pod_target.headers_sandbox)
       end
-
-      aggregate_targets.each do |aggregate_target|
-        target_support_dirs.delete(aggregate_target.support_files_dir)
-      end
-
-      target_support_dirs.each { |dir| FileUtils.rm_rf(dir) }
 
       unless sandbox_state.deleted.empty?
         title_options = { :verbose_prefix => '-> '.red }
--- a/lib/cocoapods/installer/sandbox_dir_cleaner.rb
+++ b/lib/cocoapods/installer/sandbox_dir_cleaner.rb
@@ -1,0 +1,89 @@
+module Pod
+  class Installer
+    # Cleans up the sandbox directory by removing stale target support files and headers.
+    #
+    class SandboxDirCleaner
+      # @return [Sandbox] The sandbox directory that will be cleaned.
+      #
+      attr_reader :sandbox
+
+      # @return [Array<PodTarget>]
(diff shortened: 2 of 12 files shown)


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Creates a stage sandbox and sandbox dir cleaner step.  
  [sebastianv1](https://github.com/sebastianv1)
```

</p>

lib/cocoapods/installer/sandbox_dir_cleaner.rb
```diff
@@ -0,0 +1,98 @@
...
+      attr_reader :sandbox
+
+      # @return [Array<PodTarget>]
+      #         The list of all pod targets used to analyze directories installed into the Sandbox.
+      #
+      attr_reader :pod_targets
+      # @return [Array<AggregateTarget>]
+      #         The list of all aggregate targets used to analyze directories installed into the Sandbox.
+      #
+
```
dnkoutso: nit formatting

lib/cocoapods/installer/sandbox_header_paths_installer.rb
```diff
@@ -0,0 +1,40 @@
+module Pod
+  class Installer
+    # Adds all the search paths into the sandbox HeaderStore and each pod target's HeaderStore.
+    #
+    class SandboxHeaderPathsInstaller
+      # @return [Sandbox] The sandbox to use for this analysis.
```
dnkoutso: are these copied docs?

lib/cocoapods/target/pod_target.rb
```diff
@@ -131,6 +131,34 @@ def label
       end
     end
 
+    # @return [Pathname] the pathname for headers in the sandbox.
+    #
+    def headers_sandbox
+      Pathname.new(pod_name)
+    end
+
+    # @return [Hash{FileAccessor => Hash] Hash of file accessors by header mappings.
```
dnkoutso: missing } in the doc type

lib/cocoapods/target/pod_target.rb
```diff
@@ -131,6 +131,34 @@ def label
...
+    def header_mappings_by_file_accessor
+      valid_accessors = file_accessors.reject { |fa| fa.spec.non_library_specification? }
+      Hash[valid_accessors.map do |file_accessor|
+        # Private headers will always end up in Pods/Headers/Private/PodA/*.h
+        # This will allow for `""` imports to work.
+        [file_accessor, header_mappings(headers_sandbox, file_accessor, file_accessor.headers)]
+      end]
+    end
+
+    # @return [Hash{FileAccessor => Hash] Hash of file accessors by public header mappings.
```
dnkoutso: same, although I think those were copied/pasted

lib/cocoapods/installer.rb
```diff
@@ -190,6 +192,13 @@ def download_dependencies
       end
     end
 
+    # Stages the sandbox after analysis.
+    #
+    def stage_sandbox(sandbox, pod_targets, aggregate_targets)
+      SandboxHeaderPathsInstaller.new(sandbox, pod_targets).install!
+      clean_sandbox(pod_targets, aggregate_targets)
```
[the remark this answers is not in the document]
sebastianv1: And `SandboxHeaderPathsInstaller` needs to go before analysis such that all the header search paths are set up correctly for the build settings

lib/cocoapods/installer.rb
```diff
@@ -229,6 +238,7 @@ def generate_pods_project(generator = create_generator(installation_options.gene
         end
 
         write_lockfiles
+        SandboxDirCleaner.new(sandbox, pod_targets, aggregate_targets).clean!
```
dnkoutso: this does not seem to belong in the method `generate_pods_project`, why not within `clean_sandbox`?

lib/cocoapods/installer/sandbox_dir_cleaner.rb
```diff
@@ -0,0 +1,98 @@
...
+  class Installer
+    # Cleans up the sandbox directory by removing stale target support files and headers.
+    #
+    class SandboxDirCleaner
+      # @return [Sandbox] The sandbox directory that will be cleaned.
+      #
+      attr_reader :sandbox
+
+      # @return [Array<PodTarget>]
+      #         The list of all pod targets used to analyze directories installed into the Sandbox.
```
dnkoutso: probably needs updating as it was copied from `analyzer.rb`?

lib/cocoapods/target/pod_target.rb
```diff
@@ -688,5 +716,43 @@ def uses_modular_headers?(only_if_defines_modules = true)
...
+    #         linked.
+    #
+    # @param  [Array<Pathname>] headers
+    #         The absolute paths of the headers which need to be mapped.
+    #
+    # @return [Hash{Pathname => Array<Pathname>}] A hash containing the
+    #         headers folders as the keys and the absolute paths of the
+    #         header files as the values.
+    #
+    def header_mappings(headers_sandbox, file_accessor, headers)
```
dnkoutso: if this method lives in `pod_target` I dont think you need `headers_sandbox` as a parameter?

lib/cocoapods/installer.rb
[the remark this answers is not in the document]
sebastianv1: Ended up redoing the `generate_pods_project` function. Let me know what y'all think

lib/cocoapods/installer/sandbox_dir_cleaner.rb
sebastianv1: Caught in the act :)

lib/cocoapods/installer.rb
```diff
@@ -190,6 +191,22 @@ def download_dependencies
...
 
+    # Stages the sandbox after analysis.
+    #
+    # @param [Sandbox] sandbox
+    #        The sandbox to stage.
+    #
+    # @param [Array<PodTarget>] pod_targets
+    #        The list of all pod targets.
+    #
+    # @param [Array<AggregateTarget>] aggregate_target
```
dnkoutso: this has a param of `aggregate_target` that is not being passed into the method. Also missing line after `return` line

lib/cocoapods/target/pod_target.rb
sebastianv1: Good point

lib/cocoapods/installer.rb
```diff
@@ -206,7 +221,15 @@ def create_generator(generate_multiple_pod_projects = false)
 
     # Generates the Xcode project(s) that go inside the `Pods/` directory.
     #
-    def generate_pods_project(generator = create_generator(installation_options.generate_multiple_pod_projects))
+    def generate_pods_project
+      stage_sandbox(sandbox, pod_targets)
+      clean_sandbox(pod_targets, aggregate_targets)
+      create_and_save_projects
+      write_lockfiles
+      SandboxDirCleaner.new(sandbox, pod_targets, aggregate_targets).clean!
```
amorde: nit: it's a bit odd that _some_ of the cleaning happens in `clean_sandbox` and some of it happens through `SandboxDirCleaner` - is there any way to consolidate that? I'm not familiar with how the two code paths differ

lib/cocoapods/target/pod_target.rb
```diff
@@ -688,5 +716,43 @@ def uses_modular_headers?(only_if_defines_modules = true)
...
+        sub_dir = dir
+        if header_mappings_dir
+          relative_path = header.relative_path_from(file_accessor.path_list.root + header_mappings_dir)
+          sub_dir += relative_path.dirname
+        end
+        mappings[sub_dir] ||= []
+        mappings[sub_dir] << header
+      end
+      mappings
+    end
```
amorde: not that this is an issue, but note that this will now be part of the public API of `PodTarget` - does it need to be? it feels like something the sandbox should handle

lib/cocoapods/installer.rb
sebastianv1: Yeah, so `clean_sandbox` cleans up the directories for pod targets that will be generated and `SandboxDirCleaner` handles cleaning up directories for pod targets that were removed.

lib/cocoapods/target/pod_target.rb
sebastianv1: That's a good point. I'm open to hearing what others think about moving it into the sandbox. I don't really have a strong preference about where it lives, I just moved it here since everywhere we needed the header mappings we were already iterating over the pod targets, so it made sense in my head at the time to just put it in here.

lib/cocoapods/installer.rb
sebastianv1: `clean_sandbox` and `SandboxDirCleaner` will diverge with incremental installation since `clean_sandbox` will take in the list of generated pod targets while `SandboxDirCleaner` will always take in all pod and aggregate targets in order to compute what needs to be removed.

amorde: ah okay that makes sense. a `SandboxDirPreparer` or something would be cool but not necessary for this PR 👍

lib/cocoapods/target/pod_target.rb
```diff
@@ -688,5 +716,43 @@ def uses_modular_headers?(only_if_defines_modules = true)
     def create_build_settings
       BuildSettings::PodTargetSettings.new(self)
     end
+
+    # Computes the destination sub-directory in the sandbox
+    #
+    # @param  [Pathname] headers_sandbox
```
dnkoutso: nit: remove this now that the param is removed.

lib/cocoapods/target/pod_target.rb
dnkoutso: This might explode a bit but passing `file_accessor` does not make sense as the `pod_target` is already a holder and owner of `file_accessors`.

sebastianv1: So `header_mappings` won't actually be a part of the public API of `PodTarget` since it's under `private`. It takes in a `file_accessor` since it's more of a helper used by the public functions `public_header_mappings_by_file_accessor` and `header_mappings_by_file_accessor`

lib/cocoapods/installer/sandbox_dir_cleaner.rb
```diff
@@ -0,0 +1,98 @@
...
+  class Installer
+    # Cleans up the sandbox directory by removing stale target support files and headers.
+    #
+    class SandboxDirCleaner
+      # @return [Sandbox] The sandbox directory that will be cleaned.
+      #
+      attr_reader :sandbox
+
+      # @return [Array<PodTarget>]
+      #         The list of all pod targets that will be installed into the Sandbox.
```
dnkoutso: is this correct or copy pasta?

sebastianv1: Correct in that I wrote it :) The list of pod targets that will be installed are used to compute which folders need to be deleted.

lib/cocoapods/installer.rb
```diff
@@ -318,22 +339,17 @@ def validate_build_configurations
...
-        pod_target.build_headers.implode!
-        target_support_dirs.delete(pod_target.support_files_dir)
+        pod_target.build_headers.implode_path!(pod_target.headers_sandbox)
+        sandbox.public_headers.implode_path!(pod_target.headers_sandbox)
+        FileUtils.rm_rf(pod_target.support_files_dir)
       end
 
       aggregate_targets.each do |aggregate_target|
-        target_support_dirs.delete(aggregate_target.support_files_dir)
+        FileUtils.rm_rf(aggregate_target.support_files_dir)
```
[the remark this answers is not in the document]
sebastianv1: Can remove. Any reason why we don't want that if the files are regenerated anyways?

lib/cocoapods/installer.rb
```diff
@@ -318,22 +339,17 @@ def validate_build_configurations
     #         regenerated from scratch and any file which might not be
     #         overwritten.
     #
-    # @todo   [#247] Clean the headers of only the pods to install.
-    #
-    def clean_sandbox
```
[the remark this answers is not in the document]
sebastianv1: Sets up incremental since `clean_sandbox` will only take in a subset of all pod targets to clean.

lib/cocoapods/installer.rb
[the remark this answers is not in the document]
segiddins: Because it means we _have_ to rewrite the files. Right now, we only actually write to disk if there's a change

PR #8352, opened 2018-12-13
[Analyzer] Remove duplicate tests

amorde:


spec/unit/installer/analyzer_spec.rb
```diff
@@ -353,70 +353,6 @@ module Pod
           aranalytics_target.dependent_targets.all?(&:scoped).should.be.true
         end
 
-        it 'does not mark transitive dependencies as dependent targets' do
```
amorde: this same exact test has 3 copies

dnkoutso: lol

amorde: :)

PR #8370, opened 2018-12-20
Don't update system gems on CircleCI

amorde:


segiddins:
👍🏻 in future if we need a newer version, we should pin explicitly

Issue #8399, opened 2019-01-09
Provide more information about use_frameworks! in the auto-generated Podfile


michaeleisel:
Currently, the Podfile generated (see `lib/cocoapods/command/init.rb` in this repo) has something along the lines of:
```
# Comment the next line if you're not using Swift and don't want to use dynamic frameworks
use_frameworks!
```

However, it is now possible to use static libraries containing Swift code. The comment should be updated accordingly. Also, a lot of developers are unfamiliar with the tradeoffs of static vs. dynamic libraries. In particular, the startup time can be increased by hundreds of milliseconds from using dynamic libraries (and in a way that is difficult to measure). On the other hand, dynamic libraries are better for incremental build time. So perhaps the comment should link to an article about the tradeoffs, or at least mention that you are either sacrificing startup time or incremental build performance (the two biggest tradeoffs that I can think of) when you choose whether or not to include `use_frameworks!`.

dnkoutso:
Sounds like a good PR to come and fairly easy to do so. Want to make one?

michaeleisel:
Done, see https://github.com/CocoaPods/CocoaPods/pull/8401

PR #8401, opened 2019-01-09
Provide more information about use_frameworks! in the auto-generated Podfile

michaeleisel:
Closes https://github.com/CocoaPods/CocoaPods/issues/8399 . Let me know if this message looks good, and also if the code formatting is good. I will then test it and update the tests that would be affected by it.

### DIFF
--- a/lib/cocoapods/command/init.rb
+++ b/lib/cocoapods/command/init.rb
@@ -84,10 +84,8 @@ module Pod
       def target_module(app, tests)
         target_module = "\ntarget '#{app.name.gsub(/'/, "\\\\\'")}' do\n"
 
-        app_uses_swift = app.source_build_phase.files_references.any? { |fr| File.extname(fr.path) == '.swift' }
-
         target_module << <<-RUBY
-  # Comment the following line to use static libraries.
+  # Comment the next line if you don't want to use dynamic frameworks
   use_frameworks!
 
          RUBY
--- a/lib/cocoapods/command/init.rb
+++ b/lib/cocoapods/command/init.rb
@@ -87,7 +87,7 @@ module Pod
         app_uses_swift = app.source_build_phase.files_references.any? { |fr| File.extname(fr.path) == '.swift' }
 
         target_module << <<-RUBY
-  # Comment the following line to use static libraries. With static libraries, the startup time is better, and with dynamic libraries, the incremental compilation time is better.
+  # Comment the following line to use static libraries.
   use_frameworks!
 
          RUBY
--- a/lib/cocoapods/command/init.rb
+++ b/lib/cocoapods/command/init.rb
@@ -86,19 +86,11 @@ module Pod
 
         app_uses_swift = app.source_build_phase.files_references.any? { |fr| File.extname(fr.path) == '.swift' }
 
-        target_module << if app_uses_swift
-                           <<-RUBY
-  # Comment the next line if you're not using Swift and don't want to use dynamic frameworks
+        target_module << <<-RUBY
+  # Comment the following line to use static libraries. With static libraries, the startup time is better, and with dynamic libraries, the incremental compilation time is better.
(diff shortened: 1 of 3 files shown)


lib/cocoapods/command/init.rb
```diff
@@ -86,19 +86,11 @@ def target_module(app, tests)
 
         app_uses_swift = app.source_build_phase.files_references.any? { |fr| File.extname(fr.path) == '.swift' }
 
-        target_module << if app_uses_swift
-                           <<-RUBY
-  # Comment the next line if you're not using Swift and don't want to use dynamic frameworks
+        target_module << <<-RUBY
+  # Comment the following line to use static libraries. With static libraries, the startup time is better, and with dynamic libraries, the incremental compilation time is better.
```
dnkoutso: I am not sure the last part is true, why is incremental compilation better? I think "# Comment the following line to use static libraries." is good enough here.

CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Provide more information about use_frameworks! in the auto-generated Podfile  
  [michaeleisel](https://github.com/michaeleisel)
```

</p>

lib/cocoapods/command/init.rb
michaeleisel: Incremental compilation is better because, when it's linking the main executable, it won't have to link in the cocoapods, those cocoapods are already linked into their own dynamic frameworks. I am fine with keeping or removing it. I could leave out the bit about startup time as well, although in practice I see many people that have `use_frameworks` in their Podfile and don't understand that it increases startup time

michaeleisel:
@paulb777 good call, that does trip people up

lib/cocoapods/command/init.rb
paulb777: +1 on keeping it short.  

I agree with the incremental compilation point in theory. However in practice I would suspect it is negligible.  

I think the nuances are too complicated to try to capture here and providing summary info that let's people know what to search for is sufficient.

michaeleisel: The incremental compilation point is measurable and it can have a big impact on big projects, e.g. I've worked on a project with a 25-second link time due to how many files were being included, and which would've no doubt become much smaller using dylibs. But I do understand that most projects don't have to worry about that. I'm perfectly fine removing it.

As for removing the bit about startup time, I don't think that most iOS devs search things like "static vs. dynamic libraries", and even if they did, most of the articles don't really discuss startup time. If we really don't want to mention it, perhaps we could link to some doc on an overview of static and dynamic libraries, or comment out `use_frameworks!` (and add `use_modular_headers`) by default. Otherwise, we are essentially accepting that most iOS devs will continue to be ignorant of this.

dnkoutso: But this sounds like that its only affected by big projects which I would expect people will be able to perform such an analysis. I dont think we want to provide this messaging for the majority of the projects which are small and this improvement is negligible.

michaeleisel: Even small projects can pull in lots of code via Cocoapods. But, I'd be fine with a short comment that doesn't mention any of the tradeoffs. Ideally in that case, I'd like to comment out use_frameworks, but I could live with leaving it as it is.

dnkoutso: SG, can you please update the PR to reflect the shorter comment? Then we can merge!

michaeleisel:
OK this is what i have. if this looks good, how would i test it?

dnkoutso:
CI will test it and there is no need for tests to cover this change. I believe the integration specs might fail but we can wait and see.

PR #8334, opened 2018-12-05
Fix OTHER_LDFLAGS regression

paulb777:
Fix #8333 - a regression introduced in #8314 

Let me know if this is reasonable and I'll update the integration tests

PR #8314, opened 2018-11-29
Remove unnecessary linker flags from pod targets

segiddins:
closes #8204

PR #8428, opened 2019-01-17
[CherryPick] Remove unnecessary linker flags from pod targets

dnkoutso:
Merge pull request #8314 from CocoaPods/segiddins/remove-linker-flas-from-pod-targets

This cherry-picks https://github.com/CocoaPods/CocoaPods/pull/8314 into `1-6-stable` branch to release 1.6.0 for CocoaPods.

Will need one more after which is https://github.com/CocoaPods/CocoaPods/pull/8334

### DIFF
--- a/lib/cocoapods/target/build_settings.rb
+++ b/lib/cocoapods/target/build_settings.rb
@@ -512,11 +512,15 @@ module Pod
         define_build_settings_method :frameworks, :memoized => true, :sorted => true, :uniqued => true do
           return [] if (!target.requires_frameworks? || target.static_framework?) && !test_xcconfig?
 
-          frameworks = vendored_dynamic_frameworks.map { |l| File.basename(l, '.framework') }
-          frameworks.concat vendored_static_frameworks.map { |l| File.basename(l, '.framework') } unless test_xcconfig?
-          frameworks.concat consumer_frameworks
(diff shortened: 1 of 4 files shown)


lib/cocoapods/target/build_settings.rb
```diff
@@ -538,8 +542,7 @@ def initialize(target, test_spec = nil)
 
         # @return [Array<String>]
         define_build_settings_method :weak_frameworks, :memoized => true do
-          return [] if (!target.requires_frameworks? || target.static_framework?) && !test_xcconfig?
-
+          return [] unless test_xcconfig?
```
dnkoutso: that doesnt look right

dnkoutso: well wait it might be

dnkoutso: Its correct https://github.com/CocoaPods/CocoaPods/pull/8314/files#diff-54f9c07edbb5b7fc64e04683656d1778R565

amorde: previous logic was a bit odd since `static_framework?` anyway

dnkoutso: @paulb777 does this mean this is OK?

paulb777: Yes, it's ok.  I did a self-correcting loop. :)

PR #8275, opened 2018-11-14
Allow app specs to be used to upload to the app store

segiddins:
- [PodTargetInstall] Add app spec resources directly to the app target
    Rather than using the copy resources script for them
- [PodTargetInstaller] Remove target build settings that are in the pod target xcconfig
- [BuildSettings] Allow {pod,user}_target_xcconfig to override singular settings
- [BuildSettings] Add missing plural settings
- [PodTargetInstaller] Use the correct methods for accessing app spec paths
- [PodTarget] Store non-library spec build settings

PR #8432, opened 2019-01-18
Do not generate default `Info.plist` file for test targets if one is already specified.

dnkoutso:
Given:
```
test_spec.pod_target_xcconfig = { 'INFOPLIST_FILE' => '"$(PODS_TARGET_SRCROOT)/MyInfo.plist"' }
```

CocoaPods will stomp over this setting. This takes care of this case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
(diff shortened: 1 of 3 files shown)


segiddins:
Please add a test for this fix

efirestone:
Was this fixed in another way? Can I take this over to help?

segiddins:
https://github.com/CocoaPods/CocoaPods/issues/8275 should allow pod target xcconfigs to override any build setting

PR #8445, opened 2019-01-24
Set the path of development pod groups to the top-most shared directory

amorde:
This fixes an issue that prevents new groups from being placed in the correct directory if all source files are in a nested folder.

For example, if all source code is in `MyLib/Source` directory, then new files or folders should be added within the `Source` folder instead of the `MyLib` folder. For development pods which do not have a common nested directory, the behavior will remain the same.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
+* Set the path of development pod groups to the top-most shared directory  
+  [Eric Amorde](https://github.com/amorde)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
+* Set the path of development pod groups to the top-most shared directory
```
dnkoutso: nit I know some don't and its unintuitive a bit but I still update my CHANGELOG entry to point to the PR even without an issue.

amorde: nah agreed, I prefer to do it too. will add

PR #8450, opened 2019-01-25
Merge 1-6-stable into master

dnkoutso:


Gemfile.lock
```diff
@@ -7,34 +7,34 @@ GIT
 
 GIT
   remote: https://github.com/CocoaPods/Core.git
-  revision: 167566c0150b8ef9de585cc7bacc664672dc0a54
+  revision: 6900c1b10c50f6201d33061697415d8c19d3a419
```
dnkoutso: while merging picked up the SHA from Core master which would be 1.7.0 instead of the one from `1-6-stable`

dnkoutso: that is because `1-6-stable` also points to a `1-6-stable` branch in `cocoapods-core` but in 1.7.0 we want master.

dnkoutso:
@igor-makarov we just made 1.6.0.rc.1 release but it seems when I merge the changes from 1.6 to master (1.7) a CDN test fails, do you have any insight as to why?

dnkoutso:
This https://github.com/CocoaPods/Specs/blob/master/CocoaPods-version.yml was updated as part of the release.

igor-makarov:
@dnkoutso I will look into it, probably tomorrow as it's the weekend here...

igor-makarov:
Yeah, in Core there are mocks for this. However, here it worked without and I'm curious what's changed that it doesn't now.

dnkoutso:
Yeah I do not know why it started failing exactly either but it has to do with this file.

dnkoutso:
so OK I think I found the issue. Basically the cocoapods-core in `Gemfile.lock` was never updated on master and it still points to 1.6.0.beta.2 at SHA 167566c0150b8ef9de585cc7bacc664672dc0a54. 

Because of this bump here we are dealing with this failure on line 44:
```ruby
 download_file('CocoaPods-version.yml')
```

Which happens within `cdn_source.rb` for master (1.7.0).

dnkoutso:
Fixed with commit https://github.com/CocoaPods/CocoaPods/pull/8450/commits/6082aee62ce06b748a6fc0b0e443b11736785275 as part of this PR.

Gemfile.lock
dnkoutso: figured it out.

igor-makarov:
Seems odd, but nothing seems broken functionality-wise. 🤷🏻‍♂️

igor-makarov:
Thanks for the fix!

Issue #8453, opened 2019-01-25
Pods are linked (i.e. `-l`) to vendored frameworks instead of vendored libraries


Westacular:
* [x] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

Run "pod install" with 1.6.0.rc.1, on an app that incorporates a private Pod that contains both vendored static libraries and vendored static frameworks.

For the purposes of this report, let's say the pod is named `MyPod` and contains `libLibraryFoo.a` as a vendored library and `FrameworkBar.framework` as a vendored framework.

## What did you expect to happen?

I expected the `MyPod.xcconfig` that was generated to contain
```
OTHER_LDFLAGS = $(inherited) -l"LibraryFoo" -framework "FrameworkBar"
```

## What happened instead?

The MyPod.xcconfig that was generated contains
```
OTHER_LDFLAGS = $(inherited) -l"FrameworkBar" -framework "FrameworkBar"
```

i.e., it's erroneously linking `FrameworkBar` both as a library and a framework, _and_ there's no mention of `LibraryFoo` in any of the `LDFLAGS` for `MyPod` _or_ the application.

The culprit appears to be [build_settings.rb:610](https://github.com/CocoaPods/CocoaPods/blob/1.6.0.rc.1/lib/cocoapods/target/build_settings.rb#L610), inside `define_build_settings_method :libraries`, it's doing
```ruby
libraries.concat vendored_static_frameworks.map { |l| File.basename(l, '.framework') }
```
where I believe the intended code is
```ruby
libraries.concat vendored_static_libraries.map { |l| File.basename(l, l.extname).sub(/\Alib/, '') }
```
and the error is a result of some copy and pasting from the frameworks code that never got cleaned up.

Making that change locally appears to result in the correct behaviour. I'll make a PR on the assumption that this is the correct fix.

### Stack

```
   CocoaPods : 1.6.0.rc.1
        Ruby : ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-darwin16]
    RubyGems : 2.6.11
        Host : Mac OS X 10.14.2 (18C54)
       Xcode : 10.1 (10B61)
         Git : git version 2.19.1
Ruby lib dir : /usr/local/Cellar/ruby/2.4.1_1/lib
[...]

PR #8455, opened 2019-01-26
Fix linking of vendored libraries in pod targets

Westacular:
Closes [8453](https://github.com/CocoaPods/CocoaPods/issues/8453).

Also, updates unit tests to avoid this in future.

Specifically: the build_settings unit tests _should_ have caught this issue, however, the banana-lib test project that tested the relevant functionality had both Bananalib.framework and libBananalib.a, which both have the same name when stripped down, creating an ambiguity.

I've updated the banana-lib project to instead have BananaFramework.framework and libBananaStaticLib.a, and updated all the unit tests in accordance with the new names.

I've verified that the relevant build_setting test now fails without the fix, and passes with it.

One curiosity was that the `'detects duplicate library names'` test was failing following the rename, and not raising a warning as expected. Not entirely sure why that is... in any case, I've modified that test (based on the example set in `'detects duplicate framework names'`) and it now passes again.


CHANGELOG.md
```diff
@@ -12,8 +12,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
-* None.  
-
+* Fix linking of vendored libraries and frameworks in pod targets
```
dnkoutso: nit please add 2 empty spaces in the end of this line for formatting.

Westacular: Oh, right. Done.

lib/cocoapods/target/build_settings.rb
```diff
@@ -607,7 +607,7 @@ def initialize(target, test_spec = nil)
 
           libraries = []
           if test_xcconfig? || target.requires_frameworks? && !target.static_framework?
-            libraries.concat vendored_static_frameworks.map { |l| File.basename(l, '.framework') }
+            libraries.concat vendored_static_libraries.map { |l| File.basename(l, l.extname).sub(/\Alib/, '') }
```
amorde: the logic of `MyLibrary` -> `libMyLibrary.a` also existing in `:static_libraries_to_import` and `:dynamic_libraries_to_import` - could you pull it out into a function?

spec/unit/target/build_settings/pod_target_settings_spec.rb
```diff
@@ -221,7 +221,7 @@ class BuildSettings
             @generator.spec_consumers.each { |sc| sc.stubs(:frameworks => []) }
             @generator.stubs(:dependent_targets => [pod_target])
             @generator.other_ldflags.should.
-              be == %w(-l"Bananalib" -l"xml2" -framework "Bananalib")
+              be == %w(-l"BananaStaticLib" -l"xml2" -framework "BananaFramework")
```
amorde: nice catch on this test!

lib/cocoapods/target/build_settings.rb
Westacular: Sure, done

dnkoutso:
thanks!

PR #8232, opened 2018-10-29
Add a Target::BuildType to represent how a target is built

segiddins:
This centralizes logic for static/dynamic library/framework in a single place, and will make future per-target configuration much easier.

- [x] Docs
- [x] Tests for the new class
- [x] Update integration specs
- [x] Rebase

PR #8457, opened 2019-01-26
Fix default framework type to be dynamic

dnkoutso:
This was a regression here 0b0e8dadd2dae796667e7969c95287efb02ca39f and specifically here https://github.com/CocoaPods/CocoaPods/pull/8232/files#diff-f5c3fee776e864ada6541431a1532b00R477

I do not think we should change the linking style by default to static.

Only affects master.

edit: I was wrong it is not a regression per se because we are still using `host_requires_frameworks?` everywhere else, which we should delete.

### DIFF
--- a/lib/cocoapods/installer/analyzer.rb
+++ b/lib/cocoapods/installer/analyzer.rb
@@ -480,7 +480,7 @@ module Pod
                                                                                        resolver_specs_by_target,
                                                                                        build_configurations)
 
-        build_type = target_definition.uses_frameworks? ? Target::BuildType.static_framework : Target::BuildType.static_library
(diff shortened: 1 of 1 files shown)


segiddins:
This definitely should have a new spec added

segiddins:
I don't think this change is correct -- we've had the `AggregateTarget` be static for a while

dnkoutso:
Really? I didn't know that....

dnkoutso:
So whenever we have `use_frameworks!` the aggregates are static frameworks instead? Yeah surprised as well.

segiddins:
It's so we're not copying a `Pods-`... framework into the user's app

PR #8466, opened 2019-01-28
Merge `1-6-stable` into `master`

dnkoutso:


lib/cocoapods/target/build_settings.rb
```diff
@@ -623,13 +623,24 @@ def initialize(target, non_library_spec = nil)
...
+        end
+
         # @return [Array<String>]
         define_build_settings_method :libraries, :memoized => true, :sorted => true, :uniqued => true do
           return [] if library_xcconfig? && target.build_as_static?
 
           libraries = []
           if non_library_xcconfig? || target.build_as_dynamic?
-            libraries.concat vendored_static_frameworks.map { |l| File.basename(l, '.framework') }
+            libraries.concat linker_names_from_libraries(vendored_static_libraries)
```
dnkoutso: `vendored_static_libraries` is correct compared to `vendored_static_frameworks` which was the real fix.

PR #6558, opened 2017-03-14
Only share pod target xcscheme if present during validation.

dnkoutso:
@segiddins this regressed from https://github.com/CocoaPods/CocoaPods/issues/5670. When linting pre-built pods then there is no xcscheme generated to share.

Caught in an internal project of ours and verified this fixes it. Also added a test.

PR #8473, opened 2019-01-29
Share pod target xcscheme if present during validation

icecrystal23:
This is the proper fix for 6558


icecrystal23:
Probably.  I tested it manually (using pod lib lint --no-clean and making sure the framework scheme was there).

I may have time to add a test later.

dnkoutso:
OK. will wait for a test to be added to merge this.

dnkoutso:
@icecrystal23 I didnt see you added tests here for this change!

spec/unit/validator_spec.rb
```diff
@@ -788,6 +788,28 @@ def podspec_path(name = 'JSONKit', version = '1.4')
...
           @validator.send(:add_app_project_import)
         end
+
+        it 'should share pod target xcscheme' do
+          @validator.send(:create_app_project)
+          pods_project = Xcodeproj::Project.new(@validator.validation_dir + 'Pods/Pods.xcodeproj')
+          pod_target = fixture_pod_target('coconut-lib/CoconutLib.podspec')
+          installer = stub(:pod_targets => [pod_target])
+          installer.stubs(:pods_project).returns(pods_project)
+          File.stubs(:exist?).returns(true)
```
dnkoutso: can you pass exactly what you expect here? Why are we blindly mocking this?

dnkoutso: These tests don't appear to be testing anything in particular. If we mock out this above then of course the test would return `true`. 

I think you need more of an integration test which is the problem with this class and its tests in general.

icecrystal23: Yeah, the test could be improved.  I probably won't have time in the near future to get this working as an integration test.  If we could mock the specific file name rather than mocking the whole method, would that be "good enough" for now? (I'm not sure if that's possible with these mocking tools, though.)

icecrystal23: I looked at this a little more today.  It is hard to improve this test since it doesn't actually install the pods.  The test for the initial fix for #6558 is also pretty sorry as it stubs out the method `@validator.stubs(:shares_pod_target_xcscheme?).returns(false)`

Since my fix edits the code in that file, I at least tried to stub the dependency within that method (does the file exist or not) rather than stubbing out the whole method.

lib/cocoapods/validator.rb
```diff
@@ -821,7 +821,8 @@ def translate_output_to_linter_messages(parsed_output)
     end
 
     def shares_pod_target_xcscheme?(pod_target)
-      Pathname.new(@installer.pods_project.path + pod_target.label).exist?
+      scheme_path = Xcodeproj::XCScheme.user_data_dir(@installer.pods_project.path) + "#{pod_target.label}.xcscheme"
```
dnkoutso: I still don't understand the fix here. Can you give a path example of what works today and one that doesn't?

icecrystal23:
To reproduce the issue:
Clone this git repo, for Pod "Pulley": https://github.com/52inc/Pulley
Run "pod lib lint --no-clean"
Open the example workspace.  Notice that there is no "Pulley" framework scheme shared.

djmareewish (changes requested):
@icecrystal23 #9882

icecrystal23:
> @icecrystal23 #9882

I am not following the connection @djmareewish . However, this PR is not too important for the end users as far as I recall and was just making it easier to validate things when there are issues. Let me know if you need any follow up.

Issue #8496, opened 2019-02-07
Sources under a "Resources" directory missing with a Resource Bundle


sebastianv1:
# Report
**Note**: This is based off the tip of CocoaPods master.

Xcode does not have the correct path to source files under a directory named `Resources` when its resource bundle also points to a directory named `Resources`. These are not the same directories, as the project structure looks like:
```
MyProject
-->Sources
---->Resources
------>1.h & 1.m
-->Resources
---->Images.xcassets
```
This causes the build to fail as `1.h` and `1.m` are missing.

## CocoaPods Environment
branch:`master`

## Project that demonstrates the issue
[ResourcesErrorSample.zip](https://github.com/CocoaPods/CocoaPods/files/2842173/ResourcesErrorSample.zip)

segiddins:
Any chance this is related to https://github.com/CocoaPods/CocoaPods/pull/8445/files ?

segiddins:
cc @amorde

amorde:
Ah yep, it is

amorde:
I'm thinking we shouldn't be using `reflect_file_system_structure` for resources here

https://github.com/CocoaPods/CocoaPods/blob/984af0b1c31000b30455b276bcac21d424003f40/lib/cocoapods/installer/xcode/pods_project_generator/file_references_installer.rb#L110-L115

amorde:
Have a fix for this, PR will be up soon

PR #8503, opened 2019-02-10
Fix incorrect path for Resources group in development pods

amorde:
Closes https://github.com/CocoaPods/CocoaPods/issues/8496

This PR changes the behavior to set the path of the Pod group to the Pod root, rather than the most common path. This removes a lot of ambiguity between groups that CocoaPods creates and groups created by the existence of a directory.

The downside to removing the common path handling is that folders such as `Source` or `Classes` will show up in Xcode. This seems reasonable to me because it reflects the actual structure on disk.

### DIFF
--- a/spec/unit/installer/xcode/pods_project_generator/file_references_installer_spec.rb
+++ b/spec/unit/installer/xcode/pods_project_generator/file_references_installer_spec.rb
@@ -280,6 +280,21 @@ module Pod
           end
 
           #-------------------------------------------------------------------------#
+
+          describe 'Installation With Development Pods' do
+            before do
+              @project = Project.new(config.sandbox.project_path)
+              @project.add_pod_group('BananaLib', fixture('banana-lib'), true)
+            end
+
+            it 'sets the path of the Pod group to the installation root' do
+              @installer.install!
+              group = @project.group_for_spec('BananaLib')
+              group.path.should == fixture('banana-lib').relative_path_from(config.sandbox.root).to_s
+            end
+          end
+
+          #-------------------------------------------------------------------------#
         end
       end
     end
--- a/spec/unit/installer/xcode/pods_project_generator/file_references_installer_spec.rb
+++ b/spec/unit/installer/xcode/pods_project_generator/file_references_installer_spec.rb
@@ -290,7 +290,7 @@ module Pod
             it 'sets the path of the Pod group to the installation root' do
               @installer.install!
               group = @project.group_for_spec('BananaLib')
-              group.path.should == fixture('banana-lib').relative_path_from(config.sandbox.root).to_s
+              group.path.should == '../../spec/fixtures/banana-lib'
             end
           end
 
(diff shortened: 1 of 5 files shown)


amorde:
Integration specs will need updating, but will wait for review to see if this approach is reasonable

amorde:
```diff
Bacon::Error: File comparison error `Pods/Pods.xcodeproj` for install_multiple_test_specs:
         - AppHost-TestLib-Unit-Tests-Info.plist
         - LaunchScreen.storyboard
         - main.m
-      - main.swift
       - Pod:
         - TestLib.podspec
       - Support Files:
         - TestLib.unit-unittests1.xcconfig
         - TestLib.unit-unittests2.xcconfig
         - TestLib.xcconfig
+      - TestLib:
+        - Classes:
+          - main.swift
       - UnitTests1:
-        - Test1.swift
+        - TestLib:
+          - UnitTests1:
+            - Test1.swift
       - UnitTests2:
-        - Test2.swift
+        - TestLib:
+          - UnitTests2:
+            - Test2.swift
   - Frameworks:
     - iOS:
       - Foundation.framework
```

This integration spec failure is interesting, the `UnitTests1` and `UnitTest2` duplication is a bummer. I wonder if dev pods should just remove the `Test` group in favor of matching the format on disk?

segiddins:
> I wonder if dev pods should just remove the `Test` group in favor of matching the format on disk?

But then what happens when you have multiple test targets, each coming from the same directory?

amorde:
@segiddins good point, that won't work. That might be enough of a reason to keep the `common_paths` functionality then

amorde:
Another thing I realized here is the collision with the `Resources` group / folder - because this pod contains source files inside `Source/Resources` directory, and `common_paths` strips the `Source` group, we are left with a Resources group that contains files from `Sources/Resources` *and* resource files from the `Resources` directory

amorde:
Okay, sticking with the conservative approach to get the previously desired effect from #8445 while fixing #8496. Ready for review

spec/unit/installer/xcode/pods_project_generator/file_references_installer_spec.rb
```diff
@@ -280,6 +280,21 @@ class PodsProjectGenerator
...
+          describe 'Installation With Development Pods' do
+            before do
+              @project = Project.new(config.sandbox.project_path)
+              @project.add_pod_group('BananaLib', fixture('banana-lib'), true)
+            end
+
+            it 'sets the path of the Pod group to the installation root' do
+              @installer.install!
+              group = @project.group_for_spec('BananaLib')
+              group.path.should == fixture('banana-lib').relative_path_from(config.sandbox.root).to_s
```
dnkoutso: should we be explicit on the value we expect instead of calling a method?

amorde: hmm sure can do, easy enough to change later if necessary

PR #8512, opened 2019-02-13
[Repo::Add] Fix file:// URLs on Ruby 2.6

segiddins:
Also make the tests quiter

### DIFF
--- a/lib/cocoapods/command/repo/push.rb
+++ b/lib/cocoapods/command/repo/push.rb
@@ -107,7 +107,7 @@ module Pod
         # specs to the master repo.
         #
         def check_if_master_repo
-          remotes = `git -C "#{repo_dir}" remote -v 2>&1`
+          remotes, = Executable.capture_command('git', %w(remote --verbose), :capture => :merge, :chdir => repo_dir)
           master_repo_urls = [
             'git@github.com:CocoaPods/Specs.git',
             'https://github.com/CocoaPods/Specs.git',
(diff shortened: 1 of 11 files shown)


lib/cocoapods/command/repo/push.rb
```diff
@@ -226,7 +227,7 @@ def add_specs_to_repo
         #
         def push_repo
           UI.puts "\nPushing the `#{@repo}' repo\n".yellow
-          repo_git('-C', repo_dir, 'push', 'origin', 'master')
```
dnkoutso: why is `-C` removed as an option here?

lib/cocoapods/command/repo/push.rb
```diff
@@ -160,7 +160,8 @@ def validate_podspec_files
         # @return [void]
         #
         def check_repo_status
-          clean = `git -C "#{repo_dir}" status --porcelain  2>&1` == ''
+          porcelain_status, = Executable.capture_command('git', %w(status --porcelain), :capture => :merge, :chdir => repo_dir)
```
dnkoutso: same question here

spec/spec_helper/temporary_repos.rb
```diff
@@ -19,18 +19,18 @@ def repo_make(name)
       path = repo_path(name)
       path.mkpath
       Dir.chdir(path) do
-        `git init`
+        Pod::Executable.capture_command!('git', %w(init))
         repo_make_readme_change(name, 'Added')
-        `git add .`
-        `git commit -m "Initialized."`
+        Pod::Executable.capture_command!('git', %w(add .))
+        Pod::Executable.capture_command!('git', %w(commit -nm Initialized.))
```
dnkoutso: also added option `-n` here.

lib/cocoapods/command/repo/push.rb
segiddins: `repo_git` already adds it

lib/cocoapods/command/repo/push.rb
segiddins: not needed since we have the `chdir`

spec/spec_helper/temporary_repos.rb
segiddins: yes, `-n` skips verification, since we're making the initial commit in the repo, it normally spits out an error message

PR #8368, opened 2018-12-19
Incremental pod installation option.

sebastianv1:
Adds the option to only generate the projects for pod targets that have changed since the previous installation.

PR #8524, opened 2019-02-18
Move pre-save UUID manipulations to be before the post-install hooks

igor-makarov:
Quite recently (#8368), there was a change introduced that generated predictable UUIDs for target objects.  
This has caused a regression in my post-install hook, where I'm relying on the UUID of the target to be known and final.  

The PR moves the `predictabilize_uuids` and the `TargetUUIDGenerator` calls from the writer to the installer, to be right before the call to `run_podfile_post_install_hooks`.


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Move pre-save UUID manipulations to be before the post-install hooks  
```

</p>

igor-makarov:
I've updated the integration specs in [cocoapods-integration-specs#216](https://github.com/CocoaPods/cocoapods-integration-specs/pull/216).

lib/cocoapods/installer.rb
```diff
@@ -43,6 +43,7 @@ class Installer
     autoload :SandboxHeaderPathsInstaller,  'cocoapods/installer/sandbox_header_paths_installer'
     autoload :SandboxDirCleaner,            'cocoapods/installer/sandbox_dir_cleaner'
     autoload :ProjectCache,                 'cocoapods/installer/project_cache/project_cache'
+    autoload :TargetUUIDGenerator,          'cocoapods/installer/xcode/target_uuid_generator'
```
segiddins: this auto load should be nested somewhere in the `Xcode` namespace

igor-makarov: I've noticed that the source file for `TargetUUIDGenerator` isn't correctly placed, I moved it now.

igor-makarov:
I've noticed that the source file for `TargetUUIDGenerator` isn't correctly placed, I moved it now.

lib/cocoapods/installer.rb
```diff
@@ -318,6 +323,14 @@ def create_and_save_projects(pod_targets_to_generate, aggregate_targets_to_gener
       end
     end
 
+    def predictabilize_uuids
```
sebastianv1: nit: Can we have the functions pass in a list of projects. And should they also be private?

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
+* Move pre-save UUID manipulations to be before the post-install hooks
```
dnkoutso: I'd argue this isn't needed and is encapsulated as part of new feature development of multi-xcodeproj.

igor-makarov: I wasn't talking about encapsulation though. I have a post-install script that relies on UUIDs being known at the time.  
Specifically, it uses them to mark some targets as non-migratable for Swift so that there are no warnings about targets I can't control.

igor-makarov: Oh, did you mean the changelog record?

lib/cocoapods/installer.rb
igor-makarov: What is the rationale for this, exactly? :generated_projects is an instance prop.

CHANGELOG.md
dnkoutso: yes!

igor-makarov: Sure, will remove...

CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Move pre-save UUID manipulations to be before the post-install hooks  
```

</p>

lib/cocoapods/installer.rb
sebastianv1: Injecting parameters is generally a better practice and makes it much easier to reason about each property's state especially given that it's a class instance var not assigned in the initializer. This will also make refactoring much easier in the future.

igor-makarov: So why is there an instance property? Should I remove it?

igor-makarov: Ok, I see your point. I've made the methods receive the projects as a parameter. 

They are private because they are called from a private method. Do you think that's a good reason? I don't see any other reason to call them in this context and for other contexts they can use the generator classes.

sebastianv1: Ah they're already marked private, I missed the `private` above.

igor-makarov:
💪

PR #8528, opened 2019-02-20
Installation option to skip generating pod schemes.

dnkoutso:
This is useful for very large projects where the number of schemes actually slows down Xcode. This does not change the default integration.

# Tests to be added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
+* Installation option to skip generating pod schemes.  
+  [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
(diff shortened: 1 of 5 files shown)


dnkoutso:
@segiddins how does (if at all) collide with the option `share_schemes_for_development_pods` which takes a bool or string.

I figured that `generate_pod_schemes` should just take precedence instead of us reconciling the two. WDYT?

segiddins:
If it's conflicting podfile options, the podfile validator should validate they arent incompatible

dnkoutso:
> If it's conflicting podfile options, the podfile validator should validate they aren't incompatible

They are not conflicting. `generate_pod_schemes` as it stands from this PR will take precedence and will always skip sharing schemes even if `share_schemes_for_development_pods` is set to `true`.

PR #8470, opened 2019-01-29
Copies the bcsymbolmap files of a vendored framework

dacaiguoguogmail:
Vendored frameworks might have bcsymbolmap files next to them so should copy them  https://github.com/CocoaPods/CocoaPods/issues/8461

PR #8535, opened 2019-02-21
Follow up improvements for `bcsymbolmap` paths

dnkoutso:
Follow up on https://github.com/CocoaPods/CocoaPods/pull/8470 /cc @dacaiguoguogmail

### DIFF
--- a/lib/cocoapods/target/framework_paths.rb
+++ b/lib/cocoapods/target/framework_paths.rb
@@ -34,7 +34,7 @@ module Pod
       end
 
       def all_paths
-        [source_path, dsym_path].compact
+        [source_path, dsym_path, bcsymbolmap_paths].flatten.compact
       end
     end
   end
--- a/spec/unit/target/pod_target_spec.rb
+++ b/spec/unit/target/pod_target_spec.rb
@@ -662,6 +662,50 @@ module Pod
         end
       end
 
+      describe 'resource and framework paths' do
+        before do
+          @watermelon_spec = fixture_spec('watermelon-lib/WatermelonLib.podspec')
+          @monkey_spec = fixture_spec('monkey/monkey.podspec')
+          @target_definition = Podfile::TargetDefinition.new('Pods', nil)
+          @target_definition.abstract = false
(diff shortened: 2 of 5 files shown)


spec/unit/target/pod_target_spec.rb
```diff
@@ -662,6 +662,39 @@ module Pod
         end
       end
 
+      describe 'resource and framework paths' do
```
dnkoutso: moved these to its own describe method, they should not have been under "test spec support"

lib/cocoapods/target/framework_paths.rb
```diff
@@ -34,7 +34,7 @@ def hash
       end
 
       def all_paths
-        [source_path, dsym_path].compact
+        [source_path, dsym_path, bcsymbolmap_paths].compact
```
dnkoutso: @amorde do you think we want this? It makes perfect sense for this class but this propagates to input paths of script phases...As I am typing this I think we need it and that was a bug in the original PR.

spec/unit/target/pod_target_spec.rb
```diff
@@ -662,6 +662,50 @@ module Pod
...
+            ],
+            'WatermelonLib/Tests' => [],
+            'WatermelonLib/SnapshotTests' => [],
+            'WatermelonLib/App' => [
+              Target::FrameworkPaths.new('${BUILT_PRODUCTS_DIR}/WatermelonLib/WatermelonLib.framework'),
+            ],
+          }
+        end
+
+        it 'returns correct vendored framework paths' do
```
dnkoutso: new test for vendored framework paths.

lib/cocoapods/target/framework_paths.rb
amorde: yes definitely, I had thought this was covered in the review but I guess it was just included in `hash` https://github.com/CocoaPods/CocoaPods/pull/8470#discussion_r252100005

amorde: could you add a test for this? just realized there aren't any

dnkoutso: yes sir!

dnkoutso: added

PR #8539, opened 2019-02-21
Cherry pick fixes for Xcodeproj 1.8.1 bump to 1-6-stable

dnkoutso:


amorde:
Sorry I missed the original implementation, but I'm concerned about this because it will break Podspecs that specify `OTHER_LDFLAGS` in their `pod_target_xcconfig` or `post_install` hooks that use that setting. I see Xcodeproj was already released though

amorde:
Okay, after investigation this only affects target build settings (not .xcconfigs) and may only cause a problem if a post_install hook reads and writes values as different data types

PR #8572, opened 2019-03-03
Add sibling targets to cache analysis to ensure they regenerate

### LINKED ISSUE #7173: Previously installed pods for a smaller subset of platforms can be incomplete on incremental installation
# Report

## What did you do?

This is not at all specific to the involved pod, but this allows to reproduce this issue. It must be a podspec which is on multiple platforms on which it uses different proper subsets of source files.

1. Assume initially a Podfile as:
```ruby
target "TargetA" do
  platform :ios, '10.0'
  pod "Fabric"
end
```

2. Run `pod install`
3. Add a target on a different platform using the same pod:
```diff
target "TargetA" do
  platform :ios, '10.0'
  pod "Fabric"
end

+target "TargetB" do
+  platform :macos, '10.13'
+  pod "Fabric"
+end
```

Install all pod dependencies correctly.

## What happened instead?

Pod Fabric is missing the platform-specific sources for macOS.

## Workaround

Nuke the Pods directory and run `pod install` again.

## Why does that go wrong?

This happens because the `PodDirCleaner` has cleaned the pod previously based on just including source files belonging to the selected subset of platforms, here iOS. When we add macOS as a platform, the `SandboxAnalyzer` doesn't know for what platforms this pod was installed before as this information is not present in the sandbox manifest, so that it cannot know that the pod has to be flagged as changed. So the pod will be categorized as unchanged and the installer will skip any further actions assuming there is nothing further to do.
The cache which runs the `PodDirCleaner` as well is not affected by this.

### A: Clean more forgiving

Don't clean files which are specified for platforms which are unused. That would require that the `PodDirCleaner` is initialized with the root spec for all platforms.
❎ This can blow up the size of the installation for certain cases.

### B: Add the platform to the sandbox manifest

Store together with each pod in the sandbox manifest for what platform a pod was used when it was installed. That would allow the `SandboxAnalyzer` to flag the pod as changed and the installer would put the files correctly in place again.
❎ This requires changes to Core, migration to a new version etc.

### LINKED ISSUE #8577: When using incremental installs, not all targets get generated
I've encountered an issue while using the experimental Incremental Install feature in CP 1.7.0.
The issue occurs when using a pod that results in more target.  
When one of the targets needs generation, and incremental install is enabled, only the modified target gets generated.

I've made a demo project that reproduces the issue: [link](https://github.com/igor-makarov/DemoCPIncrementalBug)

I'm attaching the excerpt from the Podfile from the project here:
```ruby
abstract_target 'All' do
  pod 'BlocksKit/Core'
  target 'App' do
    pod 'BlocksKit'
    pod 'DevPod/App', path: './DevPod/'
  end
  target 'Widget' do
    pod 'DevPod/Widget', path: './DevPod/'
  end
end
```

Adding a file to the subspec `DevPod/Widget` causes a regeneration of the relevant target. However, the sibling target `DevPod/App` gets missed.

I've already got PR #8572 that fixes the problem, however I don't have an idea of how to write the tests for it. @sebastianv1 thanks for volunteering to help!

igor-makarov:
This PR fixes #8577.

I've verified it works with my [Demo project](https://github.com/igor-makarov/DemoCPIncrementalBug) however I'm having trouble adding a test for this case. 

Can anyone help me out with using multiple targets and subspecs in `project_cache_analyzer_spec.rb`?

### DIFF
--- a/lib/cocoapods/installer/project_cache/project_cache_analyzer.rb
+++ b/lib/cocoapods/installer/project_cache/project_cache_analyzer.rb
@@ -89,6 +89,10 @@ module Pod
 
           pod_targets_to_generate = (changed_pod_targets + added_pod_targets + dirty_pod_targets).uniq
 
+          # If one target in a pod needs generation, then all of the targets do.
+          # Otherwise we end up with the rest of the targets ungenerated.
+          pod_targets_to_generate += compute_sibling_pod_targets(pod_targets, pod_targets_to_generate)
+
           # We either return the full list of aggregate targets or none since the aggregate targets go into the Pods.xcodeproj
           # and so we need to regenerate all aggregate targets when regenerating Pods.xcodeproj.
           aggregate_target_to_generate =
@@ -148,6 +152,13 @@ module Pod
                                  raise "[BUG] Unknown target type #{target}"
                                end
             support_files_dir_exists && xcodeproj_exists
+          end
+        end
+
+        def compute_sibling_pod_targets(all_pod_targets, targets)
+          pod_names = targets.map(&:pod_name)
+          all_pod_targets.select do |target|
+            pod_names.include?(target.pod_name) && !targets.include?(target)
           end
         end
       end


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Add sibling targets to cache analysis to ensure they regenerate  
```

</p>

dnkoutso:
I'll let @sebastianv1 review. tl;dr i dont think the original issue filed in 2017 referring to "incremental installation" is the same "incremental installation" launched with 1.7.0.beta.

igor-makarov:
Oops!  
I didn't notice the date on the issue and it seemed consistent with the problem I was having.  
Perhaps I ought to open a new issue to eliminate confusion?

sebastianv1:
@igor-makarov Yeah looks like an older issue. Do you mind also adding to the issue some repro steps describing what changed that resulted in the pods not being regenerated with your demo project?

dnkoutso:
If it fixes #7173 and the issue you describe is more or less the same then I do not think there is a need for a new issue to be filed. There will be breadcrumb here for people to follow

igor-makarov:
There's no real repro in that issue. I think it's best that I make a new one.

I would still like some help with the unit test, if anyone is up to it...

sebastianv1:
@igor-makarov I'd be happy to help. Once you file that ticket let me know so I can get a better idea where the analyzer fails and then we can write one around that.

igor-makarov:
It's all detailed in https://github.com/igor-makarov/DemoCPIncrementalBug actually!  
Just follow the steps listed there to reproduce.

igor-makarov:
I've opened #8577 and updated the original post to remove the link to the confusing issue.
To make sure, this PR fixes #8577.

lib/cocoapods/installer/project_cache/project_cache_analyzer.rb
```diff
@@ -89,6 +89,10 @@ def analyze
 
           pod_targets_to_generate = (changed_pod_targets + added_pod_targets + dirty_pod_targets).uniq
 
+          # If one target in a pod needs generation, then all of the targets do.
+          # Otherwise we end up with the rest of the targets ungenerated.
```
sebastianv1: I would change this comment to something more along the lines:
```
Since multi xcodeproj generation groups targets by `pod_name`, we need to also append the sibling `PodTarget` instances under the same `pod_name` if any have changed to make sure we end up with the correct set of targets.
```
Feel free to rephrase, but the main idea is that the bug occurs because of how multi xcode proj generation works and that we need to make sure the set of pod targets to generate includes those siblings.

lib/cocoapods/installer/project_cache/project_cache_analyzer.rb
```diff
@@ -150,6 +154,13 @@ def compute_dirty_targets(targets)
             support_files_dir_exists && xcodeproj_exists
           end
         end
+
+        def compute_sibling_pod_targets(all_pod_targets, targets)
```
sebastianv1: change parameters to `pod_targets` and `pod_targets_to_generate`

lib/cocoapods/installer/project_cache/project_cache_analyzer.rb
```diff
@@ -150,6 +154,13 @@ def compute_dirty_targets(targets)
...
+          pod_names = targets.map(&:pod_name)
+          all_pod_targets.select do |target|
```
sebastianv1: It might be cleaner to do something like:
```ruby
pod_targets_by_name = pod_targets.group_by { |target| target.pod_name }
sibling_targets_to_generate = pod_targets_to_generate.flat_map { |t| pod_targets_by_name[t.pod_name] }
```
And then keep the `.uniq` above where we assign `pod_targets_to_generate =`

sebastianv1:
Do you want me to add a commit onto your PR with some specs?

igor-makarov:
Sure thing!

sebastianv1:
Cool. I'll try and get some added by end of day

dnkoutso:
Will go to 1.7.0.beta.2, just needs a `CHANGELOG` entry and a test which you both said is coming :)

igor-makarov:
In what cases is a changelog entry required?  
Last time I added one I was asked to remove it so this time I wasn't too sure. 

It's a bug in the implementation detail, no?

dnkoutso:
If I recall correctly the prior time the changelog entry was removed because there was a follow up fix on a new feature.

igor-makarov:
Well, isn't incremental install also a new feature for 1.7.0'?

dnkoutso:
Correct and it launched with beta 1 and now we did a fix on top of it for beta 2.

igor-makarov:
Oh got it... I'll add it tomorrow.

dnkoutso:
Superseded by https://github.com/CocoaPods/CocoaPods/pull/8582

PR #8582, opened 2019-03-06
Generate all sibling pod targets that share the same xcode project. 

sebastianv1:
@igor-makarov Here's one spec. I'll add a couple more later. This should ideally fail the tests.

closes #8572

### DIFF
--- a/lib/cocoapods/installer/project_cache/project_cache_analyzer.rb
+++ b/lib/cocoapods/installer/project_cache/project_cache_analyzer.rb
@@ -75,28 +75,42 @@ module Pod
             return full_install_results
           end
 
+          pod_targets_to_generate = Set[]
+          aggregate_targets_to_generate = Set[]
           added_targets, removed_targets = compute_added_and_removed_targets(target_by_label,
                                                                              cache_key_by_target_label.keys,
                                                                              cache.cache_key_by_target_label.keys)
           added_pod_targets, added_aggregate_targets = added_targets.partition { |target| target.is_a?(PodTarget) }
           removed_aggregate_targets = removed_targets.select { |target| target.is_a?(AggregateTarget) }
+          pod_targets_to_generate.merge(added_pod_targets)
+          aggregate_targets_to_generate.merge(added_aggregate_targets + removed_aggregate_targets)
 
           changed_targets = compute_changed_targets_from_cache(cache_key_by_target_label, target_by_label, cache)
           changed_pod_targets, changed_aggregate_targets = changed_targets.partition { |target| target.is_a?(PodTarget) }
+          pod_targets_to_generate.merge(changed_pod_targets)
+          aggregate_targets_to_generate.merge(changed_aggregate_targets)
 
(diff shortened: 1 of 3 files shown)


dnkoutso:
For https://github.com/CocoaPods/CocoaPods/pull/8572

sebastianv1:
@igor-makarov Please take a look

spec/unit/installer/project_cache/project_cache_analyzer_spec.rb
```diff
@@ -161,6 +161,52 @@ module ProjectCache
...
+              subspec_target_2.label => TargetCacheKey.from_cache_hash('BUILD_SETTINGS_CHECKSUM' => 'Blah'),
+            }
+            cache = ProjectInstallationCache.new(cache_key_by_aggregate_target_labels, @build_configurations, @project_object_version)
+            analyzer = ProjectCacheAnalyzer.new(@sandbox, cache, @build_configurations, @project_object_version, subspec_pods, [])
+            result = analyzer.analyze
+            result.pod_targets_to_generate.should.equal(subspec_pods)
+            result.aggregate_targets_to_generate.should.equal(nil)
+          end
+
+          it 'returns sibling pod target when adding a new subspec' do
```
igor-makarov: Not sure what this second test does. It passes when the sibling computation is disabled...

igor-makarov:
Looking good!
I left a note that I'm not sure about the second test...

lib/cocoapods/installer/project_cache/project_cache_analyzer.rb
```diff
@@ -75,28 +75,42 @@ def analyze
...
           end
 
+          pod_targets_to_generate = Set[]
+          aggregate_targets_to_generate = Set[]
           added_targets, removed_targets = compute_added_and_removed_targets(target_by_label,
                                                                              cache_key_by_target_label.keys,
                                                                              cache.cache_key_by_target_label.keys)
           added_pod_targets, added_aggregate_targets = added_targets.partition { |target| target.is_a?(PodTarget) }
           removed_aggregate_targets = removed_targets.select { |target| target.is_a?(AggregateTarget) }
+          pod_targets_to_generate.merge(added_pod_targets)
```
dnkoutso: do we want all of these to be `merge!`?

dnkoutso: talked offline, no because its a `Set`

spec/unit/installer/project_cache/project_cache_analyzer_spec.rb
sebastianv1: It's independent but somewhat related to sibling targets. When adding a subspec there will be two new pod targets that the analyzer thinks are new. Just another spec case I thought of

igor-makarov: Okay, I still don't get it but it might be the late hour (2AM).

The test might need a comment.

PR #8593, opened 2019-03-08
Do not warn of `.swift-version` file being deprecated if pod does not use Swift.

dnkoutso:


lib/cocoapods/validator.rb
```diff
@@ -125,7 +125,6 @@ def validate
       $stdout.flush
 
       perform_linting
-      warn_for_dot_swift_file_deprecation
```
dnkoutso: this check was running too _early_. At that time we haven't created the `Installer` and we do not know if the pod uses Swift or not. Moved to a bit later and added a test.

PR #8638, opened 2019-03-25
Deintegrate deleted targets even if `incremental_installation` is turned on.

dnkoutso:


lib/cocoapods/installer/user_project_integrator.rb
```diff
@@ -196,8 +209,8 @@ def workspace_path
           podfile_dir   = File.dirname(podfile.defined_in_file || '')
           absolute_path = File.expand_path(path_with_ext, podfile_dir)
           Pathname.new(absolute_path)
-        elsif user_project_paths.count == 1
-          project = user_project_paths.first.basename('.xcodeproj')
+        elsif user_project_paths_to_integrate.count == 1
```
sebastianv1: Should this be the full set of `user_project_paths`?
If someone doesn't explicitly state the user project in their Podfile and rely on this feature, it seems like this will return a count of 0 causing an error. (Also worth noting it looks like it would happen today incremental enabled)

dnkoutso: ah I think you are correct. We need to take that into consideration

PR #8648, opened 2019-03-28
Fix for AppIcon on Lib Lint

Itaybre:
Xcodeproj generates new projects with a default Icon called AppIcon [here](https://github.com/CocoaPods/Xcodeproj/blob/0ec1f7f1a80ab2be8dfa60864fbac54206e25e93/lib/xcodeproj/constants.rb#L307).
So when executing `pod lib lint`, the generated Xcodeproj has a set AppIcon and then picked up by the [copy_resources_script.rb](https://github.com/CocoaPods/CocoaPods/blob/b52a70d50f9e31a247794feac25ec9b4effbd13f/lib/cocoapods/generator/copy_resources_script.rb#L226).
This change deletes the ASSETCATALOG_COMPILER_APPICON_NAME key in the build settings when linting.
I did not add a check inside copy_resources_script.rb because checking all the image assets might take too long on large projects and might hide misconfigurations on normal projects.

Fixes: 
#7617

Todo:
- [x] Tests
- [x] Update changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Sebastian Shanus](https://github.com/sebastianv1)
   [#8636](https://github.com/CocoaPods/CocoaPods/pull/8636)
 
+* Fix: AppIcon not found when executing `pod lib lint`
+  [Itay Brenner](https://github.com/itaybre)
+  [#8648](https://github.com/CocoaPods/CocoaPods/issues/8648)
+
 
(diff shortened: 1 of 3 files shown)


segiddins:
Looks good, thanks!

Itaybre:
Removed the platform check because after testing, discovered that the same issue happened in all platforms

dnkoutso:
Please point to `1-7-stable` branch if we want this fix in 1.7.0.beta.4. or rc.1

Itaybre:
@dnkoutso done

Issue #8657, opened 2019-03-30
[Bug] compute the cache about `incremental_installation` for 1.7.0.beta3


Whirlwind:
I test the `incremental_installation` in my project. But I found that there are some pod targets always be `Regenerating`.

I don't know why the cache does not work. So I debug the code, and I found that the YAML force use the `downcase` when YAML try to sort the array.

The `installation_cache.yaml`:
```
  AWEVideoFeed-b583dabf:
    BUILD_SETTINGS_CHECKSUM:
      AWEVideoFeed-b583dabf: 70d2e833e159805a67a68ee07012e312
    CHECKSUM: 3df9b79bd1c7543473679c40b930c525a61716cc
    SPECS:
      - AWEVideoFeed/ModuleInterface (0.1.0)
      - AWEVideoFeed/ModuleInterface/Interface (0.1.0)
      - AWEVideoFeed/ModuleInterface/Model (0.1.0)
      - AWEVideoFeed/MUS (0.1.0)
```

The `MUS` is the last one in the `SPECS`, it is because the `when String then value.downcase` of `sorting_string` in the `yaml_helper.rb`.
But the `MUS` is the first one when the target generate the cache key: `'SPECS' => pod_target.specs.map(&:to_s).sort`.

It is the why the cache does not work: the `SPCES` always be different.

dnkoutso:
Fixed by #8663

Issue #8659, opened 2019-03-31
[Feature] Use `Development Pods` group for `generate_multiple_pod_projects` mode


Whirlwind:
When `:generate_multiple_pod_projects=>true`, all pod will use a single project in `Dependencies` group, include the development pods. It is hard to find the development pod in 150+ pods!! On the other hand, it is hard to check whether is a local pod for a pod ( I must check the path of a source file).

Have CP a plan that move back the local pod to `Development Pods` group? Of cause, I still hope keep the single pod project for per-pod in the `Development Pods`.

sebastianv1:
I think it makes sense for subprojects to just holds a `Dependencies` group, but for `Pods.xcodeproj` we could produce a similar structure as before since it holds every pod in the workspace. Thoughts @segiddins @dnkoutso ?

amorde:
I'd be in favor of something like the `Development Pods` group that exists now, I use it in my projects

sebastianv1:
Cool, so I'm thinking the structure would look something like:
```
├── Pods.xcodeproj
│   ├── Development Pods (if they exist)
│   │   ├── Local Pod A
│   │   │   └── Dependencies
│   ├── Pods
│   │   ├── Remote Pod A
│   │   │   └── Dependencies
...
```
The `Dependencies` group still gets generated, but only for each individual subproject and *not* for `Pods.xcodeproj`.

Whirlwind:
Why do we need the `Dependencies` group? I think the old structure is fine:
```
├── Pods.xcodeproj
│   ├── Development Pods (if they exist)
│   │   ├── Local Pod A.xcodeproj
│   ├── Pods
│   │   ├── Remote Pod A.xcodeproj
```

Or, move the development pod into workspace:
```
├── Pods.xcodeproj
│   ├── Pods
│   │   ├── Remote Pod A.xcodeproj
│── Local Pod A.xcodeproj
```

sebastianv1:
`Dependencies` group is required for target dependencies to be set up correctly. A target dependency can only be added to a target if it exists in the same project or is nested as a subproject, hence the `Dependencies` group.

sebastianv1:
PR: #8731

wulongwang:
cocoapods can do that below?
> ```
> ├── Pods.xcodeproj
> │   ├── Remote Pod A.xcodeproj
> │   ├── Framework
> │── Local Pod A.xcodeproj
> ```

PR #8663, opened 2019-04-01
Order SPECS key by sorting according to downcase

sebastianv1:
closes https://github.com/CocoaPods/CocoaPods/issues/8657

* Writing to the yaml file sorts entries by downcase, so reading from the cache and then from a `PodTarget` instance would always produce a "dirty" target since the array orders would be different.
* Apply sort_by operator on files in the target cache key.


sebastianv1:
Fixes #8659

Issue #8653, opened 2019-03-29
Podfile with `integrate_targets: false` set fails on lack of SWIFT_VERSION


dfed:
* [x] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?
Ran `pod install`

## What did you expect to happen?

Download our dependencies distributed via CocoaPods

## What happened instead?
```
Analyzing dependencies
Downloading dependencies
Using CocoaLumberjack (3.4.2)
[!] Unable to determine Swift version for the following pods:

- `CocoaLumberjack` does not specify a Swift version and none of the targets (`Random-Inc`) integrating it have the `SWIFT_VERSION` attribute set. Please contact the author or set the `SWIFT_VERSION` attribute in at least one of the targets that integrate this pod.
```

### Stack

```
   CocoaPods : 1.6.1
        Ruby : ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-darwin17]
    RubyGems : 3.0.2
        Host : Mac OS X 10.14.4 (18E226)
       Xcode : 10.2 (10E125)
         Git : git version 2.21.0
Ruby lib dir : /Users/dan_federman/.rbenv/versions/2.3.1/lib
Repositories : master - https://github.com/CocoaPods/Specs.git @ d1274e7ae102fca8339dae1ec8fa9a6b1331e149
```

### Installation Source

```
Executable Path: /Users/dan_federman/.rbenv/versions/2.3.1/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.2
cocoapods-plugins     : 1.0.0
cocoapods-rome        : 1.0.1
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
cocoapods-trunk       : 1.3.1
cocoapods-try         : 1.1.0
```

### Podfile

```ruby
source 'https://github.com/CocoaPods/Specs.git'

platform :ios, '11.0'

# Only download the files, don't create Xcode projects
install! 'cocoapods', integrate_targets: false

use_frameworks!

target 'Random-Inc' do
  pod "CocoaLumberjack/Swift", '~> 3.0'
end
```

## Project that demonstrates the issue

See Podfile above

dnkoutso:
confirmed both issues. Will fix for 1.7.0.

segiddins:
[...]

PR #8672, opened 2019-04-03
Do not verify pod Swift version from integrated targets if integration is disabled.

dnkoutso:
closes https://github.com/CocoaPods/CocoaPods/issues/8653

### DIFF
--- a/lib/cocoapods/installer/xcode/target_validator.rb
+++ b/lib/cocoapods/installer/xcode/target_validator.rb
@@ -14,18 +14,20 @@ module Pod
         #
         attr_reader :pod_targets
 
-        # Create a new TargetValidator with aggregate and pod targets to
-        # validate.
+        # @return [InstallationOptions] installation_options
         #
-        # @param [Array<AggregateTarget>] aggregate_targets
-        #                                 The aggregate targets to validate.
+        attr_reader :installation_options
+
+        # Initialize a new instance
         #
-        # @param [Array<PodTarget>] pod_targets
-        #                           The pod targets to validate.
+        # @param [Array<AggregateTarget>] aggregate_targets @see #aggregate_targets
+        # @param [Array<PodTarget>] pod_targets @see #pod_targets
+        # @param [InstallationOptions] installation_options @see #installation_options
         #
-        def initialize(aggregate_targets, pod_targets)
+        def initialize(aggregate_targets, pod_targets, installation_options)
           @aggregate_targets = aggregate_targets
           @pod_targets = pod_targets
+          @installation_options = installation_options
         end
 
         # Perform the validation steps for the provided aggregate and pod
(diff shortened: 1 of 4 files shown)


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Do not verify pod Swift version if integration is disabled  
```

</p>

lib/cocoapods/installer/xcode/target_validator.rb
```diff
@@ -34,7 +36,7 @@ def initialize(aggregate_targets, pod_targets)
         def validate!
           verify_no_duplicate_framework_and_library_names
           verify_no_static_framework_transitive_dependencies
-          verify_swift_pods_swift_version
+          verify_swift_pods_swift_version if installation_options.integrate_targets?
```
segiddins: can we also add a test for what swift version should be used when integrating targets is disabled?

dnkoutso: hm good point......what version should it be set to?

segiddins: The highest version the pod supports?

dnkoutso: and what if the pod does not specify `swift_version`?

segiddins: Then we have to error, right?

dnkoutso: This class does not _set_ the swift version for the target, only validates it. I'd argue none of these checks should happen if integration is off.

amorde: Wouldn't that be a reason to fail validation?

segiddins: I think so, @amorde

spec/unit/installer/xcode/target_validator_spec.rb
```diff
@@ -42,8 +42,8 @@ def create_validator(sandbox, podfile, lockfile)
             config.repos_dir = fixture_path + 'spec-repos'
             podfile = Pod::Podfile.new do
               platform :ios, '8.0'
-              install! 'cocoapods', :integrate_targets => false
-              project 'SampleProject/SampleProject'
+              install! 'cocoapods', :integrate_targets => true
```
dnkoutso: all of these tests were "wrong" as in they were never taking into consideration `:integrate_targets => false` above and the project was never properly loaded. This is now fixed.

spec/unit/installer/xcode/target_validator_spec.rb
```diff
@@ -322,69 +321,82 @@ def create_validator(sandbox, podfile, lockfile)
             'the author or set the `SWIFT_VERSION` attribute in at least one of the targets that integrate this pod.'
           end
 
-          it 'does not raise when targets integrate the same pod but only one of the targets is a swift target' do
+          it 'does not raise when integrating targets is disabled' do
```
dnkoutso: diff is weird but this is the new test here that explicitly states `integrate_targets => false`

dnkoutso:
After much work and understanding this is not the right direction we want here.

PR #8676, opened 2019-04-03
Add an error message for :podspec pods not matching the version in Pods and on disk

orta:
Every time we update one of our `:podspec` dependencies, everyone else on the team has to run `bundle exec pod update [the-dep]`. This feels pretty reasonable, because there is an underlaying difference in the version in `Pods/Local Podspecs` and the version on disk being referenced.

What gets tricky is that the error message doesn't really give you much to work with:

```
 ▲ Example  $ bundle exec pod install
Analyzing dependencies
Fetching podspec for `Emission` from `../`
Fetching podspec for `RNSVG` from `../node_modules/react-native-svg`
Fetching podspec for `React` from `../node_modules/react-native`
Fetching podspec for `SentryReactNative` from `../node_modules/react-native-sentry`
Fetching podspec for `react-native-mapbox-gl` from `../node_modules/@mapbox/react-native-mapbox-gl`
Fetching podspec for `react-native-navigator-ios` from `../node_modules/react-native-navigator-ios`
Fetching podspec for `yoga` from `../node_modules/react-native/ReactCommon/yoga`
[!] CocoaPods could not find compatible versions for pod "tipsi-stripe":
  In snapshot (Podfile.lock):
    tipsi-stripe (from `../node_modules/tipsi-stripe/tipsi-stripe.podspec`)

  In Podfile:
    tipsi-stripe (from `../node_modules/tipsi-stripe/tipsi-stripe.podspec`)

None of your spec sources contain a spec satisfying the dependencies: `tipsi-stripe (from `../node_modules/tipsi-stripe/tipsi-stripe.podspec`), tipsi-stripe (from `../node_modules/tipsi-stripe/tipsi-stripe.podspec`)`.

You have either:
 * out-of-date source repos which you can update with `pod repo update` or with `pod install --repo-update`.
 * mistyped the name or version.
 * not added the source repo that hosts the Podspec to your Podfile.

Note: as of CocoaPods 1.0, `pod repo update` does not happen on `pod install` by default.
```

I know what I am doing, and I wasn't quite sure of what to make of that. So, I've improved the error messaging specifically around this issue:

```
 ▲ Example  $ /Users/ortatherox/dev/projects/cocoapods/Rainforest/CocoaPods/bin/pod install
Analyzing dependencies
Fetching podspec for `Emission` from `../`
Fetching podspec for `RNSVG` from `../node_modules/react-native-svg`
Fetching podspec for `React` from `../node_modules/react-native`
Fetching podspec for `SentryReactNative` from `../node_modules/react-native-sentry`
Fetching podspec for `react-native-mapbox-gl` from `../node_modules/@mapbox/react-native-mapbox-gl`
Fetching podspec for `react-native-navigator-ios` from `../node_modules/react-native-navigator-ios`
Fetching podspec for `yoga` from `../node_modules/react-native/ReactCommon/yoga`
[!] CocoaPods could not find compatible versions for pod "tipsi-stripe":
  In snapshot (Podfile.lock):
    tipsi-stripe (from `../node_modules/tipsi-stripe/tipsi-stripe.podspec`)

  In Podfile:
    tipsi-stripe (from `../node_modules/tipsi-stripe/tipsi-stripe.podspec`)

It seems like you've changed the version of the dependency `tipsi-stripe` and it differs from the version stored in `Pods/Local Podspecs`.
You should run `pod update tipsi-stripe --no-repo-update` to apply changes you've made to it.
```

Making it explicit that it's a difference between the cached version, and the local version and giving direct advice on what to do with it.

TODO:
- [ ]  I'm really struggling to make a test for this, none of the other tests rely on previous state and I'm a bit lost on how to replicate the setup


lib/cocoapods/resolver.rb
```diff
@@ -477,6 +483,8 @@ def handle_resolver_error(error)
                   'but they required a higher minimum deployment target.'
               end
             end
+            # require 'pry'
```
dnkoutso: debug code that can be removed :D

dnkoutso:
Would need a changelog entry and if you want it in 1.7.x let's point to `1-7-stable` branch instead.

CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Add an error message for :podspec pods not matching the version in Pods and on disk  
  [orta](https://github.com/orta)
```

</p>

orta:
OK, cool, getting there - accidentally dragged in some extra commits

spec/unit/resolver_spec.rb
```diff
@@ -377,6 +377,33 @@ def create_resolver(podfile = @podfile, locked_deps = empty_graph, specs_updated
...
+        config.sandbox.expects(:specification).with('Reachability').returns(spec)
+
+        resolver = create_resolver(podfile, locked_deps)
+        e = lambda { resolver.resolve }.should.raise Informative
+
+        e.message.should.include <<-EOS.strip
+It seems like you've changed the version of the dependency `Reachability`
+and it differs from the version stored in `Pods/Local Podspecs`.\nYou should run `pod update Reachability --no-repo-update` to apply
+        EOS
+      end
```
orta: I'm a little surprised I can't get this test working, what I think it does:

- Sets up Reachability as the fixtured version (3.0.0)
- Sets up a lockfile for Reachability which references the same on-disk Podspec (3.0.0)
- Tells the sandbox to actually return a version of (2.9.0)

Which I'm sure is the case for us locally, is there anything obvious I'm missing (the resolve doesn't raise in this case)

alloy:
It’s been a while since I’ve thought about this, but I think it makes sense to think through the entire process, instead of only the current result. Namely, does it make sense for `:podspec` dependency requirements to get locked?

They are _meant_ to be used as specs for ‘stable’ versions (as defined by the project that uses them–i.e. it could be a specific commit that the using project needs), they just live outside of the source code repo or a spec repo. Seeing as they point to stable versions, it _seems_ to make sense for those to be locked, as you normally don’t want different versions across `pod install` invocations.

However, the fact that they are currently being locked isn’t because the resolved podspec doesn’t match one specified in the lockfile (by means of eg a checksum comparison), but only because they are compared to a cached version of the spec that gets stored in `Pods/Local Podspecs`. In practice, this means that wiping out the `Pods` directory (and `Pods/Local Podspecs` in it) or on a clean installation (unless you check the `Pods` directory into SCM), CP will install whatever the spec resolves to.

Finally:
* With _on-disk_ podspecs, the act of updating the file could be considered as the act of updating–and this is actually what some projects out there seem to be doing for their vendored libraries (the React Native project, for instance, makes users specify `:podspec` dependencies in their podfile using [these vendored podspecs](https://github.com/facebook/react-native/tree/master/third-party-podspecs)).
* However, a `:podspec` dependency can also point to the network, in which case the contents being updated might be outside of your control and could pose a risk.

So I think that either:
1. We _always_ lock `:podspec` sources by adding checksum data to the lockfile–and improve the error, as this PR does.
2. We _never_ lock `:podspec` sources to ease the type of use-case that e.g. React Native has–but with the understanding that remote sources could update the contents outside of the user’s control.
3. We _only_ lock `:podspec` sources (as described in 1) when they point to remote sources.

(Initially I was personally leaning towards 2, but after typing it all out and thinking through the remote aspect, I’m leaning towards 3 or 1.)

orta:
For 2, we could determine if the podspec is locally pretty easily and act differently if so?

segiddins:
I'm behind (1), imho that's the least-surprising (and least-breaking) change

orta:
Paired with @dnkoutso  and we got it covered 👍

alloy:
@orta Cool. Did you also discuss a roadmap for locking podspecs as per (1)?

dnkoutso:
Any reason not to land this?

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ## Master
 
+* Add an error message for :podspec pods not matching the version in Pods and on disk
```
dnkoutso: @orta this is in the wrong section perhaps when the branch change occurred it got messed up.

lib/cocoapods/resolver.rb
```diff
@@ -441,6 +441,12 @@ def handle_resolver_error(error)
               o << "\nIt seems like you've changed the constraints of dependency `#{name}` " \
               "inside your development pod `#{local_pod_parent.name}`.\nYou should run `pod update #{name}` to apply " \
               "changes you've made."
+            elsif !conflict.possibility && conflict.locked_requirement && conflict.locked_requirement.external_source && conflict.locked_requirement.external_source[:podspec] && 
+                                           conflict.requirement && conflict.requirement.external_source && conflict.requirement.external_source[:podspec]
+              # The internal version of the Podspec doesn't match the external definition of a podspec
+              o << "\nIt seems like you've changed the version of the dependency `#{name}` " \
+              "and it differs from the version stored in `Pods/Local Podspecs`.\nk  You should run `pod update #{name} --no-repo-update` to apply " \
```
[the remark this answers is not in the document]
dnkoutso: +1 lets fix this

spec/unit/resolver_spec.rb
dnkoutso: We took care of it.

orta:
Yeah, I'm unlikely to get the time to add it to the lock file - this should be good as-is

spec/unit/resolver_spec.rb
```diff
@@ -377,6 +377,30 @@ def create_resolver(podfile = @podfile, locked_deps = empty_graph, specs_updated
         EOS
       end
 
+      # TODO: Make this work
```
segiddins: TODO

orta:
Thanks @segiddins - missed that from the rebase, OK this is good 👍

PR #8684, opened 2019-04-05
Specify which versions of Xcode are supported, fixes CocoaPods/cocoap…

fulldecent:
Fixes https://github.com/CocoaPods/cocoapods.org/issues/412

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,8 @@ CocoaPods manages dependencies for your Xcode projects.
 You specify the dependencies for your project in a simple text file: your `Podfile`. 
 CocoaPods recursively resolves dependencies between libraries, fetches 
 source code for all dependencies, and creates and maintains an Xcode 
-workspace to build your project. Xcode versions 10 and 9 are supported.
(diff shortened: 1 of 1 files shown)


dnkoutso:
While this is an OK change I am not sure we want that as we typically "forget" to update this over time and years.

fulldecent:
Good point, updated to more generic wording

README.md
```diff
@@ -11,7 +11,8 @@ CocoaPods manages dependencies for your Xcode projects.
 You specify the dependencies for your project in a simple text file: your `Podfile`. 
 CocoaPods recursively resolves dependencies between libraries, fetches 
 source code for all dependencies, and creates and maintains an Xcode 
-workspace to build your project.
+workspace to build your project. The latest released Xcode versions and the
```
dnkoutso: I would add the word "Typically, ...` atm unfortunately we never guarantee this nor do we test for it.

fulldecent: This word change can be a default commitment. If things are not working as described here then they are a bug which should be reported.

PR #8743, opened 2019-04-28
Fix an issue that prevented deployment targets from being treated as a minimum

### LINKED ISSUE #7314: pods are always installed for iOS 8 deployment target instead of iOS 9
* [🌈] I've read and understood the [*CONTRIBUTING guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

1. Create a new Xcode 9.2 project "example" for iOS Swift
2. Change deployment target to iOS 9.0
3. Use a Podfile with:

```
platform :ios, '9.0'
use_frameworks!

abstract_target 'default_pods' do
    pod 'AFNetworking', '~> 3'
    target 'example'
end
```

4. Run `pod install` or `pod update`

AFNetworking is installed with a Deployment Target of iOS 9.0.

## What happened instead?

AFNetworking is installed with a Deployment Target of iOS 8.0, which causes noticeable warnings like _/Users/coeur/example/Pods/AFNetworking/AFNetworking/AFURLSessionManager.m:163:30: 'setResumingHandler:' is only available on iOS 9.0 or newer_

Even if I manually change the Pods project Deployment Target to iOS 9.0, a simple `pod update` will reset it to iOS 8.0.

## CocoaPods Environment

* CocoaPods 1.4.0.rc.1 (also reproduced with CocoaPods 1.3.1)
* Xcode 9.2 & High Sierra

https://github.com/Coeur/CocoaPods7314

amorde:
Closes #7314

With this change, the deployment targets will be bumped to the minimum deployment target of the user targets into which they are installed.

Requires https://github.com/CocoaPods/cocoapods-integration-specs/pull/225

### DIFF
--- a/lib/cocoapods/installer/analyzer.rb
+++ b/lib/cocoapods/installer/analyzer.rb
@@ -733,7 +733,9 @@ module Pod
           archs = target_requires_64_bit ? ['$(ARCHS_STANDARD_64_BIT)'] : []
         end
         host_requires_frameworks = target_definitions.any?(&:uses_frameworks?)
-        platform = determine_platform(specs, target_definitions, host_requires_frameworks)
+        min_deployment_target = target_inspections.map { |ti| ti.platform.deployment_target }.min
+        platform_name = target_definitions.first.platform.name
+        platform = Analyzer.determine_platform(specs, platform_name, host_requires_frameworks, min_deployment_target)
         file_accessors = create_file_accessors(specs, platform)
         PodTarget.new(sandbox, host_requires_frameworks, user_build_configurations, archs, platform, specs,
                       target_definitions, file_accessors, scope_suffix, :build_type => build_type)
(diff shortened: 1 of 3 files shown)


lib/cocoapods/installer/analyzer.rb
```diff
@@ -1023,6 +999,37 @@ def requires_64_bit_archs?(platform, object_version)
             false
           end
         end
+
+        # Calculates and returns the platform to use for the given list of specs and target definitions.
```
amorde: diff looks weird because this was moved into a `class << self` block

CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Fix an issue that prevented deployment targets from being treated as a minimum  
```

</p>

segiddins:
I don’t think we want this — changing the deployment target a pod is built with can break its compilation, since different depredations will be in effect

amorde:
Couldn’t the same be said for _not_ changing it? If pods have compile time checks for the SDK availability, keeping their deployment targets lower will make them continue to use older APIs.

This change also fits the documentation of the deployment target DSL which describes it as a minimum, not an absolute

amorde:
Thought about this some more and I still think it's the right way to go. Here's an example where I think we are doing the _wrong thing_.

Given the following:

* A Podfile with a target that has a deployment target of 11.0, and one dependency `MyPod`

```ruby
target 'MyApp' do
  platform :ios, '11.0'
  pod 'MyPod'
end
```

* `MyPod` is a pod that supports iOS 10.0 and higher, via the following Podspec DSL:

```ruby
Pod::Specification.new |s|
  s.name = 'MyPod'
  s.ios.deployment_target = '10.0'
  // ...
end
```
* `MyPod` contains a file that conditionally references an API that was introduced in iOS 11.0

```swift
import UIKit

extension UIView {	
    public var topPadding: CGFloat {
        if #available(iOS 11.0, *) {
            return safeAreaInsets.top
        } else {
            return 0
        }
    }
}
```

Currently CocoaPods sets the deployment target of `MyPod` to `'10.0'`, not `'11.0'`. This means that, in this example, `topPadding` will always return 0 even though the app it is integrated into has a deployment target of 11.0. While this is a simple example, this could have significant impact on the behavior of the app at runtime.

To get around this, Pod authors have to bump their deployment targets (dropping support for older versions), or Pod consumers have to manually override the deployment target using a post_install hook or something similar.

icecrystal23:
You assertion that 

> in this example, topPadding will always return 0

is false.  `#available` is a runtime check.  `MyPod` with deployment target 10 can support iOS 10 and higher.  That's the whole point of the `#available` feature.

amorde:
While normally #available works that way a sample project I setup was behaving as described - not sure there's another issue that is causing it, but I'll upload it here when I'm back to my personal laptop.

Regardless I might close this PR either way

amorde:
Thinking about this some more I think I agree with @icecrystal23 and @segiddins on this. I'm going to close this

amorde:
The reason my project was misbehaving - I forgot the iOS simulators don't return `safeAreaInsets` properly even on iPhone X/Xs/Xr simulators 🤦‍♂apologies for the confusion there.

PR #8748, opened 2019-04-29
Skip the "install_framework" step for "Debug" configuration

icecrystal23:
The "install_framework" step is not necessary for running Debug builds.  Skipping this script in "Debug" configuration saves us a lot of time when running Debug builds, as we have a lot of frameworks included.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 * Add support for pods in abstract-only targets to be installed.  
   [Samuel Giddins](https://github.com/segiddins)
+  
+  * Skip the "install_framework" step for "Debug" configuration  
+  [icecrystal23](https://github.com/icecrystal23)
 
 ##### Bug Fixes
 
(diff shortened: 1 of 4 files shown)


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Skip the "install_framework" step for "Debug" configuration  
  [icecrystal23](https://github.com/icecrystal23)
```

</p>

dnkoutso:
>The "install_framework" step is not necessary for running Debug builds.

I don't think that is true, why do you think that is the case?

icecrystal23:
> > The "install_framework" step is not necessary for running Debug builds.
> 
> I don't think that is true, why do you think that is the case?

Because there is nothing inherently wrong with fat frameworks.  They work fine on both devices and simulators.  Apple complains about them when you upload (they don't want you shipping simulator code), which is why it is necessary to strip out the simulator code before uploading a release build.

dnkoutso:
If pods are built as frameworks they need to be copied over to the app bundle so they can be loaded. I do not see how that works if we dont copy them.

icecrystal23:
I thought the script was only doing the stripping; now I see it is actually doing the copying too.  I thought the stripping was what was taking so long, but it only saves 1 second off our builds to skip stripping (takes 9 seconds to do the copy and 10 if we copy and strip 😞).  Oh well.

Issue #8608, opened 2019-03-14
App with multiple nested targets doesn't propagate nested dependencies properly to app


lilyball:
* [X] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

When upgrading from CocoaPods 1.5.3 to 1.6.1, dependencies from nested targets aren't propagating to the app correctly anymore. In my testing it appears that if the app has multiple nested targets, it only propagates dependencies from one of them. CocoaPods 1.5.3 correctly propagated dependencies from all of them.

## What did you do?

`pod install` after upgrading CocoaPods to 1.6.1.

## What did you expect to happen?

Dependencies should be set up properly and my app should compile.

## What happened instead?

My host app target lost access to dependencies declared by its embedded framework target.

For example, in the sample project below, upgrading from CocoaPods 1.5.3 to 1.6.1 produces the following xcconfig diff:

```diff
diff --git a/Pods/Target Support Files/Pods-DepTest/Pods-DepTest.debug.xcconfig b/Pods/Target Support Files/Pods-DepTest/Pods-DepTest.debug.xcconfig
index e803c35..a59d17d 100644
--- a/Pods/Target Support Files/Pods-DepTest/Pods-DepTest.debug.xcconfig	
+++ b/Pods/Target Support Files/Pods-DepTest/Pods-DepTest.debug.xcconfig	
@@ -1,11 +1,9 @@
-ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = YES
 CODE_SIGN_IDENTITY =
-FRAMEWORK_SEARCH_PATHS = $(inherited) "${PODS_CONFIGURATION_BUILD_DIR}/PMHTTP" "${PODS_CONFIGURATION_BUILD_DIR}/PMJSON" "${PODS_CONFIGURATION_BUILD_DIR}/libcmark_gfm"
+FRAMEWORK_SEARCH_PATHS = $(inherited) "${PODS_CONFIGURATION_BUILD_DIR}/libcmark_gfm"
 GCC_PREPROCESSOR_DEFINITIONS = $(inherited) COCOAPODS=1
+HEADER_SEARCH_PATHS = $(inherited) "${PODS_CONFIGURATION_BUILD_DIR}/libcmark_gfm/libcmark_gfm.framework/Headers"
[...]

PR #8758, opened 2019-05-03
Ensure all embedded pod targets are copied over to the host target.

dnkoutso:
closes https://github.com/CocoaPods/CocoaPods/issues/8608

### DIFF
--- a/lib/cocoapods/installer/analyzer.rb
+++ b/lib/cocoapods/installer/analyzer.rb
@@ -303,7 +303,7 @@ module Pod
                 pod_target.name
               end
             end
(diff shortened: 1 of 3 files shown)


lib/cocoapods/installer/analyzer.rb
```diff
@@ -303,7 +303,7 @@ def embedded_target_pod_targets_by_host(aggregate_target, embedded_aggregate_tar
                 pod_target.name
               end
             end
-            embedded_pod_targets_by_build_config[configuration_name] = embedded_pod_targets
+            embedded_pod_targets_by_build_config[configuration_name] += embedded_pod_targets
```
dnkoutso: all of these targets were overridden by the last one we iterated on.

Issue #8649, opened 2019-03-29
pod lib lint fails for Swift-only vendored frameworks


yanniks:
* [x] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

Executed `pod lib lint` on a Podspec that only contains a reference to a vendored framework, no source files. `swift_version` is specified.

## What did you expect to happen?

A dummy project with Swift files should be created, linting should succeed. Since I specified a `swift_version` I expected that the tests run against a Swift project.

## What happened instead?

The created dummy project contains an Objective-C `main.m` file. The build fails since Xcode doesn't link to Swift libraries on Objective-C only projects.

Errors:
```
- NOTE  | [iOS] xcodebuild:  ld: warning: Could not find auto-linked library 'swiftFoundation'
    - NOTE  | [iOS] xcodebuild:  ld: warning: Could not find auto-linked library 'swiftMetal'
    - NOTE  | [iOS] xcodebuild:  ld: warning: Could not find auto-linked library 'swiftDarwin'
    - NOTE  | [iOS] xcodebuild:  ld: warning: Could not find auto-linked library 'swiftUIKit'
    - NOTE  | [iOS] xcodebuild:  ld: warning: Could not find auto-linked library 'swiftObjectiveC'
    - NOTE  | [iOS] xcodebuild:  ld: warning: Could not find auto-linked library 'swiftCoreFoundation'
    - NOTE  | [iOS] xcodebuild:  ld: warning: 
    - NOTE  | [iOS] xcodebuild:  ld: warning: Could not find auto-linked library 'swiftCoreImage'
    - NOTE  | [iOS] xcodebuild:  ld: warning: Could not find auto-linked library 'swiftQuartzCore'
    - NOTE  | [iOS] xcodebuild:  ld: warning: Could not find auto-linked library 'swiftCoreGraphics'
    - NOTE  | [iOS] xcodebuild:  ld: warning: Could not find auto-linked library 'swiftCore'
```

The issue is resolved as soon as the project contains one Swift file.

### Stack

```
   CocoaPods : 1.6.1
        Ruby : ruby 2.3.7p456 (2018-03-28 revision 63024) [universal.x86_64-darwin18]
    RubyGems : 2.5.2.3
        Host : Mac OS X 10.14.4 (18E226)
       Xcode : 10.2 (10E125)
[...]

PR #8762, opened 2019-05-04
Correctly detect that a prebuilt pod uses Swift.

dnkoutso:
closes https://github.com/CocoaPods/CocoaPods/issues/8649


dnkoutso:
Actually the thing I do not like about this change is that `s.swift_versions` DSL is currently a root attribute. This means a pod target will be considered as if the are using Swift across all of its subspecs whereas some subspecs or tests specs or app specs might not.

dnkoutso:
Will move this to 1.8.0 for now as we need to expand the DSL to be better here.

CHANGELOG.md
```diff
@@ -20,6 +20,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
   [#8815](https://github.com/CocoaPods/CocoaPods/pull/8815)
 
+* Correctly detect that a prebuilt pod uses Swift.
```
dnkoutso: needs a rebase.

dnkoutso:
Closing for now as it might not be the direction we want to go.

PR #8706, opened 2019-04-14
Fix a bug that caused the headers dir for pods containing a custom module map to be deleted

amorde:
Closes #8695 

I initially had a workaround in `SandboxDirCleaner` to make sure the directory remained, but it makes more sense for this to be reflected in the value of `headers_sandbox`

For a pod with a different module name, a module map is symlinked to `Pods/Public/[module_name]/[module_name].modulemap`. However, since `headers_sandbox` is set to `Pods/Public/[pod_name]`, this directory was not kept when cleaning the sandbox.

The point at which the directory would be marked as one to keep is in the `else` block here: https://github.com/CocoaPods/CocoaPods/blob/6bd7b5bb2bafe4d78f38bda09b81f7a6bffc407e/lib/cocoapods/installer/sandbox_dir_cleaner.rb#L50-L54

PR #8766, opened 2019-05-04
Fix 1.7.0.rc.1 static library regression for header_dir attribute

### LINKED ISSUE #8765: 1.7.0.rc1 regression on static library build HEADER_SEARCH_PATHS
#8706 breaks the static library builds of `FirebaseDatabase` and `FirebaseFirestore`. `FirebaseDatabase` fails with 

```
Firebase/Database/third_party/Wrap-leveldb/APLevelDB.mm:54:9: fatal error: 'leveldb/db.h' file not found
#import "leveldb/db.h"
```
because the public headers from the leveldb-library pod are no longer in `HEADER_SEARCH_PATHS`:

```
diff -r good-Pods/Target Support Files/FirebaseDatabase/FirebaseDatabase.xcconfig Pods/Target Support Files/FirebaseDatabase/FirebaseDatabase.xcconfig
4c4
< HEADER_SEARCH_PATHS = $(inherited) "${PODS_ROOT}/Headers/Private" "${PODS_ROOT}/Headers/Private/FirebaseDatabase" "${PODS_ROOT}/Headers/Public" "${PODS_ROOT}/Headers/Public/FirebaseAuthInterop" "${PODS_ROOT}/Headers/Public/FirebaseCore" "${PODS_ROOT}/Headers/Public/FirebaseDatabase" "${PODS_ROOT}/Headers/Public/GoogleUtilities" "${PODS_ROOT}/Headers/Public/leveldb-library"
---
> HEADER_SEARCH_PATHS = $(inherited) "${PODS_ROOT}/Headers/Private" "${PODS_ROOT}/Headers/Private/FirebaseDatabase" "${PODS_ROOT}/Headers/Public" "${PODS_ROOT}/Headers/Public/FirebaseAuthInterop" "${PODS_ROOT}/Headers/Public/FirebaseCore" "${PODS_ROOT}/Headers/Public/FirebaseDatabase" "${PODS_ROOT}/Headers/Public/GoogleUtilities"
```

To reproduce, put pod `FirebaseDatabase` in a Podfile without `use_frameworks!`, pod install, and build.

paulb777:
Fix #8765

#8706 introduced a bug building pods with the `header_dir` attribute as static libraries. This is a minimal fix for that issue.

### DIFF
--- a/lib/cocoapods/target/pod_target.rb
+++ b/lib/cocoapods/target/pod_target.rb
@@ -150,7 +150,11 @@ module Pod
     # @return [Pathname] the pathname for headers in the sandbox.
     #
     def headers_sandbox
-      Pathname.new(product_module_name)
+      if root_spec.attributes_hash.key?('header_dir')
+        Pathname.new(pod_name)
+      else
+        Pathname.new(product_module_name)
+      end
     end
 
     # @return [Hash{FileAccessor => Hash}] Hash of file accessors by header mappings.
--- a/lib/cocoapods/target/pod_target.rb
+++ b/lib/cocoapods/target/pod_target.rb
@@ -150,7 +150,7 @@ module Pod
     # @return [Pathname] the pathname for headers in the sandbox.
     #
     def headers_sandbox
-      if root_spec.attributes_hash.key?('header_dir')
+      if root_spec.consumer(platform).header_dir
         Pathname.new(pod_name)
       else
         Pathname.new(product_module_name)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,10 +12,13 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
+* Fix 1.7.0.rc.1 static library regression for pods with `header_dir` attribute  
+  [Paul Beusterien](https://github.com/paulb777)
+  [#8765](https://github.com/CocoaPods/CocoaPods/issues/8765)
+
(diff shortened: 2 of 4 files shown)


lib/cocoapods/target/pod_target.rb
```diff
@@ -150,7 +150,11 @@ def all_files
     # @return [Pathname] the pathname for headers in the sandbox.
     #
     def headers_sandbox
-      Pathname.new(product_module_name)
+      if root_spec.attributes_hash.key?('header_dir')
```
amorde: we might want something a bit more robust for this in the future (maybe `root_spec.consumer(platform).header_dir`?) but this should work for now

dnkoutso: yes can we use `consumer` here?

paulb777: My first attempt to use `consumer` generated a bunch of test failures. I'll make another attempt to work though them later on today.

paulb777: Why is using `consumer` more robust?

amorde: My thought was it would remove the reliance on knowing the structure of attributes_hash, but I can see how that’s not really a problem in practice

paulb777: I suppose it's at least a consistent style to use `consumer`. I made the update.

dnkoutso:
consumer will also take care of multi-platform and automatic default wrapper etc.

PR #8783, opened 2019-05-08
Use CDN specs repo in CI examples

amorde:
This should reduce CircleCI build times by ~2-4 minutes


amorde:
This seems to make the `pod install` step for each example actually take much longer.. I'm guessing its because it's refreshing the same specs every time?

Might close this, but want to dig into it a bit more

amorde:
Rebased and switched to cdn.cocoapods.org

PR #8794, opened 2019-05-12
Add --analyze option for the linters

### LINKED ISSUE #8792: [Feature Request] Add --analyze option to `pod lib lint`
Add a `--analyze` option to `pod lib lint` to include static analysis when linting. 

Some ideas for implementation at https://stackoverflow.com/a/39635929

paulb777:
Close #8792

### DIFF
--- a/lib/cocoapods/validator.rb
+++ b/lib/cocoapods/validator.rb
@@ -239,6 +239,10 @@ module Pod
     #
     attr_accessor :skip_tests
 
+    # @return [Bool] Whether the validator should run Xcode Static Analysis.
+    #
+    attr_accessor :analyze
+
     # @return [Bool] Whether frameworks should be used for the installation.
     #
     attr_accessor :use_frameworks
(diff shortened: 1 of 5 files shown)


segiddins:
This seems like a new feature, it probably shouldn't be targeting the stable branch?

lib/cocoapods/validator.rb
```diff
@@ -662,7 +666,19 @@ def build_pod
           if scheme.nil?
             UI.warn "Skipping compilation with `xcodebuild` because target contains no sources.\n".yellow
           else
-            output = xcodebuild('build', scheme, 'Release')
+            if analyze
+              output = xcodebuild('analyze', scheme, 'Release')
+              find_output = Executable.execute_command('find', [validation_dir, '-name', '*.html'], false)
```
segiddins: Is this a reliable way to determine whether analysis failed? Is there a reason why the `xcodebuild` call won't itself fail?

paulb777: Unfortunately not, the `xcodebuild` succeeds if the build succeeds even if there are `analyze` issues. See the discussion at https://stackoverflow.com/a/39635929.

paulb777:
Yep, it's a new feature. It seems like a safe, isolated addition that could provide value sooner in 1.7.0 with low risk.  I'm not sure of the exact policy and wanted to discuss if it could go in. I'll retarget to master if it's too late for 1.7.0.

lib/cocoapods/validator.rb
dnkoutso: should we move this into an `analyze_pod` method for consistency with `build_pod` and `test_pod`?

dnkoutso:
Given we are converging on 1.7.0 I recommend we push that to master. Seems less risky I know but no reason to push it imo.

lib/cocoapods/validator.rb
paulb777: I'm not sure, since analyze is mostly an option to build_pod.

paulb777:
Rebased to master

lib/cocoapods/validator.rb
dnkoutso: Gotcha makes sense

paulb777:
I didn't see anything close to what would be needed in the tests for other lint options, so I likely would need some pointers.

dnkoutso:
See `it 'runs xcodebuild with correct arguments when skipping import validation'`

it prints out the command line arguments expected.

PR #8804, opened 2019-05-14
Update target validator specs to point to the right project path.

dnkoutso:
No logical changes in this PR. Just updating the tests to be more correct as I have found they were not in another PR of mine.

No changelog entry needed and targeting master.


amorde:
I ran into this as some point too, thanks for fixing it 👍

PR #8809, opened 2019-05-15
Update Core dependency

dnkoutso:
~I expect the first PR to fail due to integration specs.~ It did. Updated.


dnkoutso:
Integration specs https://github.com/CocoaPods/cocoapods-integration-specs/pull/231

PR #8782, opened 2019-05-08
Don't add resources to a 2nd pod target build phase

paulb777:
Fix #8173

test_spec resources are already copied in `Copy Bundle Resources`. Don't copy them in `Copy Pods Resources`.

Issue #8173, opened 2018-10-10
Resource file is added to bundle and copied in build phase causing duplicates


mozeryansky:
In my podspec I have:
```
...
s.test_spec 'Tests' do |test_spec|
    test_spec.source_files = 'Tests/Sources/**/*'
    test_spec.resources = 'Tests/Resources/**/*'
end
... 
```

This causes the image I have in the Resources folder to be added to the 'Copy Bundle Resources' build phase, and CocoaPods adds it to the '[CP] Copy Pods Resources' phase. When building this duplicates the file.

Xcode produces this warning:
`
Showing All Messages
:-1: ignoring duplicated output file: '/Users/michael/Library/Developer/Xcode/DerivedData/MyWorkspace-aeopwxywrxoflbeustfgkzsnmojk/Build/Products/Debug-iphonesimulator/MyTarget.xctest/background.jpg' (in target 'MyTarget')`

### Stack

```
   CocoaPods : 1.5.3
        Ruby : ruby 2.3.7p456 (2018-03-28 revision 63024) [universal.x86_64-darwin17]
    RubyGems : 2.5.2.3
        Host : Mac OS X 10.13.6 (17G65)
       Xcode : 10.0 (10A255)
         Git : git version 2.17.1 (Apple Git-112)
Ruby lib dir : /System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib
Repositories : calthings - git@github.com:Calthings/CTPodSpecs.git @ e0244915947e6543c1de4116c6ad4fb72538fd3b
               master - https://github.com/CocoaPods/Specs.git @ 7ea1295dbee73cd26cc6f685fa0d4d3a78384a7a
```

### Installation Source

```
Executable Path: /usr/local/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.2
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
cocoapods-trunk       : 1.3.1
cocoapods-try         : 1.1.0
```

dnkoutso:
Can you please upload a sample app?

mozeryansky:
From building the sample project I've narrowed down the issue. It only happens when I have a custom scheme which builds the test target and the app in the same build. In the attached project there is a scheme called "All" which generates the issue.

[Issue8173.zip](https://github.com/CocoaPods/CocoaPods/files/2466312/Issue8173.zip)

[...]

PR #8812, opened 2019-05-16
1.7.0.rc.2 regression - library test spec resources

paulb777:
Resources need to be added for test specs in library builds since libraries are excluded from the resource copy part of `add_files_to_build_phases`.

This fixes a regression introduced in #8782


dnkoutso:
And the fix for https://github.com/CocoaPods/CocoaPods/issues/8173 remains, correct?

paulb777:
Correct. This is a better fix for #8173

PR #8834, opened 2019-05-22
Update LICENSE

amorde:
Time continues to pass ⏳


orta:
We should ditch the 2nd part of that timespan, and add some of the folks currently maintaining

orta:
Skip it entirely, it declares when the license was started ( see how it's not in the official template https://choosealicense.com/licenses/mit/ )

orta:
👍 Yep, thanks!

Issue #8695, opened 2019-04-11
CocoaPods 1.7.0beta3 regression with lottie-ios modulemap when using static linking


chrisballinger:
* [x] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

Update our Gemfile to point from CocoaPods 1.6.1 to 1.7.0beta3.

My guess is that this regression is related to lottie's use of custom module names conflicting with how CocoaPods handles Swift static library module maps.

Btw the issue also affects both Lottie 2.5.3 (Obj-C) and Lottie 3 rewrite.

## What did you expect to happen?

All dependencies previously compile-able are still compiled.

## What happened instead?

lottie-ios can no longer be compiled.

<unknown>:0: error: module map file `.../app/Pods/Headers/Public/Lottie/lottie-ios.modulemap' not found
<unknown>:0: error: underlying Objective-C module 'Lottie' not found

### Stack

```
   CocoaPods : 1.7.0.beta.3
        Ruby : ruby 2.6.1p33 (2019-01-30 revision 66950) [x86_64-darwin18]
    RubyGems : 3.0.2
        Host : Mac OS X 10.14.4 (18E226)
       Xcode : 10.2 (10E125)
         Git : git version 2.21.0
Ruby lib dir : /usr/local/Cellar/ruby/2.6.1/lib
Repositories : master - https://github.com/CocoaPods/Specs.git @ 4db5f7fd60ba774deb000ec442f861594a50b460
```

### Installation Source

```
Executable Path: /usr/local/lib/ruby/gems/2.6.0/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.4
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.1.0
cocoapods-trunk       : 1.3.1
cocoapods-try         : 1.1.0
```

### Podfile

```ruby
platform :ios, '11.0'
inhibit_all_warnings!

target 'Regression' do
  pod 'lottie-ios'
end
```

## Project that demonstrates the issue

~I will try to make an isolated project test case soon.~ There is an isolated test case project demonstrating the issue here: https://github.com/chrisballinger/cocoapods-issue-8695

segiddins:
Does this happen even without the post-install hook?

chrisballinger:
[...]

PR #8837, opened 2019-05-23
Fix a regression in static library header dir paths

### LINKED ISSUE #8836: Header path changed in 1.7.0

# Report

## What did you do?

Run `pod install` and built my project with Xcode. That project does work properly when using CocoaPods 1.6.1.

My project to build properly when using CocoaPods 1.7.0.

## What happened instead?

Xcode complained it could not find headers with a hyphen `-` in the pod name.

Looking at files in the `Pods/` directory, it seems pod names in `Pods/Headers/Public/` and `Pods/Headers/Private/` now have hyphens `-` replaced by underscores `_` (though they still use hyphens in `Pods/`).

I had to change for example:
```objc
#import <pantryman-ios/pantryman-ios.h>
```
to: (changing the hyphen in the directory name to an underscore, but keeping the hyphen in the file name)
```objc
#import <pantryman_ios/pantryman-ios.h>
```
for my project to continue building.

That is a breaking change that doesn't seem to be mentioned in the change log. Was that intended?

### Stack

```
   CocoaPods : 1.7.0
        Ruby : ruby 2.5.5p157 (2019-03-15 revision 67260) [x86_64-darwin18]
    RubyGems : 3.0.3
        Host : Mac OS X 10.14.5 (18F132)
       Xcode : 10.2.1 (10E1001)
         Git : git version 2.21.0
```

### Plugins

```
claide-plugins        : 0.9.2
cocoapods-deintegrate : 1.0.4
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.1.0
cocoapods-trunk       : 1.3.1
cocoapods-try         : 1.1.0
```

The project's Podfile doesn't `use_frameworks!`.

amorde:
Closes #8836

Original implementation in #8706 was a bit heavy handed - we can get the same bug fix by adjusting the behavior of the sandbox cleaner.

### DIFF
--- a/spec/unit/installer/sandbox_dir_cleaner_spec.rb
+++ b/spec/unit/installer/sandbox_dir_cleaner_spec.rb
@@ -4,13 +4,17 @@ module Pod
   describe Installer::SandboxDirCleaner do
     before do
       @pod_target = fixture_pod_target('banana-lib/BananaLib.podspec')
+      spec = fixture_spec('coconut-lib/CoconutLib.podspec')
+      spec.module_name = 'CoconutLibModule'
+      @other_pod_target = fixture_pod_target_with_specs([spec])
       @aggregate_target = AggregateTarget.new(config.sandbox, false, {}, [], Platform.ios,
                                               fixture_target_definition('MyApp'), config.sandbox.root.dirname, nil,
                                               nil, {})
-      @cleaner = Installer::SandboxDirCleaner.new(config.sandbox, [@pod_target], [@aggregate_target])
+      @cleaner = Installer::SandboxDirCleaner.new(config.sandbox, [@pod_target, @other_pod_target], [@aggregate_target])
       @project = Project.new(config.sandbox.project_path)
       @sandbox = config.sandbox
       FileUtils.mkdir_p(@sandbox.target_support_files_dir(@pod_target.name))
+      FileUtils.mkdir_p(@sandbox.target_support_files_dir(@other_pod_target.name))
       FileUtils.mkdir_p(@sandbox.target_support_files_dir(@aggregate_target.name))
     end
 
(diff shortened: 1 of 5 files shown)


dnkoutso:
@amorde this ensures both https://github.com/CocoaPods/CocoaPods/issues/8836 and work https://github.com/CocoaPods/CocoaPods/issues/8695?

spec/unit/installer/sandbox_dir_cleaner_spec.rb
```diff
@@ -32,15 +36,25 @@ module Pod
...
       FileUtils.mkdir_p(banana_lib_public_headers)
       FileUtils.mkdir_p(banana_lib_private_headers)
+      FileUtils.mkdir_p(coconut_lib_public_headers)
+      FileUtils.mkdir_p(coconut_lib_private_headers)
       @cleaner.expects(:remove_dir).with(random_pod_private_headers)
       @cleaner.expects(:remove_dir).with(random_pod_public_headers)
+      @cleaner.expects(:remove_dir).with(coconut_lib_public_headers).never
+      @cleaner.expects(:remove_dir).with(coconut_lib_private_headers)
       @cleaner.clean!
-      banana_lib_public_headers.should.exist?
```
dnkoutso: what about these assertions?

amorde:
Have tested against #8695 but not #8836 yet, will do that before merging

spec/unit/installer/sandbox_dir_cleaner_spec.rb
amorde: since `:remove_dir` is stubbed above, these assertions will always pass. they aren't testing anything

amorde:
Confirmed #8836 is fixed with this PR

Issue #8683, opened 2019-04-05
Enhancement: Add an `install!` parameter that would prevent generation of Pods.xcodeproj


dfed:
* [x] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?
Ran `pod install`

## What did you expect to happen?

Download our dependencies distributed via CocoaPods.

## What happened instead?
```
Analyzing dependencies
Downloading dependencies
Using AFNetworking (3.2.1)
Generating Pods project
Skipping User Project Integration
Sending stats
Pod installation complete! There is 1 dependency from the Podfile and 1 total pod installed.
```
In addition to downloading our dependencies distributed via CocoaPods, a `Pods/Pods.xcodeproj` file directory with a bunch of files were written to disk. Since I'm using another project-generation tool to pull in sources downloaded via Cocoapods, it'd be great not have to create `Pods/Pods.xcodeproj`. I understand that modifying the behavior of `integrate_targets: false` to stop producing these files would be a breaking change – it would be great to get a new parameter on `install!` that could prevent these support files from being created.

### Stack

```
   CocoaPods : 1.6.1
        Ruby : ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-darwin17]
    RubyGems : 3.0.2
        Host : Mac OS X 10.14.4 (18E226)
       Xcode : 10.2 (10E125)
         Git : git version 2.21.0
Ruby lib dir : /Users/dan_federman/.rbenv/versions/2.3.1/lib
Repositories : master - https://github.com/CocoaPods/Specs.git @ d1274e7ae102fca8339dae1ec8fa9a6b1331e149
```

### Installation Source

```
Executable Path: /Users/dan_federman/.rbenv/versions/2.3.1/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.2
cocoapods-plugins     : 1.0.0
cocoapods-rome        : 1.0.1
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
cocoapods-trunk       : 1.3.1
cocoapods-try         : 1.1.0
```

### Podfile

```ruby
source 'https://github.com/CocoaPods/Specs.git'

platform :ios, '11.0'

# Only download the files, don't create Xcode projects
install! 'cocoapods', integrate_targets: false

use_frameworks!

[...]

PR #8872, opened 2019-06-03
Added option to only download Dependencies

Itaybre:
This new option `just_download` skips the Pods.xcodeproj generation

**How to use**
In the Podfile add:
```
install! 'cocoapods', just_download: true
```

Fix: #8683

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
   [#8983](https://github.com/CocoaPods/CocoaPods/pull/8983)
 
+* Added option to skip Pods.xcodeproj generation
+  [Itay Brenner](https://github.com/itaybre)
+  [8872](https://github.com/CocoaPods/CocoaPods/pull/8872)
+
 ##### Bug Fixes
 
 * Make CDNSource show up in `pod repo env`  
@@ -118,7 +122,7 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
-* None.  
+* None.
 
 ##### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,7 +68,7 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
   [#8983](https://github.com/CocoaPods/CocoaPods/pull/8983)
 
-* Added option to skip Pods.xcodeproj generation
+* Added option to skip Pods.xcodeproj generation  
   [Itay Brenner](https://github.com/itaybre)
   [8872](https://github.com/CocoaPods/CocoaPods/pull/8872)
 
@@ -122,7 +122,7 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
-* None.
+* None.  
 
 ##### Bug Fixes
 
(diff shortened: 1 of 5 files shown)


dnkoutso:
You can use:
```ruby
      # Whether to integrate the installed pods into the user project
      #
      # If set to false, Pods will be downloaded and installed to the `Pods/` directory
      # but they will not be integrated into your project.
      #
      option :integrate_targets, true
```

Set it to false in your `Podfile`. I do not think we want to introduce this change.

Itaybre:
@dnkoutso I saw this issue #8683 and thought that it could be useful. Also it was labeled as as help_wanted, so I just made it.

The difference with `integrate_targets` is that this avoid the Generate Pod Project as it is unnecessary and if you have lots of dependencies that could take some time to finish .

As said in the Issue, I did not want to modify `integrate_targets` behavior since it would be a breaking change.

dnkoutso:
I would name this flag something along the lines `:skip_pods_project_generation` instead of `:just_download` which does not seem very descriptive.

Itaybre:
Makes sense, I will change the flag name

Itaybre:
@dnkoutso Done

lib/cocoapods/installer.rb
```diff
@@ -156,13 +156,27 @@ def install!
...
+      if installation_options.skip_pods_project_generation?
+        show_skip_pods_project_generation_message
+      else
+        regular_intergration
+      end
+      perform_post_install_actions
+    end
+
+    def show_skip_pods_project_generation_message
+      UI.section 'Just Download option enabled'
```
dnkoutso: please remove this message as it does not appear useful.

lib/cocoapods/installer.rb
```diff
@@ -156,13 +156,27 @@ def install!
...
+      UI.section 'Skipping Pods Project Creation'
+      UI.section 'Skipping User Project Integration'
+    end
+
+    def regular_intergration
```
dnkoutso: rename to `integrate` there is no `irregular` integration per se

lib/cocoapods/installer/installation_options.rb
```diff
@@ -181,6 +181,13 @@ def hash
       # since the previous installation.
       #
       option :incremental_installation, false
+
+      # Whether to create a `Pods.xcodeproj` or not
```
dnkoutso: missing period at the end of this.

CHANGELOG.md
```diff
@@ -8,7 +8,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
-* None.  
+* Added option to skip Pods.xcodeproj generation
```
dnkoutso: your branch needs to be rebased to latest master and this to be placed in the right spot on the `master` section (above 1.7.5 release)

lib/cocoapods/installer/installation_options.rb
dnkoutso: actually just add this as comment to match the above style of comments. This documentation will appear in the guides website.

```ruby
      # Whether to skip generating the `Pods.xcodeproj` and perform only dependency resolution and downloading.
      #
```

dnkoutso:
Almost there! Can you please also squash into a single commit? thanks!

Itaybre:
@dnkoutso there, I forgot to rename the methods in the unit testing, should be fine now

CHANGELOG.md
dnkoutso: nit please add two empty spaces after the word `generation` for markdown formatting.

CHANGELOG.md
```diff
@@ -118,7 +122,7 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
-* None.
```
dnkoutso: please do not remove this is intentional.

dfed:
Thanks folks!

Issue #8879, opened 2019-06-06
[1.7.1] Pod using custom modulemap and integrate with `modular_headers` will cause the `export *` export all Public and Private headers, where `use_frameworks!` does not


dreampiggy:
* [x] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

1. Write a CoocaPods's Pod, which have a `modulemap`, and have some `private_header_files`. Both of them as valid Podspec syntax.
2. Integrate the Pod, using `modular_headers`, it can be done using `use_modular_headers!` or specify only for this Pod
3. Run `pod install`
4. Check the generate Xcode project. The modulemap file is placed under `Pods/Headers/Private`
5. This cause the warning for clang. Because the modulemap use `export *` syntax, which will mark all `Pods/Headers/Private/*.h` as public headers.

Note: This issue does not appear, when don't use `modular_headers`, or `use_frameworks!`.

## What did you expect to happen?

When using custom modulemap, the modulemap file should be placed into `Pods/Headers/Public`, just as the default CocoaPods generated one, but not the `Pods/Headers/Private`.

## What happened instead?

When using custom modulemap, the modulemap file should be placed into `Pods/Headers/Private`, cause the `export *` syntax mismatch all the headers as public.

User who have private header files, should manualy strip the private headers, even they already declare them in `Podspec.private_header_files` syntax.

```
// Lib.modulemap

exclude header "Private1.h"
exclude header "Private2.h"

export *
module * { export * }
```

### Stack

```
   CocoaPods : 1.7.1
        Ruby : ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-darwin18]
    RubyGems : 2.7.9
        Host : Mac OS X 10.14.4 (18E226)
       Xcode : 10.2.1 (10E1001)
         Git : git version 2.20.1 (Apple Git-117)
Ruby lib dir : /Users/lizhuoli/.rvm/rubies/ruby-2.5.1/lib
Repositories : master - https://github.com/CocoaPods/Specs.git @ b287af13e912623835bbfdf04ce18b0892a058b0
```

### Installation Source

```
Executable Path: /Users/lizhuoli/.rvm/gems/ruby-2.5.1/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.4
cocoapods-plugins     : 1.0.0
[...]

PR #8888, opened 2019-06-07
Change the `module_map_path`, to make sure the custom modulemap have the same behavior between use_frameworks and modular_headers

dreampiggy:
See the issue's background and reason for chancing this:

#8879 

And our Pod who use custom modulemap, which face the issue because of this:

https://github.com/SDWebImage/SDWebImage/pull/2749

### DIFF
--- a/lib/cocoapods/target/pod_target.rb
+++ b/lib/cocoapods/target/pod_target.rb
@@ -427,9 +427,9 @@ module Pod
       basename = "#{label}.modulemap"
       if build_as_framework?
         super
-      elsif file_accessors.any?(&:module_map)
-        build_headers.root + product_module_name + basename
       else
+        # modulemap should be placed at the same directory with all Public Headers.
+        # To let `export *` works and no Private Headers is exported outside.
         sandbox.public_headers.root + product_module_name + basename
       end
     end


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Change the `module_map_path`, to make sure the custom modulemap have the same behavior between use_frameworks and modular_headers  
  [dreampiggy](https://github.com/dreampiggy)
```

</p>

lib/cocoapods/target/pod_target.rb
```diff
@@ -427,9 +427,9 @@ def module_map_path
       basename = "#{label}.modulemap"
       if build_as_framework?
         super
-      elsif file_accessors.any?(&:module_map)
```
dnkoutso: this is when a custom module map is present by the podspec? can you not place it into the folder you expect it to be placed?

segiddins:
This breaks explicitly adding `header "foo.h"` declarations when `foo.h` is a private header

dnkoutso:
ping @dreampiggy if you want to continue this PR and discussion.

dreampiggy:
@segiddins @dnkoutso

I'm now understand the case.

> The `framework` can use modulemap to reference the `PrivateHeaders` folder's header files. So our `use_modular_headers`, which still build Static Library + Headers which should obey the same rule.

But actually, my issue is that:

> The `framework` modulemap don't export the `PrivateHeaders` header in module, does not trigger clang's warning of `-W-incomplete-umbrella`.

> However, using the same modulemap, the `use_modular_headers` don't export the Private Headers (they are not in `xxx.framework/PrivateHeaders` now), do trigger clang's warning of `-W-incomplete-umbrella`

This does not sounds correct. Why this cause warning ? They use the same modulemap, same source files and headers, the only differnt is How module is structured (via a `.framework` folder, or using `.a` + headers folder).

If there are other ways can solve my problem, I think it's OK.

dreampiggy:
Maybe you can have a check of my issue #8879, to know my use case.

I'm a Pod Author and I want to support all the integration ways (via Static Library/Dynamic Library, via Static Framework/Dyanmic Framework, via Clang Module or No Module), without any extra warnings. Is this impossible ? Or any better solution ?

PR #8890, opened 2019-06-07
[PodTargetInstaller] Set Default Module for Storyboards in resource bundle targets

jtreanor:
🌈 If a storyboard or xib is included in a pod via `resource_bundles` and the bundle does not have the same name as the pod, the app can crash with an error like this:

```
2019-06-07 17:28:34.730781+0100 ResourceBundleIssue[78655:8535426] Unknown class _TtC20WordPressUIResources24FancyAlertViewController in Interface Builder file.
Could not cast value of type 'UIViewController' (0x110579a10) to 'WordPressUI.FancyAlertViewController' (0x103b95f38).
2019-06-07 17:28:34.732146+0100 ResourceBundleIssue[78655:8535426] Could not cast value of type 'UIViewController' (0x110579a10) to 'WordPressUI.FancyAlertViewController' (0x103b95f38).
```

This is because by default storyboards have this option checked for custom classes:

image

Currently, the workaround is either to name the resource bundle the same as the pod or to uncheck that box and set an explicit module for every object in the Storyboard.

This PR makes using Storyboards from resource bundle's easier by setting the "Default Module" build setting for resource bundle targets to the module of the pod itself. Here is the setting in Xcode:

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   unless repo update was specifically requested.
   [Artem Sheremet](https://github.com/dotdoom)
 
+* Set Default Module for Storyboards in resource bundle targets.  
+   [James Treanor](https://github.com/jtreanor)
+   [#8890](https://github.com/CocoaPods/CocoaPods/pull/8890)
+
 ##### Bug Fixes
 
 * Fix set `cache_root` from config file error  
--- a/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
@@ -503,6 +503,10 @@ module Pod
                     configuration.build_settings['CONFIGURATION_BUILD_DIR'] = target.configuration_build_dir('$(BUILD_DIR)/$(CONFIGURATION)$(EFFECTIVE_PLATFORM_NAME)')
                   end
 
+                  # Set the 'IBSC_MODULE' build settings for resource bundles so that Storyboards and Xibs can load
+                  # classes from the parent module.
+                  configuration.build_settings['IBSC_MODULE'] = target.product_module_name
+
                   # Set the `SWIFT_VERSION` build setting for resource bundles that could have resources that get
                   # compiled such as an `xcdatamodeld` file which has 'Swift' as its code generation language.
                   if contains_compile_phase_refs && file_accessors.any? { |fa| target.uses_swift_for_spec?(fa.spec) }
--- a/spec/unit/installer/xcode/pods_project_generator/pod_target_installer_spec.rb
+++ b/spec/unit/installer/xcode/pods_project_generator/pod_target_installer_spec.rb
@@ -857,6 +857,14 @@ module Pod
                   end
                 end
 
+                it 'sets the correct default module for storyboards in resource bundle targets' do
+                  @installer.install!
+                  @bundle_target = @project.targets.find { |t| t.name == 'BananaLib-Pods-SampleProject-banana_bundle' }
+                  @bundle_target.build_configurations.each do |bc|
+                    bc.build_settings['IBSC_MODULE'].should == 'BananaLib'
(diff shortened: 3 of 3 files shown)


CHANGELOG.md
```diff
@@ -32,6 +32,8 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   unless repo update was specifically requested.
   [Artem Sheremet](https://github.com/dotdoom)
 
+* Set Default Module for Storyboards in resource bundle targets. [James Treanor](https://github.com/jtreanor)
```
dnkoutso: please format this properly. Include two spaces after "bundle targets.  "

dnkoutso: also include an issue or PR number.

dnkoutso:
I do not see this setting in Xcode 10.x where do you see it?

LGTM either way, is this for all platforms or a subset?

CHANGELOG.md
jtreanor: 👍 I think that's correct now.

jtreanor:
@dnkoutso I see the build setting in Xcode 10.2.1. However, it only appears for targets with type "Bundle".

The setting appears in Xcode for iOS only as far as I can see. Should I exclude macOS targets or is it fine as is?

I have updated the failing integration tests here: https://github.com/CocoaPods/cocoapods-integration-specs/pull/236

dnkoutso:
If the setting appears in iOS bundles only lets scope it for that just to be safe for side effects.

jtreanor:
The platform change is done now.

CHANGELOG.md
```diff
@@ -32,6 +32,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   unless repo update was specifically requested.
   [Artem Sheremet](https://github.com/dotdoom)
 
+* Set Default Module for Storyboards in resource bundle targets.
```
dnkoutso: please add two empty spaces at the end of this line for formatting.

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -503,6 +503,12 @@ def add_resources_bundle_targets(file_accessors)
                     configuration.build_settings['CONFIGURATION_BUILD_DIR'] = target.configuration_build_dir('$(BUILD_DIR)/$(CONFIGURATION)$(EFFECTIVE_PLATFORM_NAME)')
                   end
 
+                  # Set the 'IBSC_MODULE' build settings for iOS resource bundles so that Storyboards and Xibs can load
+                  # classes from the parent module. The setting does not apply to macOS.
+                  unless target.platform == :osx
```
dnkoutso: does this appear on :watchos and :tvOS? if not lets switch that to 
```ruby
  configuration.build_settings['IBSC_MODULE'] = target.product_module_name if target.platform == :ios
```

dnkoutso:
This will go to 1.8.0 release.

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
jtreanor: Just to be fully sure I have checked all platforms and it turns out I was wrong here. The setting is present for all platforms (iOS, macOS, tvOS, watchOS), but it only appears in Xcode if there is a xib for storyboard in the target.

That's why I mistakingly thought it was not present on macOS. I will revert the platform check now.

CHANGELOG.md
jtreanor: Sorry about that. Its done now.

jtreanor:
Okay, I think this is ready now. As mentioned above, I've updated the failing integration tests in https://github.com/CocoaPods/cocoapods-integration-specs/pull/236

dnkoutso:
LG

dnkoutso:
Merged the integration specs PR, if you can bump the submodule here it should go green.

jtreanor:
Thanks! That's updated now.

dnkoutso:
Thanks!

Issue #8764, opened 2019-05-04
Linting Does Not Support Dependencies with Multiple Swift Versions


cnoon:
* [X] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

I ran into an issue when using `pod lib lint` and `pod spec lint` on frameworks across Xcode 10.1 and 10.2. We have a library ([Willow](https://github.com/Nike-Inc/Willow)) that supports both Swift 4.2 and 5.0 in [Willow 5.2.1](https://github.com/Nike-Inc/Willow/blob/5.2.1/Willow.podspec). We then have a second library that also supports Swift 4.2 and 5.0 called Surf. When linting with Xcode 10.2, everything works as expected. When linting with Xcode 10.1, Willow can never be compiled because CocoaPods is always setting the Swift version to `5.0` even though it isn't supported on Xcode 10.1.

```bash
cnoon:~/Programming/Troubleshooting/CocoaPods/LintingPods$ pod lib lint Surf.podspec

 -> Surf (1.0.0)
    - ERROR | [iOS] xcodebuild: Returned an unsuccessful exit code. You can use `--verbose` for more information.
    - NOTE  | xcodebuild:  note: Using new build system
    - NOTE  | [iOS] xcodebuild:  note: Planning build
    - NOTE  | [iOS] xcodebuild:  note: Constructing build description
    - NOTE  | [iOS] xcodebuild:  error: SWIFT_VERSION '5.0' is unsupported, supported versions are: 3.0, 4.0, 4.2. (in target 'App')
    - NOTE  | [iOS] xcodebuild:  error: SWIFT_VERSION '5.0' is unsupported, supported versions are: 3.0, 4.0, 4.2. (in target 'Willow')
    - NOTE  | [iOS] xcodebuild:  error: SWIFT_VERSION '5.0' is unsupported, supported versions are: 3.0, 4.0, 4.2. (in target 'Surf')
    - NOTE  | [iOS] xcodebuild:  warning: Skipping code signing because the target does not have an Info.plist file. (in target 'App')

[!] Surf did not pass validation, due to 1 error.
You can use the `--no-clean` option to inspect any issue.
```

By adding `--swift-version=4.2`, I can get Surf to compile, but Willow will still be set to a SWIFT_VERSION of 5.0.

```bash
[...]

PR #8897, opened 2019-06-12
Attempt to use Swift version for dependencies during lint.

dnkoutso:
closes https://github.com/CocoaPods/CocoaPods/issues/8764

During `lint`, pod authors can pass the Swift version their pod should be compiled with. However, this setting does not apply to dependencies which is fine generally. There is also no currently a way to define the `supports_swift_version` DSL since the target is completely generated and we have no way to determine the constraints unless they get passed in somehow.

This change will effectively _try_ to set the Swift version for dependencies _only if_ the original pod author has declared that they support it which should be safe to do so, otherwise we fall back to what we always did.

/cc @cnoon

### DIFF
--- a/lib/cocoapods/target/pod_target.rb
+++ b/lib/cocoapods/target/pod_target.rb
@@ -186,7 +186,7 @@ module Pod
     def swift_version
       @swift_version ||= begin
         if spec_swift_versions.empty?
-          target_definitions.map(&:swift_version).compact.uniq.first
+          target_definition_swift_version
         else
           spec_swift_versions.sort.reverse_each.find do |swift_version|
             target_definitions.all? do |td|
@@ -195,6 +195,13 @@ module Pod
           end.to_s
         end
       end
+    end
+
+    # @return [String] the Swift version derived from the target definitions that integrate this pod. This is used for
+    #         legacy reasons and only if the pod author has not specified the Swift versions their pod supports.
+    #
+    def target_definition_swift_version
+      target_definitions.map(&:swift_version).compact.uniq.first
     end
 
     # @return [Array<Version>] the Swift versions supported. Might be empty if the author has not
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
   [#8880](https://github.com/CocoaPods/CocoaPods/issues/8880)
 
+* Attempt to use Swift version for dependencies during lint.  
+  [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
+  [#8764](https://github.com/CocoaPods/CocoaPods/issues/8764)
+
 
 ## 1.7.1 (2019-05-30)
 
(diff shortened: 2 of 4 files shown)


lib/cocoapods/target/pod_target.rb
```diff
@@ -186,7 +186,7 @@ def public_header_mappings_by_file_accessor
     def swift_version
       @swift_version ||= begin
         if spec_swift_versions.empty?
-          target_definitions.map(&:swift_version).compact.uniq.first
+          target_definition_swift_version
```
dnkoutso: just extracted method, also makes it much more clear.

sprynmr:
It seems like this still won't allow dependencies to compile with a different version than the main target if their `swift_version` is different. For example, a podspec for a pod that compiles only in swift 4.2, and a dependency that only compiles in swift 4.0. It still tries to compile the dependency in 4.2.

dnkoutso:
yes this was a best effort. It only works if the version is explicitly added in the list of swift versions inside the podspec.

sprynmr:
Just for clarity, if my pod only works with swift 4.2 and has a dependency that only works with swift 4.0, I can't get it to lint properly right? It will always try and compile the dependency with 4.0?

dnkoutso:
Has the pod author of your dependency specified `s.swift_version = '4.0'` (or with 1.7 `s.swift_versions = '4.0'`)? If so then that version will _always_ be honored.

If the pod author of your dependency *doesn't* specify it then the Swift version chosen will be the default one generated by the Xcodeproj project if I recall correctly.

sprynmr:
We are the pod author of the dependencies in question (and could simply go through the process of updating them to be consistent). We have used the singular version in the dependency podspec `s.swift_version = '4.0'`. It for some reason isn't being honored with 1.7.2. I'll double check everything on our side. Not honoring would be considered a bug correct?

dnkoutso:
most probably a bug but would love an issue with clear steps to repro and perhaps a sample.

sprynmr:
Appreciate the rapid feedback! I will check all my bits to make sure I'm not missing something, and file a bug if it is indeed one.

Coeur:
> Just for clarity, if my pod only works with swift 4.2 and has a dependency that only works with swift 4.0, I can't get it to lint properly right? It will always try and compile the dependency with 4.0?

Year 2019 and it can't build with Swift 4.2? Time to fork it and forget about the original.

sprynmr:
> > Just for clarity, if my pod only works with swift 4.2 and has a dependency that only works with swift 4.0, I can't get it to lint properly right? It will always try and compile the dependency with 4.0?
> 
> Year 2019 and it can't build with Swift 4.2? Time to fork it and forget about the original.

In all reality it's not really an issue for us to just get all the dependencies up to 4.2. They are all internal / private spec that we can easily update. It was just unexpected behavior, so I started poking at it.

dnkoutso:
@Coeur no need. Things can move slower in internal environments or overall open source.

Coeur:
Sadly, I know. That reminds me of that question from two weeks ago, where the author was basically asking: ["_We're using CocoaPods 1.4.0 and we're not allowed to update it, how to deal with Swift 5?_"](https://stackoverflow.com/q/56464531).

PR #8911, opened 2019-06-16
add CDN repo caching (for more time saving)

igor-makarov:
I'll add an empty commit to this to show how this works, then will remove it.


igor-makarov:
Now that I think about it, the download size for these examples is so negligible, it doesn't really save any time. Never mind! 😂

PR #8826, opened 2019-05-20
Clean stale local podspec

parski:
**Bug**:

When installing a Pod from a local source a podspec appears in `${PODS_ROOT}/Local Podspecs/`.

Example:
```ruby
pod 'Alamofire', :path => '../Alamofire'
```
After a `pod install` leaves you with `<App>/Pods/Local Podspecs/Alamofire.podspec.json`.

Now when you install a non-local version of the same pod the local podspec remains, stale as a driftwood log.

PR #8914, opened 2019-06-17
Clean stale local podspec

dnkoutso:
Supersedes https://github.com/CocoaPods/CocoaPods/pull/8826

### DIFF
--- a/lib/cocoapods/installer.rb
+++ b/lib/cocoapods/installer.rb
@@ -514,7 +514,7 @@ module Pod
         raise StandardError, message
       end
 
-      pod_installer = PodSourceInstaller.new(sandbox, specs_by_platform, :can_cache => installation_options.clean?)
+      pod_installer = PodSourceInstaller.new(sandbox, podfile, specs_by_platform, :can_cache => installation_options.clean?)
       pod_installers << pod_installer
       pod_installer
     end
--- a/lib/cocoapods/installer/pod_source_installer.rb
+++ b/lib/cocoapods/installer/pod_source_installer.rb
@@ -14,6 +14,11 @@ module Pod
       #
       attr_reader :sandbox
 
+      # @return [Podfile] the podfile that should be integrated with the user
+      #         projects.
+      #
+      attr_reader :podfile
+
       # @return [Hash{Symbol=>Array}] The specifications that need to be
       #         installed grouped by platform.
       #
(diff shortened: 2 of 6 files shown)


amorde:
Looks like this needs a `if File.exist?(path)`

dnkoutso:
Not sure...I think it deletes more than its intending to...like pods with external sources such as `:podspec` etc.

lib/cocoapods/installer/pod_source_installer.rb
```diff
@@ -61,6 +61,7 @@ def name
       def install!
         download_source unless predownloaded? || local?
         PodSourcePreparer.new(root_spec, root).prepare! if local?
+        sandbox.remove_local_podspec(name) unless local?
```
segiddins: `unless predownloaded? || local?` ?

dnkoutso:
Problem remains when using `:podspec` in the `Podfile`. These pods are never stored anywhere https://github.com/CocoaPods/CocoaPods/blob/master/lib/cocoapods/external_sources/podspec_source.rb#L9 therefore we erroneously think the podspec should be deleted.

dnkoutso:
Alright pushed update. I think its OK now and verified it works.

spec/unit/installer/pod_source_installer_spec.rb
```diff
@@ -28,6 +29,12 @@ module Pod
           pod_folder = config.sandbox.pod_dir('BananaLib')
           pod_folder.should.exist
         end
+
+        it 'removes local podspec' do
```
dnkoutso: this test is wrong and always passes.

dnkoutso: actually no not yet but still updating this test.

lib/cocoapods/installer/pod_source_installer.rb
```diff
@@ -191,6 +199,14 @@ def local?
         sandbox.local?(root_spec.name)
       end
 
+      # @return [Boolean] whether the pod uses an external source (e.g. :podspec) in the
+      #         resolution process to retrieve its podspec.
+      #
+      def external?
+        @dependencies ||= podfile.dependencies.select(&:external?).map(&:name)
```
segiddins: will this ever actually be different from `released?`? Should the ivar be named `external_dependencies`?

lib/cocoapods/installer.rb
```diff
@@ -514,7 +514,7 @@ def create_pod_installer(pod_name)
         raise StandardError, message
       end
 
-      pod_installer = PodSourceInstaller.new(sandbox, specs_by_platform, :can_cache => installation_options.clean?)
+      pod_installer = PodSourceInstaller.new(sandbox, podfile, specs_by_platform, :can_cache => installation_options.clean?)
```
segiddins: should this be passed in the podfile dependency cache instead of the raw podfile?

lib/cocoapods/installer/pod_source_installer.rb
dnkoutso: oh interesting. I can check and see. I can rename the ivar yes.

lib/cocoapods/installer.rb
dnkoutso: I think there is none at that time. Analyzer constructs one by using the `podfile`:
```ruby
@podfile_dependency_cache = PodfileDependencyCache.from_podfile(podfile)
```

I could do the same but I do not see strong benefit from it as groups dependencies by target definition.

PR #8916, opened 2019-06-18
Print correct platform name when inferring target platform.

dnkoutso:


dnkoutso:
Integration specs https://github.com/CocoaPods/cocoapods-integration-specs/pull/237

PR #8939, opened 2019-06-27
Integrate a pod into a custom project name if specified.

dnkoutso:
Requires https://github.com/CocoaPods/Core/pull/547

### DIFF
--- a/lib/cocoapods/installer/project_cache/project_cache_analyzer.rb
+++ b/lib/cocoapods/installer/project_cache/project_cache_analyzer.rb
@@ -75,6 +75,11 @@ module Pod
             return full_install_results
           end
 
+          if project_names_changed?(pod_targets, cache)
+            UI.message 'Ignoring project cache due to project name changes.'
+            return full_install_results
+          end
+
           pod_targets_to_generate = Set[]
           aggregate_targets_to_generate = Set[]
           added_targets, removed_targets = compute_added_and_removed_targets(target_by_label,
@@ -95,23 +100,22 @@ module Pod
           pod_targets_to_generate.merge(dirty_pod_targets)
           aggregate_targets_to_generate.merge(dirty_aggregate_targets)
 
-          # Since multi xcodeproj will group targets by PodTarget#pod_name into individual projects, we
-          # need to append these "sibling" targets to the list of targets we need to generate before finalizing the total list,
+          # Since multi xcodeproj will group targets by PodTarget#project_name into individual projects, we need to
+          # append these "sibling" targets to the list of targets we need to generate before finalizing the total list,
           # otherwise we will end up with missing targets.
           #
           sibling_pod_targets = compute_sibling_pod_targets(pod_targets, pod_targets_to_generate)
           pod_targets_to_generate.merge(sibling_pod_targets)
 
-          # We either return the full list of aggregate targets or none since the aggregate targets go into the Pods.xcodeproj
-          # and so we need to regenerate all aggregate targets when regenerating Pods.xcodeproj.
+          # We either return the full list of aggregate targets or none since the aggregate targets go into the
+          # Pods.xcodeproj and so we need to regenerate all aggregate targets when regenerating Pods.xcodeproj.
 
-          total_aggregate_targets_to_generate =
-            unless aggregate_targets_to_generate.empty?
-              aggregate_targets
-            end
+          total_aggregate_targets_to_generate = unless aggregate_targets_to_generate.empty?
+                                                  aggregate_targets
+                                                end
 
-          ProjectCacheAnalysisResult.new(pod_targets_to_generate.to_a, total_aggregate_targets_to_generate, cache_key_by_target_label,
-                                         build_configurations, project_object_version)
+          ProjectCacheAnalysisResult.new(pod_targets_to_generate.to_a, total_aggregate_targets_to_generate,
+                                         cache_key_by_target_label, build_configurations, project_object_version)
         end
 
         private
@@ -155,7 +159,7 @@ module Pod
             support_files_dir_exists = File.exist? target.support_files_dir
             xcodeproj_exists = case target
                                when PodTarget
-                                 File.exist? sandbox.pod_target_project_path(target.pod_name)
+                                 File.exist? sandbox.pod_target_project_path(target.project_name)
                                when AggregateTarget
                                  File.exist? sandbox.project_path
                                else
@@ -166,8 +170,15 @@ module Pod
         end
 
         def compute_sibling_pod_targets(pod_targets, pod_targets_to_generate)
-          pod_targets_by_name = pod_targets.group_by(&:pod_name)
-          pod_targets_to_generate.flat_map { |t| pod_targets_by_name[t.pod_name] }
+          pod_targets_by_project_name = pod_targets.group_by(&:project_name)
+          pod_targets_to_generate.flat_map { |t| pod_targets_by_project_name[t.project_name] }
+        end
+
+        def project_names_changed?(pod_targets, cache)
+          pod_targets.any? do |pod_target|
+            next unless (target_cache_key = cache.cache_key_by_target_label[pod_target.label])
+            target_cache_key.project_name != pod_target.project_name
+          end
         end
       end
     end
--- a/lib/cocoapods/installer/sandbox_dir_cleaner.rb
+++ b/lib/cocoapods/installer/sandbox_dir_cleaner.rb
@@ -62,6 +62,14 @@ module Pod
 
           removed_sandbox_private_headers = sandbox_private_headers(pod_targets) - sandbox_private_headers_to_install
           removed_sandbox_private_headers.each { |path| remove_dir(path) }
+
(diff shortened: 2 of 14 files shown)


lib/cocoapods/installer/xcode/target_validator.rb
```diff
@@ -148,6 +149,10 @@ def verify_swift_pods_have_module_dependencies
           raise Informative, 'The following Swift pods cannot yet be integrated '\
                              "as static libraries:\n\n#{error_messages.join("\n\n")}"
         end
+
+        def verify_no_multiple_project_names
+          # TODO
```
dnkoutso: need to integrate this.

dnkoutso:
@sebastianv1 the biggest thing I need your help with is whether the project name should be part of the cache.

dnkoutso:
Also I need to ensure that if the project name changes for a pod then the previous one is removed.

sebastianv1:
@dnkoutso Yeah the project name should be a part of the target cache key. We'll also have to add some logic in the project cache analyzer to regenerate both the project it used to be a part of (since it was removed), in addition to the new project it is a part of.

lib/cocoapods/installer/xcode/target_validator.rb
sebastianv1: Does this verify that a single pod doesn't belong to two different projects? One possible configuration is a user wants to group all their iOS pods together into one project and their watch pods together. Do we want to allow or disallow this type of cofig?

dnkoutso: I think it already works but I will verify with a test. PodTarget are split across variants so I think this propagates through everything.

dnkoutso: > doesn't belong to two different projects?

I think you mean if the pod is split across different variants, not so much in the sense of multi-xcodeproj or `project` DSL in `Podfile` correct?

dnkoutso: Confirmed it does not throw because pod targets are already split. This means folks can group their projects by platform or anything else that pod variants support.

lib/cocoapods/installer/project_cache/project_cache_analyzer.rb
```diff
@@ -155,7 +155,7 @@ def compute_dirty_targets(targets)
             support_files_dir_exists = File.exist? target.support_files_dir
             xcodeproj_exists = case target
                                when PodTarget
-                                 File.exist? sandbox.pod_target_project_path(target.pod_name)
+                                 File.exist? sandbox.pod_target_project_path(target.project_name)
```
dnkoutso: @sebastianv1 I think this is correct but it needs a test. I might need your help to write it.

spec/unit/installer/sandbox_dir_cleaner_spec.rb
```diff
@@ -26,18 +27,18 @@ module Pod
...
     it 'cleans up stale headers and keeps pod target headers' do
-      random_pod_private_headers = @pod_target.build_headers.root.join('RandomPod')
+      random_pod_private_headers = @banana_pod_target.build_headers.root.join('RandomPod')
       random_pod_public_headers = @sandbox.public_headers.root.join('PublicPod')
       banana_lib_public_headers = @sandbox.public_headers.root.join('BananaLib')
-      banana_lib_private_headers = @pod_target.build_headers.root.join('BananaLib')
+      banana_lib_private_headers = @banana_pod_target.build_headers.root.join('BananaLib')
       coconut_lib_public_headers = @sandbox.public_headers.root.join('CoconutLibModule')
-      coconut_lib_private_headers = @pod_target.build_headers.root.join('CoconutLibModule')
+      coconut_lib_private_headers = @banana_pod_target.build_headers.root.join('CoconutLibModule')
```
dnkoutso: @sebastianv1 this appears to had been wrong, did you mean to have this as `@other_pod_target` (and in my PR `@coconut_pod_target`)?

dnkoutso: I guess it didnt matter

dnkoutso:
I believe this is actually complete!

sebastianv1:
Inside `compute_sibling_pod_targets` in `project_cache_analyzer.rb` we check for sibling pod targets that need to be regenerated by `pod_name`. This logic probably needs to be updated to check for all the targets that belong to the project that wants to be regenerated and append that set.

dnkoutso:
Great catch!

sebastianv1:
Also with incremental installation, we need to make sure the project it's coming from is being regenerated as well otherwise we would end up with a duplicate state target in the old project.

This is a little tricky since the way we specify which projects to generate are based on the set of pod targets. One hacky way to do this is grabbing the set of pod targets that belong (if any) to the old project and adding that to the list pod targets that need to be regenerated. 😬

Gemfile.lock
```diff
@@ -5,16 +5,6 @@ GIT
   specs:
     claide (1.0.2)
 
-GIT
```
dnkoutso: will revert

Gemfile
```diff
@@ -25,7 +25,8 @@ gem 'json', :git => 'https://github.com/segiddins/json.git', :branch => 'seg-1.7
 
 group :development do
   cp_gem 'claide',                'CLAide'
-  cp_gem 'cocoapods-core',        'Core'
+  # cp_gem 'cocoapods-core',        'Core'
```
dnkoutso: will revert

sebastianv1:
Oh also another thought: should we validate that you can only use `project` DSL with the multi pods project installation option enabled?

dnkoutso:
ah yes! should be fairly easy.

lib/cocoapods/installer/xcode/multi_pods_project_generator.rb
```diff
@@ -13,7 +13,9 @@ def generate!
           # Generate container Pods.xcodeproj.
           container_project = create_container_project(aggregate_targets, sandbox.project_path)
 
-          project_paths_by_pod_targets = pod_targets.group_by { |pod_target| sandbox.pod_target_project_path(pod_target.pod_name) }
+          project_paths_by_pod_targets = pod_targets.group_by do |pod_target|
+            sandbox.pod_target_project_path(pod_target.project_name)
+          end
```
sebastianv1: Ah we run into another fun little problem with split projects: `def pod_name_from_grouping` relies upon `pod_name` to determine if the pod is local or not. This gates the behavior of putting the subproject into the `Development Pods` or `Pods` group. My opinion is just to put it in the `Pods` group and only put it in the `Development Pods` group if all are local. What do you think?

dnkoutso:
Left todo:

- [ ] Do not validate project names if multi xcodeproj option is OFF
- [ ] Test for case in which a project name change should cause a full reinstallation (can be improved further in 1.8.1 or 1.9.x)
- [ ] Handle case in which a dev pod project name and a non dev pod project name is shared.

lib/cocoapods/installer/xcode/target_validator.rb
```diff
@@ -36,6 +38,7 @@ def validate!
           verify_no_static_framework_transitive_dependencies
           verify_swift_pods_swift_version
           verify_swift_pods_have_module_dependencies
+          verify_no_multiple_project_names if installation_options.generate_multiple_pod_projects?
```
dnkoutso: @sebastianv1 this is now handled.

dnkoutso:
@sebastianv1 completed "Test for case in which a project name change should cause a full reinstallation (can be improved further in 1.8.1 or 1.9.x)" and added a test for it.

dnkoutso:
@sebastianv1 it seems that if if a project name is specified for a local pod and the same one for a remote pod then the alphabetic order of which we install and treat pod targets is used to group them.

This means a local pod that meant to be under `Development Pods` actually ends up in `Pods` but compilation and all paths still work properly.

I am inclined to leave this behavior as the same happens for source files (although that is between two local pods).

lib/cocoapods/installer/xcode/multi_pods_project_generator.rb
dnkoutso: Answered below.

lib/cocoapods/installer/sandbox_dir_cleaner.rb
```diff
@@ -62,6 +62,12 @@ def clean!
 
           removed_sandbox_private_headers = sandbox_private_headers(pod_targets) - sandbox_private_headers_to_install
           removed_sandbox_private_headers.each { |path| remove_dir(path) }
+
+          project_dir_names_to_install = pod_targets.map { |pod_target| sandbox.root + "#{pod_target.project_name}.xcodeproj" }
```
amorde: should this be `sandbox.pod_target_project_path(pod_target.name)`?

dnkoutso: ah yeah nice.

dnkoutso: I would still use `sandbox.pod_target_project_path(pod_target.project_name)` though.

Issue #8940, opened 2019-06-27
pod lint doesn't respect swift_version of dependencies


sprynmr:
* [x] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

Ran `pod lib lint mypod`

## What did you expect to happen?

Dependencies with different versions of swift to compile with their respective versions and pass linting.

Per discussion here: https://github.com/CocoaPods/CocoaPods/pull/8897#issuecomment-506399057

## What happened instead?

Linting failed as it was using the swift version from the main Podspec to compile the dependencies

### Stack

```
   CocoaPods : 1.7.2
        Ruby : ruby 2.3.7p456 (2018-03-28 revision 63024) [universal.x86_64-darwin18]
    RubyGems : 2.5.2.3
        Host : Mac OS X 10.14.5 (18F203)
       Xcode : 10.2.1 (10E1001)
         Git : git version 2.20.1 (Apple Git-117)
Ruby lib dir : /System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib
Repositories : master - https://github.com/CocoaPods/Specs.git @ a551a2a5f25a47ddf704f7937f8c617dd119f56b
          
```

### Installation Source

```
Executable Path: /usr/local/bin/pod
```

### Plugins

```
claide-plugins                   : 0.9.2
cocoapods-deintegrate            : 1.0.4
cocoapods-plugins                : 1.0.0
cocoapods-search                 : 1.0.0
cocoapods-static-swift-framework : 0.4
cocoapods-stats                  : 1.1.0
cocoapods-trunk                  : 1.3.1
cocoapods-try                    : 1.1.0
```

## Project that demonstrates the issue

https://twstaticfiles.s3.amazonaws.com/specbug.zip

dnkoutso:
@sprynmr that is a confirmed bug. Will ship in 1.7.3.

The PR #8897 actually regressed it. Thanks again for filing.

dnkoutso:
@sprynmr have you tried CocoaPods 1.7.1?

sprynmr:
I haven't, but I will test it now.

dnkoutso:
Fix https://github.com/CocoaPods/CocoaPods/pull/8941

dnkoutso:
[...]

PR #8941, opened 2019-06-27
Honor the Swift version set by a dependency pod during lint.

dnkoutso:
closes https://github.com/CocoaPods/CocoaPods/issues/8940

This was a regression in 1.7.2 in which we never honored the `swift_version` set by the dependency spec and always defaulted to target definition Swift version.

The reason it was not caught in the prior PR it was because the test was changed also and got a false positive.

### DIFF
--- a/lib/cocoapods/validator.rb
+++ b/lib/cocoapods/validator.rb
@@ -612,7 +612,7 @@ module Pod
                             else
                               pod_target.spec_swift_versions.map(&:to_s).find do |v|
                                 v == derived_swift_version
-                              end || pod_target.target_definition_swift_version
+                              end || pod_target.swift_version
                             end
             build_configuration.build_settings['SWIFT_VERSION'] = swift_version
           end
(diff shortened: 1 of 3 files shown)


lib/cocoapods/validator.rb
```diff
@@ -612,7 +612,7 @@ def configure_pod_targets(targets, target_installation_results, deployment_targe
                             else
                               pod_target.spec_swift_versions.map(&:to_s).find do |v|
                                 v == derived_swift_version
-                              end || pod_target.target_definition_swift_version
+                              end || pod_target.swift_version
```
dnkoutso: this will internally check if the pod has a swift version set in its spec otherwise return the one from the target definition.

dnkoutso: and this was the regression from 1.7.1 -> 1.7.2

amorde:
Oh never mind, we don't run lint on the examples

PR #8943, opened 2019-06-27
[Installer] Add 'treat_warnings_as_errors' option

### LINKED ISSUE #8155: Feature Request: Add option to `pod install` to fail if any warnings are emitted
When doing a pod install, I often get warnings like the following
```
Pod installation complete! There are 33 dependencies from the Podfile and 63 total pods installed.

[!] `<PBXResourcesBuildPhase UUID=`2E14AAE21E7C267E00C2D465`>` attempted to initialize an object with an unknown UUID. `1A2F35A41F28EE0100BC1A6D` for attribute: `files`. This can be the result of a merge and  the unknown UUID is being discarded.

[!] `<PBXResourcesBuildPhase UUID=`2E14AAE21E7C267E00C2D465`>` attempted to initialize an object with an unknown UUID. `2F062F561F54828700588292` for attribute: `files`. This can be the result of a merge and  the unknown UUID is being discarded.
```

This is often the result of a bad merge during development. It would be nice if I could make `pod install` fail if warnings like this come up, so my dev teams can correct them immediately.

ezraberch:
This PR adds a treat_warnings_as_errors installer option, false by default. When enabled warnings will cause the installer to immediately terminate, just as errors do.

Closes #8155

### DIFF
--- a/lib/cocoapods/user_interface.rb
+++ b/lib/cocoapods/user_interface.rb
@@ -12,6 +12,7 @@ module Pod
     @title_level       =  0
     @indentation_level =  2
     @treat_titles_as_messages = false
+    @treat_warnings_as_errors = false
     @warnings = []
 
     class << self
@@ -20,6 +21,7 @@ module Pod
       attr_accessor :indentation_level
       attr_accessor :title_level
       attr_accessor :warnings
+      attr_accessor :treat_warnings_as_errors
 
       # @return [IO] IO object to which UI output will be directed.
       #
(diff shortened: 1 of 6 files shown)


lib/cocoapods/user_interface.rb
```diff
@@ -372,7 +374,11 @@ def gets
       # return [void]
       #
       def warn(message, actions = [], verbose_only = false)
-        warnings << { :message => message, :actions => actions, :verbose_only => verbose_only }
+        if treat_warnings_as_errors
+          raise Informative, message
```
dnkoutso: this might not be the right place to add this. This is meant to add all warnings and display them in the end.

I would push that to the `installer.rb` instead.

ezraberch: The problem with that approach is that the warnings which may oe encountered during installation aren't just in installer.rb. The warning in the originating issue, for example, is in [Xcodeproj/lib/xcodeproj/project/object.rb](https://github.com/CocoaPods/Xcodeproj/blob/master/lib/xcodeproj/project/object.rb).

It would be possible to check afterward in installer.rb whether any warnings have triggered, but that would violate the expected behavior that errors cause an immediate halt in execution.

PR #8950, opened 2019-06-29
Use `git rev-parse` to build CocoaPods with the correct Core branch

igor-makarov:
This solves an issue where, for example, the `1-7-stable` branch of `cocoapods` was running release scripts with the `master` branch of `cocoapods-core`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,7 @@ jobs:
             }
             trap bundle_env ERR
 
-            bundle check || bundle install --path .bundle
+            bundle install --jobs=3 --retry=3 --deployment --path .bundle
       - save_cache:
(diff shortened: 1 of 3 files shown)


igor-makarov:
I managed to get it to build on Circle and on Travis, but the usability for the maintainers might be less than perfect. This might be a yak shave.

dnkoutso:
Could we possibly just make it part of Rainforest release process once a branch is cut to ensure some (I doubt for all) critical dependencies are updated? Or detect that a branch was created and update `Gemfile.lock` ?

igor-makarov:
I don't nearly understand Rainforest enough to do it myself. However, I am more convinced now that I'm shaving a yak here.

PR #8974, opened 2019-07-08
Use `HEAD` of Bacon for graceful `Interrupt` handling

igor-makarov:
When running the full spec suite, it is impossible to interrupt it because [Bacon](https://github.com/leahneukirchen/bacon) doesn't handle `Interrupt` gracefully.

This issue was fixed in the Bacon repo, however, no releases of Bacon were made since 2012.

This PR moves to use the `HEAD` version of Bacon so that local testing can be less painful.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -39,7 +39,7 @@ group :development do
 
   gem 'cocoapods-dependencies', '~> 1.0.beta.1'
 
-  gem 'bacon'
+  gem 'bacon', :git => 'https://github.com/leahneukirchen/bacon.git'
   gem 'mocha'
   gem 'mocha-on-bacon'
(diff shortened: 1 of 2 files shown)


dnkoutso:
Its a bit sad development has halted. I am OK with this change but would love a release instead.

igor-makarov:
Yeah, I know it's the better way but 7 years I doubt it'll happen.

BTW why aren't CP using something more common like RSpec?

amorde:
We just have a lot of stuff built on Bacon, and migrating to RSpec would take a lot of time. I looked at it a bit for Core but didn't get super far. RSpec has some features I'd really like to use.

PR #8983, opened 2019-07-10
Do not re-write sandbox files if they have not changed.

dnkoutso:
A fix here is that the `Manifest.lock` file is used as an input file for the `Check Pods Manifest Lock` script phase. Each time `pod install` runs this file is re-written causing the script phase to run again.

### DIFF
--- a/lib/cocoapods/installer.rb
+++ b/lib/cocoapods/installer.rb
@@ -740,13 +740,15 @@ module Pod
       @lockfile = generate_lockfile
 
       UI.message "- Writing Lockfile in #{UI.path config.lockfile_path}" do
+        # No need to invoke Sandbox#update_changed_file here since this logic already handles checking if the
+        # contents of the file are the same.
         @lockfile.write_to_disk(config.lockfile_path)
       end
 
       UI.message "- Writing Manifest in #{UI.path sandbox.manifest_path}" do
-        sandbox.manifest_path.open('w') do |f|
-          f.write config.lockfile_path.read
-        end
+        # No need to invoke Sandbox#update_changed_file here since this logic already handles checking if the
+        # contents of the file are the same.
+        @lockfile.write_to_disk(sandbox.manifest_path)
       end
     end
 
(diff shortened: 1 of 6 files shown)


lib/cocoapods/sandbox.rb
```diff
@@ -444,6 +444,27 @@ def local_podspec(name)
...
+    # @return [void]
+    #
+    def self.update_changed_file(path, contents)
+      if path.exist?
+        content_stream = StringIO.new(contents)
+        identical = File.open(path, 'rb') { |f| FileUtils.compare_stream(f, content_stream) }
+        return if identical
+      end
+      File.open(path, 'w') { |f| f.write(contents) }
+    end
```
amorde: could also use this in `TargetInstallerHelper#update_changed_file`

lib/cocoapods/installer.rb
```diff
@@ -740,13 +740,13 @@ def write_lockfiles
...
+        # No need to invoke Sandbox#update_changed_file here since this logic already handles checking if the
+        # contents of the file are the same.
         @lockfile.write_to_disk(config.lockfile_path)
       end
 
       UI.message "- Writing Manifest in #{UI.path sandbox.manifest_path}" do
-        sandbox.manifest_path.open('w') do |f|
-          f.write config.lockfile_path.read
-        end
+        Sandbox.update_changed_file(sandbox.manifest_path, config.lockfile_path.read)
```
amorde: couldn't this also use `Lockfile#write_to_disk`? like `@lockfile.write_to_disk(sandbox.manifest_path)`

segiddins:
Neat!

dnkoutso:
Been thinking on making this an instance method in `sandbox.rb` like `Sandbox#write_file` that also performs a check to ensure the file is written within the sandbox and not anywhere.

This means we need to pass `sandbox` in a few classes but I might be OK with it.

lib/cocoapods/sandbox.rb
dnkoutso: This is actually a bit different as each generator has its own `save_as` method

dnkoutso:
Decided not to go with this yet it can be improved again in the future.

Issue #8868, opened 2019-05-30
Unit test targets no longer embed transitive dependencies, fail to load at runtime


lilyball:
* [X] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

CocoaPods 1.6 introduced a regression where unit test targets without a host app no longer embed the dependencies of the framework target that they inherit search paths from.

In CocoaPods 1.5.x and earlier, a unit test target with `inherit! :search_paths` and no host app will embed the frameworks whose search paths it inherits. In CocoaPods 1.6.x (I tested 1.6.2) this changed so unit test targets that use `inherit! :search_paths` do not embed any inherited dependencies, which means the test bundle cannot be loaded into the test runner and the tests fail.

The default `Podfile` created with `pod init` uses `inherit! :search_paths` for these unit test targets, which tells me that this is the expected configuration.

I've tested this on both CocoaPods 1.6.2 and CocoaPods 1.7.1.

## Project that demonstrates the issue

[PodInherit.zip](https://github.com/CocoaPods/CocoaPods/files/3238661/PodInherit.zip)

This sample project was created by making  new iOS Framework w/Unit Tests project in Xcode, running `pod init` (with 1.7.1), and adding the single `pod 'PMJSON'` line to the `Podfile`. I then added some code to the project to actually use the dependency.

Running `pod install` with CocoaPods 1.5.3 properly adds the `[CP] Embed Pods Frameworks` build phase to the unit test target, but running `pod install` with CocoaPods 1.6.x or later eliminates that build phase entirely (assuming the unit test target has no direct dependencies of its own).

dnkoutso:
looks like a bug yes. Not sure if its intentional or not. Need someone to spend time with it.

dnkoutso:
Regressed here which was meant to fix another regression https://github.com/CocoaPods/CocoaPods/pull/7992

segiddins:
I think you need to not use `inherit! :search_paths` if you're not linking the target thats being inherited from

dnkoutso:
Actually that might be true ^

lilyball:
But I am linking it. This is a unit test target, it's absolutely linking the library it's testing.

segiddins:
Maybe I wasn’t understanding the scenario, I’ll try to take another look

lilyball:
[...]

PR #8984, opened 2019-07-10
Fix inheriting search paths for test targets in `init` command.

dnkoutso:
closes https://github.com/CocoaPods/CocoaPods/issues/8868

If a unit test target is part of a target that does not support embedding then the test target itself must embed the frameworks in order for it to run successfully. Otherwise, in common cases the app target will embed the frameworks and the unit test target loads them from there.

For UI test targets we _always_ embed because the test bundle gets wrapped in a _different_ app and not the app that is being tested.

### DIFF
--- a/lib/cocoapods/command/init.rb
+++ b/lib/cocoapods/command/init.rb
@@ -48,7 +48,7 @@ module Pod
       private
 
       # @param  [Xcodeproj::Project] project
-      #         The xcode project to generate a podfile for.
+      #         The Xcode project to generate a podfile for.
       #
       # @return [String] the text of the Podfile for the provided project
       #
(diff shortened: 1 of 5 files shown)


lib/cocoapods/command/init.rb
```diff
@@ -75,38 +73,32 @@ def podfile_template(project)
...
-      # @param  [[Xcodeproj::PBXTarget]] targets
-      #         An array which always has a target as its first item
-      #         and may optionally contain related test targets
+      # @return [String] the text for the target module.
       #
-      # @return [String] the text for the target module
-      #
-      def target_module(app, tests)
-        target_module = "\ntarget '#{app.name.gsub(/'/, "\\\\\'")}' do\n"
+      def target_module(host, tests)
```
amorde: nit: docs for parameters

dnkoutso:
docs added, will merge on green.

PR #9002, opened 2019-07-13
Add support for UI test targets

### LINKED ISSUE #9001: Supporting UITest bundles for test specifications
I tried to add UI test support by adding a `test_spec` which uses an `app_spec` as host app in the following configuration:

```
    s.app_spec "MyApp" do |app|
    end
    
    s.test_spec 'UITests' do |ui_tests|
      ui_tests.requires_app_host = true
      ui_tests.app_host_name = "MyPod/MyApp"
      ui_tests.dependency "MyPod/MyApp"
    end
```

There are two limitations with current integration:
1- CocoaPods generates a test bundle having product type `com.apple.product-type.bundle.unit-test`. For UITest bundles, this should be `com.apple.product-type.bundle.ui-testing`.

2- CocoaPods adds the following build settings to generated test target:
https://github.com/CocoaPods/CocoaPods/blob/714b103d2c68d80b3fa5730995d00cd16f9c1053/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_dependency_installer.rb#L143-L144

However, for a UITest bundle, these build settings should be empty. Rather, Xcode adds the following when you create a UITest bundle and select a particular host app:

```
'TEST_TARGET_NAME'=$APP_NAME   # $APP_NAME is literal name of the host app
```

---

To tweak these limitations, I provided 'TEST_TARGET_NAME' manually as part of `pod_target_xcconfig` for `test_spec` definition. And added the following `post_install` hook to `Podfile`:

```ruby
post_install do |installer|
    installer.generated_projects.each do |project|
            pbxproj = project.root_object
            pbxproj.targets.each do |target|
                if target.symbol_type == :unit_test_bundle
                    if target.name.include?('UITests')
                        target.product_type = 'com.apple.product-type.bundle.ui-testing'
                            config.build_settings.delete('TEST_HOST')
                        end
                    end
                end
            end
    end
end
```

Then I could run UI tests included in `test_spec` successfully.

To add a proper support, the following should be done:

1- Ability to set up test type for a `test_spec`. This should be easy, we can make `test_type` writer public.
2- Set proper build settings according to test type.
3- Set proper product type when creating corresponding xcodeproj.

Am I missing anything else which might be problematic?

manuyavuz:
requires: https://github.com/CocoaPods/Core/pull/562

This PR enables CP to generate a UI test bundle target when a test specification is defined as follows:

```ruby
s.test_spec 'UITests' |tests|
  # we have to set `test_type` to `:ui`, as `:unit` is the default value
  test.test_type = :ui
  test.requires_app_host = true
end
```

Resolves #9001.

### DIFF
--- a/spec/fixtures/watermelon-lib/UITests/WatermelonUITests.m
+++ b/spec/fixtures/watermelon-lib/UITests/WatermelonUITests.m
@@ -1,23 +1,30 @@
-#import <KIF/KIF.h>
+#import <XCTest/XCTest.h>
 
-@interface WatermelonUITests : KIFTestCase
+
+@interface WatermelonUITests : XCTestCase
 
 @end
 
 @implementation WatermelonUITests
 
 - (void)setUp {
-    [super setUp];
     // Put setup code here. This method is called before the invocation of each test method in the class.
+
+    // In UI tests it is usually best to stop immediately when a failure occurs.
+    self.continueAfterFailure = NO;
+
+    // UI tests must launch the application that they test. Doing this in setup will make sure it happens for each test method.
+    [[[XCUIApplication alloc] init] launch];
+
+    // In UI tests it’s important to set the initial state - such as interface orientation - required for your tests before they run. The setUp method is a good place to do this.
 }
 
 - (void)tearDown {
     // Put teardown code here. This method is called after the invocation of each test method in the class.
-    [super tearDown];
 }
 
 - (void)testExample {
-    // This is an example of a functional test case.
+    // Use recording to get started writing UI tests.
     // Use XCTAssert and related functions to verify your tests produce the correct results.
 }
 
--- a/spec/fixtures/watermelon-lib/UITests/WatermelonUITests.m
+++ b/spec/fixtures/watermelon-lib/UITests/WatermelonUITests.m
@@ -1,0 +1,24 @@
+#import <KIF/KIF.h>
+
+@interface WatermelonUITests : KIFTestCase
+
+@end
+
+@implementation WatermelonUITests
+
+- (void)setUp {
+    [super setUp];
+    // Put setup code here. This method is called before the invocation of each test method in the class.
+}
+
+- (void)tearDown {
+    // Put teardown code here. This method is called after the invocation of each test method in the class.
+    [super tearDown];
+}
+
+- (void)testExample {
+    // This is an example of a functional test case.
+    // Use XCTAssert and related functions to verify your tests produce the correct results.
+}
+
+@end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
+* Add support for UI test specs with `test_type` value `:ui`  
+  [Yavuz Nuzumlali](https://github.com/manuyavuz)
+  [#9002](https://github.com/CocoaPods/CocoaPods/pull/9002)
+  [Core#562](https://github.com/CocoaPods/Core/pull/562)
+
 * Replace git-based `MasterSource` with CDN-based `TrunkSource`  
   [Igor Makarov](https://github.com/igor-makarov)
   [#8923](https://github.com/CocoaPods/CocoaPods/pull/8923)
--- a/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_dependency_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_dependency_installer.rb
@@ -137,11 +137,16 @@ module Pod
             app_host_target_names = app_native_target.resolved_build_setting('PRODUCT_NAME', true)
             test_native_target.build_configurations.each do |configuration|
               app_host_target_name = app_host_target_names[configuration.name] || target.name
-              test_host = "$(BUILT_PRODUCTS_DIR)/#{app_host_target_name}.app/"
-              test_host << 'Contents/MacOS/' if pod_target.platform == :osx
-              test_host << app_host_target_name.to_s
-              configuration.build_settings['BUNDLE_LOADER'] = '$(TEST_HOST)'
-              configuration.build_settings['TEST_HOST'] = test_host
+              case test_native_target.symbol_type
+              when :unit_test_bundle
+                test_host = "$(BUILT_PRODUCTS_DIR)/#{app_host_target_name}.app/"
+                test_host << 'Contents/MacOS/' if pod_target.platform == :osx
+                test_host << app_host_target_name.to_s
+                configuration.build_settings['BUNDLE_LOADER'] = '$(TEST_HOST)'
+                configuration.build_settings['TEST_HOST'] = test_host
+              when :ui_test_bundle
+                configuration.build_settings['TEST_TARGET_NAME'] = app_host_target_name
+              end
             end
             target_attributes = project.root_object.attributes['TargetAttributes'] || {}
             target_attributes[test_native_target.uuid.to_s] = { 'TestTargetID' => app_native_target.uuid.to_s }
--- a/lib/cocoapods/target/pod_target.rb
+++ b/lib/cocoapods/target/pod_target.rb
@@ -428,7 +428,13 @@ module Pod
       end
     end
 
-    # @return [String] The label to use for the given test type.
+    # Returns the label to use for the given test type.
+    # This is used to generate native target names for test specs.
+    #
+    # @param  [Symbol] test_type
+    #         The test type to map to provided by the test specification DSL.
+    #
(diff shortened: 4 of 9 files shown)


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Add support for UI test targets  
```

</p>

manuyavuz:
This needs a change in spec fixtures in `WatermelonApp.podspec`.

```ruby
  s.test_spec 'SnapshotTests' do |test_spec|
    test_spec.test_type = :ui     ### This line needed
    test_spec.requires_app_host = true
    test_spec.source_files = 'SnapshotTests/*.{h,m}'
    test_spec.dependency 'iOSSnapshotTestCase/Core'
  end
```

dnkoutso:
Why change the spec and not add another `test_spec` of type UI?

manuyavuz:
I did your way at first, but we would have to change several tests in this block unrelated to this change:
https://github.com/CocoaPods/CocoaPods/blob/8cfd882c36250a1c314b8f940500b7675fc12a89/spec/unit/target/pod_target_spec.rb#L772-L826

If we are ok with that, I can revert.

dnkoutso:
I kinda dont mind that, either update WatermelonLib.podspec to add a test_spec with type `:ui` or add a brand new lib with a single `:ui` `test_spec`

dnkoutso:
Need to add a dummy source file for UI tests test spec.

Maybe a dummy resource as well for other tests to appear meaningful (e.g. for resource paths)

For most of the part this LGTM.

manuyavuz:
@dnkoutso, I don’t like current target naming behavior after adding UI option.

When you don’t provide a name to a test spec, it names spec as `Tests`, and target name as `$PodName-Unit-$SpecName`, which results in sth like `AFNetworking-Unit-Tests` that is good.

For UI type, I changed `Unit` to `UI`, and if there is only on test spec with no name specified, it will have `AFNetworking-UI-Tests` that is again good.

However, if you have multiple test specs one being ui other being unit, one will have a bad naming for sure. You can keep one of them with no name, and it will have a good name, but for other, you have to specify a name different than `Tests` and you lose standard there, and get probably redundancy such as having a target name `AFNetworking-UI-UITests`.

I'm thinking about a good solution for this, but it's hard without breaking change.

1. We can add spec attribute `name` or `target_name`, users can use it to name however they want. But it would create confusion with name of the spec itself if we do `name`, and `target_name` feels too specific as we would use that name for all file names or similar. 
2. We can change naming policy, and if test_spec does not have a name specified, it would be named considering test_type. For example:
```ruby
# This spec will have name `UnitTests`, and target name would be `PodUnitTests` or `Pod-UnitTests`
s.test_spec do |test|
  test.test_type :unit
end

# This spec will have name `UITests`, and target name would be `PodUITests` or `Pod-UITests`
s.test_spec do |test|
  test.test_type :ui
end
```
This option feels good to me, but it would change behavior compared to previous versions. I'm not sure if we are counting this as a breaking change. Their code would still work, but will generate a different target name, which may break their CI systems for example.

3. Force users to explicitly provide a name for test specs, and use provided name to name target name.
This is most flexible option, but it's a breaking change. We can put a warning suggesting change, but still, a breaking change.

What's your thoughts on this? cc @segiddins

dnkoutso:
I like 2. as well but I agree its a breaking change as a CI system might be going off the target name to build.

Can we do two only if multiple test specs across test types are found? Noone should be in this position because there is only one test_type now.

manuyavuz:
We can do it, but that would create inconsistency in the naming if someone starts with one test spec, and after some time later add another one. The naming of their first spec would be changed even though they don't make any changes.

I think 2 is actually not that bad if we warn users of the change during pod install. But we should do warn only if they have an existing target with old naming. And I'm not sure if it's easy now considering new changes such as multi project installation, incremental installation.

dnkoutso:
> We can do it, but that would create inconsistency in the naming if someone starts with one test spec, and after some time later add another one. The naming of their first spec would be changed even though they don't make any changes.

You are correct that is a crappy experience. 

I am hesitant to intro the breaking change, perhaps we can land this PR and continue as a follow up for naming?

pointrlabs:
Sounds good 👍

manuyavuz:
Sounds good.

dnkoutso:
LGTM.

dnkoutso:
We can land this as last enhancement for 1.8.0.

dnkoutso:
@manuyavuz here's a diff to add that fixes some of the failing tests:
```diff
diff --git a/lib/cocoapods/target/pod_target.rb b/lib/cocoapods/target/pod_target.rb
index ef57889d..8e37f348 100644
--- a/lib/cocoapods/target/pod_target.rb
+++ b/lib/cocoapods/target/pod_target.rb
@@ -428,6 +428,19 @@ module Pod
       end
     end
 
+    # @return [String] The label to use for the given test type.
+    #
+    def label_for_test_type(test_type)
+      case test_type
+      when :unit
+        'Unit'
+      when :ui
+        'UI'
+      else
+        raise ArgumentError, "Unknown test type `#{test_type}`."
+      end
+    end
+
     # @return [Specification] The root specification for the target.
     #
     def root_spec
@@ -491,13 +504,7 @@ module Pod
     # @return [String] The derived name of the test target.
     #
     def test_target_label(test_spec)
-      type_string = case test_spec.test_type
-                    when :unit
-                      'Unit'
-                    when :ui
-                      'UI'
-                    end
-      "#{label}-#{type_string}-#{subspec_label(test_spec)}"
+      "#{label}-#{label_for_test_type(test_spec.test_type)}-#{subspec_label(test_spec)}"
     end
 
     # @param  [Specification] app_spec
@@ -520,7 +527,7 @@ module Pod
       if app_spec
         [app_target.name, app_target.app_target_label(app_spec)]
       elsif test_spec.consumer(platform).requires_app_host?
-        [name, "AppHost-#{label}-#{test_spec.test_type.capitalize}-Tests"]
+        [name, "AppHost-#{label}-#{label_for_test_type(test_spec.test_type)}-Tests"]
       end
     end
 
diff --git a/spec/unit/installer/xcode/multi_pods_project_generator_spec.rb b/spec/unit/installer/xcode/multi_pods_project_generator_spec.rb
index cf6dbe20..e5475e26 100644
--- a/spec/unit/installer/xcode/multi_pods_project_generator_spec.rb
+++ b/spec/unit/installer/xcode/multi_pods_project_generator_spec.rb
@@ -5,6 +5,7 @@ module Pod
     class Xcode
       describe MultiPodsProjectGenerator do
         describe 'Generating Multi Pods Project' do
+          #shit
           before do
             @ios_platform = Platform.new(:ios, '6.0')
             @osx_platform = Platform.new(:osx, '10.8')
@@ -268,16 +269,20 @@ module Pod
             ]
 
             watermelon_project.targets.map(&:name).sort.should == [
+              'AppHost-WatermelonLib-iOS-UI-Tests',
               'AppHost-WatermelonLib-iOS-Unit-Tests',
+              'AppHost-WatermelonLib-macOS-UI-Tests',
               'AppHost-WatermelonLib-macOS-Unit-Tests',
               'WatermelonLib-iOS',
               'WatermelonLib-iOS-App',
+              'WatermelonLib-iOS-UI-UITests',
               'WatermelonLib-iOS-Unit-SnapshotTests',
               'WatermelonLib-iOS-Unit-Tests',
               'WatermelonLib-iOS-WatermelonLibExampleAppResources',
               'WatermelonLib-iOS-WatermelonLibTestResources',
               'WatermelonLib-macOS',
               'WatermelonLib-macOS-App',
+              'WatermelonLib-macOS-UI-UITests',
               'WatermelonLib-macOS-Unit-SnapshotTests',
               'WatermelonLib-macOS-Unit-Tests',
               'WatermelonLib-macOS-WatermelonLibExampleAppResources',
```

manuyavuz:
I could just find time working on fixing them :) Thanks!

spec/fixtures/watermelon-lib/UITests/WatermelonUITests.m
```diff
@@ -0,0 +1,24 @@
+#import <KIF/KIF.h>
+
+@interface WatermelonUITests : KIFTestCase
```
dnkoutso: hm why wouldnt those be XCUITestCase? KIF is meant for `:unit` test bundles

dnkoutso: updated.

dnkoutso:
@manuyavuz I will merge on green! Let's follow up on naming although I thought about it a bit more and I am still very hesitant to break existing CI setups.............maybe we just go through this ¯\\\_(ツ)_/¯

PR #9009, opened 2019-07-16
Fix regenerating all aggregate targets if one is removed.

sebastianv1:
This fixes a bug during incremental installation where if an aggregate target is removed, we will not generate `Pods.xcodeproj` since the `AggregateTarget` instance doesn't exist and thus will produce a nil value in an array thereby making it empty. A result of this getting shipped was because the aggregate target test fixtures weren't setup correctly and shared the same target definition in the cache. Verified that the fixed test fails with the old code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
+* Fix regenerating aggregate targets for incremental installation.  
+  [Sebastian Shanus](https://github.com/sebastianv1)
+  [#9009](https://github.com/CocoaPods/CocoaPods/pull/9009)
+
 * Fix heuristic for determining whether the source URL to be added is CDN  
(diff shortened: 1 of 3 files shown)


sebastianv1:
Damn, every time

dnkoutso:
Also remove master specs repo submodule entry `spec/fixtures/spec-repos/master`

The master specs repo is removed on `master` (1.8.x release).

dnkoutso:
Just rubocop 

```
Offenses:
lib/cocoapods/installer/project_cache/project_cache_analyzer.rb:111:56: C: Don't use parentheses around the condition of an unless.
          total_aggregate_targets_to_generate = unless (aggregate_targets_to_generate.empty? && removed_aggregate_target_labels.empty?)
                                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Issue #9013, opened 2019-07-17
inhibit_all_warnings! breaks with Swift files in Xcode11 beta 4


icecrystal23:
* [x] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

Run `pod install`, tried to compile my project with Xcode 11 beta 4

## What did you expect to happen?

The project compiles successfully

## What happened instead?

I got a swift compilation error

    <unknown>:0: error: unknown argument: '-w'
    <unknown>:0: error: unknown argument: '-Xanalyzer'
    <unknown>:0: error: unknown argument: '-analyzer-disable-all-checks'
    Command CompileSwiftSources failed with a nonzero exit code

## CocoaPods Environment

Podfile option `inhibit_all_warnings!` seems to be causing this, when applied to Swift files.

## Project that demonstrates the issue

TODO: Create sample project.  Should be easy: include pod Pulley with inhibit_all_warnings!
and try to build.

## Note about a potential fix

Search for "-w -Xanalyzer -analyzer-disable-all-checks" (currently showing up in `pod_target_installer.rb` in my code base, but it could be outdated) for a clue as to where a fix needs to be worked in.  These need to be skipped for Swift files.  Potentially another solution for ignoring warnings in Swift code will need to be applied.

dnkoutso:
Whats the way to inhibit warnings for Swift files? Definitely something we can fix.

icecrystal23:
Set this on the target:

    SWIFT_SUPPRESS_WARNINGS = YES;

dnkoutso:
Has this been the case for Xcode 10? Sorry I do not know from the top of my head.

dnkoutso:
Looks like Xcode 8 https://miqu.me/blog/2016/07/31/xcode-8-new-build-settings-and-analyzer-improvements/

dnkoutso:
Hmmm we are already removing `.swift` files here https://github.com/CocoaPods/CocoaPods/blob/master/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb#L282-L296

dnkoutso:
Was reading it wrong. We do not remove `.swift` files there.

dnkoutso:
PR is out.

icecrystal23:
Cool, thanks for the quick turnaround!

[...]

PR #9014, opened 2019-07-18
Do not pass inhibit warnings compiler flags for Swift source files.

dnkoutso:
closes https://github.com/CocoaPods/CocoaPods/issues/9013

It will probably be for 1.7.5 release and will add a milestone.

Passing those flags in Xcode 11 Beta 4 fails to compile.

### DIFF
--- a/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
@@ -287,12 +287,18 @@ module Pod
               {
                 true => file_accessor.arc_source_files,
                 false => file_accessor.non_arc_source_files,
-              }.each do |arc, files|
-                next if files.empty?
-                files = files - headers - other_source_files
-                flags = compiler_flags_for_consumer(consumer, arc)
-                regular_file_refs = project_file_references_array(files, 'source')
-                native_target.add_file_references(regular_file_refs, flags)
+              }.each do |arc, source_files|
+                next if source_files.empty?
+                source_files = source_files - headers - other_source_files
+                swift_source_files, non_swift_source_files = source_files.partition { |file| file.extname == '.swift' }
+                {
+                  :objc => non_swift_source_files,
+                  :swift => swift_source_files,
+                }.each do |language, files|
+                  compiler_flags = compiler_flags_for_consumer(consumer, arc, language)
+                  file_refs = project_file_references_array(files, 'source')
+                  native_target.add_file_references(file_refs, compiler_flags)
+                end
               end
 
               header_file_refs = project_file_references_array(headers, 'header')
@@ -780,11 +786,17 @@ module Pod
           #         The consumer for the specification for which the compiler flags
           #         are needed.
           #
+          # @param  [Boolean] arc
+          #         Whether the arc is enabled or not.
+          #
+          # @param  [Symbol] language
+          #         The language these compiler warnings are for. Can be either :objc or :swift.
+          #
           # @return [String] The compiler flags.
           #
-          def compiler_flags_for_consumer(consumer, arc)
+          def compiler_flags_for_consumer(consumer, arc, language)
             flags = consumer.compiler_flags.dup
-            if !arc
+            if !arc && language == :objc
               flags << '-fno-objc-arc'
             else
               platform_name = consumer.platform_name
@@ -793,7 +805,7 @@ module Pod
                 flags << '-DOS_OBJECT_USE_OBJC=0'
               end
             end
-            if target.inhibit_warnings?
+            if target.inhibit_warnings? && language == :objc
               flags << '-w -Xanalyzer -analyzer-disable-all-checks'
             end
             flags * ' '
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
-* None.  
+* Do not pass inhibit warnings compiler flags for Swift source files.  
+  [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
+  [#9013](https://github.com/CocoaPods/CocoaPods/issues/9013)
 
 
 ## 1.7.4 (2019-07-09)
--- a/spec/unit/installer/xcode/pods_project_generator/pod_target_installer_spec.rb
+++ b/spec/unit/installer/xcode/pods_project_generator/pod_target_installer_spec.rb
@@ -1196,25 +1196,29 @@ module Pod
 
               it 'adds -w per pod if target definition inhibits warnings for that pod' do
                 @installer.target.target_definitions.first.stubs(:inhibits_warnings_for_pod?).returns(true)
-                flags = @installer.send(:compiler_flags_for_consumer, @spec.consumer(:ios), true)
-
+                flags = @installer.send(:compiler_flags_for_consumer, @spec.consumer(:ios), true, :objc)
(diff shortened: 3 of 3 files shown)


lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -780,9 +786,15 @@ def create_build_phase_to_symlink_header_folders(native_target)
...
+          # @param  [Symbol] language
+          #         The language these compiler warnings are for. Can be either :objc or :swift.
+          #
           # @return [String] The compiler flags.
           #
-          def compiler_flags_for_consumer(consumer, arc)
+          def compiler_flags_for_consumer(consumer, arc, language)
             flags = consumer.compiler_flags.dup
             if !arc
               flags << '-fno-objc-arc'
```
dnkoutso: I wasnt sure if this makes sense for Swift files but Xcode 11 Beta 4 allows it.

amorde: huh, didn't even think of that. that's interesting

dnkoutso:
We also pass `-suppress-warnings` for Swift https://github.com/CocoaPods/CocoaPods/blob/1-7-stable/lib/cocoapods/target/build_settings.rb#L785 so no need to use `SWIFT_SUPPRESS_WARNINGS`

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
dnkoutso: actually that seems to be a lie. investigating more.

dnkoutso: nope, not allowed either.

dnkoutso: updated with a test to remove `-fno-objc-arc` for Swift source files.

segiddins:
Should file a radar for this regression

dnkoutso:
Will do but I do not expect that to be fixed anytime soon. Maybe that is my answer _not_ to point to `1-7-stable` and have it on `master` for 1.8.x.

dnkoutso:
https://openradar.appspot.com/radar?id=4947476331102208

mrgrauel:
A fast release - 1.7.5 - would be awesome 👍

dnkoutso:
We are waiting a day or two to hear from Apple if this change was intended or a regression from their part. 

If people are blocked please use https://bundler.io/ and point to this branch or commit to be unblocked.

We do not want to release a 1.7.5 and _then_ release 1.7.6 _again_ if this is a regression from Apple.

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -287,12 +287,18 @@ def add_files_to_build_phases(native_target, test_native_targets, app_native_tar
...
-                files = files - headers - other_source_files
-                flags = compiler_flags_for_consumer(consumer, arc)
-                regular_file_refs = project_file_references_array(files, 'source')
-                native_target.add_file_references(regular_file_refs, flags)
+              }.each do |arc, source_files|
+                next if source_files.empty?
+                source_files = source_files - headers - other_source_files
+                swift_source_files, non_swift_source_files = source_files.partition { |file| file.extname == '.swift' }
+
+                non_swift_compiler_flags = compiler_flags_for_consumer(consumer, arc, :objc)
```
segiddins: seems fine, would be nice to do this in a loop since we're duplicating this 3 lines of logic here

dnkoutso: done

lilyball:
This doesn't seem like an Xcode regression. It seems like an Xcode bugfix. I just compiled my project using Xcode 10.2.1 and checked, and it did not pass the per-file compiler flags to swift files. In Xcode 11 beta 4 it's now passing the per-file compiler flags to swift files.

The fact is, these flags *never worked* on Swift. We just didn't notice before because Xcode wasn't passing them along. Now that it is passing them along, it's obvious that the flags shouldn't be applied to swift files.

I strongly recommend publishing a 1.7.5 for this instead of waiting.

dnkoutso:
Shipped with 1.7.5.

Coeur:
> We do not want to release a 1.7.5 and _then_ release 1.7.6 _again_

@dnkoutso 
Well, see https://github.com/CocoaPods/CocoaPods/issues/9013#issuecomment-513701454

barrypress:
@Coeur is right about #9013. The attached project is a clean iOS project, then including the Down pod, and fails to compile.
[DownTest.zip](https://github.com/CocoaPods/CocoaPods/files/3424461/DownTest.zip)

dnkoutso:
This happens because the podspec specifies `compiler_flags` https://github.com/iwasrobbed/Down/blob/master/Down.podspec#L20

We will have to re-think this through as a relatively bigger change to handle compiler flags for different sources.

lilyball:
In the short term we should probably just stop passing per-file compilation flags to Swift files. It apparently never worked before, so omitting them entirely restores the old behavior. Then we can explore adding a separate option for Swift compilation flags.

amorde:
That's not true, there are per-file complication flags for Swift that do work, the problem was CocoaPods was including Objective-C-specific flags when it shouldn't have been.

There's the `-Xcc` option for passing flags down to the Objective-C/C/C++ compiler, maybe we should be using that instead?

lilyball:
`swiftc` has some flags that make sense to pass (though I'm not sure which ones are appropriate as per-file flags), but it's clear that Xcode prior to Xcode 11 Beta 4 wasn't passing the per-file flags to `swiftc`. This is really easy to test; just create a new project in Xcode 10, add a dummy per-file compilation flag to a Swift file, and build. Not only does the build succeed despite the garbage flag, but if you inspect the build log, the flag was never passed.

But more than that, given that compilation flags are unique to each compiler, it simply doesn't make sense for CocoaPods to have a single property that it applies to all source files. It should have a property for `clang` flags, and a property for `swiftc` flags. I'm also not sure why it's even using per-file compilation flags at all instead of just passing these in the appropriate per-language build setting.

amorde:
Yes that's the issue, the `compiler_flags` DSL was introduced before Swift was released and it was never updated. We might add a `swift_compiler_flags` to match but no concrete plans for that yet

barrypress:
Perhaps I’m being naive, and in any event I owe you a big thanks for being so responsive. That said, I didn’t see a plan for what the fix will be, nor for  when users might expect it. Is it unreasonable to ask that?

Coeur:
@barrypress :
All things related to the issue are written here between your two messages.
Feel free to do a Pull Request adopting either amorde idea (_add a `swift_compiler_flags`_) or lilyball idea (_stop passing per-file compilation flags to Swift files_) if you need a fix urgently. Otherwise, tell your userbase to avoid Xcode 11 beta 4+ for a few weeks?

lilyball:
@barrypress You can also submit a PR to Down that removes that flag from the podspec as per [this comment](https://github.com/iwasrobbed/Down/issues/162#issuecomment-515081112). As the comment says, this will likely introduce some warnings in the C code, but you can then handle that by adding the appropriate `#pragma`s around the code in question to disable the warning in a targeted fashion.

Coeur:
@barrypress as there wasn't much visible move on that in the past two weeks, I would suggest that you open a specific GitHub issue in order to track that Xcode 11 issue with pods requiring `compiler_flags` that are incompatible with Swift.

PR #9016, opened 2019-07-18
Make CDNSource show up in `pod repo env`

igor-makarov:
When users submit issues, they are requested to run `pod repo env`.  
`CDNSource` should show up on that list.


igor-makarov:
Will add tests

igor-makarov:
Reused existing test to check for correct repo printing.

PR #9029, opened 2019-07-22
Only return library itself as a framework path for library specs.

dnkoutso:
This was a bug when `app_specs` were introduced.

### DIFF
--- a/lib/cocoapods/target/pod_target.rb
+++ b/lib/cocoapods/target/pod_target.rb
@@ -383,7 +383,7 @@ module Pod
                                 end
             FrameworkPaths.new(framework_source, dsym_source, bcsymbolmap_paths)
           end
-          if !file_accessor.spec.test_specification? && should_build? && build_as_dynamic_framework?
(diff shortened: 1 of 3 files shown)


lib/cocoapods/target/pod_target.rb
```diff
@@ -383,7 +383,7 @@ def framework_paths
                                 end
             FrameworkPaths.new(framework_source, dsym_source, bcsymbolmap_paths)
           end
-          if !file_accessor.spec.test_specification? && should_build? && build_as_dynamic_framework?
+          if file_accessor.spec.library_specification? && should_build? && build_as_dynamic_framework?
```
dnkoutso: with the intro of `app_specs` we were not properly ruling them out here.

dnkoutso: also from docs of this method:
```
    # @return [Hash{String=>Array<FrameworkPaths>}] The vendored and non vendored framework paths this target
    #         depends upon keyed by spec name. For the root spec and subspecs the framework path of the target itself
    #         is included.
```

PR #9037, opened 2019-07-24
Fix resources prefix for app specs.

dnkoutso:
Another small regression when app specs were introduced that we didn't account for.

### DIFF
--- a/lib/cocoapods/target/pod_target.rb
+++ b/lib/cocoapods/target/pod_target.rb
@@ -398,10 +398,15 @@ module Pod
     def resource_paths
       @resource_paths ||= begin
         file_accessors.each_with_object({}) do |file_accessor, hash|
-          resource_paths = file_accessor.resources.map { |res| "${PODS_ROOT}/#{res.relative_path_from(sandbox.project_path.dirname)}" }
-          resource_paths = [] if file_accessor.spec.non_library_specification? && build_as_framework?
+          resource_paths = if file_accessor.spec.non_library_specification? && build_as_framework?
+                             []
+                           else
+                             file_accessor.resources.map do |res|
+                               "${PODS_ROOT}/#{res.relative_path_from(sandbox.project_path.dirname)}"
+                             end
+                           end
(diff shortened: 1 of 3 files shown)


lib/cocoapods/target/pod_target.rb
```diff
@@ -398,10 +398,13 @@ def framework_paths
     def resource_paths
       @resource_paths ||= begin
         file_accessors.each_with_object({}) do |file_accessor, hash|
-          resource_paths = file_accessor.resources.map { |res| "${PODS_ROOT}/#{res.relative_path_from(sandbox.project_path.dirname)}" }
-          resource_paths = [] if file_accessor.spec.non_library_specification? && build_as_framework?
+          resource_paths = if file_accessor.spec.non_library_specification? && build_as_framework?
```
dnkoutso: minor speed boost here, no need to process `file_accessor.resources` (which hits the file system) if we know we will throw it away.

dnkoutso:
Btw the shell script phase was doing the right thing this fix is for input paths.

Screen_Shot_2019-07-23_at_10_38_27_PM

lib/cocoapods/target/pod_target.rb
```diff
@@ -398,10 +398,15 @@ def framework_paths
...
+                             []
+                           else
+                             file_accessor.resources.map do |res|
+                               "${PODS_ROOT}/#{res.relative_path_from(sandbox.project_path.dirname)}"
+                             end
+                           end
           prefix = Pod::Target::BuildSettings::CONFIGURATION_BUILD_DIR_VARIABLE
-          prefix = configuration_build_dir unless file_accessor.spec.test_specification?
+          prefix = configuration_build_dir unless file_accessor.spec.non_library_specification?
```
dnkoutso: this is the fix in this PR.

dnkoutso:
Uh wait a minute let me verify something.

dnkoutso:
Was super wrong here. This works as expected. I am repurposing this PR for the speed up only.

PR #9040, opened 2019-07-24
Use correct deployment target for test specs and app specs.

dnkoutso:
Even though subspecs _can_ define `deployment_target` we always pick the _max_ between library specs. That is because subspecs do not generate their own target in Xcode and are always merged into a single library target (framework or static library) so we need to pick something and we pick the _max_.

However, app specs and test specs _do_ generate an Xcode target for each (1-to-1), therefore we can honor the deployment target that is set for them. If it's not set then we pickup the one from the parent.

This PR does that.

### DIFF
--- a/lib/cocoapods/installer/analyzer.rb
+++ b/lib/cocoapods/installer/analyzer.rb
@@ -778,7 +778,11 @@ module Pod
         end
       end
 
-      # Calculates and returns the platform to use for the given list of specs and target definitions.
+      # Calculates and returns the platform to use for the given list specs and target definitions.
+      #
+      # @note The platform is only determined by all library specs and ignores non library ones. Subspecs are always
+      #       integrated in the same target as the root spec therefore the max deployment target is always returned
+      #       across the specs passed.
       #
       # @param [Array<Specification>] specs
       #        the specs to inspect and calculate the platform for.
@@ -792,10 +796,11 @@ module Pod
       # @return [Platform]
       #
       def determine_platform(specs, target_definitions, host_requires_frameworks)
+        library_specs = specs.select(&:library_specification?)
         platform_name = target_definitions.first.platform.name
         default = Podfile::TargetDefinition::PLATFORM_DEFAULTS[platform_name]
-        deployment_target = specs.map do |spec|
-          Version.new(spec.deployment_target(platform_name) || default)
+        deployment_target = library_specs.map do |library_spec|
+          Version.new(library_spec.deployment_target(platform_name) || default)
         end.max
         if platform_name == :ios && host_requires_frameworks
           minimum = Version.new('8.0')
--- a/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
@@ -331,7 +331,9 @@ module Pod
               name = target.test_target_label(test_spec)
               platform_name = target.platform.name
               language = target.uses_swift_for_spec?(test_spec) ? :swift : :objc
-              test_native_target = project.new_target(product_type, name, platform_name, deployment_target, nil, language)
+              test_native_target = project.new_target(product_type, name, platform_name,
+                                                      target.deployment_target_for_non_library_spec(test_spec), nil,
+                                                      language)
               test_native_target.product_reference.name = name
 
               target.user_build_configurations.each do |bc_name, type|
@@ -407,7 +409,7 @@ module Pod
               spec_name = app_spec.parent.name
               subspec_name = target.subspec_label(app_spec)
               app_target_label = target.app_target_label(app_spec)
-              platform = target.platform
+              platform = Platform.new(target.platform.symbolic_name, target.deployment_target_for_non_library_spec(app_spec))
               info_plist_entries = spec_consumer.info_plist
               app_native_target = AppHostInstaller.new(sandbox, project, platform, subspec_name, spec_name,
                                                        app_target_label, :add_main => false,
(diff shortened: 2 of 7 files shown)


lib/cocoapods/installer/analyzer.rb
```diff
@@ -744,7 +744,7 @@ def generate_pod_target(target_definitions, target_inspections, specs, scope_suf
           archs = target_requires_64_bit ? ['$(ARCHS_STANDARD_64_BIT)'] : []
         end
         host_requires_frameworks = target_definitions.any?(&:uses_frameworks?)
-        platform = determine_platform(specs, target_definitions, host_requires_frameworks)
+        platform = determine_platform(specs.select(&:library_specification?), target_definitions, host_requires_frameworks)
```
amorde: I would opt for putting this into `determine_platform` instead so the logic isn't spread out

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -497,8 +499,11 @@ def add_resources_bundle_targets(file_accessors)
                 target.user_build_configurations.each do |bc_name, type|
                   resource_bundle_target.add_build_configuration(bc_name, type)
                 end
-                resource_bundle_target.deployment_target = deployment_target
-
+                resource_bundle_target.deployment_target = if file_accessor.spec.non_library_specification?
```
amorde: is this covered by tests?

lib/cocoapods/installer/analyzer.rb
dnkoutso: hm sure sg

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
dnkoutso: eeeeeh kind of tbh but I will add one.

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -407,7 +409,7 @@ def add_app_targets
               spec_name = app_spec.parent.name
               subspec_name = target.subspec_label(app_spec)
               app_target_label = target.app_target_label(app_spec)
-              platform = target.platform
+              platform = Platform.new(target.platform.symbolic_name, target.deployment_target_for_non_library_spec(app_spec))
```
segiddins: maybe add a `platform_for_spec` method or somethign like that?

dnkoutso: will do.

spec/unit/installer/xcode/pods_project_generator/pod_target_installer_spec.rb
```diff
@@ -723,56 +620,6 @@ class PodsProjectGenerator
...
-                                            [@spec], [@target_definition], [file_accessor])
-                @installer = PodTargetInstaller.new(config.sandbox, @project, @pod_target)
-              end
-
-              after do
-                FileUtils.remove_entry(@tmpdir) if Dir.exist?(@tmpdir)
-                FileUtils.remove_entry(@symlink_dir) if File.symlink?(@symlink_dir)
-              end
-
-              it 'headers are public if podspec directory is symlinked for static lib' do
```
amorde: is this included elsewhere? were they bad tests?

dnkoutso: this one was a bad test and its the only one. It was very artificial so I removed it.

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -870,7 +875,7 @@ def custom_module_map
           def project_file_references_array(files, file_type)
             files.map do |sf|
               project.reference_for_path(sf).tap do |ref|
-                raise Informative, "Unable to find #{file_type} ref for #{sf} for target #{target.name}." unless ref
+                raise Informative, "Unable to find #{file_type} ref for `#{sf.basename}` for target `#{target.name}`." unless ref
```
dnkoutso: minor change as part of test refactoring here.

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
dnkoutso: @amorde _really_ thank you for pushing here as I discovered this is not working properly because of my test.

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -538,6 +539,11 @@ def add_resources_bundle_targets(file_accessors)
 
                 remove_pod_target_xcconfig_overrides_from_target(target.build_settings_for_spec(file_accessor.spec), resource_bundle_target)
```
dnkoutso: @segiddins this line continues to be a problem in which a bunch of settings _including_ `<PLATFORM>_DEPLOYMENT_TARGET` gets removed.

dnkoutso: I had to move `resource_bundle_target.deployment_target = deployment_target` below that line for things to function.

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
amorde: yay for tests :)

dnkoutso:
I will need to re-do some of the target level build settings before I can land this. I expect build to go red due to failing test.

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
dnkoutso: Well.....kind of nevermind this is a different issue where app specs xcconfigs are merged into the resource bundles and our Watermelon podspec includes `IPHONEOS_DEPLOYMENT_TARGET` which is removed by the resource bundle target.

dnkoutso:
@amorde I wrote a test that passes and I will make a follow up PR to fix some of the `resource_bundles` in `app_specs`

Issue #8319, opened 2018-11-30
RFC: Incremental Pod Installation (Phase 2)


sebastianv1:
# Incremental Pod Installation (Phase 2)
**author:** [sebastianv1](https://github.com/sebastianv1)  
**date:** 11/16/2018

## Summary
[Summary](https://github.com/CocoaPods/CocoaPods/issues/8253) from Phase 1 Doc.

## Motivation
[Motivation](https://github.com/CocoaPods/CocoaPods/issues/8253) from Phase 1 Doc.

### Installation Option and Flags
Enabling incremental pod installation will be gated by the installation option `incremental_installation` that depends on `generate_multiple_pod_projects` also being enabled. We want to enforce this constraint because incremental installation will optimize the project generation step by only regenerating the individual Xcode projects for targets that have changed.

In addition to the installation option, we will add a new installation flag `--clean-install` that can be used to ignore the contents of the cache and force a complete installation.

### Project Caching
In order to enable _only_ regenerating the Xcode project files for pod targets that have changed since the previous installation, we will create a cache inside the sandbox directory storing:
1. A __target cache key__ per target used to determine if it is dirty.
2. __Target metadata__ used to recreate itself as a target dependency for parent targets.

The cache will exist under the `Pods/.project_cache` directory and store files for the two cases listed above (`installation_cache` and `metadata_cache`) in addition to a `cache_version` file that is stored for backwards compatibility if changes are made in the future to the structure of the cache that would require flushing its contents.

##### `TargetCacheKey`
The `TargetCacheKey` is responsible for uniquely identifying a `Target` and determining if it has changed.

For each `PodTarget`, we can mark it as dirty based on a difference in the following criteria:
- Podspec checksum values.
- Build settings
- Set of specs to integrate.
- Set of files tracked (exclusive to local pods).
- Checkout options (if they exist for the pod).

Each `PodTarget` will store in the `installation_cache` file:
- Podspec checksum.
- List of specification names.
- Checksum of build settings.
- List of all files tracked (exclusive to local pods).
- Checkout options (if they exist for the pod).

For each `AggregateTarget`, we can mark it as dirty based on a difference in just the build settings. This is because we only care about a difference in the set of its pod targets or the way it's built, which are encapsulated in the aggregate target's build settings.

Each `AggregateTarget` will store in the `installation_cache` file:
- Checksum of build settings.

The `TargetCacheKey` public interface will be:

```ruby
class TargetCacheKey
[...]

PR #9041, opened 2019-07-24
Store relative project and file paths in the incremental cache.

dnkoutso:
All paths stored for the cache are now relative to the `sandbox.root`.

This way if a project is generated with incremental installation and is shared to someone else it will continue to work and the cache won't change.

See https://github.com/CocoaPods/CocoaPods/issues/8319#issuecomment-513093481


lib/cocoapods/installer/xcode/pods_project_generator/pod_target_dependency_installer.rb
```diff
@@ -14,10 +18,6 @@ class PodTargetDependencyInstaller
         #
         attr_reader :metadata_cache
 
-        # @return [Sandbox] The sandbox used for this installation.
```
dnkoutso: this was just moved up since its the first parameter of the intializer.

PR #9042, opened 2019-07-24
Add header mappings spec for macOS

drcapulet:
Accompanying PR to CocoaPods/cocoapods-integration-specs#241.


amorde:
thanks for this!

Issue #8519, opened 2019-02-15
Vendored Frameworks not linked to Dynamic Frameworks


cnoon:
# Report

## What did you do?

Updated to CocoaPods 1.6.0 (from CocoaPods 1.6.0-beta.2) and ran `bundle exec pod update`.

## What did you expect to happen?

That everything would update as expected without issue.

## What happened instead?

The `frameworks` and `vendored_frameworks` defined in podspecs for other dynamic frameworks were no longer linked in the `OTHER_LDFLAGS` in the `xcconfig` files for the dynamic frameworks.

![screen shot 2019-02-15 at 10 38 24 am](https://user-images.githubusercontent.com/169110/52877094-e0f50280-310d-11e9-834c-90756c905b50.png)

## Troubleshooting

I've been trying to track down exactly what happened here to trigger this regression. It appears that the original issue was introduced in #8314 from what I can tell. It stripped both `frameworks` and `vendored_frameworks` from being linked to dynamic frameworks.

Then it appears that the `frameworks` linking issue was resolved in #8498. That change was merged into `1-6-stable`. I tested this morning with the `1-6-stable` branch and I can confirm that that change does restore the `frameworks`, but not the `vendored_frameworks`.

As of right now, specifying a `vendored_framework` as a dependency of a dynamic framework does not properly link the `vendored_framework` in the dynamic frameworks `OTHER_LDFLAGS` and causes the build to fail.

We've worked around this issue temporarily by manually specifying the `-framework "VendoredFrameworkName"` in the `OTHER_LDFLAGS` of the dynamic framework, but this is obviously less than ideal. I can see that the issue lies in the `build_settings.rb` logic, but I can't quite get my head around all the logic in there to put up a PR. If someone could point me in the right direction, I'd be happy to push up the fix.

> It would be great to have this resolved in 1.6.1 if possible so we can revert our workarounds.
> I'm happy to help in any way to help make that happen.

## CocoaPods Environment

```bash
cnoon:~/Programming/Nike/Apps/Omega$ bundle exec pod env 
```

### Stack

```
   CocoaPods : 1.6.0
        Ruby : ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-darwin18]
    RubyGems : 2.7.6
        Host : Mac OS X 10.14.2 (18C54)
       Xcode : 10.1 (10B61)
         Git : git version 2.17.2 (Apple Git-113)
[...]

Issue #8204, opened 2018-10-19
CocoaPods 1.6.0.beta.1 (and .2) link both app targets and frameworks to vendored_libraries


Westacular:
* [x] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).
🌈
# Report

## What did you do?

The situation:
We have multiple podspec frameworks that have `vendored_libraries` and publicly expose the headers for the static library, in effect wrapping its functionality into a dynamic framework.

(This is largely done as a workaround to allow the use of these libraries by both other frameworks and by app targets.)

## What did you expect to happen?

In CocoaPods 1.5.3, only the framework that is vendoring the library is linked to it.

## What happened instead?

In CocoaPods 1.6.0.beta.1 and 1.6.0.beta.2, both the framework AND the app target are linked to the vendored libraries.

Specifically:
In the `Pods-$app_target_name.$config_name.xcconfig` files, `OTHER_LDFLAGS` now includes `-l"$vendored_library_name"` for each of the vendored libraries, whereas it did not it v1.5.3.

For libraries which have had their headers publicly exposed by the frameworks containing them, this leads to runtime duplicate class warnings of the sort described in #7155.

### Stack

```
   CocoaPods : 1.6.0.beta.2
        Ruby : ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-darwin16]
    RubyGems : 2.6.11
        Host : Mac OS X 10.13.6 (17G65)
       Xcode : 10.0 (10A255)
         Git : git version 2.19.0
Ruby lib dir : /usr/local/Cellar/ruby/2.4.1_1/lib
```

### Installation Source

```
Executable Path: /usr/local/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.2
cocoapods-packager    : 1.5.0
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
cocoapods-trunk       : 1.3.1
cocoapods-try         : 1.1.0
```

## Workaround

I've devised a workaround through a post_install in the Podfile:

```ruby
post_install do |installer|
  installer.aggregate_targets.each do |aggregate_target|
    if aggregate_target.name.start_with?('Pods-')
      aggregate_target.xcconfigs.each do |config_name, config_file|
[...]

PR #9045, opened 2019-07-25
Include dependent vendored frameworks in linker flags

drcapulet:
Accompanying PR to CocoaPods/cocoapods-integration-specs#242.


dnkoutso:
I'd like @segiddins to review this since he made the change in https://github.com/CocoaPods/CocoaPods/pull/8314

amorde:
+1 on @dnkoutso's comment, we had a bunch of regressions when changing this so would like to be 100% sure we get this right

drcapulet:
Of course! Based on some cursory glancing, it should fix #8519.

dnkoutso:
Does it break https://github.com/CocoaPods/CocoaPods/issues/8204 which is what #8314 fixed?

drcapulet:
No, I don't think so - that was specifically related to libraries instead of frameworks. This only targets frameworks that wouldn't otherwise be compiled since those will get included in the "Link Binary With Libraries" phase, but I'm definitely no expert on all this stuff.

.gitmodules
```diff
@@ -6,7 +6,7 @@
 	url = https://github.com/tonymillion/Reachability.git
 [submodule "spec/cocoapods-integration-specs"]
 	path = spec/cocoapods-integration-specs
-	url = https://github.com/CocoaPods/cocoapods-integration-specs.git
```
segiddins: will need to point back to master

lib/cocoapods/target/build_settings.rb
```diff
@@ -558,9 +558,8 @@ def initialize(target, non_library_spec = nil)
           if library_xcconfig? && (target.should_build? && target.build_as_dynamic?)
             frameworks.concat vendored_static_frameworks.map { |l| File.basename(l, '.framework') }
           end
-          if non_library_xcconfig?
-            frameworks.concat dependent_targets_to_link.flat_map { |pt| pt.build_settings.frameworks_to_import }
-          end
+          targets_to_link = library_xcconfig? ? dependent_targets.reject(&:should_build?) : dependent_targets_to_link
```
segiddins: for unbbuilt dependent targets, should we only be linking dynamic frameworks here, instead of dynamic and static?

drcapulet: Yes, updated.

dnkoutso:
@drcapulet needs a rebase to fix CHANGELOG conflict.

dnkoutso:
Also would need to rebase off of `1-8-stable` and point to that branch instead to have this change released for 1.8.0.beta.2.

lib/cocoapods/target/build_settings.rb
```diff
@@ -559,12 +559,16 @@ def initialize(target, non_library_spec = nil)
...
           frameworks = []
           frameworks.concat consumer_frameworks
-          if library_xcconfig? && (target.should_build? && target.build_as_dynamic?)
-            frameworks.concat vendored_static_frameworks.map { |l| File.basename(l, '.framework') }
-          end
-          if non_library_xcconfig?
+          if library_xcconfig?
+            # We know that this library target is being built dynamically based
+            # on the guard above, so include any vendored static frameworks.
+            frameworks.concat vendored_static_frameworks.map { |l| File.basename(l, '.framework') } if target.should_build?
```
dnkoutso: does this mean that a vendored static framework will be linked _twice_? Once for this dynamic framework and once on the app target itself?

drcapulet: This line is just moved around, but no - since this is a dynamic library target, when it gets compiled it needs to include any vendored static frameworks, and then the whole dynamic library will get linked into it's parent (either app or another library).

dnkoutso: If I recall correctly CocoaPods will link that vendored static framework to the app target as well not just the dynamic library target.

From its early days CocoaPods had a very (kinda annoying) naive approach to just link everything to the app level target.

dnkoutso: I can verify my own question this week.

dnkoutso: I guess you are right this was already happening before. I will still verify though

dnkoutso:
@drcapulet needs a rebase. I will try to land it for a beta.2.

.gitmodules
dnkoutso: yes @drcapulet will need to revert this

dnkoutso:
@drcapulet can you rebase to fix the `CHANGELOG` conflict and we can merge?

ryang1428:
Still having issues installing vendors frameworks in 1.8.0 :/

ryang1428:
@dnkoutso should I open a new issue? I am still having issues installing vendored frameworks via 1.8.1.. Getting the same compile time errors:

```
Undefined symbols for architecture x86_64:
  "_OBJC_CLASS_$_CardIO", referenced from:
      objc-class-ref in AppCoordinator.o
      objc-class-ref in VoiceController.o
      objc-class-ref in CheckController.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

icecrystal23:
I think this may be causing us to have duplicate symbol errors, as most of our vendored frameworks are static and therefore we do NOT want them included in linker flags outside the project vendoring them.  

I will investigate and create an issue (and maybe a PR to fix) if I can repro for you.

PR #9066, opened 2019-07-31
[Analyzer] Do not implicitly whitelist transitive dependencies

### LINKED ISSUE #6838: configuration not work for dependency
CP 1.2.0 and CP 1.2.1, I have a Podfile: 
```
pod "DebugTool", :configuration=>"Debug"
```

After `pod install`, the Podfile.lock is 
```
 - DebugTool
    - Aspects
    - CocoaLumberjack
    - FLEX
    - LumberjackConsole
    - MGBenchmark
    - MMPlaceHolder
    - NBUCore
    - NBULog
    - XLForm
```

And the `MGBenchmark` will not be used by other pods. So I hope the `MGBenchmark` will be link under `Release` configuration.

But I found the release.xcconfig:
```
OTHER_LDFLAGS = $(inherited) -ObjC -l"Aspects" -l"CocoaLumberjack"  -l"FLEX" -l"LumberjackConsole" -l"MGBenchmark" -l"MMPlaceHolder"  -l"NBUCore" -l"NBULog" -l"XLForm"
```
The `DebugTool`'s dependencies all be link besides the `DebugTool`. I think it is not correct.

amorde:
Closes #6838

This resolves an issue which caused the transitive dependencies of a dependency with a whitelist (ex. `:configurations => ['Debug']`) to be linked with the host target even for configurations not included in the whitelist.

This should also provide a small build time improvement for projects using the new build system that contain these dependencies, when using recent versions of Xcode that detect implicit dependencies based on linker flags (Xcode 10.2 or higher).


lib/cocoapods/resolver/resolver_specification.rb
```diff
@@ -17,10 +17,28 @@ class ResolverSpecification
...
-      def initialize(spec, used_by_non_library_targets_only, source)
+      # @return [Bool] whether this resolved specification is a transitive dependency that was not directly included
+      #         by the target that depends on it
+      #
+      attr_reader :transitive
+      alias :transitive? transitive
+
+      # @return [Array<Specification>] specifications which depend on this specification
+      #
+      attr_reader :incoming_vertices
```
amorde: could use some suggestions for a better name

lib/cocoapods/resolver.rb
```diff
@@ -85,9 +85,12 @@ def initialize(sandbox, podfile, locked_dependencies, sources, specs_updated,
     #
     def resolve
       dependencies = @podfile_dependency_cache.target_definition_list.flat_map do |target|
-        @podfile_dependency_cache.target_definition_dependencies(target).each do |dep|
-          next unless target.platform
-          @platforms_by_dependency[dep].push(target.platform)
+        @podfile_dependency_cache.target_definition_dependencies(target).tap do |deps|
```
amorde: took the opportunity to make this (very tiny) optimization

spec/unit/installer/analyzer_spec.rb
```diff
@@ -1687,6 +1687,32 @@ module Pod
...
+            target 'SampleProject' do
+              pod 'AFNetworking'
+              pod 'RxCocoa', '4.0.0', :configurations => ['Debug']
+            end
+          end
+          @analyzer = Installer::Analyzer.new(config.sandbox, @podfile)
+          target_definition = @podfile.target_definitions['SampleProject']
+          aggregate_target = @analyzer.analyze.targets.find { |t| t.target_definition == target_definition }
+
+          aggregate_target.pod_targets_for_build_configuration('Debug').map(&:name).
```
dnkoutso: nit extract `aggregate_target.pod_targets_for_build_configuration('Debug')` to make the rest of lines single lines?

dnkoutso: something like 
```ruby
debug_pod_targrets.map(&:name).sort.should == [
 'PodA',
 'PodB'
]
```

dnkoutso: I'd suspect `AFNetworking` is also meant to be for release here right?

lib/cocoapods/resolver/resolver_specification.rb
dnkoutso: `dependent_specs`?

spec/unit/installer/analyzer_spec.rb
amorde: yup can do

lib/cocoapods/resolver/resolver_specification.rb
amorde: yes. nice

amorde:
Ugh rubocop is being very picky, and not behaving the same way on my machine

amorde:
Would like 1 more review before merging

lib/cocoapods/installer/analyzer.rb
```diff
@@ -554,11 +563,20 @@ def filter_pod_targets_for_target_definition(target_definition, pod_targets_by_t
             whitelists = dependencies.map do |dependency|
               target_definition.pod_whitelisted_for_configuration?(dependency.name, configuration_name)
             end.uniq
+            if resolver_spec.transitive?
```
segiddins: instead of introducing a whole new property, could you check `target_dependency.dependencies.any? {|d| d.root_name == resolver_spec.name }` or somethign like that?

lib/cocoapods/installer/analyzer.rb
```diff
@@ -554,11 +563,20 @@ def filter_pod_targets_for_target_definition(target_definition, pod_targets_by_t
...
+              # Include the whitelist config for the source of the transitive dependency, so we can exclude
+              # the transitive dependencies of whitelisted dependencies
+              whitelists += resolver_spec.dependent_specs.map do |dependent_spec|
```
segiddins: I don't think this check is correct, since a spec can come in transitively through multiple different specs, and we need to make sure we're only merging things where the only specs depending on that spec have been excluded

e.g.

A -> B
C -> B

A whitelisted, C not, we still need to include B

lib/cocoapods/resolver.rb
```diff
@@ -132,7 +137,9 @@ def resolver_specs_by_target
               payload = vertex.payload
               non_library = !used_by_aggregate_target_by_spec_name.fetch(vertex.name)
               spec_source = payload.respond_to?(:spec_source) && payload.spec_source
-              ResolverSpecification.new(payload, non_library, spec_source)
+              transitive = !explicitly_included_by_aggregate_by_spec_name.fetch(vertex.name, false)
+              incoming_spec_vertices = vertex.incoming_edges.map { |edge| edge.origin.payload }
```
segiddins: does this need to filter by platform, potentially?

lib/cocoapods/installer/analyzer.rb
amorde: good catch, definitely need to address this

lib/cocoapods/resolver.rb
amorde: ah potentially yes, I think if a specification has a platform specific dependency then that could be an issue. we support that correct? `s.ios.dependency 'SomePod'`

lib/cocoapods/installer/analyzer.rb
```diff
@@ -554,11 +563,20 @@ def filter_pod_targets_for_target_definition(target_definition, pod_targets_by_t
...
+                target_definition.pod_whitelisted_for_configuration?(dependent_spec.name, configuration_name)
+              end
+              whitelists.uniq!
+            end
 
             case whitelists
             when [], [true] then nil
             when [false] then next
             else
+              next if resolver_spec.transitive?
```
amorde: this is incorrect, we need to allow `[false, true]` iff the spec is `transitive?` rather than skipping with `next`. this would resolve the above issue for when one dependent is whitelisted and the other is not

lib/cocoapods/installer/analyzer.rb
amorde: we could use `resolver_spec.spec.dependencies` and whitelist any names included there if the direct dependency itself gets whitelisted, but that would require multiple iterations over the list of pod targets to make sure we don't bail early if the whitelisted dependent comes after the one which is not whitelisted. I think that would increase the complexity of this implementation

lib/cocoapods/installer/analyzer.rb
amorde: also uncovered a case where the subspec of a pod depends on the current pod, it will always be whitelisted. this needs to filter `dependent_specs` to only include the target definition's direct dependencies which have been whitelisted for the current configuration

amorde:
Having trouble getting this to cover all the edge cases - closing for now.

PR #9069, opened 2019-07-31
Do not add CocoaPods script phases to targets that have no paths to embed.

dnkoutso:
We already do this when we integrate the user targets, but we don't do it for generated pod targets.

Integration specs https://github.com/CocoaPods/cocoapods-integration-specs/pull/244

### DIFF
--- a/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
@@ -614,9 +614,11 @@ module Pod
                 pod_target.resource_paths.values_at(*spec_paths_to_include).flatten.compact
               end
             end
-            generator = Generator::CopyResourcesScript.new(resource_paths_by_config, target.platform)
-            update_changed_file(generator, path)
-            add_file_to_support_group(path)
+            unless resource_paths_by_config.values.all?(&:empty?)
+              generator = Generator::CopyResourcesScript.new(resource_paths_by_config, target.platform)
+              update_changed_file(generator, path)
+              add_file_to_support_group(path)
+            end
           end
 
           # Creates a script that embeds the frameworks to the bundle of the test target.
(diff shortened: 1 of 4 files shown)


lib/cocoapods/installer/xcode/pods_project_generator/pod_target_integrator.rb
```diff
@@ -60,32 +60,36 @@ def add_copy_resources_script_phase(native_target, spec)
 
             input_paths_by_config = {}
             output_paths_by_config = {}
-            if use_input_output_paths
```
dnkoutso: this is no longer behind this if statement because we need to calculate the paths to determine if we should remove or add the script phase.

spec/unit/installer/xcode/pods_project_generator/pod_target_integrator_spec.rb
```diff
@@ -12,33 +12,32 @@ class PodTargetIntegrator
...
+                                                                                 [], [@test_native_target])
             end
 
             describe '#integrate!' do
-              it 'integrates test native targets with frameworks and resources script phases' do
-                PodTargetIntegrator.new(@target_installation_result).integrate!
-                @test_native_target.build_phases.count.should == 2
+              it 'integrates test native targets with framework script phase' do
+                PodTargetIntegrator.new(@coconut_target_installation_result).integrate!
+                @test_native_target.build_phases.count.should == 1
```
dnkoutso: this is correct. There are only frameworks to embed here but no resources since there are no resource bundles to copy.

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -614,9 +614,11 @@ def create_test_target_copy_resources_script(test_spec)
                 pod_target.resource_paths.values_at(*spec_paths_to_include).flatten.compact
               end
             end
-            generator = Generator::CopyResourcesScript.new(resource_paths_by_config, target.platform)
-            update_changed_file(generator, path)
-            add_file_to_support_group(path)
+            unless resource_paths_by_config.values.all?(&:empty?)
```
segiddins: nit: `.each_value.all?(&:empty?)` to avoid the intermediary array

dnkoutso: nice will update.

PR #9073, opened 2019-08-01
Use the new Lockfile method option to warn about the old git spec repo

igor-makarov:
To be merged after https://github.com/CocoaPods/Core/pull/574 to unlock its function.

### DIFF
--- a/lib/cocoapods/config.rb
+++ b/lib/cocoapods/config.rb
@@ -205,7 +205,7 @@ module Pod
     # @return [Nil] If no Lockfile is available.
     #
     def lockfile
-      @lockfile ||= Lockfile.from_file(lockfile_path) if lockfile_path
+      @lockfile ||= Lockfile.from_file(lockfile_path, true) if lockfile_path
     end
 
     # Returns the path of the Podfile.


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Use the new Lockfile method option to warn about the old git spec repo  
```

</p>

dnkoutso:
Should we provide an installation_option for folks who want to continue using the master specs repo?

igor-makarov:
Does CP provide installation options for all warnings? It doesn't fail the build...

dnkoutso:
Most warnings, but I think its useful to provide it and have it off by default. I can see cases where folks will continue to use the `git` based repo.

amorde:
This feels a bit early, we haven't yet made a release which uses the CDN by default. Would this make more sense for 1.9?

igor-makarov:
I think that because 1.8 is when CDN becomes the default, a lot of users would expect their project to migrate and would be surprised it didn't. This provides a message for them.

dnkoutso:
So lets say these steps happen:

I use 1.7.5 today
Have a Podfile.lock with some pods from master specs repo
I edit Podfile and remove source of the master specs repo or replace it with source 'https://cdn.cocoapods.org/'
I delete the git based master specs repo via pod repo remove master
I run pod install
Does it all work and the Podfile.lock gets updated to specify trunk instead of master? Do I get a warning then?

igor-makarov:
I just checked the scenario you've described and it seems that it resolves itself with existing functionality.  
Perhaps this whole warning thing is redundant. If a user really *wants* to keep the old master repo, why do we need to tell them anything?

Issue #9043, opened 2019-07-24
Embed Pods Frameworks incorrectly removed


revolter:
* [x] 🌈 I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

- Removed the `TwitterKit` pod from the `Podfile`;
- Ran `pod install`.

## What did you expect to happen?

The `[CP] Embed Pods Frameworks` build phases should not be removed.

## What happened instead?

They were completely removed, even if the `*-input-files.xcfilelist` file was not empty after `pod install`.

### Stack

```
   CocoaPods : 1.7.5
        Ruby : ruby 2.6.3p62 (2019-04-16 revision 67580) [universal.x86_64-darwin19]
    RubyGems : 3.0.3
        Host : Mac OS X 10.15 (19A512f)
       Xcode : 11.0 (11M374r)
         Git : git version 2.21.0 (Apple Git-120)
Ruby lib dir : /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib
Repositories : master - https://github.com/CocoaPods/Specs.git @ 39fd35e1393010fcdce3e5c74ee947b9be01ffb9
```

### Installation Source

```
Executable Path: /Users/revolt/.gems/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.4
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.1.0
cocoapods-trunk       : 1.3.1
cocoapods-try         : 1.1.0
```

### Podfile

```ruby
# frozen_string_literal: true

source 'https://github.com/CocoaPods/Specs.git'

platform :ios, '10.0'
inhibit_all_warnings!

install!('cocoapods')

def defaults
    pod 'FLEX', git: 'https://github.com/Flipboard/FLEX', commit: 'ab9515c', configurations: %w[Debug Staging]
    pod 'Fingertips', configurations: %w[Debug Staging]

    pod 'FBSDKLoginKit'
    pod 'FBSDKShareKit'
    pod 'GoogleSignIn'
    # pod 'TwitterKit'

    pod 'Crashlytics'
    pod 'Flurry-iOS-SDK/FlurrySDK'

    pod 'SQLCipher'

    pod 'PureLayout'
[...]

PR #9100, opened 2019-08-12
Fix last framework not being removed from the script

revolter:
Removing the last pod dependency that vendors a framework and running
`pod install` caused the `install_framework "${PODS_ROOT}/*.framework"`
lines not to get removed from the CocoaPods generated
`Pods-*-frameworks.sh` script.

### DIFF
--- a/lib/cocoapods/installer/xcode/pods_project_generator/aggregate_target_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/aggregate_target_installer.rb
@@ -30,7 +30,7 @@ module Pod
               # cause an App Store rejection because frameworks cannot be
               # embedded in embedded targets.
               #
-              create_embed_frameworks_script if target.includes_frameworks? && !target.requires_host_target?
+              create_embed_frameworks_script if !target.requires_host_target?
               create_bridge_support_file(native_target)
               create_copy_resources_script if target.includes_resources?
               create_acknowledgements


revolter:
I think that this is also related to #9043 (For example, https://github.com/CocoaPods/CocoaPods/issues/9043#issuecomment-514910394).

lib/cocoapods/installer/xcode/pods_project_generator/aggregate_target_installer.rb
```diff
@@ -30,7 +30,7 @@ def install!
               # cause an App Store rejection because frameworks cannot be
               # embedded in embedded targets.
               #
-              create_embed_frameworks_script if target.includes_frameworks? && !target.requires_host_target?
+              create_embed_frameworks_script if !target.requires_host_target?
```
amorde: I don't think this is the way to go - `target.includes_frameworks?` is the place that should be deciding if there are still any frameworks which require the embed frameworks script.

The logic for that is here:

https://github.com/CocoaPods/CocoaPods/blob/af83e4713771bc98dd7f903cc81c4fbc8866812c/lib/cocoapods/target/aggregate_target.rb#L227-L232

I recommend looking at that section of code to see if we are incorrectly determining that an aggregate target still has framework dependencies

revolter: There is nothing wrong with that method. It's not about incorrectly detecting the number of frameworks. Simply removing the last framework, the modified line of code makes it skip the script file update, so the framework related line of code doesn't get removed.

revolter: For example, you install `TwitterKit`, it adds:

```rb
  install_framework "${PODS_ROOT}/TwitterCore/iOS/TwitterCore.framework"
  install_framework "${PODS_ROOT}/TwitterKit/iOS/TwitterKit.framework"
fi
if [[ "$CONFIGURATION" == "Release" ]]; then
  install_framework "${PODS_ROOT}/TwitterCore/iOS/TwitterCore.framework"
  install_framework "${PODS_ROOT}/TwitterKit/iOS/TwitterKit.framework"
fi
```

you then remove `TwitterKit`, I think it removes it from the project(s), and then the `includes_frameworks?` method returns `false` and the `TwitterCore.framework` and `TwitterKit.framework` references incorrectly get left in the script.

revolter: @amorde, Was I clear enough? 😃

dnkoutso:
Can you include a test for this change?

revolter:
Ok. And I also have to find out a solution for `does not create an embed frameworks script, if the target does not have frameworks to embed`. I think it should be something like:

```ruby
if target.included_frameworks_before? && !target.includes_frameworks?
  remove_embed_frameworks_script
elsif target.includes_frameworks? && !target.requires_host_target?
  create_embed_frameworks_script
end
```

instead.

dnkoutso:
@revolter I verified this is a bug indeed. Not a very bad one since we do the right thing as far as script phases is concerned. But the files _do_ remain on the disk and are generated even if they are unused.

By the way the same applies to resources scripts as you said.

dnkoutso:
@revolter hi there! I know this is pretty old but I am reviving the project a bit, would you like to rebase and continue this PR?

CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Fix last framework not being removed from the script  
  [revolter](https://github.com/revolter)
```

</p>

revolter:
@dnkoutso, luckily, I found this again, and after looking at it again, and at the conflict during rebase (🙈), I kinda agree with @amorde.

Meaning:
- if the last pod that vendors a framework was removed, and
- if that shell script (if I'm not mistaken) only gets created if there's a pod that vendors a framework
  > `target.includes_frameworks?` is the place that should be deciding if there are still any frameworks which require the embed frameworks script

then it sounds to me like we simply have to remove the entire script, instead of keeping it without doing anything.

---

And only now I fully understand this:
> But the files do remain on the disk and are generated even if they are unused.

because, indeed, the run script phase that called that shell script does get removed.

Issue #9110, opened 2019-08-16
Project generated from app spec has bad codesigning config


watt:
* [x] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

Create a new pod with an app spec. Run `pod gen` to create a project for that app spec. Open in Xcode, and try to enable codesigning so that I can build & install to a device.

## What did you expect to happen?

I can enable pretty easily enable codesigning with my already-installed iOS Developer cert, without mucking in the build settings.

## What happened instead?

The settings in Xcode look like this:

<img width="909" alt="1 signing disabled" src="https://user-images.githubusercontent.com/100192/63192455-ffd24400-c01f-11e9-8fe8-e4e00182a7dd.png">

(Clicking "Enable Development Signing" turns on "Automatically manage signing" but doesn't set a cert.)

Go into build settings and delete the config lines shown here.

<img width="864" alt="2 lines to delete" src="https://user-images.githubusercontent.com/100192/63192690-8e46c580-c020-11e9-97a0-d1822cbf4ff8.png">

Now the signing set up looks like this, which is how I would have expected it to look fresh out of the generator.

<img width="904" alt="3 signing ready to enable" src="https://user-images.githubusercontent.com/100192/63192712-9868c400-c020-11e9-954a-529e0236ee05.png">

After picking a profile, my local iOS Developer cert is picked up automatically.

After making this change manually, these lines were deleted from the generated `Pods.xcodeproj/project.pbxproj`:

```diff
                46EB2E00000210 /* Release */ = {
                        isa = XCBuildConfiguration;
                        baseConfigurationReference = 46EB2E00000310 /* Foo.example.xcconfig */;
                        buildSettings = {
                                ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
                                CODE_SIGNING_ALLOWED = YES;
                                CODE_SIGNING_REQUIRED = YES;
                                CODE_SIGN_IDENTITY = "iPhone Developer";
-                               "CODE_SIGN_IDENTITY[sdk=appletvos*]" = "";
-                               "CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "";
-                               "CODE_SIGN_IDENTITY[sdk=watchos*]" = "";
                                CURRENT_PROJECT_VERSION = 1;
[...]

PR #9122, opened 2019-08-22
Fix iOS app spec code signing.

dnkoutso:
closes https://github.com/CocoaPods/CocoaPods/issues/9110

Today one cannot build an app spec for device due to code signing settings being broken.

I still ultimately want to re-do the target build settings but not for the 1.8.x beta.

cc @watt

### DIFF
--- a/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
@@ -445,6 +445,13 @@ module Pod
                 end
                 # For macOS we do not code sign the appbundle because we do not code sign the frameworks either.
                 configuration.build_settings['CODE_SIGN_IDENTITY'] = '' if target.platform == :osx
+                # For iOS, we delete the target_installer empty values that get set for libraries since CocoaPods will
+                # code sign the libraries manually but for apps this is not true.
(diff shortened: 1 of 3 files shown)


lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -445,6 +445,13 @@ def add_app_targets
                 end
                 # For macOS we do not code sign the appbundle because we do not code sign the frameworks either.
                 configuration.build_settings['CODE_SIGN_IDENTITY'] = '' if target.platform == :osx
+                # For iOS, we delete the target_installer empty values that get set for libraries since CocoaPods will
+                # code sign the libraries manually but for apps this is not true.
+                if target.platform == :ios
+                  configuration.build_settings.delete('CODE_SIGN_IDENTITY[sdk=appletvos*]')
```
justinseanmartin: Is there a better way to enumerate these than hardcoding the different SDKs here?

dnkoutso: nope there is no current better way.

dnkoutso:
Integration specs https://github.com/CocoaPods/cocoapods-integration-specs/pull/247

PR #9125, opened 2019-08-23
Add USE_RECURSIVE_SCRIPT_INPUTS_IN_SCRIPT_PHASES = YES to all xcconfigs

### LINKED ISSUE #8073: XCode10 (new build system) - if incremental build, Embed script doesn't run

igor-makarov:
Closes #8073 
Requires https://github.com/CocoaPods/cocoapods-integration-specs/pull/248


lib/cocoapods/target/build_settings.rb
```diff
@@ -204,6 +204,11 @@ def save_as(path)
         '${PODS_BUILD_DIR}/$(CONFIGURATION)$(EFFECTIVE_PLATFORM_NAME)'
       end
 
+      # @return [String]
+      define_build_settings_method :use_recursive_script_inputs_in_script_phases, :build_setting => true do
```
dnkoutso: I think this can be only under AggregateSettings class in here. i.e there is no need for xcconfig files from PodTargets to  have this flag. I guess it could be harmless and one could have custom script phases though...

igor-makarov: Are you saying there are no cases in which CP has script phases in Pod targets?

dnkoutso: not currently. Again unless there is a custom user `script_phase` specified in the `podspec`, in which case users will have to remember to enable `USE_RECURSIVE_SCRIPT_INPUTS_IN_SCRIPT_PHASES` so maybe its better what we have now..........

igor-makarov: It doesn't hurt anyone anyway - it's a sensible default everywhere and Apple is just being overly cautious.

dnkoutso: yeah I think we can keep it, its harmless since most pods do not specify custom user script phases.

igor-makarov:
Let me know if I modified the `CHANGELOG` correctly?

CHANGELOG.md
```diff
@@ -12,6 +12,22 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
+* None.  
+
+## 1.8.0.beta.2
```
dnkoutso: no need for this its get added during release. Just put it under `master` bug fixes for now.

dnkoutso:
@igor-makarov can you bump the submodule sha now that the PR was merged? Then I will merge this.

igor-makarov:
Bumped the integration submodule.

PR #9131, opened 2019-08-27
Move custom `Copy Headers` script phase for header mappings before `Compile Sources`.

dnkoutso:
The custom "Copy Headers" phase added to preserve header mappings within a `.framework` are executed _after_ `Compile Sources`. This is generally fine but there is a case in which it can fail if the module itself imports its own header at an expected location. The header would not be copied by then causing a build failure.

This change moves these custom script phases to _always_ execute before `Compile Sources` so all headers are in the right spot before compilation.

- [x] add tests

### DIFF
--- a/spec/unit/installer/xcode/pods_project_generator/pod_target_installer_spec.rb
+++ b/spec/unit/installer/xcode/pods_project_generator/pod_target_installer_spec.rb
@@ -1075,13 +1075,17 @@ module Pod
                   snake-umbrella.h
                 )
 
-                copy_files_build_phases = target.copy_files_build_phases.sort_by(&:name)
-                copy_files_build_phases.map(&:name).should == [
-                  'Copy . Public Headers',
+                target.build_phases.map(&:display_name).should == [
+                  'Headers',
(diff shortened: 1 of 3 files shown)


spec/unit/installer/xcode/pods_project_generator/pod_target_installer_spec.rb
```diff
@@ -1075,13 +1075,17 @@ class PodsProjectGenerator
                   snake-umbrella.h
                 )
 
-                copy_files_build_phases = target.copy_files_build_phases.sort_by(&:name)
-                copy_files_build_phases.map(&:name).should == [
-                  'Copy . Public Headers',
+                target.build_phases.map(&:display_name).should == [
```
dnkoutso: updated test. The test here no longers perform `sort` and checks against _all_ script phase names instead of just the `copy_files_build_phases`.

Verified my change work and verified the test fails without my change in terms of build phase ordering.

dnkoutso:
Will merge on green!

PR #9142, opened 2019-09-04
Make APPLICATION_EXTENSION_API_ONLY not break when performing a cached install

### LINKED ISSUE #8967: Use 'generate_multiple_pod_projects' project error
Version
Cocoapods 1.7.3
Ruby 2.6.2

Error Info
image

I am using a search engine to find a solution to this problem
Add configuration in podfile `config.build_settings['APPLICATION_EXTENSION_API_ONLY'] = 'NO'`, but did not solve the problem

Podfile
```

install! 'cocoapods',
         :generate_multiple_pod_projects => true,
         :incremental_installation => true

platform :ios, '9.0'

target 'Action Extension' do
  pod 'SVProgressHUD'
end

post_install do |installer|
    installer.pods_project.build_configuration_list.build_configurations.each do |configuration|
        configuration.build_settings['CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES'] = 'YES'
    end
            config.build_settings['ENABLE_BITCODE'] = 'NO'
            config.build_settings['APPLICATION_EXTENSION_API_ONLY'] = 'NO'
        end
    end
end

```

It seems that this configuration does not take effect in the third-party library.
```
            config.build_settings['ENABLE_BITCODE'] = 'NO'
            config.build_settings['APPLICATION_EXTENSION_API_ONLY'] = 'NO'
        end
    end
```
image

image
image

### LINKED ISSUE #9141: When performing an incremental pod install, APPLICATION_EXTENSION_API_ONLY isn't set
I believe the issue in https://github.com/CocoaPods/CocoaPods/issues/8967 is the same I'm encountering.

#### Steps to reproduce:

1. Use a project with an iOS app target and an extension target.
2. Integrate a local development pod that is linked to both using `pod install` with incremental install enabled. At this point,`APPLICATION_EXTENSION_API_ONLY` is set correctly.
3. Modify the local pod in some way.
4. Perform an incremental `pod install`. At this point,`APPLICATION_EXTENSION_API_ONLY` isn't set.

igor-makarov:
Closes #9141 
Closes #8967 

The change also moves the `APPLICATION_EXTENSION_API_ONLY` flag from the .pbxproj to the .xcconfig.


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Make APPLICATION_EXTENSION_API_ONLY not break when performing a cached incremental install  
```

</p>

lib/cocoapods/target.rb
```diff
@@ -46,6 +46,10 @@ class Target
     attr_reader :build_type
     private :build_type
 
+    # @return [Boolean] whether the target .
```
dnkoutso: incomplete doc

lib/cocoapods/target.rb
```diff
@@ -46,6 +46,10 @@ class Target
...
+    #
+    attr_accessor :application_extension_api_only
```
dnkoutso: we've tried hard to make those `attr_reader` instead of setting those fields at random places in the code. Can you update that to be an `attr_reader`?

dnkoutso:
Needs a changelog entry

dnkoutso:
cc @sebastianv1

igor-makarov:
Made the property readonly, added changelog, fixed the code comments.

lib/cocoapods/installer/analyzer.rb
```diff
@@ -438,6 +438,18 @@ def generate_targets(resolver_specs_by_target, target_inspections)
...
         end
+
+        aggregate_targets.each do |aggregate_target|
+          is_app_extension = !(aggregate_target.user_targets.map(&:symbol_type) &
+              [:app_extension, :watch_extension, :watch2_extension, :tv_extension, :messages_extension]).empty?
+          is_app_extension ||= aggregate_target.user_targets.any? { |ut| ut.common_resolved_build_setting('APPLICATION_EXTENSION_API_ONLY') == 'YES' }
+
+          next unless is_app_extension
+
+          aggregate_target.mark_application_extension_api_only
```
sebastianv1: Can the logic for determining `is_app_extension` live inside `AggregateTarget`. It looks like you're only using properties off `aggregate_target`. This allows you to remove `mark_application_extension_api_only` which is kinda ugly and get rid of the default `false` value.

sebastianv1:
Please add a spec for `project_cache_analyzer` thats verifies it'll mark the target as dirty for incremental installs. I'm surprising this doesn't work since we mark a target as dirty if any of the build settings change.

lib/cocoapods/installer/analyzer.rb
igor-makarov: While the logic for `is_app_extension` can be put in AggregateTarget, `mark_application_extension_api_only` cannot be done away with - PodTarget also needs to be marked and it has no idea who its parent is.

igor-makarov:
The cache analyzer worked correctly, the problem was that the app extension flag was being set after generation in some obscure corner which was bypassed in an incremental cached install.

lib/cocoapods/installer/analyzer.rb
sebastianv1: Ah bummer.

igor-makarov: I had a chat with @segiddins about it and it makes sense for the Analyzer to be the single one responsible for this weird flag.

spec/unit/installer/analyzer_spec.rb
```diff
@@ -2029,6 +2029,149 @@ module Pod
...
+
+          it 'configures APPLICATION_EXTENSION_API_ONLY for watch app extension targets' do
+            @user_project = Xcodeproj::Project.open(SpecHelper.create_sample_app_copy_from_fixture('Sample Extensions Project'))
+            targets = @user_project.targets
+            targets.delete(targets.find('Sample Extensions Project').first)
+            extension_target = targets.find('Today Extension').first
+            extension_target.product_type = 'com.apple.product-type.watchkit2-extension'
+            extension_target.name = 'watchOS Extension Target'
+            extension_target.symbol_type.should == :watch2_extension
+            @user_project.save
```
dnkoutso: does this actually leave the file system dirty after this test runs?

igor-makarov: It doesn't - this is a temp copy, see above.

PR #9150, opened 2019-09-08
Exclude the local spec-repos directory from Time Machine backups

### LINKED ISSUE #8308: Exclude local spec-repos directory from backups
The size of the [CocoaPods master repo](https://github.com/CocoaPods/Specs) on disk is well over 1 GB right now (see also #7046). Not only that, the repo is also included in Time Machine backups as can be verified by `tmutil isexcluded ~/.cocoapods/repos/master`.

Since CocoaPods repos are merely caches, I’d suggest excluding the local spec-repos directory from backups. This can be achieved using `tmutil addexclusion ~/.cocoapods/repos` which adds the extended attribute `com.apple.metadata:com_apple_backup_excludeItem`. This attribute not only helps Time Machine but can also be picked up by other backup applications.

If there’s any interest in this feature, I’d be happy to give the implementation a try myself.

---

✓ I’ve read and understood the [contributing guidelines](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md) and have done my best effort to follow 🌈

sonicdoe:
Excludes the local spec-repos directory (`~/.cocoapods/repos`, by default) from Time Machine Backups using `tmutil addexclusion`. Resolves #8308.

Alternatively (and ideally), the local spec-repos directory could be placed under the `~/Library/Caches` directory (which is usually excluded by Time Machine as well as many other backup applications). However, that would arguably be a more drastic change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Paul Beusterien](https://github.com/paulb777)
   [#9632](https://github.com/CocoaPods/CocoaPods/pull/9632)
 
+* Exclude the local spec-repos directory from Time Machine Backups.  
+  [Jakob Krigovsky](https://github.com/sonicdoe)
+  [#8308](https://github.com/CocoaPods/CocoaPods/issues/8308)
+
 ##### Bug Fixes
 
 * Re-implement `bcsymbolmap` copying to avoid duplicate outputs.  
--- a/lib/cocoapods/config.rb
+++ b/lib/cocoapods/config.rb
@@ -314,6 +314,18 @@ module Pod
       nil
     end
 
+    # Excludes the given dir from Time Machine backups.
+    #
+    # @param  [Pathname] dir
+    #         The directory to exclude from Time Machine backups.
+    #
+    # @return [void]
+    #
+    def exclude_from_backup(dir)
+      return if Gem.win_platform?
+      system('tmutil', 'addexclusion', dir.to_s, %i(out err) => File::NULL)
+    end
+
     public
 
     #-------------------------------------------------------------------------#
--- a/spec/functional/command/repo/update_spec.rb
+++ b/spec/functional/command/repo/update_spec.rb
@@ -39,6 +39,16 @@ module Pod
       (repo2 + 'README').read.should.include 'Updated'
     end
 
+    # Conditionally skip the test if `tmutil` is not available.
+    has_tmutil = system('tmutil', 'version', :out => File::NULL)
+    cit = has_tmutil ? method(:it) : method(:xit)
+    cit.call 'excludes the spec-repo from Time Machine backups' do
+      repo_make('repo1')
+      repo_clone('repo1', 'repo2')
+      run_command('repo', 'update', 'repo2')
+      `tmutil isexcluded #{config.repos_dir + 'repo2'}`.chomp.should.start_with?('[Excluded]')
+    end
+
     it 'repo updates do not fail when executed in parallel' do
       repo1 = repo_make('repo1')
       repo_clone('repo1', 'repo2')
--- a/lib/cocoapods/command/repo/update.rb
+++ b/lib/cocoapods/command/repo/update.rb
@@ -21,6 +21,17 @@ module Pod
         def run
           show_output = !config.silent?
           config.sources_manager.update(@name, show_output)
+          exclude_repos_dir_from_backup
+        end
+
+        private
+
+        # Excludes the repos directory from backups.
+        #
+        # @return [void]
+        #
+        def exclude_repos_dir_from_backup
+          config.exclude_from_backup(config.repos_dir)
         end
       end
     end
--- a/spec/unit/config_spec.rb
+++ b/spec/unit/config_spec.rb
@@ -289,6 +289,26 @@ module Pod
         end
       end
 
+      describe '#exclude_from_backup' do
+        # Conditionally skip the test if `tmutil` is not available.
+        has_tmutil = system('tmutil', 'version', :out => File::NULL)
+        cit = has_tmutil ? method(:it) : method(:xit)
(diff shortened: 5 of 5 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
+* Exclude the local spec-repos directory from Time Machine Backups.
```
dnkoutso: please add two empty spaces at the end of this sentence for markdown formatting.

lib/cocoapods/command/repo/add.rb
```diff
@@ -67,6 +68,13 @@ def create_repos_dir
             "#{e.class.name}: #{e.message}"
         end
 
+        # Excludes the repos directory from backups.
+        #
+        # @return [void]
```
dnkoutso: nit add extra line with `#` after this.

lib/cocoapods/config.rb
```diff
@@ -314,6 +314,16 @@ def podfile_path_in_dir(dir)
       nil
     end
 
+    # Excludes the given dir from Time Machine backups.
+    #
+    # @param  [Pathname] dir
+    #         The directory to exclude from Time Machine backups.
+    #
+    # @return [void]
```
dnkoutso: nit: same here.

lib/cocoapods/command/repo/add.rb
```diff
@@ -45,6 +45,7 @@ def run
           section << " (branch `#{@branch}`)" if @branch
           UI.section(section) do
             create_repos_dir
+            exclude_repos_dir_from_backup
```
dnkoutso: what would happen for existing repos?

sonicdoe: Since `exclude_repos_dir_from_backup` excludes the whole local spec-repos directory from Time Machine backups, all existing repos would get excluded as well.

In other words, the first time you run `pod repo add` (with these changes), `~/.cocoapods/repos` (which includes all existing repos) would get excluded from backups.

dnkoutso: @sonicdoe hmm I can imagine many systems already having created the repos they need. What if we put that in the `pod repo update` (`update.rb`) instead?

sonicdoe: In my eyes, `pod repo update` doesn’t seem like the right place since this should be a one-time operation when the repos directory is created. However, if it helps users coming from older versions and you’re fine with it, I’d be glad to change this part.

Relatedly, what would you suggest if we were to move the repos directory to `~/Library/Caches` (as I’ve suggested in the [original post](https://github.com/CocoaPods/CocoaPods/pull/9150#issue-315284701))? Would you provide some sort of migration for existing users or just keep existing repos where they are?

dnkoutso: > Relatedly, what would you suggest if we were to move the repos directory to ~/Library/Caches (as I’ve suggested in the original post)? Would you provide some sort of migration for existing users or just keep existing repos where they are?

I say it will be somewhat complicated to do now, perhaps for a 2.0 release. We _do_ use trunk now so things are even better in 1.8.x but the migration might be a bit painful. Needless to say the git repo is so huge people who have managed to clone it in bad network conditions probably don't want to try again.

> In my eyes, pod repo update doesn’t seem like the right place since this should be a one-time operation when the repos directory is created. However, if it helps users coming from older versions and you’re fine with it, I’d be glad to change this part.

I kind of agree with you but I can't find a better spot to ensure this always happens and migrates the old users as well.

dnkoutso: Perhaps we can add it to the `installer.rb`?

sonicdoe: I’ve now moved it to `pod repo update`. I think this is a good trade-off between almost never being called by existing users (`pod repo add`) and being called quite often and outside of the context of spec-repos (`pod install`).

spec/functional/command/repo/update_spec.rb
```diff
@@ -39,6 +39,16 @@ module Pod
       (repo2 + 'README').read.should.include 'Updated'
     end
 
+    # Conditionally skip the test if `tmutil` is not available.
+    has_tmutil = system('tmutil version', :out => File::NULL)
```
segiddins: shoudl not be using a shell here

sonicdoe: Thanks, changed to `system('tmutil', 'version')`.

dnkoutso:
This needs a rebase

sonicdoe:
Done, rebased.

lib/cocoapods/config.rb
```diff
@@ -314,6 +314,17 @@ def podfile_path_in_dir(dir)
...
+    #
+    def exclude_from_backup(dir)
+      system('tmutil', 'addexclusion', dir.to_s, %i(out err) => File::NULL)
```
dnkoutso, line 325: should we check `Gem.win_platform?` before invoking this?

sonicdoe: We could also explicitly check for macOS but it might require pulling in [the OS gem](https://github.com/rdp/os) because [reliably detecting the operating system isn’t that easy](https://stackoverflow.com/q/170956). As we don’t care about the command’s exit status (and are basically just hoping for the best), invoking `tmutil` shouldn’t hurt. I agree, though, that it doesn’t make much sense to call it on any other operating system since `tmutil` definitely won’t be there.

dnkoutso: I dont mind personally if its best effort. We have used `Gem.win_platform?` in a couple of other spots like in `header_store.rb`.

sonicdoe: Done, I’ve added `return if Gem.win_platform?`.

sonicdoe:
Thank you for reviewing 👍 Rebased!

PR #9153, opened 2019-09-10
Correctly set deployment target for non library specs even if the root spec does not specify one.

dnkoutso:
Deployment target for non library spec is explicitly used if one is specified, otherwise the code used to use the one from the parent spec. However, is it not guaranteed that the parent spec has a deployment target specified.

The `analyzer` sets a default deployment target if that is the case on the `PodTarget` itself. This PR ensures we use that if the deployment target for a non library spec is `nil`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
+* Correctly set deployment target for non library specs even if the root spec does not specify one.  
+  [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
+  [#9153](https://github.com/CocoaPods/CocoaPods/pull/9153)
+
 * Make `APPLICATION_EXTENSION_API_ONLY` build setting not break when performing a cached incremental install.  
   [Igor Makarov](https://github.com/igor-makarov)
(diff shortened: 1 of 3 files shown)


spec/unit/target/pod_target_spec.rb
```diff
@@ -753,6 +753,14 @@ module Pod
           @pod_target.deployment_target_for_non_library_spec(@watermelon_spec.test_specs.first).to_s.should == '8.0'
           @pod_target.deployment_target_for_non_library_spec(@watermelon_spec.app_specs.first).to_s.should == '8.0'
         end
+
+        it 'returns the determined deployment target even if the podspec does not explicitly specify one' do
```
dnkoutso: verified this goes red without my change.

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
+* Correctly set deployment target for non library specs even if the root spec does not specify one.  
+  [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
+  [#9153](https://github.com/CocoaPods/CocoaPods/pull/9153)g
```
paulb777: Fix typo at end of line

dnkoutso: thanks done!

Issue #9151, opened 2019-09-09
CocoaPoads client does not support any authentication, even simple one by login/password during fetching Spec files and metadata files(like all_pods.txt) from CDN source(s).


igorudovika:
I am trying new CDN approach in restricted zones. To work with custom created CDN source or to work with public https://cdn.cocoapods.org one via proxy server. I had a look in source code https://github.com/CocoaPods/Core/blob/master/lib/cocoapods-core/cdn_source.rb and found that during get request CocoaPods client does not  try to set login/password and always works with anonymous user. I can see rest client Fingertips/nap in source code and it can use login/password. In same time seems it is possible to use authentication for fetching pod libraries sources.

dnkoutso:
cc @igor-makarov

igor-makarov:
The server does not require authentication. It's public. 

Using a proxy is done by setting the `http_proxy` environment variable, as is customary in Ruby.

igorudovika:
> The server does not require authentication. It's public.
> 
> Using a proxy is done by setting the `http_proxy` environment variable, as is customary in Ruby.

@igor-makarov thanks for answer, yep, I am familiar `http_proxy` environment variable. I talked more about different kinds of proxying repository managers.  In same time as far CDN approach is becoming main and default one I am thinking how to work with private Specs repositories.

igor-makarov:
Private spec repos can use the original git method, it wasn't deprecated and is still fully supported. We decided they tend to be much smaller and therefore not in need of CDN. 

What size are your private spec repos? Care to elaborate on your use cases?

igorudovika:
Yep, I am understanding that main target of CDN approach is resolving problem with huge size of master Spec repo and that for now it is still possible to use git approach. I just would like to see CDN approach more flexible and seems now(on stage of first implementations) is good time to do it. So in future for different use cases we all will not need to choose different approaches because each of them have limitations that was skipped initially.

igor-makarov:
I think using the CDN is an optimization based on the main registry's size. 

git is much more suited for everything else, precisely because it's a well-tested and flexible tool that can give users all the authentication and proxying options they need.

jlstephens89:
[...]

PR #9164, opened 2019-09-18
Check CDN repo URL procedure changed with common one

igorudovika:
Very small improvement that will simplify resolve https://github.com/CocoaPods/CocoaPods/issues/9151 issue

### DIFF
--- a/lib/cocoapods/sources_manager.rb
+++ b/lib/cocoapods/sources_manager.rb
@@ -68,7 +68,7 @@ module Pod
       #         The URL of the source.
       #
       def cdn_url?(url)
-        url =~ %r{^https:\/\/} &&
+        url =~ %r{^https?:\/\/} &&
           Pod::HTTP.validate_url(url + '/all_pods.txt', nil).ok?
       rescue => e
         raise Informative, "Couldn't determine repo type for URL: `#{url}`: #{e}"
--- a/lib/cocoapods/sources_manager.rb
+++ b/lib/cocoapods/sources_manager.rb
@@ -69,7 +69,7 @@ module Pod
       #
       def cdn_url?(url)
         url =~ %r{^https:\/\/} &&
-          REST.head(url + '/all_pods.txt').ok?
+          Pod::HTTP.validate_url(url + '/all_pods.txt', nil).ok?
       rescue => e
         raise Informative, "Couldn't determine repo type for URL: `#{url}`: #{e}"
       end


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Check CDN repo URL procedure changed with common one  
  [igorudovika](https://github.com/igorudovika)
```

</p>

dnkoutso:
cc @igor-makarov

igor-makarov:
Seems good, didn't know about that one!

dnkoutso:
@igor-makarov any strong opinions on shipping this with 1.8 or 1.9?

dnkoutso:
I guess its related to #9151 also and when we want to ship that.

igor-makarov:
1.9 seems more on the safe side since it's part of some bigger work.

dnkoutso:
Sounds good.

dnkoutso:
This is good to merge right @igor-makarov? i.e it does not depend on something else landing?

igor-makarov:
Doesn't depend on other PRs as far as I see.

dnkoutso:
We are using latest master and this change actually returns a legitimate `git` repo as if it were CDN.

dnkoutso:
This change now makes CocoaPods follows redirects which can cause a false positive in which a 200 result is returned and making CocoaPods think that the URL that is meant to be a `git` repo is CDN.

@igorudovika @igor-makarov can you weigh in on how this can be fixed? We hit this case internally.

igor-makarov:
Can you be more specific? What gets redirected where?

dnkoutso:
We use Stash internally and the URL for our source repo is `https://.../something/something.git`

In the code we do `https://.../something/something.git/all_pods.txt` which then redirects to another page due to this change here and we get a false positive which then fails in other ways later in the process.

igor-makarov:
I've got no quick fix so I guess revert and let's try to figure out a grand unified HTTP client refactor later 🙃

igorudovika:
I can work with fix for this issue.
But problem is still not clear for me. Seems `https://.../something/something.git/all_pods.txt` matches for old regex too. So mine revert of mine fix will not help

igor-makarov:
The problem is, old logic was using an incorrect test but it didn't follow redirect so no false positive.

New logic follows redirects and therefore has a false positive.

But fundamentally the test for CDN is not correct.

igor-makarov:
Perhaps check for content type header?

dnkoutso:
I am going to revert temporarily and then we can discuss a proper fix.

igorudovika:
Thanks @igor-makarov  I got you, so problem is in `validate_url` it tries to get `actual_url` by taking into account redirects.

dnkoutso:
yeap. And I think just checking for the 200 status code is still not sufficient.

igor-makarov:
Maybe it should GET the CP meta YML and actually parse it?

igorudovika:
Do you mean CocoaPods-version.yml file?
if yes, what tokens I need to check in order to validate content? Existing of:
`min:`, `last:` and `prefix_lengths:` tokens?

igor-makarov:
Yeah, that seems decently robust.

igorudovika:
ok, then, let me prepare new fix.

dnkoutso:
Just try to parse the whole file as YAML document might be sufficient.

igor-makarov:
Yeah, the parser will throw if it isn't a valid meta file.

igorudovika:
ok, I will do

igorudovika:
Added new PR https://github.com/CocoaPods/CocoaPods/pull/9212
Seems for now this change does not contain any logic that I am interesting but perhaps you will find it useful for project. After Typhoeus lib introducing in Core repo I will try to change this logic again with new PR

PR #9191, opened 2019-09-26
Move target build settings into the generated .xcconfig files

amorde:
This moves most of the custom build settings applied during installation into the target's `.xcconfig` files.

Reasons to do this:

1. Consistency - we already put customized build settings in `.xcconfig`s
2. Troublshooting - it's much easier for issue authors to provide the contents of a pod target's `.xcconfig` (and to diff between two versions) than to pull build settings from a project file
3. Duplication - there were a couple settings that were being placed in both `.xcconfig` and `.pbxproj` files

### DIFF
--- a/lib/cocoapods/installer/xcode/pods_project_generator/aggregate_target_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/aggregate_target_installer.rb
@@ -6,6 +6,11 @@ module Pod
         # project and the relative support files.
         #
         class AggregateTargetInstaller < TargetInstaller
+
+          # @return [AggregateTarget] the aggregate target to be installed
+          #
+          attr_reader :target
+
           # Creates the target in the Pods project and the relative support files.
           #
           # @return [TargetInstallationResult] the result of the installation of this target.
@@ -15,13 +20,12 @@ module Pod
               native_target = add_target
               create_support_files_dir
               create_support_files_group
-              create_xcconfig_file(native_target)
               if target.build_as_framework?
                 create_info_plist_file(target.info_plist_path, native_target, target.version, target.platform)
-                create_module_map(native_target)
+                create_module_map
                 create_umbrella_header(native_target)
               elsif target.uses_swift?
-                create_module_map(native_target)
+                create_module_map
                 create_umbrella_header(native_target)
               end
               # Because embedded targets live in their host target, CocoaPods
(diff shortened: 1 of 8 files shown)


lib/cocoapods/installer/xcode/pods_project_generator/aggregate_target_installer.rb
```diff
@@ -6,6 +6,11 @@ class PodsProjectGenerator
         # project and the relative support files.
         #
         class AggregateTargetInstaller < TargetInstaller
+
+          # @return [AggregateTarget] the aggregate target to be installed
+          #
+          attr_reader :target
```
amorde: Added just for the YARD docs, this is inherited

lib/cocoapods/installer/xcode/pods_project_generator/aggregate_target_installer.rb
```diff
@@ -35,6 +39,8 @@ def install!
               create_copy_resources_script if target.includes_resources?
               create_acknowledgements
               create_dummy_source(native_target)
+              remove_build_setting_overrides(config_hash, native_target)
+              create_xcconfig_file(native_target, Xcodeproj::Config.new(config_hash))
```
amorde: this is moved to the end so each installation step has a chance to modify `config_hash`

lib/cocoapods/installer/xcode/pods_project_generator/aggregate_target_installer.rb
```diff
@@ -50,21 +56,27 @@ def target_definition
...
           def custom_build_settings
             settings = {
               'CODE_SIGN_IDENTITY[sdk=appletvos*]' => '',
               'CODE_SIGN_IDENTITY[sdk=iphoneos*]'  => '',
               'CODE_SIGN_IDENTITY[sdk=watchos*]'   => '',
+              'PRODUCT_BUNDLE_IDENTIFIER'          => 'org.cocoapods.${PRODUCT_NAME:rfc1034identifier}',
+              'SKIP_INSTALL'                       => 'YES',
               'MACH_O_TYPE'                        => 'staticlib',
               'OTHER_LDFLAGS'                      => '',
               'OTHER_LIBTOOLFLAGS'                 => '',
```
amorde: these are left as build settings due to some checks we have for user projects that override these

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -132,19 +132,6 @@ def install!
...
-          # Adds the target for the library to the Pods project with the
-          # appropriate build configurations.
-          #
-          # @note   Overrides the superclass implementation to remove settings that are set in the pod target xcconfig
-          #
-          # @return [PBXNativeTarget] the native target that was added.
-          #
-          def add_target
-            super.tap do |native_target|
-              remove_pod_target_xcconfig_overrides_from_target(target.build_settings, native_target)
```
amorde: this now happens at the end of installation, so no longer needed here

Issue #9216, opened 2019-10-02
pod repo update crash on 1.8.x


jcesarmobile:
* [x] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

I know CocoaPods 1.8 use the CDN by default, but since `pod setup` does nothing, if you have a Podfile with `source 'https://github.com/CocoaPods/Specs.git'`, when you run `pod repo update` it will fail.

`pod setup` should continue working or there should be a new command to get the git based source downloaded.

## What did you do?

I have an autogenerated Podfile with `source 'https://github.com/CocoaPods/Specs.git'`, when I run `pod repo update` it fails.

## What did you expect to happen?

To get the git sources as on previous versions

## What happened instead?

Got this error:
No such file or directory ~/.cocoapods/repos/Spec_Lock

### Stack

```
   CocoaPods : 1.8.0
        Ruby : ruby 2.3.7p456 (2018-03-28 revision 63024) [x86_64-darwin17]
    RubyGems : 2.5.2.3
        Host : Mac OS X 10.14.6 (18G95)
       Xcode : 11.0 (11A420a)
         Git : git version 2.21.0 (Apple Git-122)
Ruby lib dir : /Users/jcesarmobile/.rbenv/versions/2.3.7/lib
Repositories: 
```
### Plugins

```
cocoapods-deintegrate : 1.0.4
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.1.0
cocoapods-trunk       : 1.4.0
cocoapods-try         : 1.1.0
```

dnkoutso:
hmmmmmm can you list out exactly all steps you do from a clean CocoaPods install? You may have found a bug.

`pod setup` intentionally won't do something anymore as it has nothing to do. However, the error you've seen I've seen in another repo so I think you have uncovered a bug.

jcesarmobile:
Yeah, is a bug, there is no way of setting up the local copy on cocopadods 1.8.x, but it will try to use it if you have the github url in the Podfile.
I guess a workaround is to manually clone the repo, but would be good if there was a command for it.

dnkoutso:
@jcesarmobile so I can't reproduce your issue. It seems you might be hitting this issue https://github.com/CocoaPods/CocoaPods/issues/9209

Can you paste your `pod repo list`?

[...]

PR #9234, opened 2019-10-09
Do not crash if the repos dir is not setup.

dnkoutso:
closes https://github.com/CocoaPods/CocoaPods/issues/9216

### DIFF
--- a/lib/cocoapods/sources_manager.rb
+++ b/lib/cocoapods/sources_manager.rb
@@ -113,9 +113,13 @@ module Pod
         end
 
         changed_spec_paths = {}
-        # Ceate the Spec_Lock file if needed and lock it so that concurrent
+
+        # Do not perform an update if the repos dir has not been setup yet.
+        return unless repos_dir.exist?
+
+        # Create the Spec_Lock file if needed and lock it so that concurrent
         # repo updates do not cause each other to fail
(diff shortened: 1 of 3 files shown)


lib/cocoapods/sources_manager.rb
```diff
@@ -113,9 +113,13 @@ def update(source_name = nil, show_output = false)
...
         changed_spec_paths = {}
-        # Ceate the Spec_Lock file if needed and lock it so that concurrent
+
+        # Do not perform an update if the repos dir has not been setup yet.
+        return unless repos_dir.exist?
+
+        # Create the Spec_Lock file if needed and lock it so that concurrent
         # repo updates do not cause each other to fail
-        File.open("#{Config.instance.repos_dir}/Spec_Lock", File::CREAT) do |f|
+        File.open("#{repos_dir}/Spec_Lock", File::CREAT) do |f|
```
paulb777: Why this change?

dnkoutso: I figured thats what the `sources_manager` get initialized with. I dont think we will ever have this case but if you had 2 `sources_manager` instances with two different folders they would have to wait for each other for the operation to be completed but they shouldn't have to?

PR #9241, opened 2019-10-10
Add user projects to workspace during incremental install.

### LINKED ISSUE #9237: There is not any projects in xcworkspace if recreate the xcworkspace
After `pod install`, everything is OK. If I delete the xcworkspace bundler, then rerun the `pod install`, I will get a new xcworkspace bundler. But there is not any user projects in the xcworkspace, only a `Pods` project.

I check the code in `user_project_integrator.rb`, I find that:
```
      def user_project_paths_to_integrate
        targets_to_integrate.map(&:user_project_path).compact.uniq
      end
      def create_workspace
        all_projects = user_project_paths_to_integrate.sort.push(sandbox.project_path).uniq
        file_references = all_projects.map do |path|
          relative_path = path.relative_path_from(workspace_path.dirname).to_s
          Xcodeproj::Workspace::FileReference.new(relative_path, 'group')
        end

        if workspace_path.exist?
          workspace = Xcodeproj::Workspace.new_from_xcworkspace(workspace_path)
          new_file_references = file_references - workspace.file_references
          unless new_file_references.empty?
            new_file_references.each { |fr| workspace << fr }
            workspace.save_as(workspace_path)
          end

        else
          UI.notice "Please close any current Xcode sessions and use `#{workspace_path.basename}` for this project from now on."
          workspace = Xcodeproj::Workspace.new(*file_references)
          workspace.save_as(workspace_path)
        end
      end
```

Because I have the all cached pod projects, include the Pods.xcodeproj. So the `targets_to_integrate` is empty, it will get an empty array from the `user_project_paths_to_integrate`, and will not add user project into the xcworkspace.

It seems that it is a small bug.

I am using CocoaPods v1.7.

sebastianv1:
closes #9237

During incremental installation, if the set of `targets_to_generate` was empty and the workspace was missing, we wouldn't add the user projects to the workspace and be missing.

This changes `user_project_paths_to_integrate` to map over `targets` instead of the subset `targets_to_generate` to ensure all user projects will be added. There is existing logic to make sure that duplicate file references to user projects are removed.

### DIFF
--- a/lib/cocoapods/installer/user_project_integrator.rb
+++ b/lib/cocoapods/installer/user_project_integrator.rb
@@ -92,7 +92,7 @@ module Pod
       # @return [void]
       #
       def create_workspace
-        all_projects = user_project_paths_to_integrate.sort.push(sandbox.project_path).uniq
+        all_projects = user_project_paths.sort.push(sandbox.project_path).uniq
         file_references = all_projects.map do |path|
           relative_path = path.relative_path_from(workspace_path.dirname).to_s
           Xcodeproj::Workspace::FileReference.new(relative_path, 'group')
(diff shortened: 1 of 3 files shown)


lib/cocoapods/installer/user_project_integrator.rb
```diff
@@ -227,7 +227,7 @@ def workspace_path
       # @note   Empty target definitions are ignored.
       #
       def user_project_paths_to_integrate
-        targets_to_integrate.map(&:user_project_path).compact.uniq
+        targets.map(&:user_project_path).compact.uniq
```
dnkoutso: this method name was renamed generally to explicitly specify the `to_integrate`, can we inline this now? Feels weird to to have to `_integrate` but the full list of targets there.

PR #9265, opened 2019-10-16
Fix app spec frameworks script

segiddins:
The bug was that we were passing in the config name instead of the type to a method that was expecting the type, and because we were constructing arrays via splats instead of `concat`, it was silently coercing `nil` to `[]`. Integration specs change shows this is fixed, exception added in dep lookup will mean this cant silently start failing again.

Fixing a regression introduced on master, hence no CHANGELOG.

### DIFF
--- a/lib/cocoapods/target/pod_target.rb
+++ b/lib/cocoapods/target/pod_target.rb
@@ -712,7 +712,7 @@ module Pod
         hash[nil] = hash.each_value.reduce(Set.new, &:|).to_a
         hash
       end
-      @recursive_dependent_targets[configuration]
+      @recursive_dependent_targets.fetch(configuration) { raise ArgumentError, "No configuration #{configuration} for #{self}, known configurations are #{@recursive_dependent_targets.keys}" }
     end
 
     def _add_recursive_dependent_targets(set, configuration: nil)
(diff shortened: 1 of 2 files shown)


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Fix app spec frameworks script  
  [segiddins](https://github.com/segiddins)
```

</p>

lib/cocoapods/target/pod_target.rb
```diff
@@ -712,7 +712,7 @@ def recursive_dependent_targets(configuration: nil)
         hash[nil] = hash.each_value.reduce(Set.new, &:|).to_a
         hash
       end
-      @recursive_dependent_targets[configuration]
+      @recursive_dependent_targets.fetch(configuration) { raise ArgumentError, "No configuration #{configuration} for #{self}, known configurations are #{@recursive_dependent_targets.keys}" }
```
dnkoutso: does the same fix need to be applied to `recursive_test_dependent_targets` below?

segiddins: nope, because that's only called by methods that also call `recursive_dependent_targets`

Issue #9023, opened 2019-07-19
Cannot get ObjC and Swift to interop


michaeleisel:
* [ ] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report
I want to call some C/ObjC functions from within Swift, and I don't want to expose these functions to consumers of my cocoapod. When I try to use a bridging header, Xcode tells me "error: using bridging headers with framework targets is unsupported". When I try to use a custom module map, cocoapods tells me that it won't work for static libraries, and I'm not sure if it works that way anyways. What should I do?

## What did you do?

Run `pod install`, with a brand new .xcodeproj that depends on this library and see that there's a build error

## What did you expect to happen?

It builds

## What happened instead?

It does not build

### Stack

```
   CocoaPods : 1.7.3
        Ruby : ruby 2.6.0p0 (2018-12-25 revision 66547) [universal.x86_64-darwin19]
    RubyGems : 3.0.1
        Host : Mac OS X 10.15 (19A471t)
       Xcode : 11.0 (11M336w)
         Git : git version 2.21.0
Ruby lib dir : /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib
Repositories : artsy - https://github.com/artsy/Specs.git @ bdd1edc21cd8e834919636d4ace59e4bf362f2fc
               master - https://github.com/CocoaPods/Specs.git @ 3ec921a1a1f1cd8a1304d33bfdee5947e201c75f
```

### Installation Source

```
Executable Path: /usr/local/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.4
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.1.0
cocoapods-trunk       : 1.3.1
cocoapods-try         : 1.1.0
```

### Podfile

```ruby
# Uncomment the next line to define a global platform for your project
# platform :ios, '9.0'

target 'ZippyRun' do
  # Comment the next line if you don't want to use dynamic frameworks
  # use_frameworks!
  pod 'ZippyJSON', :git => 'https://github.com/michaeleisel/zippyjson'


  # Pods for ZippyRun

  target 'ZippyRunTests' do
    inherit! :search_paths
    # Pods for testing
  end

end
```

## Project that demonstrates the issue

[...]

Issue #7594, opened 2018-04-06
[1.5.0, static libs] Mixed Objective-C/Swift pod fails to import Swift from Objective-C


MaximeLM:
* [🌈] I've read and understood the [*CONTRIBUTING guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

First, a lot of thanks for this feature ❤️

# Report

## What did you do?

* Update to CocoaPods 1.5.0
* Remove `use_frameworks!` from Podfile

Pod setup: one Pod `TestLib` with mixed Swift/Objective-C code. Objective-C files import Swift code with `#import <TestLib/TestLib-Swift.h>`.

## What did you expect to happen?

The project would build.

## What happened instead?

The project failed to build with error `'TestLib/TestLib-Swift.h' file not found`.

### Stack

```
   CocoaPods : 1.5.0
        Ruby : ruby 2.2.1p85 (2015-02-26 revision 49769) [x86_64-darwin14]
    RubyGems : 2.4.6
        Host : Mac OS X 10.13.4 (17E199)
       Xcode : 9.3 (9E145)
         Git : git version 2.15.1 (Apple Git-101)
Ruby lib dir : /Users/maximelemoine/.rvm/rubies/ruby-2.2.1/lib
Repositories : master - https://github.com/CocoaPods/Specs.git @ 4ddb0bcf3b52ed9ff3e40a475fb06d9dc264bcfe
```

### Installation Source

```
Executable Path: /Users/maximelemoine/.rvm/gems/ruby-2.2.1/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.2
cocoapods-packager    : 1.5.0
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.0.0
cocoapods-trunk       : 1.3.0
cocoapods-try         : 1.1.0
```

### Podfile

```ruby
# Uncomment the next line to define a global platform for your project
platform :osx, '10.10'

#use_frameworks!

target 'TestSwiftImport' do
  
  pod 'TestLib', :path => '.'

end
```

## Project that demonstrates the issue

https://github.com/MaximeLM/TestSwiftImport

dnkoutso:
This is a known issue you can workaround it via `#import "TestLib-Swift.h"` instead.

Zmeu213:
... or just use cocoapods 1.4

[...]

PR #9363, opened 2019-11-23
[Pod Target Installer] Make private Objective-C files available to Swift

abiligiri:
🌈

This is a proposal to allow ObjC and Swift to interop in pod target

https://github.com/CocoaPods/CocoaPods/issues/9023
https://github.com/CocoaPods/CocoaPods/issues/7594

These changes do make the private Objective-C files available to Swift but they are also exposed outside the pod target with a _Private suffix. If this limitation is acceptable, I can add more tests.

Suggestions to eliminate the limitation are appreciated.

### DIFF
--- a/lib/cocoapods/generator/module_map.rb
+++ b/lib/cocoapods/generator/module_map.rb
@@ -40,11 +40,10 @@ module Pod
       #
       # @param  [PodTarget, AggregateTarget] target @see target
       #
-      def initialize(target)
+      def initialize(target, skip_umbrella = false, private_module_map = false)
         @target = target
-        @headers = [
-          Header.new(target.umbrella_header_path.basename, true),
-        ]
+        @headers = skip_umbrella ? [] : [Header.new(target.umbrella_header_path.basename, true)]
+        @private_module_map = private_module_map
       end
 
       # Generates and saves the Info.plist to the given path.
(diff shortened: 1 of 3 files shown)


dnkoutso:
@abiligiri why wouldn't we want to add a `s.private_module_map` DSL instead? Would it solve the issue?

booiiing:
@dnkoutso if I understand this PR correctly, it "only" adds another generated module map to make Obj-C accessible from Swift. so it would solve the interop part of #9023.
however, before an option for a private_module_map would be introduced (which is a custom module map, while this is generated), custom module maps would need to be supported for static targets in general. I believe this is another aspect.

@abiligiri I would still suggest an improvement: instead of adding the "_Private" suffix, the generated module map could be added to the XCode build setting "MODULEMAP_PRIVATE_FILE".

dnkoutso:
@booiiing thanks for the clarification!

abiligiri:
will look into this. thanks @booiiing and @dnkoutso

abiligiri:
@booiiing I tried your suggestion of removing the _Private suffix and setting the MODULEMAP_PRIVATE_FILE. It doesn't work. The headers are not visible with in the pod target

abiligiri:
@dnkoutso I created a sample project here https://github.com/abiligiri/SwiftObjCMixer

master branch which compiles was integrated with changes in this PR
private_module_map branch which doesn't compile was integrated based on @booiiing 's suggestion

amorde:
Private module maps unfortunately don't get exposed to Swift as far as I remember. I ran into this myself at one point.

My biggest concern here is exposing this by default doesn't seem ideal. This effectively changes the access level of existing Objective-C code that may have been intentionally unexposed. Maybe this would make more sense as a plugin?

Ikaberry202 (approved):
Good to go

abiligiri:
> Good to go

What does this mean? Will it be merged?

Issue #9379, opened 2019-12-02
Unable to customize schemes before projects are written.


yusufoos:
🌈 We would like to modify the visibility of specific schemes before our projects are written/saved to disk.

Unfortunately, this is not supported by podfile post install hooks because `projects_writer.write!` overwrites the visibility of all the schemes.

https://github.com/CocoaPods/CocoaPods/blob/c1292cca13a8afa067e5157856b73cb0a65d6da4/lib/cocoapods/installer.rb#L322-L326

https://github.com/CocoaPods/CocoaPods/blob/c1292cca13a8afa067e5157856b73cb0a65d6da4/lib/cocoapods/installer/xcode/pods_project_generator/pods_project_writer.rb#L38-L59

Ideally we would `run_podfile_post_install_hooks` after the PodsProjectWriter calls `recreate_user_schemes` on all the projects but before it calls save on all the projects.

I'd be willing to make a PR if this approach is acceptable.

Thanks!

dnkoutso:
We do not want to move that currently as it will break a ton of projects which they will need to add `project.save` to their hooks. 

Additionally we actually prefer the user to be able to modify the project _before_ `project.save` from CocoaPods is called because that would only write the project once which can be a slow operation and also cause various Xcode issues while its open.

yusufoos:
I think you misunderstood, I'm proposing moving the hook to _before_ project.save is called but _after_ `recreate_user_schemes` is called. It would looks something like this:

```
def write! 
   cleanup_projects(projects) 
  
   projects.each do |project| 
     UI.message "- Writing Xcode project file to #{UI.path project.path}" do 
       library_product_types = [:framework, :dynamic_library, :static_library] 
       results_by_native_target = Hash[pod_target_installation_results.map do |_, result| 
         [result.native_target, result] 
       end] 
       project.recreate_user_schemes(false) do |scheme, target| 
         next unless target.respond_to?(:symbol_type) 
         next unless library_product_types.include? target.symbol_type 
         installation_result = results_by_native_target[target] 
         next unless installation_result 
         installation_result.test_native_targets.each do |test_native_target| 
           scheme.add_test_target(test_native_target) 
         end 
       end 
     end 
   end

   run_podfile_post_install_hooks

   projects.each do |project| 
     project.save
   end

end 
```

dnkoutso:
ah...hmmm I think that would work we can try opening up a PR and see if tests fail for starters.

PR #9382, opened 2019-12-03
Move run_podfile_post_install_hooks call to execute right before projects are saved

yusufoos:
This will allow us to make changes to the project and schema without it getting overwritten by the pods_project_writer. #9379

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
+* Move `run_podfile_post_install_hooks` call to execute right before projects are saved.  
+  [Yusuf Sobh](https://github.com/yusufoos)
+  [#9379](https://github.com/CocoaPods/CocoaPods/issues/9379)
+
 * Do not apply header mapping copy if the spec does not provide a header mappings dir.  
   [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
   [#9308](https://github.com/CocoaPods/CocoaPods/issues/9308)
--- a/lib/cocoapods/installer.rb
+++ b/lib/cocoapods/installer.rb
@@ -319,11 +319,11 @@ module Pod
         predictabilize_uuids(generated_projects) if installation_options.deterministic_uuids?
         stabilize_target_uuids(generated_projects)
 
-        run_podfile_post_install_hooks
-
         projects_writer = Xcode::PodsProjectWriter.new(sandbox, generated_projects,
                                                        target_installation_results.pod_target_installation_results, installation_options)
-        projects_writer.write!
+        projects_writer.write! do
+          run_podfile_post_install_hooks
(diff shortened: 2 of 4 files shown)


dnkoutso:
PR fails on `in runs the post-install hooks before serializing the Pods project`

Error:
```
Bacon::Error: not all expectations were satisfied
unsatisfied expectations:
- expected exactly once, not yet invoked: #<Pod::Installer:0x7c9b358>.run_podfile_post_install_hooks(any_parameters); in sequence 'hooks'
- expected exactly once, not yet invoked: #<AnyInstance:Pod::Installer::Xcode::PodsProjectWriter>.write!(any_parameters); in sequence 'hooks'
satisfied expectations:
```

CHANGELOG.md
```diff
@@ -35,6 +35,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
+* Move `run_podfile_post_install_hooks` call to execute right before projects are saved.
```
dnkoutso: please add two empty spaces at the end of this line for markdown formatting.

yusufoos:
I have a dependent PR that fixes the failing integration test:
https://github.com/CocoaPods/cocoapods-integration-specs/pull/258

yusufoos:
@dnkoutso Should be good to go now, all green.

dnkoutso:
@yusufoos why did the integration specs had to change? [This](https://github.com/CocoaPods/cocoapods-integration-specs/compare/f53f3b354e586cbeb473c05995c98a20d29ec9e2...02b224fabc934a7fa6d119ab2572946adbe283b7#diff-c43121a98702cf7b7665215d89eb1322) feels like a breaking change to me?

yusufoos:
The method `cleanup_projects` sorts the projects and is now called _before_ we run the the post install hooks. The integration test was assuming the "first" target to be "Reachability" which was the case when the targets were not sorted. Now that the targets are sorted the "first" target is "Pods-SampleApp". I changed the integration test so that it does not make any assumptions about the order of `installer.pods_project.targets`

dnkoutso:
I see but can we not update all targets in that test and instead maintain it?
```ruby
  target = installer.pods_project.targets.find { |t| t.name == 'Reachability' }
    config.build_settings['GCC_ENABLE_OBJC_GC'] = 'supported'
  end
```

dnkoutso:
I am still skeptical about:
> The integration test was assuming the "first" target to be Reachability which was the case when the targets were not sorted. Now that the targets are sorted, the "first" target is Pods-SampleApp

An update to 1.9.x will break users who have made this assumption...

yusufoos:
@dnkoutso I've updated the PR to move the sort call after post install hooks is called to ensure we don't break users who made assumptions about the ordering.

dnkoutso:
Nice thanks I think that should be good. Sorry for being pedantic but we never know about how users use those :)

Issue #9308, opened 2019-10-31
No longer able to build sample app with Cocoapods 1.8.x


chrismrideos:
* [Y] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

Steps:
1) Upgraded from CocoaPods `1.7.5` to `1.8.4`
2) Ran `pod install` on [this sample project](https://github.com/rideOS/RideOsApiTestApp)
3) Opened the generated `RideOsApiTestApp.xcworkspace` and attempted to build the app

## What did you expect to happen?

The app to build and run correctly, as it did when running `pod install` using CocoaPods `1.7.5`

## What happened instead?

Got a lot of build errors in `RideOsApi-umbrella.h`

### Stack

```
   CocoaPods : 1.8.4
        Ruby : ruby 2.6.3p62 (2019-04-16 revision 67580) [universal.x86_64-darwin19]
    RubyGems : 3.0.3
        Host : Mac OS X 10.15 (19A603)
       Xcode : 11.1 (11A1027)
         Git : git version 2.21.0
Ruby lib dir : /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib
Repositories : master - git - https://github.com/CocoaPods/Specs.git @ 935d7ef0d772a97b5d0b82232c76fedfdc712366
               trunk - CDN - https://cdn.cocoapods.org/
```

### Installation Source

```
Executable Path: /usr/local/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.4
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.1.0
cocoapods-trunk       : 1.4.1
cocoapods-try         : 1.1.0
```

### Podfile

```ruby
platform :ios, '13.0'
use_frameworks!

target 'RideOsApiTestApp' do
  pod 'RideOsApi'
end
```

## Project that demonstrates the issue

https://github.com/rideOS/RideOsApiTestApp

[...]

PR #9383, opened 2019-12-03
Do not apply header mapping copy if the spec does not provide a header mappings dir.

dnkoutso:
closes https://github.com/CocoaPods/CocoaPods/issues/9308

@drcapulet can you also take a look that this seems correct to you?


lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -942,7 +942,7 @@ def add_header(file_accessor, build_file, public_headers, private_headers, nativ
                     'Project'
                   end
 
-            if target.build_as_framework? && any_header_mapping_dirs? && acl != 'Project'
+            if target.build_as_framework? && header_mappings_dir(file_accessor) != nil && acl != 'Project'
```
dnkoutso: this is the fix. Do not add a copy headers phase unless there is a header mappings dir specified for the header.

This means that if a subspecs has a header mappings dir but the root spec doesn't we would get build errors with the generated umbrella header not pointing to the correct paths.

PR #9407, opened 2019-12-14
Split out unit and integration tests on CI

amorde:


amorde:
Unfortunately this actually makes it slower because Travis won't assign all VMs when there's too much demand.

PR #9431, opened 2019-12-24
Fix a bug where an incremental install missed library resources

igor-makarov:
I ran into an issue where having a (local) pod that has resources behaves incorrectly when performing an incremental install:

If a resource file (png, xib, storyboard etc.) is renamed/moved, the containing target gets regenerated. However, the aggregate target does not.

This leads to a state where the generated resource install script contains incorrect resource paths. The script fails silently, causing the build to complete and the app to crash at runtime.

This PR fixes this issue. I've added specs for the case I ran into.

### DIFF
--- a/lib/cocoapods/installer/project_cache/target_cache_key.rb
+++ b/lib/cocoapods/installer/project_cache/target_cache_key.rb
@@ -51,9 +51,12 @@ module Pod
               return :project if (other.key_hash.keys - key_hash.keys).any?
               return :project if other.key_hash['CHECKSUM'] != key_hash['CHECKSUM']
               return :project if other.key_hash['SPECS'] != key_hash['SPECS']
-              return :project if other.key_hash['FILES'] != key_hash['FILES']
               return :project if other.key_hash['PROJECT_NAME'] != key_hash['PROJECT_NAME']
             end
+
+            this_files = key_hash['FILES']
+            other_files = other.key_hash['FILES']
+            return :project if this_files != other_files
 
             this_build_settings = key_hash['BUILD_SETTINGS_CHECKSUM']
             other_build_settings = other.key_hash['BUILD_SETTINGS_CHECKSUM']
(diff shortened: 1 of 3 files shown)


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Fix a bug where an incremental install missed library resources  
```

</p>

sebastianv1:
Ah interesting. In this case it might make more sense to store a checksum of the contents of the aggregate target script phases (i.e. resources, embed frameworks). I imagine that's how it's materializing as a bug.

igor-makarov:
I've checked and it won't be possible. The live cache keys (the ones compared to the saved ones) are generated from an `AggregateTarget`, which doesn't have the scripts. They are created much later. 

In fact, I think that is actually a good thing because generating scripts takes time and avoiding it is good.

dnkoutso:
How do we want to proceed here?

igor-makarov:
I've been using this patch in our project and can confirm it has no side effects I know of.

lib/cocoapods/installer/project_cache/target_cache_key.rb
```diff
@@ -51,10 +51,13 @@ def key_difference(other)
               return :project if (other.key_hash.keys - key_hash.keys).any?
               return :project if other.key_hash['CHECKSUM'] != key_hash['CHECKSUM']
               return :project if other.key_hash['SPECS'] != key_hash['SPECS']
-              return :project if other.key_hash['FILES'] != key_hash['FILES']
               return :project if other.key_hash['PROJECT_NAME'] != key_hash['PROJECT_NAME']
             end
 
+            this_files = key_hash['FILES']
+            other_files = other.key_hash['FILES']
+            return :project if this_files != other_files
```
sebastianv1: Why did this get pulled out?

igor-makarov: Because we need to compare file lists for all targets now,

sebastianv1: Cool

sebastianv1:
The change seems good. This seems to target one particular scenario, and doesn't address what looks to be a bigger issue of determining if an `AggregateTarget` is dirty (not a slam on this PR, it's likely been a bug since incremental launched. Thank you for finding this bug and fixing it @igor-makarov). I'll spend some time looking more into what other ways `AggregateTarget`s can become dirty (maybe `xcframeworks`?).

igor-makarov:
Yeah, frameworks are being missed by the cache.

dnkoutso:
just needs a rebase and we can land!

igor-makarov:
Rebased 💪🏻

PR #9435, opened 2019-12-25
Performance improvements for incremental install

igor-makarov:
* Only clean pod sources if new pods have been installed
* Lock pod sources only if they have been previously unlocked

### DIFF
--- a/lib/cocoapods/installer.rb
+++ b/lib/cocoapods/installer.rb
@@ -565,6 +565,7 @@ module Pod
     #
     def clean_pod_sources
       return unless installation_options.clean?
+      return if installed_specs.empty?
       pod_installers.each(&:clean!)
     end
 
--- a/lib/cocoapods/installer/pod_source_installer.rb
+++ b/lib/cocoapods/installer/pod_source_installer.rb
@@ -90,7 +90,8 @@ module Pod
       #
       def lock_files!(file_accessors)
         return if local?
-        FileUtils.chmod('u-w', source_files(file_accessors))
+        unlocked_files = source_files(file_accessors).reject { |f| (File.stat(f).mode & 0o200).zero? }
+        FileUtils.chmod('u-w', unlocked_files)
       end
 
       # Unlocks the source files if appropriate.


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* [DNM] Incremental performance improvements  
```

</p>

lib/cocoapods/sources_manager.rb
```diff
@@ -72,6 +71,8 @@ def create_source_with_url(url)
       #
       def cdn_url?(url)
         if url =~ %r{^https?:\/\/}
+          require 'typhoeus'
```
dnkoutso: can this change land separately?

igor-makarov:
I'm beginning to think @segiddins is right and this needs to land in `master`.

lib/cocoapods/sources_manager.rb
igor-makarov: Do you mean a separate PR?

dnkoutso: maybe? or separate commit

igor-makarov: sure thing

igor-makarov:
* Spun out the Typhoeus fix
* Rebased on `master`

lib/cocoapods/installer/pod_source_installer.rb
```diff
@@ -90,7 +90,8 @@ def clean!
       #
       def lock_files!(file_accessors)
         return if local?
-        FileUtils.chmod('u-w', source_files(file_accessors))
+        unlocked_files = source_files(file_accessors).reject { |f| (File.stat(f).mode & 0o200).zero? }
```
segiddins: this seems like it would be slower to me, since it's 2 syscalls per file (1 to stat and one to lock), vs just 1 for only doing the `chmod`

igor-makarov: Here's the thing though - and I don't know why or under what circumstances - `fstat` seems much cheaper than `chmod` so in the case of an incremental install I measured a significant performance benefit.

dnkoutso:
We can try this in our project and quickly measure any negative or positive impact.

dnkoutso:
cc @sebastianv1 maybe?

lib/cocoapods/installer/pod_source_installer.rb
dnkoutso: @igor-makarov should we cross reference if `unlock_files!` or `lock_files!` in `installer.rb` to use the list of generated pod targets instead?

```ruby
    # Unlocks the sources of the Pods.
    #
    def unlock_pod_sources
      pod_installers.each do |installer|
        pod_target = pod_targets.find { |target| target.pod_name == installer.name }
        installer.unlock_files!(pod_target.file_accessors)
      end
    end

    # Locks the sources of the Pods if the config instructs to do so.
    #
    def lock_pod_sources
      return unless installation_options.lock_pod_sources?
      pod_installers.each do |installer|
        pod_target = pod_targets.find { |target| target.pod_name == installer.name }
        installer.lock_files!(pod_target.file_accessors)
      end
    end
```

Change the `.find` to be against `generated_pod_targets` and if its not found then dont do anything?

dnkoutso: Something like:
```ruby
    def unlock_pod_sources
      pod_installers.each do |installer|
        pod_target = generated_pod_targets.find { |target| target.pod_name == installer.name }
        installer.unlock_files!(pod_target.file_accessors) if pod_target
      end
    end

    # Locks the sources of the Pods if the config instructs to do so.
    #
    def lock_pod_sources
      return unless installation_options.lock_pod_sources?
      pod_installers.each do |installer|
        pod_target = generated_pod_targets.find { |target| target.pod_name == installer.name }
        installer.lock_files!(pod_target.file_accessors) if pod_target
      end
    end
```

igor-makarov: This seems like a nice idea but I don't feel confident in that part of the code and it feels like a bigger change to me?

The `fstat` change is a local optimization so I feel better about it.

igor-makarov: Additionally, I feel like coupling it with the incremental install feature could introduce UB, for example, a user unlocking a file in Xcode and it never getting locked again because the project doesn't get regenerated.

lib/cocoapods/installer.rb
```diff
@@ -563,6 +563,7 @@ def install_source_of_pod(pod_name)
     #
     def clean_pod_sources
       return unless installation_options.clean?
+      return if installed_specs.empty?
       pod_installers.each(&:clean!)
```
segiddins, line 567: will `pod_installers` not be empty when there are no installed specs?

igor-makarov: This is a very old PR so I had to check.

The `pod_installers` ivar contains all the installers, regardless of what gets (incrementally) generated.

Issue #7795, opened 2018-05-30
Cocoapods 1.5.2 does not support symbolic link 


wqyfavor:
We develop tools to enhance CP. One mode is to create symbolic link instead of copying frameworks.
On cocoapods 1.5.2, it doesn't work. I debugged into source code and find that in path_list.rb read_file_system method. If root points to a soft-link directory, this method reads nothing to @files.

While CP 1.2.0 works fine.

wqyfavor:
You may test with these codes which I extract from Cocoapods sources.



```
#!/usr/bin/ruby
require 'find'
require 'pathname'

# cocoapods 1.5.2
def read_file_system(root)
	dirs = []
	files = []
	root_length = root.length + 1
	Find.find(root.to_s) do |f|
		directory = File.directory?(f)
		f = f.slice(root_length, f.length - root_length)
		next if f.nil?

		(directory ? dirs : files) << f
	end

	dirs.sort_by!(&:upcase)
	files.sort_by!(&:upcase)

	puts "#{files}"
end

# cocoapods 1.2.0
def escape_path_for_glob(path)
	result = path.to_s
	characters_to_escape = ['[', ']', '{', '}', '?', '*']
	characters_to_escape.each do |character|
		result.gsub!(character, "\\#{character}")
	end
	Pathname.new(result)
end

def read_file_system1_2(root)
	escaped_root = escape_path_for_glob(root)

	absolute_paths = Pathname.glob(escaped_root + '**/*', File::FNM_DOTMATCH).lazy
	dirs_and_files = absolute_paths.reject { |path| path.basename.to_s =~ /^\.\.?$/ }
	dirs, files = dirs_and_files.partition { |path| File.directory?(path) }

	root_length = root.length + 1
	sorted_relative_paths_from_full_paths = lambda do |paths|
		relative_paths = paths.lazy.map do |path|
			path_string = path.to_s
			path_string.slice(root_length, path_string.length - root_length)
		end
		relative_paths.sort_by(&:upcase)
	end
	
	puts "#{sorted_relative_paths_from_full_paths.call(files)}"
end

puts "Read by 1.5.2"
read_file_system "~/SomeBundle"
read_file_system "~/SomeBundleSoftLink"

puts "\nRead by 1.2.0"
read_file_system1_2 "~/SomeBundle"
read_file_system1_2 "~/SomeBundleSoftLink"
```

[...]

PR #9451, opened 2020-01-02
[PathList] Add symlink follow support

### LINKED ISSUE #6603: Pathname.glob() doesn't traverse symlinks

manuyavuz:
Adds support to follow symlink folders under `PathList` root once.

This might also fix #6603, #7761, #7795.

I needed this change because currently I can't put a file which exists under a symlink folder as a resource to my pod. The main reason is that `**/*` glob pattern does not follow symlinks in Ruby for Dir.glob. Result is the same with `Find.find` module, it does not traverse symlinks.

There is a discussion in a ruby forum with a workaround:
http://blade.nagaokaut.ac.jp/cgi-bin/scat.rb/ruby/ruby-talk/206308

It basically suggests to use `**/*/**/*` to follow symlinks for once. Even though I don't know the exact technical detail why that works, I assume that when ruby glob resolver sees `**` after `*`, it tries to glob recursively again from files matching with first `*`.

This expands symbolic links, but misses immediate children for the root folder. To fix this, I applied `**{,/*/**}/*` to collect matches for both `**/*` and ``**/*/**/*`. This solves missing immediate children problem, but may result in duplicate records, so a `uniq!` is required.

One problem is that, `**/*/**/*` pattern also expands `.` that exists in root folder. Therefore, we get duplicate records for top level files, such as `foo.h` and `./foo.h`. To drop latter ones, a used to final glob pattern as follows:

`**{,/[^.]*/**}/*`

This seems to work for various cases that I tried. I also fixed tests as we now support symlink following, and tests were trying to assure that we are not following symlinks.

What's your feeling about this change? I think this solves a big chunk of symlink following problem. 

However, it's not a complete solution as it only follows symlinks once. So for a directory tree like this:
```
.
├── a
│   ├── b
│   ├── d
│   │   └── e
│   └── f -> d
├── b -> a/b
└── c -> a
```
I get following output:
```
irb(main):019:0> Dir.glob('**{,/*/**}/*').uniq!
=> ["a", "a/f", "a/d", "a/d/e", "a/b", "c", "x", "b", "a/f/e", "c/f", "c/d", "c/d/e", "c/b"]
```

Here `c/d/e` is resolved, but `c/f/e` is not because it didn't follow `f` symlink in folder `a`.

One very dirty solution would be to add `{,/*/**}` part multiple times into glob pattern. Each addition would add an extra level of symlink following.

Another solution would be to traverse all found files from `**/*` and try to follow the path if it's a link and directory, and do this recursively. Even though this would be a solution, this complicates the implementation and possible performance.

A completely different solution would be to give full control to podspec developer, and run glob patterns provided by them directly on the file system, rather than trying to match them in a pre generated list of files and directories which is collected with `**` which does not have a notion to follow symlinks by default. I think this can be a great solution if we don't observe performance penalties. This would make us to completely get rid of PathList possibly.

### DIFF
--- a/spec/unit/sandbox/path_list_spec.rb
+++ b/spec/unit/sandbox/path_list_spec.rb
@@ -14,8 +14,12 @@ module Pod
         end
         expected = %w(
           Banana.modulemap
+          BananaFramework.framework/Headers/BananaFramework.h
+          BananaFramework.framework/Headers/SubDir/SubBananaFramework.h
           BananaFramework.framework/Versions/A/Headers/BananaFramework.h
           BananaFramework.framework/Versions/A/Headers/SubDir/SubBananaFramework.h
+          BananaFramework.framework/Versions/Current/Headers/BananaFramework.h
+          BananaFramework.framework/Versions/Current/Headers/SubDir/SubBananaFramework.h
           BananaLib.podspec
           Classes/Banana.h
           Classes/Banana.m
@@ -56,11 +60,14 @@ module Pod
         dirs.sort.should == %w(
           BananaFramework.framework
           BananaFramework.framework/Headers
+          BananaFramework.framework/Headers/SubDir
           BananaFramework.framework/Versions
           BananaFramework.framework/Versions/A
           BananaFramework.framework/Versions/A/Headers
           BananaFramework.framework/Versions/A/Headers/SubDir
           BananaFramework.framework/Versions/Current
+          BananaFramework.framework/Versions/Current/Headers
+          BananaFramework.framework/Versions/Current/Headers/SubDir
           Classes
           Resources
           Resources/Base.lproj
@@ -336,12 +343,12 @@ module Pod
         @path_list.dirs.map(&:to_s).should.include?('Classes/symlinked')
       end
 
-      it 'doesn\'t include file in symlinked dir' do
+      it 'includes file in symlinked dir' do
         @path_list.instance_variable_set(:@files, nil)
         @path_list.instance_variable_set(:@dirs, nil)
         File.symlink(@tmpdir, @symlink_dir)
 
-        @path_list.files.map(&:to_s).should.not.include?('Classes/symlinked/someheader.h')
+        @path_list.files.map(&:to_s).should.include?('Classes/symlinked/someheader.h')
       end
     end
 
--- a/lib/cocoapods/sandbox/path_list.rb
+++ b/lib/cocoapods/sandbox/path_list.rb
@@ -55,7 +55,7 @@ module Pod
         files = []
         root_length = root.cleanpath.to_s.length + File::SEPARATOR.length
         escaped_root = escape_path_for_glob(root)
-        Dir.glob(escaped_root + '**/*', File::FNM_DOTMATCH).each do |f|
+        Dir.glob(escaped_root + '**{,/[^.]*/**}/*', File::FNM_DOTMATCH).each do |f|
           directory = File.directory?(f)
           # Ignore `.` and `..` directories
           next if directory && f =~ /\.\.?$/
@@ -66,8 +66,8 @@ module Pod
           (directory ? dirs : files) << f
         end
 
-        dirs.sort_by!(&:upcase)
-        files.sort_by!(&:upcase)
+        dirs.sort_by!(&:upcase).uniq!
+        files.sort_by!(&:upcase).uniq!
 
         @dirs = dirs
         @files = files


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* [PathList] Add symlink follow support  
```

</p>

spec/unit/sandbox/path_list_spec.rb
```diff
@@ -14,8 +14,12 @@ module Pod
         end
         expected = %w(
           Banana.modulemap
+          BananaFramework.framework/Headers/BananaFramework.h
+          BananaFramework.framework/Headers/SubDir/SubBananaFramework.h
           BananaFramework.framework/Versions/A/Headers/BananaFramework.h
           BananaFramework.framework/Versions/A/Headers/SubDir/SubBananaFramework.h
+          BananaFramework.framework/Versions/Current/Headers/BananaFramework.h
+          BananaFramework.framework/Versions/Current/Headers/SubDir/SubBananaFramework.h
```
amorde: I don't think we want this - these are effectively the same file, `Current` is just a directory symlink to `A`. Would this cause build issues due to "duplicate" headers?

dnkoutso:
Any update here on the comments? If not we can push this out of 1.10.

spec/unit/sandbox/path_list_spec.rb
dnkoutso: any updates to this question?

RichardFevrier:
Is [this](https://github.com/ruby/dev-meeting-log/blob/master/DevelopersMeeting20210216Japan.md#feature-17548-need-simple-way-to-include-symlink-directories-in-dirglob-nobu) related ?

PR #9478, opened 2020-01-09
Fix an issue that caused stale framework artifacts when archiving

### LINKED ISSUE #9458: While using XCFramework as a vendored_framework, Instead of IOS Archive, It is generating the Generic Xcode Archive.
Hi, 

# Report

## What did you do?

ℹ I have created XCFramework, that one I added as `vendored_frameworks` and distributed in cocoapods.

ℹ Once we install that pod into the client's IOS projects, It is working fine when running into a simulator or real device. But when we create the archive it is expected to create` IOS Archive`.

## What happened instead?

ℹ Instead of `IOS Archive`, It is generating the `Generic Xcode Archive`. When we use .framework file, it is working fine. Problem with XCFramework only. 

## CocoaPods Environment

ℹ CocoaPods version **1.9.0.beta.2**
ℹ  Xcode Version - 11.2.1

ℹ I have verified this repo also [cocoapods integration repo](https://github.com/CocoaPods/cocoapods-integration-specs/tree/master/install_vendored_xcframework). The Same problem.

ℹ Try with this also `pod 'CoconutLib'`

@amorde Please help me to resolve this issue. I looking forward to hearing from you. If you need more information let me know.

amorde:
Closes #9458 

Requires https://github.com/CocoaPods/cocoapods-integration-specs/pull/264

When archiving, the `.framework` slice that was placed in `Prepare Artifacts` was left in the `Products/Applications` directory of the archive because that is what `${TARGET_BUILD_DIR}` is set to when archiving. During a non-install build, the value matches that of `${CONFIGURATION_BUILD_DIR}`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Paul Beusterien](https://github.com/paulb777)
   [#9465](https://github.com/CocoaPods/CocoaPods/pull/9465)
 
+* Fix an issue that caused iOS archives to be invalid when including a vendored XCFramework  
+  [Eric Amorde](https://github.com/amorde)
+  [#9458](https://github.com/CocoaPods/CocoaPods/issues/9458)
+
 
 ## 1.9.0.beta.2 (2019-12-17)
 
--- a/lib/cocoapods/generator/prepare_artifacts_script.rb
+++ b/lib/cocoapods/generator/prepare_artifacts_script.rb
@@ -117,7 +117,7 @@ module Pod
             fi
 
             local record_artifact=${2:-true}
-            local destination="${TARGET_BUILD_DIR}"
+            local destination="${CONFIGURATION_BUILD_DIR}"
 
             if [ -L "${source}" ]; then
               echo "Symlinked..."


lib/cocoapods/generator/prepare_artifacts_script.rb
```diff
@@ -117,7 +117,7 @@ def script
             fi
 
             local record_artifact=${2:-true}
-            local destination="${TARGET_BUILD_DIR}"
+            local destination="${CONFIGURATION_BUILD_DIR}"
```
dnkoutso: shouldnt we use something like `BUILT_PRODUCTS_DIR`?

amorde: Hmm not sure, they seem to be the same in my tests. Here's the docs for each

> Per-configuration Build Products Path (CONFIGURATION_BUILD_DIR)
The base path where build products will be placed during a build for a given configuration. By default, this is set to $(BUILD_DIR)/$(CONFIGURATION).

> BUILT_PRODUCTS_DIR
Identifies the directory under which all the product’s files can be found. This directory contains either product files or symbolic links to them. Run Script build phases can use the value of this build setting as a convenient way to refer to the product files built by one or more targets even when these files are scattered throughout a directory hierarchy (for example, when Deployment Location (DEPLOYMENT_LOCATION) is set to YES.

We've used `${CONFIGURATION_BUILD_DIR}` elsewhere such as https://github.com/CocoaPods/CocoaPods/blob/42741d897eb8ab34f9a1524988f59ff3f7760cc4/lib/cocoapods/generator/embed_frameworks_script.rb#L63

dnkoutso: sg!

CHANGELOG.md
```diff
@@ -10,14 +10,18 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 * PathList optimizations related to file system reads.  
   [manuyavuz](https://github.com/manuyavuz)
-  [#issue_number](https://github.com/CocoaPods/CocoaPods/issues/4927)
+  [#4927](https://github.com/CocoaPods/CocoaPods/issues/4927)
```
dnkoutso: I can fix this in a separate commit!

lib/cocoapods/generator/prepare_artifacts_script.rb
dnkoutso: I think its good because I think `CONFIGURATION_BUILD_DIR` is on a per-target basis while `BUILT_PRODUCTS_DIR` is global.

dnkoutso:
Merged integration specs, needs rebase and update of submodule sha and we can merge!

lib/cocoapods/generator/prepare_artifacts_script.rb
amorde: oh nice, TIL

dnkoutso: well this is how I read it from what you pasted above :D

amorde: I initially read "the product's file" as a single product but I realize now it mentions other targets

PR #9500, opened 2020-01-21
Apply Xcode 11 `XCTUnwrap` fix to library and framework targets.

dnkoutso:
See https://developer.apple.com/documentation/xcode_release_notes/xcode_11_release_notes

Integration specs: https://github.com/CocoaPods/cocoapods-integration-specs/pull/266

This fixes it for both user targets that have this issue and fixes also for pod targets.

### DIFF
--- a/lib/cocoapods/target/build_settings.rb
+++ b/lib/cocoapods/target/build_settings.rb
@@ -492,9 +492,15 @@ module Pod
         attr_reader :test_xcconfig
         alias test_xcconfig? test_xcconfig
 
+        # @return [Boolean]
+        #   whether settings are being generated for an application bundle
+        #
         attr_reader :app_xcconfig
         alias app_xcconfig? app_xcconfig
 
+        # @return [Boolean]
+        #   whether settings are being generated for an library bundle
+        #
         attr_reader :library_xcconfig
         alias library_xcconfig? library_xcconfig
 
@@ -683,6 +689,12 @@ module Pod
           file_accessors.flat_map(&:vendored_xcframeworks).map { |path| Xcode::XCFramework.new(path) }
         end
 
+        # @return [Array<String>]
+        define_build_settings_method :system_framework_search_paths, :build_setting => true, :memoized => true, :sorted => true, :uniqued => true do
+          return ['$(PLATFORM_DIR)/Developer/usr/lib'] if should_apply_xctunwrap_fix?
+          []
+        end
+
         #-------------------------------------------------------------------------#
 
         # @!group Libraries
@@ -737,16 +749,18 @@ module Pod
 
         # @return [Array<String>]
         define_build_settings_method :library_search_paths, :build_setting => true, :memoized => true, :sorted => true, :uniqued => true do
-          return [] if library_xcconfig? && target.build_as_static?
+          library_search_paths = should_apply_xctunwrap_fix? ? ['$(PLATFORM_DIR)/Developer/usr/lib'] : []
+          return library_search_paths if library_xcconfig? && target.build_as_static?
 
-          vendored = library_search_paths_to_import.dup
-          vendored.concat dependent_targets.flat_map { |pt| pt.build_settings[@configuration].vendored_dynamic_library_search_paths }
+          library_search_paths.concat library_search_paths_to_import.dup
+          library_search_paths.concat dependent_targets.flat_map { |pt| pt.build_settings[@configuration].vendored_dynamic_library_search_paths }
           if library_xcconfig?
-            vendored.delete(target.configuration_build_dir(CONFIGURATION_BUILD_DIR_VARIABLE))
+            library_search_paths.delete(target.configuration_build_dir(CONFIGURATION_BUILD_DIR_VARIABLE))
           else
-            vendored.concat(dependent_targets.flat_map { |pt| pt.build_settings[@configuration].library_search_paths_to_import })
+            library_search_paths.concat(dependent_targets.flat_map { |pt| pt.build_settings[@configuration].library_search_paths_to_import })
           end
-          vendored
+
+          library_search_paths
         end
 
         # @return [Array<String>]
@@ -837,6 +851,7 @@ module Pod
         define_build_settings_method :swift_include_paths, :build_setting => true, :memoized => true, :sorted => true, :uniqued => true do
           paths = dependent_targets.flat_map { |pt| pt.build_settings[@configuration].swift_include_paths_to_import }
           paths.concat swift_include_paths_to_import if non_library_xcconfig?
+          paths.concat ['$(PLATFORM_DIR)/Developer/usr/lib'] if should_apply_xctunwrap_fix?
           paths
         end
 
@@ -964,6 +979,20 @@ module Pod
           else
             target.spec_consumers.select { |sc| sc.spec.spec_type == @xcconfig_spec_type }
           end
+        end
+
+        # Xcode 11 causes an issue with frameworks or libraries before 12.2 deployment target that link or are part of
+        # test bundles that use XCTUnwrap. Apple has provided an official work around for this.
+        #
+        # @see https://developer.apple.com/documentation/xcode_release_notes/xcode_11_release_notes
+        #
+        # @return [Boolean] Whether to apply the fix or not.
+        #
+        define_build_settings_method :should_apply_xctunwrap_fix?, :memoized => true do
+          library_xcconfig? &&
+            target.platform.name == :ios &&
+            Version.new(target.platform.deployment_target) < Version.new('12.2') &&
+            frameworks_to_import.include?('XCTest')
         end
 
         #-------------------------------------------------------------------------#
--- a/spec/unit/target/build_settings/aggregate_target_settings_spec.rb
+++ b/spec/unit/target/build_settings/aggregate_target_settings_spec.rb
@@ -13,13 +13,18 @@ module Pod
         end
 
         before do
-          @target_definition = fixture_target_definition
+          project_path = SpecHelper.create_sample_app_copy_from_fixture('SampleProject')
+          @project = Xcodeproj::Project.open(project_path)
+          @project.save
+          @native_target = @project.targets.find { |t| t.name == 'SampleProject' }
+          @target_definition = fixture_target_definition('SampleProject')
           @specs = specs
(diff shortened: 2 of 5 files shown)


spec/unit/target/build_settings/aggregate_target_settings_spec.rb
```diff
@@ -13,13 +13,18 @@ def pod_target(spec, target_definition)
         end
 
         before do
-          @target_definition = fixture_target_definition
+          project_path = SpecHelper.create_sample_app_copy_from_fixture('SampleProject')
```
dnkoutso: this test setup was not using an actual project to the aggregate target. Now it does.

spec/unit/target/build_settings/aggregate_target_settings_spec.rb
```diff
@@ -190,7 +195,7 @@ def pod_target(spec, target_definition)
             end
 
             it 'links the pod targets with the aggregate target' do
-              @xcconfig.to_hash['OTHER_LDFLAGS'].should.include '-l"BananaLib-Pods"'
+              @xcconfig.to_hash['OTHER_LDFLAGS'].should.include '-l"BananaLib-Pods-SampleProject"'
```
dnkoutso: this is correct since we are now integrating on a real project and a real user target.

lib/cocoapods/target/build_settings.rb
```diff
@@ -1246,6 +1271,11 @@ def user_target_xcconfig_values_by_consumer_by_key
           merged_xcconfigs(user_target_xcconfig_values_by_consumer_by_key, :user_target_xcconfig)
         end
 
+        # @see BuildSettings#should_apply_xctunwrap_fix?
+        def should_apply_xctunwrap_fix?
+          target.library? && target.user_targets.any? { |ut| Version.new(ut.deployment_target) <= Version.new('12.1') }
```
dnkoutso: I am open to different heuristics on _when_ to apply this fix for aggregate targets.

The same question also applies for pod targets which I am implementing as we speak and will update the PR when its ready.

Is the answer "we should apply this always" ?

lib/cocoapods/target/build_settings.rb
```diff
@@ -966,6 +999,12 @@ def pod_target_xcconfig_values_by_consumer_by_key
           end
         end
 
+        # @see BuildSettings#should_apply_xctunwrap_fix?
+        def should_apply_xctunwrap_fix?
+          non_library_specs = target.test_specs + target.app_specs
```
dnkoutso: same question as aggregate targets, what heuristic do we use to apply (or not) the fix?

lib/cocoapods/target/build_settings.rb
justinseanmartin: Using the deployment target seems reasonable to me, as it will at least phase out the fix over time. The only alternative that I see would be checking the version of the currently selected version of Xcode and only apply it if Xcode 11 is selected, but I'm not sure there is any precedent for doing that.

justinseanmartin: I think we could tighten this up only be applied to libraries that explicitly define a dependency on XCTest in their podspec via an explicit `s.framework = 'XCTest'`? I think that effectively matches what is called out as the problem in Apple's documentation. That would prevent adding a bunch of additional search paths that aren't used to every library that gets built.

dnkoutso: Yes. I think the rules are:

1) Is library/framework
2) Deployment target < 12.1
3) XCTest is a framework linked.

lib/cocoapods/target/build_settings.rb
```diff
@@ -1246,6 +1292,17 @@ def user_target_xcconfig_values_by_consumer_by_key
           merged_xcconfigs(user_target_xcconfig_values_by_consumer_by_key, :user_target_xcconfig)
         end
 
+        # @see BuildSettings#should_apply_xctunwrap_fix?
+        def should_apply_xctunwrap_fix?
```
segiddins: make this a `define_build_settings_method, :memoized => true` so it doesn't need to be re-computed every time it gets called

lib/cocoapods/target/build_settings.rb
```diff
@@ -1246,6 +1292,17 @@ def user_target_xcconfig_values_by_consumer_by_key
...
+          target.library? && target.user_targets.any? do |ut|
```
segiddins: can we check the platform on the target definitions instead?

lib/cocoapods/target/build_settings.rb
```diff
@@ -1246,6 +1292,17 @@ def user_target_xcconfig_values_by_consumer_by_key
...
+            next false unless Version.new(ut.deployment_target) < Version.new('12.2')
```
segiddins: should this only be for iOS?

lib/cocoapods/target/build_settings.rb
```diff
@@ -1246,6 +1292,17 @@ def user_target_xcconfig_values_by_consumer_by_key
...
+            flags = (ut.resolved_build_setting('OTHER_LDFLAGS', :resolve_against_xcconfig => true) || {}).values.flatten.uniq
+            parsed_flags = Xcodeproj::Config::OtherLinkerFlagsParser.parse(flags)
+            all_frameworks = (parsed_flags[:frameworks] + parsed_flags[:weak_frameworks]).uniq
+            all_frameworks.include?('XCTest') || all_frameworks.include?('"XCTest"')
```
segiddins: why can't we check whether the build setting includes `XCTest` in it's own set of frameworks?

lib/cocoapods/target/build_settings.rb
dnkoutso: oh sure

lib/cocoapods/target/build_settings.rb
dnkoutso: I think so. I can limit it to that but I will check if it works or not on macOS first.

dnkoutso: Seems to work without the need for the fix on macOS. Will scope it on iOS only.

lib/cocoapods/target/build_settings.rb
dnkoutso: you mean check if XCTest is present in any of the `pod_targets` inside the aggregate target?

dnkoutso: actually checking against the pod targets wouldn't necessarily fix this target. We basically do what the user should do for their own target automatically here. It is possible the user target is a framework or library that depends on 10 pods that don't use XCTest but the user target itself does and would need this fix.

segiddins: but they'd have that issue without cocoapods as well, then?

dnkoutso: Thats correct they would. The above as it is currently is would fix it automatically for the user without them having to manually apply their fix on their own target.

dnkoutso: Are you suggesting to _also_ check if any of the `pod_targets` within the aggregate have `XCTest` then to apply the fix or _replace_ it?

dnkoutso: Alternatively we can skip this check entirely for aggregate targets and let uses apply the Apple fix manually for their own targets. Happy to remove it entirely.

segiddins: > Alternatively we can skip this check entirely for aggregate targets and let uses apply the Apple fix manually for their own targets. Happy to remove it entirely.


I think I'd be in favor of that

lib/cocoapods/target/build_settings.rb
```diff
@@ -966,6 +998,14 @@ def pod_target_xcconfig_values_by_consumer_by_key
           end
         end
 
+        # @return [Boolean]
+        define_build_settings_method :should_apply_xctunwrap_fix?, :memoized => true do
+          library_xcconfig? &&
+            target.platform.name == :ios &&
+            Version.new(target.platform.deployment_target) < Version.new('12.2') &&
+            target.library_spec_consumers.flat_map(&:frameworks).include?('XCTest')
```
segiddins: should this be checking the `frameworks` method in build settings rather than going into the spec consumers?

lib/cocoapods/target/build_settings.rb
dnkoutso: will remove.

dnkoutso: removed.

lib/cocoapods/target/build_settings.rb
dnkoutso: actually I avoided that because `:frameworks`build settings method includes those of test specs and app specs. In my sample this has not been an issue if the test bundle itself brings in XCTest but a library we depend on doesn't.

segiddins: > actually I avoided that because :frameworksbuild settings method includes those of test specs and app specs

but in this case, you already have a conditional check for `library_xcconfig?` in this `&&` chain

dnkoutso: If a pod target has 2 specs, 1 library spec and 1 test spec and the test spec defines `test_spec.frameworks = 'XCTest'` I wouldn't want to take that into consideration when deciding whether the xcconfig for the library spec should apply this fix or not. Using `:frameworks` from build settings includes that and we are only interested if the library itself brings in the `XCTest` framework

segiddins: Right, but the `frameworks` method on build settings is already scoped to the library specs when `library_xcconfig?` is true

dnkoutso: I am not sure that is entirely true for dynamic framework targets https://github.com/CocoaPods/CocoaPods/blob/master/lib/cocoapods/target/build_settings.rb#L570

This seems to be adding all consumer frameworks from all specs for dynamic framework targets.

dnkoutso: oooooh and this is what I was missing:
```ruby
    define_build_settings_method :spec_consumers, :memoized => true do
          if non_library_xcconfig?
            target.spec_consumers.select { |sc| non_library_spec == sc.spec }
          else
            target.spec_consumers.select { |sc| sc.spec.spec_type == @xcconfig_spec_type }
          end
        end
```
You are correct. I am going to update it.

PR #9520, opened 2020-02-05
Bump minimum Xcodeproj gem version to 1.15.0

revolter:
Needed for this: https://github.com/CocoaPods/Xcodeproj/pull/706

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -35,10 +35,10 @@ GIT
 
 GIT
   remote: https://github.com/CocoaPods/Xcodeproj.git
-  revision: ec5b43be5f76b7421af41b5b27c33782d53d132e
+  revision: 415cf26a26778fd417028de166930b0d0349066c
   branch: master
   specs:
-    xcodeproj (1.14.0)
+    xcodeproj (1.15.0)
       CFPropertyList (>= 2.3.3, < 4.0)
       atomos (~> 0.1.3)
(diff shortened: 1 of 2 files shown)


dnkoutso:
@revolter we dont strictly need it do we? The requirement accepts `>= 1.14.0`. Is there a new feature in 1.15.0 that CocoaPods needs in order to function?

revolter:
Yes, the one I linked in the description.

revolter:
Oh, so running `bundle update cocoapods` will also update `Xcodeproj` to `1.15.0`, without this PR?

dnkoutso:
Yes! because CocoaPods accepts anything Xcodeproj 1.14.0 and above.

Its harmless to land this but we dont strictly need it because we do not use something new that is required.

revolter:
Then I say we should close this, as it's not strictly needed, as you said, and could be confusing in the future.

dnkoutso:
SGTM!

dnkoutso:
@revolter you can just run `bundle update xcodeproj` btw.

revolter:
Oh, even if it's not in my `Gemfile`? (only `cocoapods` is in my `Gemfile`)

dnkoutso:
Even prior versions should continue to work btw as they accept `>=`.

Yes `bundle update xcodeproj` should work.

PR #9574, opened 2020-02-29
Fix an issue when including XCFrameworks in macOS apps

amorde:
${EFFECTIVE_PLATFORM_NAME} will only be set if applicable for the current platform.

Closes #9572

PR #9576, opened 2020-03-01
Add Vendored XCFramework example to CI example projects

amorde:
The project won't build successfully until #9575 and #9574 are merged.

Also updates the examples to use Xcode 11.3.1


Rakefile
```diff
@@ -295,6 +295,8 @@ begin
 
             project = Xcodeproj::Project.open(project_path)
             target = project.targets.first
+            scheme_target = project.targets.find { |t| t.name == scheme_name }
+            target = scheme_target unless scheme_target.nil?
```
amorde: we were always using the first target to determine the platform, which doesn't work for projects that have multiple targets with different platforms

amorde:
Gonna clean up a few files before merging

Issue #9392, opened 2019-12-07
Add option to select test specs for `pod lib lint`


paulb777:
Currently, it is only possible to test all of the test_specs or none with `--skip-tests`.  It would be nice to be able to choose which test specs to run.

For example, for a pod that has both Swift and Objective C tests, it would be useful to be able to only run the Objective C tests when `--use-libraries` is also set.

dnkoutso:
Yeap, should be easy to add!

PR #9632, opened 2020-03-21
Add --use-static-frameworks lint option

paulb777:
Also fix a bad merge in the CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,14 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Juanjo López](https://github.com/juanjonol)
   [#9232](https://github.com/CocoaPods/CocoaPods/issues/9232)
 
+* Option to lint a specified set of test_specs  
+  [Paul Beusterien](https://github.com/paulb777)
+  [#9392](https://github.com/CocoaPods/CocoaPods/pull/9392)
+
+* Add --use-static-frameworks lint option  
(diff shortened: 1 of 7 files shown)


paulb777:
Rebased

dnkoutso:
I continue to feel that the proliferation of these options is making things much worse for maintainers imo but I don't see a clear way to do manage this effectively.

Basically an author can now publish a pod under very specific set of options (modular headers, use libraries and now static frameworks) which means a consumer of a pod must somehow know that the pod works only works in a specific way.

Do you have any thoughts or ideas about this?

paulb777:
@dnkoutso I'd be fine to remove the option from push.

The reason I want this option is that the boat has already sailed on these options available in Podfile's and it would be a nice for authors to have an easy way to set up CI to test all the options before publishing.  

Specifically, I'd like to use it here for example - https://github.com/firebase/firebase-ios-sdk/blob/master/.github/workflows/storage.yml#L95

CHANGELOG.md
```diff
@@ -91,10 +99,6 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
-* Option to lint a specified set of test_specs
```
amorde, line 94: is this correct?

paulb777: Yes, #9392 is in master and not 1.9.1

PR #9640, opened 2020-03-24
Add --strict flag.

ileitch:
From time to time our project accumulates warnings, typically from a bad rebase. We'd like to catch these warnings before they're merged into our master branch by catching them in CI when we test a pull-request.

This change adds a `--strict` flag that'll cause a non-zero exit status if any warnings are emitted.

I've had to side-step CLAide as we don't have access to the parsed options yet. Any ideas how to improve this?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [lucasmpaim](https://github.com/lucasmpaim)
   [#7432](https://github.com/CocoaPods/CocoaPods/issues/7432)
 
+* Add `--strict` flag
+  [ileitch](https://github.com/ileitch)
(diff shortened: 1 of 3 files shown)


amorde:
Nice! looks like this needs a rebase (& the changelog entry moved to the most recent section) and this will be good to go.

Thanks for the contribution & thanks for you patience on the review times 🙏

dnkoutso:
@ileitch can you rebase this? Thanks!

dnkoutso:
Is this PR the same as https://github.com/CocoaPods/CocoaPods/pull/8943?

amorde:
> Is this PR the same as #8943?

Interesting, yes but implemented slightly differently. This PR makes it so the process returns a non-zero exit code if any warnings were emitted, but doesn't interrupt the process on the first warning.

I'd say this one supercedes #8943

PR #9687, opened 2020-04-04
Honor prefix_header_file=false for subspecs

paulb777:
I noticed that GoogleUtilities installations included a pch even though its [podspec](https://github.com/CocoaPods/Specs/blob/master/Specs/0/8/4/GoogleUtilities/6.5.2/GoogleUtilities.podspec.json#L23) includes  `"prefix_header_file": false`

The fix is to check prefix_header_file on the root podspec. Subspecs do not have separate pch files.


dnkoutso:
nice!

PR #9688, opened 2020-04-04
Fix deduplicate_targets and Swift crash

paulb777:
Prevent a crash. There seems to be something more fundamental with the pod target not being properly initialized with the swift version when `deduplicate_targets` is set in the Podfile, but I wasn't able to figure that out. This might be an incremental improvement.

To reproduce add the following to a Podfile with a Swift pod (I used this [one](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseStorageSwift.podspec))

```
install! 'cocoapods',
          :deduplicate_targets => false
```

```
$ pod install
Analyzing dependencies
Downloading dependencies

――― MARKDOWN TEMPLATE ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――

### Command

```
/Users/paulbeusterien/.rbenv/versions/2.6.5/bin/pod install
```

### Report

* What did you do?

* What did you expect to happen?

* What happened instead?


### Stack

```
   CocoaPods : 1.9.1
        Ruby : ruby 2.6.5p114 (2019-10-01 revision 67812) [x86_64-darwin18]
    RubyGems : 3.0.3
        Host : Mac OS X 10.15.2 (19C57)
       Xcode : 11.3.1 (11C505)
         Git : git version 2.24.0.525.g8f36a354ae-goog
Ruby lib dir : /Users/paulbeusterien/.rbenv/versions/2.6.5/lib
Repositories : cpdc-eap-spec - git - sso://cpdc-eap/spec @ 5ea80f78a3981fd5527e3361a4b74debadc3a819

               cpdc-internal-firebase - git - sso://cpdc-internal/firebase @ eb12c904fbfa33e5d87f9f84808336e6ab319bca

               cpdc-test-mlkit - git - sso://cpdc-test/mlkit @ 2b3d6463b0c1bc3df1aa0a2d9433cccf8ac4a6ca

               firebase - git - https://github.com/firebase/SpecsStaging.git @ c40ca90b1e31514aa1c401cf5bf0e24c68f06bbf

               master - git - https://github.com/CocoaPods/Specs.git @ 766a557b9e9e60587136ca4d549d3940ac9f7ec4

               PrivateSpecRepo - git - https://github.com/tripleCC/PrivateSpecRepo.git @ 1389676849943e81066525ea090043647660089d

               staging - git - git@github.com:firebase/SpecsStaging.git @ c40ca90b1e31514aa1c401cf5bf0e24c68f06bbf

[...]

PR #9689, opened 2020-04-04
Fix setting `swift_version` when deduplicate targets is turned off.

dnkoutso:


dnkoutso:
We could consider a 1.9.3 for this...and some other fixes. WDYT?

dnkoutso:
It does not seem pressing though as I think noone is using dedup false otherwise we would see those reports coming in thousands!

dnkoutso:
ah yes will do!

paulb777:
Thanks! And I don't know of a reason for this to drive a 1.9.3. I just stumbled across it while investigating another issue.

dnkoutso:
Updated test. Works well, will merge on green.

acecilia:
I experience the same issue as explained in https://github.com/CocoaPods/CocoaPods/pull/9688: 
```
NoMethodError - undefined method `empty?' for nil:NilClass
/Users/paulbeusterien/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/cocoapods-1.9.1/lib/cocoapods/installer/xcode/target_validator.rb:118:in `block in verify_swift_pods_swift_version'
```

I tested installing cocoapods from master (which includes this PR), and I still see the crash.

The trigger is the following code:

```
pre_install do |installer|
  installer.analysis_result.specifications.each do |s|
    s.swift_versions << '4.2' if s.swift_versions.empty?
  end
end
```

You can reproduce the crash by doing `pod install` on the following Podfile:
```
use_frameworks!
platform :ios, '11.0'

pod 'InputMask', '4.2.0'

pre_install do |installer|
  installer.analysis_result.specifications.each do |s|
    s.swift_versions << '4.2' if s.swift_versions.empty?
  end
end
```

dnkoutso:
Please file a new issue and upload a sample app along with it so we can debug/diagnose if we have regressed or not handling something.

Your Podfile that you pasted also doesn't seem to specify the option that turn off deduplication which this bug was for.

acecilia:
@dnkoutso 

> Your Podfile that you pasted also doesn't seem to specify the option that turn off deduplication which this bug was for.

Correct, thats why I mentioned it in this PR. I will open an issue then, thanks

Issue #9570, opened 2020-02-28
[bug] v1.9.0 - lint failing for podspec that depends on XCTest framework


g-mark:
🌈
* [x] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

With the latest version (1.9.0) of CocoaPods, running `pod lib lint` on a podspec that includes `XCTest` as a framework dependency results in the error:

```
ld: warning: Could not find or use auto-linked library 'XCTestSwiftSupport'
```

I have a CocoaPod with two podspecs: one for the framework, and another for unit testing utilities.  The framework is here: https://github.com/ww-tech/lasso .

The [primary podspec (Lasso.podspec)](https://github.com/ww-tech/lasso/blob/develop/Lasso.podspec) contains the library files.  There's [another podspec (LassoTestUtilities.podspec)](https://github.com/ww-tech/lasso/blob/develop/LassoTestUtilities.podspec) that depends on the first pod as well as the XCTest framework.

Running `pod lib lint` on the primary podspec works as expected.
Running `pod lib lint` on the test utilities podspec results in the error.

This only happens with CodoaPods v1.9.0: v1.8.4 works great.

This also happens for other libraries with similar setups - for example [RxSwift](https://github.com/ReactiveX/RxSwift/blob/5.0.1/RxTest.podspec) (note: I tried linting RxTest.podspec in v5.0.1 of RxSwift).

## What did you do?

`pod lib lint LassoTestUtilities.podspec`

## What did you expect to happen?

I expected the podspec to successfully lint (it lints ok with CocoaPods v1.8.4).

## What happened instead?

I get the following error in the lint result:

```
 -> LassoTestUtilities (1.0.1)
    - ERROR | [iOS] xcodebuild: Returned an unsuccessful exit code.
    - NOTE  | xcodebuild:  note: Using new build system
    - NOTE  | [iOS] xcodebuild:  note: Planning build
    - NOTE  | [iOS] xcodebuild:  note: Constructing build description
    - NOTE  | xcodebuild:  ld: warning: Could not find or use auto-linked library 'XCTestSwiftSupport'
    - NOTE  | xcodebuild:  clang: error: linker command failed with exit code 1 (use -v to see invocation)
[...]

PR #9716, opened 2020-04-13
Add quotes around SWIFT_INCLUDE_PATHS in xcconfig

NickEntin:
This fixes an issue where Xcode doesn't read the SWIFT_INCLUDE_PATHS config correctly for pods that rely on XCTest.

### DIFF
--- a/lib/cocoapods/target/build_settings.rb
+++ b/lib/cocoapods/target/build_settings.rb
@@ -40,6 +40,13 @@ module Pod
         WARNING_LDFLAGS
       ).to_set.freeze
 
+      # @return [Set<String>]
+      #   The build settings that are arrays, but should be wrapped in quotes.
+      #
+      QUOTE_WRAPPED_SETTINGS = %w(
+        SWIFT_INCLUDE_PATHS
+      ).to_set.freeze
+
       # @return [String]
       #   The variable for the configuration build directory used when building pod targets.
       #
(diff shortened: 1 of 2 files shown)


NickEntin:
I think this should resolve #9570.

lib/cocoapods/target/build_settings.rb
```diff
@@ -371,7 +378,16 @@ def add_inherited_to_plural(hash)
           if PLURAL_SETTINGS.include?(key)
             raise ArgumentError, "#{key} is a plural setting, cannot have #{value.inspect} as its value" unless value.is_a? Array
 
-            value = "$(inherited) #{quote_array(value)}"
+            if QUOTE_WRAPPED_SETTINGS.include?(key)
+              if value.empty?
+                value = "$(inherited)"
+              else
+                value_to_wrap = "$(inherited) #{quote_array(value)}".strip
```
segiddins, line 385: do the quotes from `quote_array` need to be escaped?

NickEntin: I don't think so. It would make sense to need that, but this doesn't seem to work the same as the other settings anyway. From testing by manually changing the xcconfig to see what works...

Quotes around the individual path only doesn't build:
```
SWIFT_INCLUDE_PATHS = $(inherited) "$(PLATFORM_DIR)/Developer/usr/lib"
```

No quotes doesn't build:
```
SWIFT_INCLUDE_PATHS = $(inherited) $(PLATFORM_DIR)/Developer/usr/lib
```

Quotes around the whole paths array, but no quotes internally does build:
```
SWIFT_INCLUDE_PATHS = "$(inherited) $(PLATFORM_DIR)/Developer/usr/lib"
```

Adding quotes around the quoted paths does build:
```
SWIFT_INCLUDE_PATHS = "$(inherited) "$(PLATFORM_DIR)/Developer/usr/lib""
```

From looking at the diff in `cocoapods-integration-specs` after running `bundle exec rake spec:rebuild_integration_fixtures`, we're outputting the last one.

segiddins: that's... super super weird. thanks for checking!!

segiddins:
Looks like this is causing at least one of the examples to fail to compile: https://circleci.com/gh/CocoaPods/CocoaPods/4837?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

NickEntin:
This causes one of the examples to fail. Closing in favor of #9736.

PR #9761, opened 2020-05-02
Remove defunct step from CircleCI

sonicdoe:
Since 4d21f1c964896a1a1999c0b286d6c03f33a51132, this step no longer has any effect because [CircleCI images using macOS 10.15 (Xcode 11.2) and later no longer have the auto-switching feature enabled](https://circleci.com/docs/2.0/testing-ios/#using-ruby).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,6 +7,9 @@ jobs:
     steps:
       - checkout
 
+      - run:
+          name: Set Ruby Version
+          command:  echo "2.4" > ~/.ruby-version
       - restore_cache:
           keys:
             - v1-gems-{{ checksum "Gemfile.lock" }}
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,7 @@ jobs:
 
       - run:
           name: Set Ruby Version
-          command:  echo 'chruby ruby-2.4' >> ~/.bash_profile
+          command:  echo "2.4" > ~/.ruby-version
       - restore_cache:
           keys:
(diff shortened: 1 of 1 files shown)


amorde:
Can we update it to use the newly recommended way to achieve this instead?

> To select a version of Ruby to use, add the chruby function to ~/.bash_profile

sonicdoe:
We can’t easily switch back to Ruby 2.4 because the [Xcode 11.3.1 CircleCI image](https://circle-macos-docs.s3.amazonaws.com/image-manifest/v2244/index.html) only makes Ruby 2.5.7, 2.6.5, and 2.7.0 available to `chruby`.

Wouldn’t it be preferable to stay with Ruby 2.6 since it’s been in use since 4d21f1c964896a1a1999c0b286d6c03f33a51132 and [support for Ruby 2.4 has already ended](https://www.ruby-lang.org/en/news/2020/04/05/support-of-ruby-2-4-has-ended/)?

amorde:
CocoaPods supports old Ruby versions because Apple tends to ship older Ruby versions. For example, Mojave ships with `2.3.7p456`, so even if official support has ended we still need to support it.

That said, I'm personally ok with updating CircleCI to use something newer since CircleCI only builds example projects

sonicdoe:
Makes sense 👍 I’ve now reverted the pull request to just remove the defunct step.

kapooklovely:
https://github.com/CocoaPods/CocoaPods/pull/9761#issuecomment-624991554

PR #9720, opened 2020-04-16
XCFramework Rework

amorde:
Closes https://github.com/CocoaPods/CocoaPods/issues/9528
Closes https://github.com/CocoaPods/CocoaPods/issues/9670
Closes https://github.com/CocoaPods/CocoaPods/issues/9525

Requires https://github.com/CocoaPods/cocoapods-integration-specs/pull/280

This PR reworks how xcframework are handled by moving the selection of the correct "slice" to the pod target instead of the aggregate target, similar to what was done in https://github.com/CocoaPods/CocoaPods/pull/9547

This also adds support for xcframeworks that wrap static libraries, as described in https://github.com/CocoaPods/CocoaPods/issues/9528.

## Implementation Details

A new script phase `Embed XCFrameworks` is added to the *Pod Target* that contains the vendored xcframework. This script copies the correct slice (either a `.framework` or static lib + headers) to `${PODS_XCFRAMEWORKS_BUILD_DIR}/[xcframework_name]/`. The existing `Embed Frameworks` script will then import the framework from `${PODS_XCFRAMEWORKS_BUILD_DIR}/[xcframework_name]/` to the target build directory.

`${FRAMEWORK_SEARCH_PATHS}` and/or `${LIBRARY_SEARCH_PATHS}` now include `${PODS_XCFRAMEWORKS_BUILD_DIR}/[xcframework_name]/` if a relevant xcframework exists for a given host target.

`${HEADER_SEARCH_PATHS}` now includes `${PODS_XCFRAMEWORKS_BUILD_DIR}/[xcframework_name]/Headers` if relevant static-library-based xcframeworks exist for a given host target.

Issue #9569, opened 2020-02-28
'Prepare Artifacts' stage failed to find .xcframework slice for 'armv7k arm64_32' architectures


bcosescu:
* [x] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report
Cocoapods script for artifacts fails to correctly find the slice for 'armv7k arm64_32' architectures in a .xcframework.

## What did you do?
1. Created a framework for watchOS with valid architectures: armv7k and arm64_32
2. Created an .xcframework this framework for both the simulator and the architectures above
     The resulted .xcframework had the following structure:
        MyFramework.xcframework
            - Info.plist
            - watchos-armv7k_arm64_32
            - watchos-i386-simulator
3. Created a local podspec called MyFramework.podspec to serve the .xcframework
4. Created a sample app project with watch extension and added the following line for the WatchKit Extension target:
       pod 'MyFramework', :path => "<PATH_TO_MY_PODSPEC>" 
5. Run `pod install`, everything went smooth.
6. Run the WatchKit Extension on a device

## What did you expect to happen?
The test app should have run on the watch device.

## What happened instead?

Watch app crashed:
dyld: dependent dylib '@rpath/MyFramework.framework/MyFramework' not found for '/private/var/containers/Bundle/Application/[UUID]/TestFrm2 WatchKit App.app/PlugIns/TestFrm2 WatchKit Extension.appex/TestFrm2 WatchKit Extension', tried but didn't find: ...

Build log showed:
[CP] Unable to find matching .xcframework slice in 'true watchos-armv7k_arm64_32/MyFramework.framework watchos-i386-simulator/MyFramework.framework' for the current build architectures (armv7k arm64_32).

### Stack

```
   CocoaPods : 1.9.0
        Ruby : ruby 2.6.3p62 (2019-04-16 revision 67580) [universal.x86_64-darwin19]
    RubyGems : 3.0.3
        Host : Mac OS X 10.15.3 (19D76)
       Xcode : 11.3.1 (11C504)
         Git : git version 2.21.1 (Apple Git-122.3)
Ruby lib dir : /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib
Repositories : master - git - https://github.com/CocoaPods/Specs.git @ b33b78a48a78fa1ddff1fdb4b1ddcf673ad2837e

[...]

PR #9790, opened 2020-05-14
Fix XCFramework slice selection when having more archs in slice than requested with $ARCHS

jerbob92:
Currently, the XCFramework slice selection uses the whole folder path to select the correct slice.
This works fine when you are building the exact same ARCHS as the one in the slice. But this fails when you build less ARCHS than are in the slice, for example:

Your slice folder name is `ios-armv7_armv7s_arm64_arm64e/CoconutLib.framework`, you `$ARCHS` is `armv7 arm64`. The current code replaces all spaces with underscores, so it becomes `armv7_arm64`, it then tries to find `armv7_arm64` in the folder name, but because `armv7s` is in between, it won't find it.

This PR splits the `ARCHS` on whitespace and then tries to find a slice that contains all separate `ARCHS` in the folder name. This is a continuation of work done in #9569, #9575 and #9720.

Note that this still has an issue that the current code also has: if you are building `armv7`, and your slice doesn't contain `armv7`, but does contain `armv7s`, it will still match it.
So my suggestion for a more robust slice matching in the future is using the actual `Info.plist` slice information, but I'm not sure how much work this is.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,6 +96,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Kenji KATO](https://github.com/katoken-0215)
   [#9706](https://github.com/CocoaPods/CocoaPods/pull/9706)
 
+* Fix XCFramework slice selection when having more slices than requested  
+  [jerbob92](https://github.com/jerbob92)
+  [#9790](https://github.com/CocoaPods/CocoaPods/pull/9790)
+
 ## 1.9.1 (2020-03-09)
 
 ##### Enhancements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,7 +96,7 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Kenji KATO](https://github.com/katoken-0215)
   [#9706](https://github.com/CocoaPods/CocoaPods/pull/9706)
 
-* Fix XCFramework slice selection when having more slices than requested  
+* Fix XCFramework slice selection when having more archs in slice than requested with $ARCHS  
   [jerbob92](https://github.com/jerbob92)
   [#9790](https://github.com/CocoaPods/CocoaPods/pull/9790)
 
--- a/lib/cocoapods/generator/copy_xcframework_script.rb
+++ b/lib/cocoapods/generator/copy_xcframework_script.rb
@@ -82,7 +82,7 @@ select_slice() {
   local target_path=""
 
   # Split archs on space so we can find a slice that has all the needed archs
-  local target_archs=$(echo $ARCHS | tr " " "\n")
+  local target_archs=$(echo $ARCHS | tr " " "\\n")
 
   local target_variant=""
   if [[ "$PLATFORM_NAME" == *"simulator" ]]; then
@@ -104,7 +104,7 @@ select_slice() {
       # We match the following: -armv7_, _armv7s_, _arm64_ and _arm64e/.
       # If we have a specific variant: ios-i386_x86_64-simulator/CoconutLib.framework
       # We match the following: -i386_ and _x86_64-
-      local target_arch_regex="[_\-]${target_arch}[\/_\-]"
+      local target_arch_regex="[_\\-]${target_arch}[\\/_\\-]"
       if ! [[ "${paths[$i]}" =~ $target_arch_regex ]]; then
         matched_all_archs="0"
       fi
--- a/lib/cocoapods/generator/copy_xcframework_script.rb
+++ b/lib/cocoapods/generator/copy_xcframework_script.rb
@@ -98,7 +98,14 @@ select_slice() {
       if ! [[ "${paths[$i]}" == *"$target_variant"* ]]; then
         matched_all_archs="0"
       fi
-      if ! [[ "${paths[$i]}" == *"$target_arch"* ]]; then
+
+      # This regex matches all possible variants of the arch in the folder name:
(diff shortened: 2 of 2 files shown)


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Fix XCFramework slice selection when having more slices than requested  
  [jerbob92](https://github.com/jerbob92)
```

</p>

jerbob92:
The build is failing because it's comparing some script, can anyone tell me where I can find this?

CHANGELOG.md
```diff
@@ -96,6 +96,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Kenji KATO](https://github.com/katoken-0215)
   [#9706](https://github.com/CocoaPods/CocoaPods/pull/9706)
 
+* Fix XCFramework slice selection when having more slices than requested
```
amorde, line 99: nit: I think you mean `archs` here, `slice` refers to each individual `.framework` that exists in the xcframework bunde

lib/cocoapods/generator/copy_xcframework_script.rb
```diff
@@ -93,7 +92,18 @@ def script
...
   fi
   for i in ${!paths[@]}; do
-    if [[ "${paths[$i]}" == *"$target_arch"* ]] && [[ "${paths[$i]}" == *"$target_variant"* ]]; then
+    local matched_all_archs="1"
+    for target_arch in $target_archs
+    do
+      if ! [[ "${paths[$i]}" == *"$target_variant"* ]]; then
+        matched_all_archs="0"
+      fi
+      if ! [[ "${paths[$i]}" == *"$target_arch"* ]]; then
```
amorde, line 101: since we are checking each architecture individually, could we remove the `*` and check that they are exactly equal? I think that would resolve the `armv7` case

amorde:
I also wanted to use the Info.plist but parsing that within a shell scrip is a bit messy, even when using something like `plutil`.

lib/cocoapods/generator/copy_xcframework_script.rb
amorde: ah nevermind I see why that woulnd't work now, we're still checking the whole list against the target

jerbob92: We could check if we match one of `-$target_arch_`, `_$target_arch_` or `_$target_arch-`, then it's a proper check. But still quite nasty.

jerbob92:
> I also wanted to use the Info.plist but parsing that within a shell script is a bit messy, even when using something like `plutil`.

Yeah I get that, tried to see whether any xcode tool can easily return info about the xcframework, but doesn't seem like it.

lib/cocoapods/generator/copy_xcframework_script.rb
amorde: maybe we could use `grep` with a regex? I don't mean to delay this improvement but I also don't want to swap one bug for another 😅

amorde: its possible a framework _only_ supports 1 arch, and therefore wouldn't have `_` as a prefix & suffix

jerbob92: Then we also need to add -$target_arch- to the list. It't not really swapping bugs, this problem already exists in the current implementation. I will check whether we can fix it with a regex :)

jerbob92: @amorde I updated the PR to use a regex. I think this is a solid solution for now :)

jerbob92: It's not completely working, for some reason x86_64 doesn't work in the regex

jerbob92: Fixed! :)

lib/cocoapods/generator/copy_xcframework_script.rb
```diff
@@ -93,11 +92,31 @@ def script
...
+        matched_all_archs="0"
+        break
+      fi
+
+      # This regex matches all possible variants of the arch in the folder name:
+      # Let's say the folder name is: ios-armv7_armv7s_arm64_arm64e/CoconutLib.framework
+      # We match the following: -armv7_, _armv7s_, _arm64_ and _arm64e/.
+      # If we have a specific variant: ios-i386_x86_64-simulator/CoconutLib.framework
+      # We match the following: -i386_ and _x86_64-
+      local target_arch_regex="[_\\-]${target_arch}[\\/_\\-]"
```
amorde, line 108: I think we might need to account for something like `-ios-armv7/CoconutLib.framework`, so should the second `[_\-]` get `/` added as well?

jerbob92: I'm not sure what you are talking about? That regex also matches `-armv7/`

amorde: ah you're right, sorry the escapes through me off 👍

amorde:
integration specs were updated, builds should pass one this PR updates the submodule

jerbob92:
> integration specs were updated, builds should pass one this PR updates the submodule

Done and everything's green :)

amorde:
@jerbob92 thank you!

Issue #9536, opened 2020-02-14
[Podspec] Explicit setting of 'CFBundleIdentifier' in s.info_plist causes a warning


giofid:
As discussed [here]( #3032), when in the .podspec file we explicitly set the bundle identifier in this way

```
# in .podspec
s.info_plist = {
  'CFBundleIdentifier' => 'com.myorg.mylib'
}
```

XCode 11 gives a warning 

`warning: User-supplied CFBundleIdentifier value 'com.myorg.mylib' in the Info.plist must be the same as the PRODUCT_BUNDLE_IDENTIFIER build setting value 'org.cocoapods.mylib'`

dnkoutso:
Probably a minor bug yes, care to make a PR?

amorde:
For anyone interested in contributing, this is currently set here:

https://github.com/CocoaPods/CocoaPods/blob/a1c94aaaa21b3da318208c7426433753c0332716/lib/cocoapods/target/build_settings.rb#L901

This could potential be moved to somewhere a bit higher level such as here: 

https://github.com/CocoaPods/CocoaPods/blob/a1c94aaaa21b3da318208c7426433753c0332716/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb#L199-L220

PR #9792, opened 2020-05-17
[Pod Target Installer] Add PRODUCT_BUNDLE_IDENTIFIER to build settings when info_plist defines a bundle id

seanreinhardtapps:
Closes CocoaPods/CocoaPods#9536

Took a shot at resolving the warning.  I tried not setting the `PRODUCT_BUNDLE_IDENTIFIER` build setting when there is a `CFBundleIdentifier` set in the spec's info_plist, however an empty value will still log the warning, since the Empty String doesn't match the value in CFBundleIdentifier.

I did resolve the warning by using the `CFBundleIdentifier`  to set the build setting `PRODUCT_BUNDLE_IDENTIFIER`.  I'm unsure if this is an acceptable alteration of the target, but putting it up for review.  There's also an edge case where the Podspec configures the bundle id with both `pod_target_xcconfig` and `info_plist` and with different values; in this case the value set in `pod_target_xcconfig` will win and there will be the same warning logged.

Added a warning with a hint to set the bundle id using `pod_target_xcconfig`, since that appears to be the correct way to set the id with Xcode 11.

### DIFF
--- a/spec/unit/installer/xcode/pods_project_generator/pod_target_installer_spec.rb
+++ b/spec/unit/installer/xcode/pods_project_generator/pod_target_installer_spec.rb
@@ -83,9 +83,48 @@ module Pod
               end
             end
 
+            it 'adds bundle identifier from info_plist to build settings' do
+              @banana_spec.info_plist = {
+                'CFBundleIdentifier' => 'some.bundle.id',
+              }
+              @installer.install!
+              @project.targets.first.build_configurations.each do |config|
+                config.resolve_build_setting('PRODUCT_BUNDLE_IDENTIFIER').should == 'some.bundle.id'
+              end
+            end
+
             it 'cleans up temporary directories' do
               @installer.expects(:clean_support_files_temp_dir).once
               @installer.install!
+            end
+
+            describe 'info_plist_bundle_id' do
+              it 'cache and return CFBundleIdentifier value when present in info_plist hash' do
+                @installer.target.root_spec.info_plist = { 'CFBundleIdentifier' => 'CocoaPods.test.id' }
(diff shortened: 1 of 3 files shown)


lib/cocoapods/target.rb
```diff
@@ -298,6 +298,23 @@ def info_plist_entries
       {}
     end
 
+    # @return [String] Bundle Identifier found in the custom Info.plist entries
+    #
+    def info_plist_bundle_id
```
amorde, line 303: can this be move to `pod_target_installer`? I think the logic here (including the warning) makes more sense to be localized to the installer and not exposed on every instance of `PodTarget`

spec/unit/installer/xcode/pods_project_generator/pod_target_installer_spec.rb
```diff
@@ -83,6 +83,16 @@ class PodsProjectGenerator
               end
             end
 
+            it 'adds bundle identifier from info_plist to build settings' do
```
amorde, line 86: thanks for adding a test! 🎉

amorde:
This LGTM with 1 potential change

lib/cocoapods/target.rb
seanreinhardtapps: @amorde, thanks for the input, moved the method to `pod_target_installer`.

amorde: looks like it's now in `pod_target`, not `pod_target_installer`

seanreinhardtapps: Apologies for that!  I've made the expected update now.

dnkoutso:
@amorde anything else here?

PR #9847, opened 2020-06-05
Stop adding app spec dependencies to library target in incremental installation

segiddins:
@sebastianv1 any pointers where I should look to add a test for this? I couldn't see any covering dependency wiring

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -117,6 +117,12 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [jerbob92](https://github.com/jerbob92)
   [#9790](https://github.com/CocoaPods/CocoaPods/pull/9790)
 
+* Don't add app spec dependencies to the parent library's target in Xcode,
+  which was happening when the dependency's project was not being regenerated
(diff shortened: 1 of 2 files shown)


dnkoutso:
`pod_target_dependency_installer` was extracted as a common class as part of https://github.com/CocoaPods/CocoaPods/pull/8368 but test cases that were related to it did not get extracted.

Tests that cover its functionality are covered within 
ttps://github.com/CocoaPods/CocoaPods/blob/master/spec/unit/installer/xcode/single_pods_project_generator_spec.rb and  https://github.com/CocoaPods/CocoaPods/blob/master/spec/unit/installer/xcode/multi_pods_project_generator_spec.rb

If you want to add a test it should be in the latter because incremental installation requires multiple Xcode projects.

A bit surprised that nothing failed but maybe there is nothing that covers this case today.

sebastianv1:
Yeah, what @dnkoutso, wish I had come up with an easy way to also create an independent unit test case, but I think I extracted it originally because the logic started to get unruly.

Otherwise, changes look fine to me.

PR #9851, opened 2020-06-07
Bump minimum Ruby to 2.3.3

### LINKED ISSUE #9821: Drop support for Ruby 2.0
* [X] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md). 🌈

macOS 10.13.3 (High Sierra) ships Ruby 2.3.3, and this is the required OS version for Xcode 10. Since we typically support [current Xcode] and [previous Xcode], this is safe to drop.

This is a prerequisite for https://github.com/CocoaPods/CocoaPods/issues/9404, which also unblocks dependabot support https://github.com/dependabot/dependabot-core/pull/731

amorde:
Closes #9821


lib/cocoapods/command/spec/lint.rb
```diff
@@ -56,7 +56,7 @@ def initialize(argv)
           @swift_version   = argv.option('swift-version', nil)
           @skip_import_validation = argv.flag?('skip-import-validation', false)
           @skip_tests      = argv.flag?('skip-tests', false)
-          @test_specs      = argv.option('test-specs', nil).try(:split, ',')
+          @test_specs      = argv.option('test-specs', nil)&.split(',')
```
amorde, line 59: this was using an `activesupport` function that doesn't seem to be available now, but now it isn't needed since we can use `&.` :D

amorde:
Need to remove some uses of `try`

Gemfile
```diff
@@ -38,26 +38,26 @@ group :development do
 
   gem 'cocoapods-dependencies', '~> 1.0.beta.1'
 
-  gem 'activesupport', '>= 4.0.2', '< 5', '!= 4.2.11.12' # Pinned < 5 to ensure we're speccing 4.x.x
+  gem 'activesupport', '> 5', '< 6' # Pinned < 6 because 6 requires Ruby 2.5.0
```
Kaspik, line 41:
```suggestion
if Gem::Version.create(RUBY_VERSION) < Gem::Version.create("2.5.0")
  gem 'activesupport', '> 5', '< 6'
else		
  gem 'activesupport', '> 5'
end
```

cocoapods.gemspec
```diff
@@ -39,8 +39,8 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'molinillo',             '~> 0.6.6'
   s.add_runtime_dependency 'xcodeproj',             '>= 1.14.0', '< 2.0'
 
-  ## Version 5 needs Ruby 2.2, so we specify an upper bound to stay compatible with system ruby
-  s.add_runtime_dependency 'activesupport', '>= 4.0.2', '< 5'
+  ## Version 6 needs Ruby 2.5.0, so we specify an upper bound to stay compatible with system ruby
+  s.add_runtime_dependency 'activesupport', '> 5', '< 6'
```
Kaspik, line 43: The same as in gemfile can be here then?

segiddins: Conditionals in the gemspec dont work, since whichever branch is taken will be the only one serialized when pushing the gem to rubygems.org

segiddins: Ideally, we remove this constraint entirely and tell people to use bundler if they need to pin to a specific version of activesupport

PR #9863, opened 2020-06-15
[Target Integrator] check compatibilityVersion to decide when to use xcfilelists.

### LINKED ISSUE #9140: pod install does not respect decreasing Xcode compatibility version
## What did you do?

1. open project: https://github.com/Speakus/CocoapodIssueChangeCompatibilityVersion
2. bundle install
3. bundle exec pod install
I already changed compatibility version to 8.0 but cocoapod system ignore this change.
the commit since changing compatibility version here:
https://github.com/Speakus/CocoapodIssueChangeCompatibilityVersion/commit/bd645ac777d9f620fad9854c606b81f48d8e6a07

cocoapod will respect compatibility version

## What happened instead?

Pods project still use compatibility version 10.0 and use `.xcfilelist` inside build phase:
https://github.com/Speakus/CocoapodIssueChangeCompatibilityVersion/blob/master/CocoapodIssueChangeCompatibilityVersion.xcodeproj/project.pbxproj#L166

## CocoaPods Environment

ℹ Please replace these two lines with the output of `pod env`.
e.g. via `pod env | pbcopy`

### Stack

```
   CocoaPods : 1.9.1
        Ruby : ruby 2.3.7p456 (2018-03-28 revision 63024) [x86_64-darwin18]
    RubyGems : 2.5.2.3
        Host : Mac OS X 10.14.6 (18G3020)
       Xcode : 11.3 (11C29)
         Git : git version 2.26.0
Ruby lib dir : /Users/speakus/.rbenv/versions/2.3.7/lib
Repositories : artsy - git - https://github.com/artsy/Specs.git @ 285dab6864a02da926cb0786f58ee3d800ffe372

               AtlasCocoapods - git - https://bitbucket.org/firstatlas/AtlasCocoapods.git @ 7dcbeaefdc11b7a1f120b8a1a20d274d9d57d975

               bitbucket-simpletechs-avoplayer-cocoapods - git - https://bitbucket.org/simpletechs/avoplayer-cocoapods.git @ 1786e8d99ad65259519565988d6f402e8b8fe4de

               cocoapods - git - https://github.com/CocoaPods/Old-Specs @ 6e256ccc84aad851d401fabb79b2c0f9e09bb875

               goinstant - git - https://github.com/goinstant/pods-specs-public @ 03600012d72355c5ac7f5f1060c023c18c2035ae

               master - git - https://github.com/CocoaPods/Specs.git @ b3447f0ca8350c384edef56d6bfb8f227f4bf19c

               trunk - CDN - https://cdn.cocoapods.org/
               twilio - git - https://github.com/twilio/cocoapod-specs @ 9bf659ed7246ba34625b81754031fbd5ac20fe68

```

### Installation Source

```
Executable Path: /Users/speakus/.rbenv/versions/2.3.7/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.4
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.1.0
cocoapods-trunk       : 1.4.1
cocoapods-try         : 1.1.0
```

### Podfile

```ruby
# Podfile syntax here: https://guides.cocoapods.org/syntax/podfile.html
minTarget = '10.0'

def crashPods()
    # for getting useful info about crashes
    pod 'Crashlytics' # Commercial
    pod 'Fabric' # Commercial
end

platform :ios, minTarget
use_frameworks!

# ignore all warnings from all pods
inhibit_all_warnings!

target 'CocoapodIssueChangeCompatibilityVersion' do
  crashPods
  pod 'PKHUD'
end
```

https://github.com/Speakus/CocoapodIssueChangeCompatibilityVersion

seanreinhardtapps:
integration specs PR: https://github.com/CocoaPods/cocoapods-integration-specs/pull/285

Closes #9140.
This PR changes the logic for using file lists to check the `compatibilityVersion` of the Xcode Project, instead of the `objectVersion`. 
I did notice in my own testing that changing the compatibility version from the Xcode UI did change both values.  It seems to me that only manually updating the Compatibility Version in the pbxproj file would result in a situation where either value doesn't correctly describe whether file lists can be used.
Also, the previous logic enabled file lists whenever the `objectVersion` was >= 50 (Xcode 9.3), but to the best of my knowledge .xcfilelists were introduced in Xcode 10.0 (objectVersion 51).

### DIFF
--- a/lib/cocoapods/installer/user_project_integrator/target_integrator.rb
+++ b/lib/cocoapods/installer/user_project_integrator/target_integrator.rb
@@ -60,6 +60,14 @@ module Pod
           'Prepare Artifacts'.freeze,
         ].freeze
 
+        # @return [float] Returns Minimum Xcode Compatibility version for FileLists
+        #
+        MIN_FILE_LIST_COMPATIBILITY_VERSION = 9.3
+
+        # @return [String] Returns Minimum Xcode Object version for FileLists
+        #
+        MIN_FILE_LIST_OBJECT_VERSION = 50
+
         # @return [AggregateTarget] the target that should be integrated.
         #
         attr_reader :target
@@ -90,7 +98,14 @@ module Pod
           #         should be stored in a file list file.
           #
           def input_output_paths_use_filelist?(object)
-            object.project.object_version.to_i >= 50
+            unless object.project.root_object.compatibility_version.nil?
+              version_match = object.project.root_object.compatibility_version.match(/Xcode ([0-9]*\.[0-9]*)/).to_a
+            end
+            if version_match&.at(1).nil?
+              object.project.object_version.to_i >= MIN_FILE_LIST_OBJECT_VERSION
+            else
+              Pod::Version.new(version_match[1]) >= Pod::Version.new(MIN_FILE_LIST_COMPATIBILITY_VERSION)
+            end
           end
 
           # Sets the input & output paths for the given script build phase.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,11 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
   [#9871](https://github.com/CocoaPods/CocoaPods/pull/9871)
 
+* Use User Project's compatibilityVersion instead of objectVersion when
+  deciding when to use xcfilelists.  
+  [Sean Reinhardt](https://github.com/seanreinhardtapps)
+  [#9140](https://github.com/CocoaPods/CocoaPods/issues/9140)
+  
 * add a `--configuration` option to `pod lib lint` and `pod spec lint`.  
   [Gereon Steffens](https://github.com/gereons)
   [#9686](https://github.com/CocoaPods/CocoaPods/issues/9686)
--- a/spec/unit/installer/xcode/pods_project_generator/pod_target_integrator_spec.rb
+++ b/spec/unit/installer/xcode/pods_project_generator/pod_target_integrator_spec.rb
@@ -86,6 +86,33 @@ module Pod
               end
 
               it 'integrates test native targets with frameworks and resource script phase input and output file lists' do
+                @project.root_object.stubs(:compatibility_version).returns('Xcode 9.10')
+                framework_paths = [Pod::Xcode::FrameworkPaths.new('${PODS_ROOT}/Vendored/Vendored.framework')]
+                resource_paths = ['${PODS_CONFIGURATION_BUILD_DIR}/TestResourceBundle.bundle']
+                @coconut_pod_target.stubs(:framework_paths).returns('CoconutLib' => framework_paths)
+                @coconut_pod_target.stubs(:resource_paths).returns('CoconutLib' => resource_paths)
+                PodTargetIntegrator.new(@coconut_target_installation_result, :use_input_output_paths => true).integrate!
+                @test_native_target.build_phases.count.should == 2
+                @test_native_target.build_phases.map(&:display_name).should == [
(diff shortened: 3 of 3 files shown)


dnkoutso:
> I did notice in my own testing that changing the compatibility version from the Xcode UI did change both values.

sorry can you explain why this isnt working then from CocoaPods side?

seanreinhardtapps:
The issue opener had the expectation that they could lower the Compatibility Version directly in their project.pbxproj file, and remove File Lists from their project.  However, CocoaPods bases use of File Lists on the Project's objectVersion.
This normally wouldn't matter - they are two different values, but updating the Compatibility Version in Xcode's UI changes both values.
Essentially, the question is whether we want to control File Lists on the objectVersion (in which case we close this PR) or Compatibility Version.

dnkoutso:
So this fixes it that if it either happens manually by editing `.pbxproj` file _or_ the UI it would work properly if I understand correctly...

seanreinhardtapps:
@dnkoutso correct, this works with Xcode UI-based updates, or manual `.pbxproj` updates.

dnkoutso:
I guess that should be OK...I am not a huge fan of parsing the string to be honest as it maybe can change ...

seanreinhardtapps:
@dnkoutso we could have it fall back to `objectVersion` if it fails to parse the `compatibilityVersion` string for an Xcode version.  Id be happy to include that for stability purposes.

dnkoutso:
@seanreinhardtapps perhaps yes that is a good idea for fallback! I am also seeing it being available in release notes for Xcode 10 but honestly I think we had it working on the prior version, hence why it was 50.

lib/cocoapods/installer/user_project_integrator/target_integrator.rb
```diff
@@ -90,7 +98,15 @@ class << self
           #         should be stored in a file list file.
           #
           def input_output_paths_use_filelist?(object)
-            object.project.object_version.to_i >= 50
+            unless object.project.root_object.compatibility_version.nil?
+              version_match = object.project.root_object.compatibility_version.match(/Xcode ([0-9]*.[0-9]*)/)
+              compatibility_version = version_match[1].to_f unless version_match.nil? || version_match.length < 2
+            end
+            if compatibility_version.to_f > 0
+              compatibility_version.to_f >= MIN_FILE_LIST_COMPATIBILITY_VERSION
```
amorde, line 106: I'd like to avoid floating point comparisons here - can we compare `Pod::Version.new(compatibility_version) >= Pod::Version.new(MIN_FILE_LIST_COMPATIBILITY_VERSION)` instead?

lib/cocoapods/installer/user_project_integrator/target_integrator.rb
```diff
@@ -90,7 +98,15 @@ class << self
           #         should be stored in a file list file.
           #
           def input_output_paths_use_filelist?(object)
-            object.project.object_version.to_i >= 50
+            unless object.project.root_object.compatibility_version.nil?
+              version_match = object.project.root_object.compatibility_version.match(/Xcode ([0-9]*.[0-9]*)/)
+              compatibility_version = version_match[1].to_f unless version_match.nil? || version_match.length < 2
```
amorde, line 103: I think we also don't want `to_f` here - ex. `'10.10'` should keep the `.10` suffix and not result in `.1`

amorde: of course I hope we never have 10+ minor updates to a major Xcode release.. but you never know :)

seanreinhardtapps: Thanks for the catch, updated.

Issue #9794, opened 2020-05-17
Pod with XCFrameworks embedded libraries not compiling for "Generic iOS Device"


remizorrr:
Setup is:
Xcode 11.4.1,
Cocapods top of tree master.

Local pod, with embedded (vendored_frameworks) xcframeworks.
the project, referencing local pod.

compiling and running on the iPhone - everything works.
compiling and running on the Simulator - everything works.
Compiling for "Generic iOS Device" - Shows "No such module 'FrameworkName'"
for the embedded XCFramework.

A project with same xcframework, embedded using Xcode compiles fine.

Example projects attached:
[Project-Not-Working.zip](https://github.com/CocoaPods/CocoaPods/files/4640013/Project-Not-Working.zip)
[Project-Working.zip](https://github.com/CocoaPods/CocoaPods/files/4640014/Project-Working.zip)

related to the previously closed issue: https://github.com/CocoaPods/CocoaPods/issues/9785

dnkoutso:
Did you try cocoapods master?

remizorrr:
Yes. the issue is on master.

remizorrr:
@dnkoutso any updates?

dnkoutso:
No updates here, I am sorry I do not currently have the time to look into this. Best to investigate further yourself and perhaps make a PR so we can fix it for everyone if its an issue.

remizorrr:
@dnkoutso I had a peak in the code. It will take a lot of time for me to find a cause of the issue. Is there anyone that could get it done? Looks like I am no the only one having issues with xcframeworks.

amorde:
We have a lot of changes to xcframeworks that will be released in the next version, but since you tried master would need to investigate further.

Personally have been taking a bit of a break from CocoaPods and have started to fall behind on other things, so I'm not sure when I will have time to look in to this. The example projects are super helpful, thank you for including them 🙏

remizorrr:
@amorde Thank you a lot. I will greatly appreciate if you could look into that. It is currently blocking our library consumers from deploying their apps to the app store. Let me know if you need any help. I would love to give you a hand here.

amorde:
Off the top of my head I would check the architectures and whether it was built with the BUILD_LIBRARY_FOR_DISTRIBUTION.

What is the difference between the working and not working examples you uploaded?

seanreinhardtapps:
[...]

PR #9866, opened 2020-06-17
remove some non-required architectures from xcframework select_slice

seanreinhardtapps:
After looking at #9794, I had a couple thoughts about the copy_xcframework_script.
There are a couple of arm architectures, arm64e and armv7s, that could be part of the app's Valid Architectures, but as far as I'm aware are not required to have a complete FAT binary. (arm64e is backwards compatible to arm64 and same with armv7s and armv7).  I'm wondering if these should be excluded from the required architectures in select_slice().

I also wanted to propose returning code 1 when an architecture couldn't be found, to help point out the error.

### DIFF
--- a/lib/cocoapods/generator/copy_xcframework_script.rb
+++ b/lib/cocoapods/generator/copy_xcframework_script.rb
@@ -84,6 +84,10 @@ select_slice() {
   # Split archs on space so we can find a slice that has all the needed archs
   local target_archs=$(echo $ARCHS | tr " " "\\n")
 
+  # arm64e and arm7vs are not required architectures for a device slice
+  target_archs=( "${target_archs[@]/arm64e}" )
+  target_archs=( "${target_archs[@]/armv7s}" )
+  
   local target_variant=""
   if [[ "$PLATFORM_NAME" == *"simulator" ]]; then
     target_variant="simulator"
@@ -158,7 +162,7 @@ install_xcframework_library() {
   local target_path="$SELECT_SLICE_RETVAL"
   if [[ -z "$target_path" ]]; then
     echo "warning: [CP] Unable to find matching .xcframework slice in '${paths[@]}' for the current build architectures ($ARCHS)."
-    return
+    exit 1
   fi
 
   install_framework "$basepath/$target_path" "$name"
@@ -175,7 +179,7 @@ install_xcframework() {
   local target_path="$SELECT_SLICE_RETVAL"
   if [[ -z "$target_path" ]]; then
     echo "warning: [CP] Unable to find matching .xcframework slice in '${paths[@]}' for the current build architectures ($ARCHS)."
-    return
+    exit 1
   fi
   local source="$basepath/$target_path"
 


lib/cocoapods/generator/copy_xcframework_script.rb
```diff
@@ -158,7 +162,7 @@ def script
   local target_path="$SELECT_SLICE_RETVAL"
   if [[ -z "$target_path" ]]; then
     echo "warning: [CP] Unable to find matching .xcframework slice in '${paths[@]}' for the current build architectures ($ARCHS)."
-    return
+    exit 1
```
amorde, line 165: if we're changing this we should also change the above `warning:` to `error:` so it shows a proper error in Xcode

lib/cocoapods/generator/copy_xcframework_script.rb
```diff
@@ -84,6 +84,10 @@ def script
   # Split archs on space so we can find a slice that has all the needed archs
   local target_archs=$(echo $ARCHS | tr " " "\\n")
 
+  # arm64e and arm7vs are not required architectures for a device slice
```
amorde, line 87: while this may be true, aren't there certain devices that require these architectures? I don't think we want to ignore the checking completely but I might be misunderstanding what this is actually doing

amorde: I think `armv7s` might be safe to drop, unsure about `arm64e` which is new for pointer authentication

seanreinhardtapps: For one perspective on it, the App Store *at the moment* strips arm64e from submissions: https://developer.apple.com/documentation/security/preparing_your_app_to_work_with_pointer_authentication?language=objc
and `$ARCHS_STANDARD` is currently `arm64 armv7`.  
On the other hand, I realized after creating this that the ARCHS of the Pod Target is whats really being considered here.  That might make this entire concern semi-irrelevant (I need to look into the logic around ARCHS when the Pod target is created).

amorde: CocoaPods does attempt to infer architectures from the user's project but its not perfect, and I don't think we have a way to determine the value of `ARCHS_STANDARD` when generating the pods project(s)

amorde:
@seanreinhardtapps would you like an invite to the CocoaPods Slack?

seanreinhardtapps:
sure, sounds good.

lib/cocoapods/generator/copy_xcframework_script.rb
seanreinhardtapps: Quick correction to a previous statement, `$ARCHS_STANDARD` is currently `arm64 armv7` for deployment targets < iOS 11.0.
With a deployment target of 11.0, `$ARCHS_STANDARD` it becomes just `arm64`.
So, `$ARCHS`, `$VALID_ARCHS`, and potentially deployment target (if $ARCHS_STANDARD is used) are all relevant to what the final result of `$ARCHS` will be when `copy_xcframework_script` sees it.

This is a factor in #9794, where the XCFramework could be installed manually (the App's deployment target is 13.4), but could not be installed by the Pod Target (deployment target of 10.0).

amorde: ah sounds like https://github.com/CocoaPods/CocoaPods/issues/7314 we don't bump the deployment target of pods that support older versions in case that introduces warnings at build time.

it's still sounding to me like this xcframework should be updated to include the additional architectures OR the app should refine its architectures

seanreinhardtapps: I have definitely concluded that this change to the xcframework script is incorrect.
For all intents and purposes, the response to an issue like  #9794 should be "Add the missing armv7 architecture, or set your Podspec's deployment target to 11.0 if you want to ship 64 bit only".

Before I put this to rest, my last question is should the `$VALID_ARCHS` of the Pod Target be set to reflect the contents of the XCFramework (from it's info.plist)?  Right now the App and the Pod target agree on `ARCHS` and `VALID_ARCHS`, but ultimately resolve to something different due to the difference in deployment target and its impact on `$ARCHS_STANDARD`.

Appreciate all the feedback you've provided on this!

dnkoutso:
Is this PR no longer needed?

amorde:
Thanks again for investigating here @seanreinhardtapps!

PR #9871, opened 2020-06-18
Warn users to delete the master specs repo if its not explicitly used.

dnkoutso:


spec/unit/installer_spec.rb
```diff
@@ -351,6 +351,59 @@ def @installer.clean_pod_sources
...
+          install! 'cocoapods', :integrate_targets => false
+          platform :ios
+        end
+        @installer = Installer.new(config.sandbox, podfile)
+        master_source = Source.new(Pod::Installer::MASTER_SPECS_REPO_GIT_URL)
+        master_source.stubs(:url).returns(Pod::Installer::MASTER_SPECS_REPO_GIT_URL)
+        config.sources_manager.stubs(:all).returns([master_source])
+        @installer.send(:warn_for_removing_git_master_specs_repo)
+        UI.warnings.should.include 'Your project does not explicitly specify the CocoaPods master specs repo. Since CDN is now used as the' \
+          ' default, you may safely remove it from your repos directory via `pod repo remove master`.' \
```
dnkoutso, line 365: @igor-makarov is `pod repo remove master` correct to delete the old master specs repo?  We do not do any migration correct? 

Basically we keep `master` and add `trunk` for users who upgrade from an older version of CP to one that uses CDN.

lib/cocoapods/installer/installation_options.rb
```diff
@@ -153,6 +153,11 @@ def hash
       #
       option :warn_for_multiple_pod_sources, true
 
+      # Whether to emit a warning if a project is not explicitly specifying the git based master specs repo and can
+      # instead use CDN which is the default.
+      #
+      option :warn_for_unused_master_specs_repo, true
```
dnkoutso, line 159: I added this option because its possible that users might have multiple projects. One that uses master specs repo and one that doesn't. This will allow them not to have warnings and not ask them to delete the master specs repo.

lib/cocoapods/installer.rb
```diff
@@ -49,6 +49,8 @@ class Installer
 
     include Config::Mixin
 
+    MASTER_SPECS_REPO_GIT_URL = 'https://github.com/CocoaPods/Specs.git'.freeze
```
amorde, line 52: do we not have this defined anywhere else? I would think we still do but I guess it's possible we don't since we switched to CDN by default

spec/unit/installer_spec.rb
amorde: looks right to me

lib/cocoapods/installer.rb
dnkoutso: we dont! neither in Core!

dnkoutso:
@paulb777 yes! I updated `validator.rb` to reflect that and ensure CI systems or other commands that use this dont output the warning.

PR #9877, opened 2020-06-21
[Command] Allow gem to run as root

### LINKED ISSUE #8929: please add the option to run as root
I only know of cocoapods because some cordova plugin decided to integrate your library. so what happens is I build the whole Ionic project, the said cordova plugin compiles and then it all fails because you guys decided running sudo is unsafe.

ok, well the problem is npm, ionic and cordova, all three involved in the build process, disagree because all of them require privileges in order to compile correctly so after 3 crashes or so i decided to chown/chmod my whole computer which defeats the purpose of the supposed safety provided by supressing root user.

Please do understand that many of us know what they're doing or don't even care about the safety of their computer.. (I am running my iOS builds on a VM, so yeah I really don't care) so please give us the option to get on with our day and don't be the bottleneck of the big picture.

Thanks.

seanreinhardtapps:
Closes #8929

Adds argument flag `--allow-root`  and environment variable `COCOAPODS_ALLOW_ROOT`

### DIFF
--- a/spec/unit/command_spec.rb
+++ b/spec/unit/command_spec.rb
@@ -47,6 +47,47 @@ module Pod
       end
     end
 
+    describe 'ensure_not_root_or_allowed!' do
+      it 'passes check with ENV["COCOAPODS_ALLOW_ROOT"]' do
+        stored_env = ENV['COCOAPODS_ALLOW_ROOT']
+        ENV['COCOAPODS_ALLOW_ROOT'] = '1'
+        lambda { Command.ensure_not_root_or_allowed!([], 123, false) }.should.not.raise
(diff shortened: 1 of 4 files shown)


spec/unit/command_spec.rb
```diff
@@ -46,7 +46,46 @@ module Pod
         end
       end
     end
+    describe 'ensure_not_root_or_allowed!' do
```
dnkoutso, line 49: nit extra line between `end` above and `describe`

dnkoutso:
Looks good to me. I do not know of the reason CocoaPods had intentionally blocked this.

dnkoutso:
@seanreinhardtapps needs a rebase and we can land.

PR #9918, opened 2020-07-17
Bump minimum version of cocoapods-downloader

dnkoutso:
In preparation of 1.10 release.

Integration specs: https://github.com/CocoaPods/cocoapods-integration-specs/pull/288

### DIFF
--- a/spec/integration.rb
+++ b/spec/integration.rb
@@ -51,6 +51,7 @@ require 'colored2'
 
 require 'cocoapods-core/lockfile'
 require 'cocoapods-core/yaml_helper'
+require 'cocoapods-downloader'
 require 'fileutils'
 require 'integration/file_tree'
 require 'integration/xcodeproj_project_yaml'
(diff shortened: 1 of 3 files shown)


spec/integration.rb
```diff
@@ -146,6 +146,7 @@
     s.replace_pattern /#{Dir.tmpdir}\/[\w-]+/io, 'TMPDIR'
     s.replace_pattern /\d{4}-\d\d-\d\d \d\d:\d\d:\d\d [-+]\d{4}/, '<#DATE#>'
     s.replace_pattern /\(Took \d+.\d+ seconds\)/, '(Took <#DURATION#> seconds)'
+    s.replace_pattern /\b#{Regexp.escape(Pod::VERSION)}\b/, '<#POD::VERSION#>'
```
segiddins, line 149: You'll also need to replace the cocoapods-downloader version, it's also in the output

dnkoutso: ah sure, fixing.

PR #9926, opened 2020-07-20
Match Core gem activesupport version requirements.

dnkoutso:


segiddins:
If core already has an activesupport requirement, is there any reaosn for us to duplicate it in cocoapods?

dnkoutso:
Good question, I could also remove it.

amorde:
Seems like if we were to only pick one, it would be CocoaPods and not Core. but I guess it doesn't really matter

amorde:
This also came up on my PR for bumping minimum Ruby version https://github.com/CocoaPods/CocoaPods/pull/9851#discussion_r439585296

Issue #9943, opened 2020-07-26
Latest XCFramework rework on Cocoapods 1.10 is embedding static frameworks


BernardGatt:
# Report

## What did you do?

The [latest rework](https://github.com/CocoaPods/CocoaPods/pull/9720) fixes the issue where X86_64 were not being picked up from XCFrameworks.

But when archiving for release, static frameworks are being embedded with the IPA which causes a Mach-O error.

Looking at the `Pods-Target-frameworks.sh` in the `Debug` and `Release` section the static framework is being included for bundling.

```sh
install_framework "${PODS_XCFRAMEWORKS_BUILD_DIR}/StaticFramework/StaticFramework.framework"
```

### Note
In 1.09 there was an embed flag which is now removed.

## What did you expect to happen?

Continue with signing and publishing of IPA.

## What happened instead?

![image](https://user-images.githubusercontent.com/346012/88482597-9e6e6100-cf62-11ea-9004-6ca26fc80252.png)

## CocoaPods Environment

1.10.0.beta.1

dnkoutso:
cc @amorde

PR #9945, opened 2020-07-27
Added static binary check before embedding framework

BernardGatt:
Static binaries are currently being embedded into the app, this causes a Mach-O error.

[Apple Documentation](https://developer.apple.com/library/archive/technotes/tn2435/_index.html#//apple_ref/doc/uid/DTS40017543-CH1-TROUBLESHOOTING_BUNDLE_ERRORS-EMBEDDED_STATIC_LIBRARIES)

Addresses issue: https://github.com/CocoaPods/CocoaPods/issues/9943

### DIFF
--- a/lib/cocoapods/generator/embed_frameworks_script.rb
+++ b/lib/cocoapods/generator/embed_frameworks_script.rb
@@ -158,7 +158,7 @@ code_sign_if_enabled() {
           end
         end
         xcframeworks_by_config.each do |config, xcframeworks|
-          xcframeworks.each do |xcframework|
+          xcframeworks.select { |xcf| xcf.build_type.dynamic_framework? }.each do |xcframework|
             name = xcframework.name
             contents_by_config[config] << %(  install_framework "#{Target::BuildSettings::XCFRAMEWORKS_BUILD_DIR_VARIABLE}/#{name}/#{name}.framework"\n)
           end
--- a/spec/integration.rb
+++ b/spec/integration.rb
@@ -318,6 +318,11 @@ describe_cli 'pod' do
                             'install --no-repo-update --no-verbose'
     end
 
+    describe 'Integrates a Pod using a vendored static xcframework' do
+      behaves_like cli_spec 'install_vendored_static_xcframework',
+                            'install --no-repo-update'
+    end
+
     describe 'Integrates a Pod using a vendored xcframework' do
       behaves_like cli_spec 'install_vendored_xcframework',
                             'install --no-repo-update'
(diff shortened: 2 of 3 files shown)


BernardGatt:
Is there anything I can do to get this merged into beta? The current version is broken without this fix.

dnkoutso:
cc @amorde because I am not very familiar with XCFrameworks. Seems correct fix given that a static framework does not need to be embedded and is linked directly to binary executable?

amorde:
Yep saw the issue and forgot to look at this. Will review this as soon as I can

lib/cocoapods/generator/embed_frameworks_script.rb
```diff
@@ -71,6 +71,14 @@ def script
 # Copies and strips a vendored framework
 install_framework()
 {
+  local basename
+  basename="$(basename -s .framework "$1")"
+  source_binary="$1/${basename}"
+  if [[ $(file "$source_binary" | grep 'current ar archive') ]]; then
```
amorde, line 77: We don't need to wait until this script runs to do this check - we know at the time this script is created during `pod install` and can skip it entirely. We had logic for this but it must have gotten lost during some refactoring.

Here's the relevant code that could be updated to fix this:

https://github.com/CocoaPods/CocoaPods/blob/11b34726a83067ecc66b364e2c5d620fd303513e/lib/cocoapods/generator/embed_frameworks_script.rb#L160-L165

changing this portion: 
```ruby
xcframeworks.each do |xcframeworks|
```

to be this:

```ruby
xcframeworks.select { |xcf| xcf.build_type.dynamic_framework? }.each do |xcframework|
```

should do the trick. That will remove the call to `install_framework` entirely

dnkoutso: awesome, if we do that @BernardGatt can you add a test also?

BernardGatt: Agreed, makes more sense.

@dnkoutso working on a test.

spec/integration.rb
```diff
@@ -318,6 +318,11 @@
                             'install --no-repo-update --no-verbose'
     end
 
+    describe 'Integrates a Pod using a vendored static xcframework' do
```
dnkoutso, line 321: looks like a test cannot be found for this. You can add a unit test also which should be sufficient. Look into `embed_fraameworks_script_spec.rb` file.

dnkoutso: adding such an example though should be good as well.

BernardGatt: There's a PR open on the integration tests repo here: https://github.com/CocoaPods/cocoapods-integration-specs/pull/289

dnkoutso: ah my bad!

amorde:
Thank you @BernardGatt!

PR #9977, opened 2020-08-13
Fix broken install link in CHANGELOG

mikehardy:
Previous link went to malfunctioning heroku app behind docs.cocoapods.org, this link harvested via cocoapods.org then clicking through to the using section


mikehardy:
Just poking around to see how to test 1.10 and noticed a busted link, no issues with me if it's fixed in any other way and this is closed of course

Thanks for cocoapods!

dnkoutso:
nice thanks!

PR #9986, opened 2020-08-17
Incremental installation works for a test host in a separate (not-regenerated) project

segiddins:
Unfortunately, much like https://github.com/CocoaPods/CocoaPods/pull/9847 there's no existing test coverage for this area of the code

### DIFF
--- a/lib/cocoapods/installer/project_cache/project_metadata_cache.rb
+++ b/lib/cocoapods/installer/project_cache/project_metadata_cache.rb
@@ -54,6 +54,10 @@ module Pod
           installation_results.each do |installation_result|
             native_target = installation_result.native_target
             target_label_by_metadata[native_target.name] = TargetMetadata.from_native_target(sandbox, native_target)
+            # app targets need to be added to the cache because they can be used as app hosts for test targets, even if those test targets live inside a different pod (and thus project)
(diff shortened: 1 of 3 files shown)


lib/cocoapods/installer/project_cache/project_metadata_cache.rb
```diff
@@ -54,6 +54,10 @@ def update_metadata!(pod_target_installation_results, aggregate_target_installat
           installation_results.each do |installation_result|
             native_target = installation_result.native_target
             target_label_by_metadata[native_target.name] = TargetMetadata.from_native_target(sandbox, native_target)
+            # app targets need to be added to the cache because they can be used as app hosts for test targets, even if those test targets live inside a different pod (and thus project)
+            installation_result.app_native_targets.each_value do |app_target|
```
dnkoutso, line 58: do we want to store all native targets? Like test ones or resource ones? Might be out of scope of this PR but was wondering if its a good opportunity here.

segiddins: In theory, we don't allow cross-project dependencies on test or resource targets, and this helps keep the size of the cache file down

Issue #9524, opened 2020-02-06
pod lib lint - ERROR | [watchOS] unknown: Encountered an unknown error (CocoaPods could not find compatible versions for pod "FirebaseCore"


Bino90:
# Report

## What did you do?

pod lib lint MYPOD.podspec --allow-warnings

## What did you expect to happen?

no errors

## What happened instead?

- ERROR | [watchOS] unknown: Encountered an unknown error (CocoaPods could not find compatible versions for pod "FirebaseCore":

In my .podspec i set that firebase should only be available on all other platforms .....

Can anyone help?

## CocoaPods Environment

```
   CocoaPods : 1.8.4
        Ruby : ruby 2.6.3p62 (2019-04-16 revision 67580) [universal.x86_64-darwin19]
    RubyGems : 3.0.3
        Host : Mac OS X 10.15.3 (19D76)
       Xcode : 11.3.1 (11C504)
         Git : git version 2.21.1 (Apple Git-122.3)
Ruby lib dir : /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib
Repositories : master - git - https://github.com/CocoaPods/Specs.git @ a9f329b380ba8b6cc66a7d9542d0d82d0c913b40
               trunk - CDN - https://cdn.cocoapods.org/
```

### Installation Source

```
Executable Path: /usr/local/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.4
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.1.0
cocoapods-trunk       : 1.4.1
cocoapods-try         : 1.1.0
```

paulb777:
The error message points in the right direction. FirebaseCore requires at least watchOS 6.0 and the podspec is at 3.0.  See https://github.com/CocoaPods/Specs/blob/master/Specs/8/b/d/FirebaseCore/6.6.1/FirebaseCore.podspec.json#L21

Bino90:
yeah right but the firebase dependency is only for iOS, tvOS and OSX(macOS) not for watchOS
    sp.ios.dependency 'Firebase/Crashlytics'
    sp.tvos.dependency 'Firebase/Crashlytics'
    sp.osx.dependency 'Firebase/Crashlytics'

dnkoutso:
@paulb777 any further updates here?

paulb777:
[...]

PR #9999, opened 2020-08-20
Honor test spec deployment target during validation.

dnkoutso:
If a test spec has a different deployment target than the library spec during validation this was not honored by the validator.

### DIFF
--- a/lib/cocoapods/validator.rb
+++ b/lib/cocoapods/validator.rb
@@ -556,6 +556,8 @@ module Pod
       validation_dir.rmtree
     end
 
+    # @return [String] The deployment targret of the library spec.
+    #
     def deployment_target
       deployment_target = spec.subspec_by_name(subspec_name).deployment_target(consumer.platform_name)
       if consumer.platform_name == :ios && use_frameworks
@@ -711,7 +713,7 @@ module Pod
             UI.warn "Skipping compilation with `xcodebuild` because target contains no sources.\n".yellow
           else
             if analyze
-              output = xcodebuild('analyze', scheme, 'Release')
+              output = xcodebuild('analyze', scheme, 'Release', :deployment_target => deployment_target)
               find_output = Executable.execute_command('find', [validation_dir, '-name', '*.html'], false)
               if find_output != ''
                 message = 'Static Analysis failed.'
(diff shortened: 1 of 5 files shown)


lib/cocoapods/validator.rb
```diff
@@ -757,7 +759,7 @@ def test_pod
               UI.warn "Skipping test spec `#{test_spec.name}` on platform `#{consumer.platform_name}` since it is not supported.\n".yellow
             else
               scheme = @installer.target_installation_results.first[pod_target.name].native_target_for_spec(test_spec)
-              output = xcodebuild('test', scheme, 'Debug')
+              output = xcodebuild('test', scheme, 'Debug', test_spec.deployment_target(consumer.platform_name))
```
dnkoutso, line 762: this method will also use the parent spec if none is explicitly specified.

lib/cocoapods/validator.rb
```diff
@@ -1048,7 +1050,7 @@ def parse_xcodebuild_output(output)
     # @return [String] Executes xcodebuild in the current working directory and
     #         returns its output (both STDOUT and STDERR).
     #
-    def xcodebuild(action, scheme, configuration)
+    def xcodebuild(action, scheme, configuration, deployment_target)
```
amorde, line 1053: lol love how this is the only change to the method. variable shadowing ftw?

paulb777:
Does this fix #9524?

lib/cocoapods/validator.rb
segiddins: nit: use a named parameter ?

spec/unit/validator_spec.rb
```diff
@@ -600,6 +600,39 @@ def podspec_path(name = 'JSONKit', version = '1.4')
         validator.results_message.strip.should.be.empty
       end
 
+      it 'uses the deployment target of the current test spec' do
```
segiddins, line 603: it's not super clear to me what in here is validating that the correct deployment target is being used for xcodebuild ?

lib/cocoapods/validator.rb
dnkoutso: love it

spec/unit/validator_spec.rb
amorde: it pulls a destination from `Fourflusher` once requesting iOS 8 for building the library and once requesting iOS 9 to run the test. at least that was my understanding. maybe would be better to add an expectation that `xcodebuild` is provided the correct deployment targets instead

dnkoutso: yeap. Those tests are super convoluted and hard to follow. I verified without my change this test fails.

PR #10016, opened 2020-08-28
Force a clean install if installation options are changed.

sebastianv1:
Incremental installation does not track any installation options in the cache. Some installation options do affect project generation, so it's safer to force a clean install if any of the podfile options have changed.

### DIFF
--- a/lib/cocoapods/installer/project_cache/project_cache_analyzer.rb
+++ b/lib/cocoapods/installer/project_cache/project_cache_analyzer.rb
@@ -34,6 +34,10 @@ module Pod
         #
         attr_reader :aggregate_targets
 
+        # @return [Hash<Symbol, Object>] Hash of installation options.
+        #
+        attr_reader :installation_options
+
         # @return [Bool] Flag indicating if we want to ignore the cache and force a clean installation.
         #
         attr_reader :clean_install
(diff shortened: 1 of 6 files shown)


lib/cocoapods/installer/project_cache/project_cache_analyzer.rb
```diff
@@ -78,7 +84,8 @@ def analyze
           # Bail out early since these properties affect all targets and their associate projects.
           if cache.build_configurations != build_configurations ||
               cache.project_object_version != project_object_version ||
-              YAMLHelper.convert(cache.podfile_plugins) != YAMLHelper.convert(podfile_plugins)
+              YAMLHelper.convert(cache.podfile_plugins) != YAMLHelper.convert(podfile_plugins) ||
+              YAMLHelper.convert(cache.installation_options) != YAMLHelper.convert(installation_options)
```
sebastianv1, line 88: I forget if this is necessary. I thought ruby hash equality didn't care about key orderings, but it might have with the podfile plugins?

segiddins: maybe it had to do with hash with indifferent access vs normal hash for plugins, in which case i think the concern might still apply? not 100% sure

PR #6535, opened 2017-03-06
Fix miss cache file problem on Jenkins concurrent install.

yanzhiwei147:
Fix [miss cache file problem on Jenkins concurrent install](https://github.com/CocoaPods/CocoaPods/issues/6280)

PR #10018, opened 2020-08-31
[Cache] Ensure cache integrity on concurrent installations.

### LINKED ISSUE #10013: Cache gets corrupted when running `pod install` concurrently
* [x] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md). 🌈

# Report

## What did you do?

Ran `pod install` concurrently from different jobs on the build server.

(We don't know for sure that concurrency is the cause, but it's intermittency and the fact that multiple processes are writing to the same location at the same time, suggests that it probably is.)

That the cache and pod installations would be fine.

## What happened instead?

Sometimes the cache gets corrupted and caches for some pods were emptied and pod install now installs those pods as empty directories.

## CocoaPods Environment

(Not allowed to share the whole thing as it reveals details of our project. Below is what looks relevant and safe to share.)

### Stack

```
   CocoaPods : 1.8.3
        Ruby : ruby 2.7.1p83 (2020-03-31 revision a0c7c23c9c) [x86_64-darwin19]
    RubyGems : 3.1.2
        Host : Mac OS X 10.15.6 (19G73)
       Xcode : 11.1 (11A1027)
         Git : git version 2.18.0
Ruby lib dir : /opt/jenkins_tools/ruby/2.7.1/lib
```

### Installation Source

```
Executable Path: /Users/jenkins/.bundle/ruby/2.7.0/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.4
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.1.0
cocoapods-trunk       : 1.4.1
cocoapods-try         : 1.1.0
```

## Additional Background

This is a long standing issue in our environment. We run multiple jobs in parallel per build server to better utilise our capacity. Most of the time, everything is fine and works as expected, but every now and then the jobs will fail because of missing files in the `Pods/` directory. Once a build server has a corrupted cache, the builds will fail until we delete the cache, so that the pods are re-downloaded and copied correctly to the `Pods/` directory.

We don't know for sure, but our theory is that this is because we are running multiple instances of `pod install` on the same machine using the same cache and that under certain circumstances that corrupts the cache. Until now, since it is relatively infrequent, we've just been manually clearing the caches when they got corrupted, but we recently saw an increase in frequency of cache corruptions and decided to try to fix it.

Work is already under way to refactor our build scripts and pipelines to eliminate this concurrency and we're hoping that will fix the cache corruptions. Because there are multiple apps and multiple teams, someone could inadvertently add a `pod install` in their build script that risks running concurrently and break the cache anyway. So we can really only decrease the probability of future cache corruptions.

The best solution would be if CocoaPods had built in support for running `pod install` concurrently. It should take care of locking the cache, so that multiple instances don't try to write to the cache at the same time. I suppose we don't know for sure that concurrency is the issue, but it seems like the most likely cause. If there's some other error condition that corrupts the cache, we'd obviously like that to be fixed.

We'd be very interested in hearing about whether others have experienced the same problem with caches getting corrupted.

codiophile:
Still doesn't handle concurrent installations with different versions of CocoaPods, but that's a significantly bigger change.

Fixes #10013

### DIFF
--- a/lib/cocoapods/downloader/cache.rb
+++ b/lib/cocoapods/downloader/cache.rb
@@ -100,8 +100,10 @@ module Pod
       end
 
       # Creates a .lock file at `location`, aquires a lock of type
-      # `lock_type`, and executes `block` while holding on to that lock. The
-      # block is passed the location for convenience.
+      # `lock_type`, checks that it is valid, and executes `block` while holding
+      # on to that lock. Afterwards, the .lock file is deleted, which is why
+      # validation of the lock is necessary, as you might have a lock on a file
+      # that doesn't exist on the filesystem anymore.
       #
       # @param  [Pathname] location
       #         the path to require a lock for.
@@ -115,12 +117,40 @@ module Pod
       #
       # @return [Void]
       #
-      def self.lock(location, lock_type, block = Proc.new)
-        lockfile = "#{location}.lock"
-        File.open(lockfile, File::CREAT, 0o644) do |f|
+      def self.lock(destination, lock_type, block = Proc.new)
+        lockfile = "#{destination}.lock"
+        f = nil
+        loop do
+          f.close if f
+          f = File.open(lockfile, File::CREAT, 0o644)
           f.flock(lock_type)
-          block.call location
+          break if Cache.validate_lock(f, lockfile)
         end
+        block.call destination
+        if lock_type == File::LOCK_SH
+          f.flock(File::LOCK_EX)
+          File.delete(lockfile) if Cache.validate_lock(f, lockfile)
+        else
+          File.delete(lockfile)
+        end
+        f.close
+      end
+
+      # Checks that the lock is on a file that still exists on the filesystem.
+      #
+      # @param  [File] file
+      #         the actual file that we have a lock for.
+      #
+      # @param  [String] filename
+      #         the filename of the file that we have a lock for.
+      #
+      # @return [Boolean]
+      #         true if `filename` still exists and is the same file as `f`
+      #
+      def self.validate_lock(file, filename)
+        file.stat.ino == File.stat(filename).ino
+      rescue Errno::ENOENT
+        false
       end
 
       private
--- a/lib/cocoapods/downloader/cache.rb
+++ b/lib/cocoapods/downloader/cache.rb
@@ -75,13 +75,13 @@ module Pod
       # @param  [Pathname] location
       #         the path to require a lock for.
       #
-      # @param  [Proc] block
+      # @param  [block] &block
       #         the block to execute inside the lock.
       #
       # @return [void]
       #
-      def self.read_lock(location, block = Proc.new)
-        Cache.lock(location, File::LOCK_SH, block)
+      def self.read_lock(location, &block)
+        Cache.lock(location, File::LOCK_SH, &block)
       end
 
       # Convenience method for acquiring an exclusive lock to safely write to
@@ -90,20 +90,20 @@ module Pod
       # @param  [Pathname] location
       #         the path to require a lock for.
       #
-      # @param  [Proc] block
+      # @param  [block] &block
       #         the block to execute inside the lock.
       #
       # @return [void]
       #
-      def self.write_lock(location, block = Proc.new)
-        Cache.lock(location, File::LOCK_EX, block)
+      def self.write_lock(location, &block)
+        Cache.lock(location, File::LOCK_EX, &block)
       end
 
       # Creates a .lock file at `location`, aquires a lock of type
-      # `lock_type`, checks that it is valid, and executes `block` while holding
-      # on to that lock. Afterwards, the .lock file is deleted, which is why
-      # validation of the lock is necessary, as you might have a lock on a file
-      # that doesn't exist on the filesystem anymore.
+      # `lock_type`, checks that it is valid, and executes passed block while
+      # holding on to that lock. Afterwards, the .lock file is deleted, which is
+      # why validation of the lock is necessary, as you might have a lock on a
+      # file that doesn't exist on the filesystem anymore.
       #
       # @param  [Pathname] location
       #         the path to require a lock for.
@@ -112,26 +112,24 @@ module Pod
       #         the type of lock, either exclusive (File::LOCK_EX) or shared
       #         (File::LOCK_SH).
       #
-      # @param  [Proc] block
-      #         the block to execute inside the lock.
-      #
       # @return [void]
       #
-      def self.lock(destination, lock_type, block = Proc.new)
-        lockfile = "#{destination}.lock"
+      def self.lock(location, lock_type)
+        raise ArgumentError, 'no block given' unless block_given?
+        lockfile = "#{location}.lock"
         f = nil
         loop do
           f.close if f
           f = File.open(lockfile, File::CREAT, 0o644)
           f.flock(lock_type)
-          break if Cache.validate_lock(f, lockfile)
+          break if Cache.valid_lock?(f, lockfile)
         end
         begin
-          block.call destination
+          yield location
         ensure
           if lock_type == File::LOCK_SH
             f.flock(File::LOCK_EX)
-            File.delete(lockfile) if Cache.validate_lock(f, lockfile)
+            File.delete(lockfile) if Cache.valid_lock?(f, lockfile)
           else
             File.delete(lockfile)
           end
@@ -148,9 +146,9 @@ module Pod
       #         the filename of the file that we have a lock for.
       #
       # @return [Boolean]
-      #         true if `filename` still exists and is the same file as `f`
+      #         true if `filename` still exists and is the same file as `file`
       #
-      def self.validate_lock(file, filename)
+      def self.valid_lock?(file, filename)
         file.stat.ino == File.stat(filename).ino
       rescue Errno::ENOENT
         false
--- a/lib/cocoapods/downloader/cache.rb
+++ b/lib/cocoapods/downloader/cache.rb
@@ -78,7 +78,7 @@ module Pod
       # @param  [Proc] block
       #         the block to execute inside the lock.
       #
-      # @return [Void]
+      # @return [void]
       #
       def self.read_lock(location, block = Proc.new)
         Cache.lock(location, File::LOCK_SH, block)
@@ -93,7 +93,7 @@ module Pod
       # @param  [Proc] block
       #         the block to execute inside the lock.
       #
-      # @return [Void]
+      # @return [void]
       #
       def self.write_lock(location, block = Proc.new)
         Cache.lock(location, File::LOCK_EX, block)
@@ -115,7 +115,7 @@ module Pod
       # @param  [Proc] block
       #         the block to execute inside the lock.
       #
-      # @return [Void]
+      # @return [void]
       #
       def self.lock(destination, lock_type, block = Proc.new)
         lockfile = "#{destination}.lock"
@@ -126,14 +126,17 @@ module Pod
           f.flock(lock_type)
           break if Cache.validate_lock(f, lockfile)
         end
-        block.call destination
-        if lock_type == File::LOCK_SH
-          f.flock(File::LOCK_EX)
-          File.delete(lockfile) if Cache.validate_lock(f, lockfile)
-        else
-          File.delete(lockfile)
+        begin
+          block.call destination
+        ensure
+          if lock_type == File::LOCK_SH
+            f.flock(File::LOCK_EX)
+            File.delete(lockfile) if Cache.validate_lock(f, lockfile)
+          else
+            File.delete(lockfile)
+          end
+          f.close
         end
-        f.close
       end
 
       # Checks that the lock is on a file that still exists on the filesystem.
--- a/lib/cocoapods/downloader/cache.rb
+++ b/lib/cocoapods/downloader/cache.rb
@@ -69,6 +69,60 @@ module Pod
         end
       end
 
+      # Convenience method for acquiring a shared lock to safely read from the
+      # cache. See `Cache.lock` for more details.
+      #
+      # @param  [Pathname] location
+      #         the path to require a lock for.
+      #
+      # @param  [Proc] block
+      #         the block to execute inside the lock.
+      #
+      # @return [Void]
+      #
+      def self.read_lock(location, block = Proc.new)
+        Cache.lock(location, File::LOCK_SH, block)
+      end
+
+      # Convenience method for acquiring an exclusive lock to safely write to
+      # the cache. See `Cache.lock` for more details.
+      #
+      # @param  [Pathname] location
+      #         the path to require a lock for.
+      #
+      # @param  [Proc] block
+      #         the block to execute inside the lock.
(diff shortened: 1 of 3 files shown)


dnkoutso:
@codiophile the change looks good, do you think this will impact performance, any way to test perfomance?

dnkoutso:
I guess I dont suspect it will be that much different. This is to help with concurrent `pod install` running...I assume most cases run a single`pod install` command...

lib/cocoapods/downloader/cache.rb
```diff
@@ -69,6 +69,60 @@ def cache_descriptors_per_pod
...
+      #         the type of lock, either exclusive (File::LOCK_EX) or shared
+      #         (File::LOCK_SH).
+      #
+      # @param  [Proc] block
+      #         the block to execute inside the lock.
+      #
+      # @return [Void]
+      #
+      def self.lock(location, lock_type, block = Proc.new)
+        lockfile = "#{location}.lock"
```
dnkoutso, line 119: is this file removed afterwords?

codiophile: It isn't, and it wasn't in #6535 either. I was experimenting with that, but I'm not sure how to ensure that it's safe to delete the file and that no other process is waiting to lock on that file. I didn't get it to work reliably, and I thought it probably doesn't matter much anyway. It's empty files that don't take up almost any space in a directory that isn't supposed to be read by humans, so it's hard to imagine that it would really bother anyone.

If you know how to safely delete a lock file without messing with other process' locking attempts, I would be happy to add that. :)

dnkoutso: Can we possibly use the `tmp` folder that the OS can then handle by itself?

dnkoutso: Maybe use `Tempfile.new`?

codiophile:
> @codiophile the change looks good, do you think this will impact performance, any way to test performance?

Considering that this code is executed by several tests and that there is no noticeable impact on test execution time, I think it's safe to say that any performance penalty is negligible. Judging from the numbers in Travis, it's actually slightly faster. ;)

lib/cocoapods/downloader/cache.rb
codiophile: `Tempfile.new` wouldn't work since it creates a new file with a random name each time, so each process would get its own file and would hence not use the same lock.

codiophile: Since you seem quite keen, I did some more research. There is a potential solution here:
https://www.ruby-forum.com/t/how-do-you-use-flock-and-clean-up-lock-files/64817/7

It seems a bit convoluted, but the logic looks sound. After acquiring the lock, you check if the file still exists on the filesystem and if it's the same file that you currently have a lock on, otherwise you abort and try again. Once you have a lock on a file that exists on the filesystem, you do your business and then you delete it before releasing the lock. Any processes that were waiting will detect that the file they got the lock for doesn't exist anymore and and try again as described above.

Anyway, it's getting quite late in my timezone, so I will probably not try it out until tomorrow.

codiophile: Actually, that logic doesn't consider shared locks, which I use for reading the cache. It gets quite complicated to accomodate those, since you need an exclusive lock to be sure that it's safe to delete the file. Unless you think it's absolutely necessary, I'd rather leave it as it is. Also, from a performance point of view, not deleting and recreating the lock file all the time should be advantageous.

lib/cocoapods/downloader/cache.rb
```diff
@@ -115,12 +117,40 @@ def self.write_lock(location, block = Proc.new)
...
+        lockfile = "#{destination}.lock"
+        f = nil
+        loop do
+          f.close if f
+          f = File.open(lockfile, File::CREAT, 0o644)
           f.flock(lock_type)
-          block.call location
+          break if Cache.validate_lock(f, lockfile)
         end
+        block.call destination
```
dnkoutso, line 129: should we wrap this in `begin /ensure`?

lib/cocoapods/downloader/cache.rb
codiophile: @dnkoutso I changed it so that the lock files are deleted after use, but now Inch doesn't like my documentation. No clue why. I added more documentation and somehow it's not good enough. Changing it back to what it was doesn't help, but deleting the documentation makes Inch happy. Seems like Inch is pretty useless. Any clue how I can keep the documentation and keep Inch happy?

lib/cocoapods/downloader/cache.rb
codiophile: Sounds sensible.

codiophile: Done, and strangely enough, Inch likes my documentation now. It seems a bit temperamental to me. 🤷‍♂️

codiophile:
By the way @dnkoutso, is there any reason why this is marked for 1.11.0 and not 1.10.0?

We are eagerly waiting for this to be released and solve our cache issues. :)

dnkoutso:
@codiophile not sure to be honest regarding the version. Something tells me that we wouldn't want to ship with 1.10.0 as we would want to give a bit more time to test this change and 1.10.0 final is about to be released.

Having said that, landing this change on master shouldn't block you from using it. I recommend using https://bundler.io/ to point to the master branch of CocoaPods that should include this change once it lands.

We do it internally and hasn't caused any trouble for us.

codiophile:
@dnkoutso Thank you for the approval! :)

> Having said that, landing this change on master shouldn't block you from using it.

Speaking of landing this change on master, it was approved, but not merged. When should I expect it on master?

> I recommend using https://bundler.io/ to point to the master branch of CocoaPods that should include this change once it lands.

Can I point bundler to my branch in the meantime?

lib/cocoapods/downloader/cache.rb
```diff
@@ -69,6 +69,93 @@ def cache_descriptors_per_pod
...
+      # Checks that the lock is on a file that still exists on the filesystem.
+      #
+      # @param  [File] file
+      #         the actual file that we have a lock for.
+      #
+      # @param  [String] filename
+      #         the filename of the file that we have a lock for.
+      #
+      # @return [Boolean]
+      #         true if `filename` still exists and is the same file as `f`
```
alloy, line 151: ?

```suggestion
      #         true if `filename` still exists and is the same file as `file`
```

lib/cocoapods/downloader/cache.rb
```diff
@@ -69,6 +69,93 @@ def cache_descriptors_per_pod
...
+      #
+      def self.validate_lock(file, filename)
```
alloy, line 153: Perhaps slightly more Rubyesque; nothing important, though.

```suggestion
      def self.has_lock?(file, filename)
```

lib/cocoapods/downloader/cache.rb
```diff
@@ -69,6 +69,93 @@ def cache_descriptors_per_pod
...
+      # @param  [locking_constant] lock_type
+      #         the type of lock, either exclusive (File::LOCK_EX) or shared
+      #         (File::LOCK_SH).
+      #
+      # @param  [Proc] block
+      #         the block to execute inside the lock.
+      #
+      # @return [void]
+      #
+      def self.lock(destination, lock_type, block = Proc.new)
```
alloy, line 120: Is there really a point in using these APIs if no block is given? How about adding something like:

```suggestion
      def self.lock(destination, lock_type, block)
        raise ArgumentError unless block_given?
```

(And then also update the convenience methods accordingly.)

lib/cocoapods/downloader/cache.rb
```diff
@@ -69,6 +69,93 @@ def cache_descriptors_per_pod
...
+      #
+      # @param  [Pathname] location
+      #         the path to require a lock for.
+      #
+      # @param  [Proc] block
+      #         the block to execute inside the lock.
+      #
+      # @return [void]
+      #
+      def self.read_lock(location, block = Proc.new)
```
alloy, line 83:
```suggestion
      def self.read_lock(location, &block)
```

lib/cocoapods/downloader/cache.rb
```diff
@@ -69,6 +69,93 @@ def cache_descriptors_per_pod
...
+        Cache.lock(location, File::LOCK_SH, block)
```
alloy, line 84:
```suggestion
        Cache.lock(location, File::LOCK_SH, &block)
```

lib/cocoapods/downloader/cache.rb
```diff
@@ -69,6 +69,93 @@ def cache_descriptors_per_pod
...
+      #
+      # @param  [Pathname] location
+      #         the path to require a lock for.
+      #
+      # @param  [Proc] block
+      #         the block to execute inside the lock.
+      #
+      # @return [void]
+      #
+      def self.write_lock(location, block = Proc.new)
```
alloy, line 98:
```suggestion
      def self.write_lock(location, &block)
```

lib/cocoapods/downloader/cache.rb
```diff
@@ -69,6 +69,93 @@ def cache_descriptors_per_pod
...
+        Cache.lock(location, File::LOCK_EX, block)
```
alloy, line 99:
```suggestion
        Cache.lock(location, File::LOCK_EX, &block)
```

lib/cocoapods/downloader/cache.rb
```diff
@@ -69,6 +69,93 @@ def cache_descriptors_per_pod
...
+        lockfile = "#{destination}.lock"
+        f = nil
+        loop do
+          f.close if f
+          f = File.open(lockfile, File::CREAT, 0o644)
+          f.flock(lock_type)
+          break if Cache.validate_lock(f, lockfile)
+        end
+        begin
+          block.call destination
```
alloy, line 130: Yielding is generally considered to be the more performant option (or at least it was when I was still doing Ruby day to day).

```suggestion
          yield destination
```

alloy (approved):
Very nice stuff! 👌

I approve of the PR as-is, although I think there are some slight improvements that could be made.

lib/cocoapods/downloader/cache.rb
codiophile: Oops 😄

lib/cocoapods/downloader/cache.rb
codiophile: Well, it has a lock, the question is if the lock is on a file that exists on the filesystem. I suppose `valid_lock?` might be a better name, though.

lib/cocoapods/downloader/cache.rb
codiophile: I was originally using yield, but then I wasn't able to pass the block between methods. Using `block = Proc.new` was the suggested solution that I found and it worked.

lib/cocoapods/downloader/cache.rb
codiophile: Does `Proc.new` still work, even if there is no block passed? I would have assumed that it would throw an exception, but maybe not. I suppose it's preferable to explicitly throw an `ArgumentError ` than relying on the behaviour of `Proc.new`. 🤷‍♂️

alloy: I see. Yeah I think for readers of the code, aside from runtime behaviour, it will be more clear to raise an `ArgumentError` yourself.

lib/cocoapods/downloader/cache.rb
alloy: When capturing the block (with the ampersand) it should work:

```
irb(main):002:0> def foo(&block)
irb(main):003:1>   bar(&block)
irb(main):004:1> end
=> :foo
irb(main):005:0> def bar(&block)
irb(main):006:1>   yield :ok
irb(main):007:1> end
=> :bar
irb(main):008:0> foo { |x| p x }
:ok
=> :ok
```

dnkoutso:
@codiophile 

> Can I point bundler to my branch in the meantime?

of course! You can even point to your own fork, any branch and any sha! https://bundler.io/v1.5/git.html

lib/cocoapods/downloader/cache.rb
codiophile: I suppose I should try that out. :)

I was reading: https://joeyates.info/2014/06/30/how-to-forward-blocks-in-ruby/

In attempt 2, he was using the ampersand syntax, but said that "we can no longer pass a block to the generalized method", because it resulted in the following error:

    wrong number of arguments (1 for 2)

But he didn't use ampersand in both methods, so maybe that was the problem.

Actually just tried this and your way works fine. :)

lib/cocoapods/downloader/cache.rb
codiophile: I just tried it and the current behaviour is that if you don't pass a block you get the following error:

    tried to create Proc object without a block (ArgumentError)

To me it's pretty clear that my mistake was that I didn't pass a block. :)

alloy: Absolutely, but you need to run that code to know that, it isn’t necessarily clear to anyone reading the code—like me 😄

dnkoutso:
From the suggested changes/open comments I cant tell if this is done or not. Is all feedback addressed @codiophile ?

codiophile:
@alloy I did implement all your changes locally, but Inch doesn't like the documentation anymore. If I use yield, `rubocop` will complain if I leave `&block` parameter unused, but if I remove the `&block` parameter, `inch` will complain about the documentation even if I remove any reference to `block`.

I think Inch might be more harmful than useful. It just gives you a grade and doesn't tell you what's wrong. I will read a bit about Inch and see if I can figure anything out.

The changes are anyway minor. I just wanted to include them to show my appreciation for your comments. I don't think it makes any real world difference, if they are included or not. :)

codiophile:
> From the suggested changes/open comments I cant tell if this is done or not. Is all feedback addressed @codiophile ?

@dnkoutso I have addressed everything locally, but I'm struggling with Inch again. No point in pushing something that I know will break the build. Any advice on how to deal with Inch?

dnkoutso:
You can run it locally with `inch` I think...I've had a few cases where I had to fix it. Just be pedantic and add every single parameter and a description to work around it.

I wouldnt let `inch` block your changes if they are good and part of review, fixing `inch` should be relatively straightforward.

codiophile:
@dnkoutso I found `inch show` :)

I should be able to sort it now.

alloy:
Yup, like I said, I approve of the PR as-is 👍

codiophile:
There was a legitimate problem with the documentation, so `inch` was genuinely useful. It's just the output is not that useful. It was a copy paste error. I used a different parameter name in my test project, where I was simulating concurrency, and inadvertently changed the name so that it didn't match the documentation anymore.

All sorted now. :)

amorde (approved):

dnkoutso:
awesome will merge on green!

PR #10033, opened 2020-09-04
[CONTRIBUTING.md] Fix incorrect documentation.

codiophile:
If you run `bundle exec rake spec` in the root of Rainforest repo, you get the following error:

    Could not locate Gemfile or .bundle/ directory

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -76,9 +76,8 @@ Code/comments should adhere to the following rules:
   work. (There are a few exceptions to this, which can be recognised by there
   not being any coverage for similar code.)
   * CocoaPods uses [bacon](https://github.com/chneukirchen/bacon) as a test runner.
-    To run all tests, use `bundle exec rake spec` in the root of the 
-    [Rainforest repo](https://github.com/CocoaPods/Rainforest). If you want to run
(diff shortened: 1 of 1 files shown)


codiophile:
I wouldn't mind making more improvements. This just stuck out, as it is completely incorrect.

`CONTRIBUTING.md` refers to https://guides.cocoapods.org/contributing/contribute-to-cocoapods.html for how to set up the environment, but that page is outdated and incomplete. Obviously, that page should be updated, but `CONTRIBUTING.md` could have a section for setting up the environment as well. What do you think?

We could (and should) update the Rainforest repo, so that it works again, but it mostly looks like overhead to me. I don't see the point in setting up the Rainforest repo for an individual contributor. It's more complicated and it doesn't add any value, so I think not pointing new contributors to the Rainforest repo is sensible.

We should probably also add a section about expected test failures locally. I have git and time machine related failures locally that I ignore. Ideally, we should fix the failures, but until then it would be helpful if it was mentioned in the docs. Also, some documentation about `rubocop` and `inch` would be helpful, as I could have run those locally, before pushing, if I only knew about them.

I went through the pain of following outdated documentation to set up my environment last weekend. In the end, the Rakefiles were the best documentation. While I have that process relatively fresh in memory, I should take the opportunity to update the docs to make it easier for the next guy. I'd just like your view on how it should be organised. Thanks.

PR #10035, opened 2020-09-05
[Cache] Don't delete cache for other versions of CocoaPods.

### LINKED ISSUE #10030: Allow different versions of CocoaPods to use the same cache (at the same time).
* [x] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md). 🌈

# Report

## What did you do?

Ran `pod install` with different versions of CocoaPods concurrently from different jobs on the build server.

That they would both be able to use the same cache, use different caches by default, or at least not delete the cache while another process is using it.

## What happened instead?

One process is deleting the cache while the other process is using it.

## CocoaPods Environment

This is a theoretical issue based on the code on the master branch.

## Additional Backround

After fixing #10013 we were discussing how to handle the fact that different projects use different versions of CocoaPods. One option is to set a different `$CP_HOME_DIR` for each project. That way, different versions of CocoaPods in different projects will not interfere with each other, but if you're building a hotfix for an old version of the same project, you could still have problems with different CocoaPods versions deleting the cache for each other.

I'm thinking that CocoaPods could handle this internally. First of all, the version of the cache should be disconnected from the version of CocoaPods. If the format of the cache is exactly the same between two versions of CocoaPods, there is no reason to delete the cache. Secondly, I think that CocoaPods could maintain multiple versions of its cache and only use the cache that corresponds to the currently running version. This could be accomplished by adding the cache version to the cache path.

One potential downside to this is that old versions of the cache would be left around, even if they aren't used anymore. Although I'm not sure the format of the cache is actually changing that often. I think deleting the cache generated by a different version of CocoaPods is just a precaution and not actually necessary. Although, I don't know the full history of the cache implementation, it looks like it's been pretty much the same for the past 5 years. It looks likely that the `ensure_matching_version` method hasn't been necessary.

Anyway, I'd like your opinion on this. It makes sense to me to have the cache versioned independently from the CocoaPods version. I suppose it might be tied to the version of the spec format, so it may not be as straight forward as I think. In an environment where we build multiple projects concurrently on the same machines, this would be very helpful. Please let me know if you see any issues with this, otherwise I would be happy to start working on it.

codiophile:
This is the simplest and safest way to accomplish this. Still waiting for feedback on how and whether we could version the cache independently from CocoaPods, so that all compatible versions of CocoaPods would use the same cache.

Fixes #10030

### DIFF
--- a/lib/cocoapods/downloader/cache.rb
+++ b/lib/cocoapods/downloader/cache.rb
@@ -7,7 +7,7 @@ module Pod
     # them in a cache directory.
     #
     class Cache
-      # @return [Pathname] The root directory where this cache store its
+      # @return [Pathname] The root directory where this cache stores its
       #         downloads.
       #
       attr_reader :root
@@ -18,8 +18,8 @@ module Pod
       #         see {#root}
       #
       def initialize(root)
-        @root = Pathname(root)
-        ensure_matching_version
+        @root = Pathname(root) + Pod::VERSION
+        @root.mkpath
       end
 
       # Downloads the Pod from the given `request`
@@ -155,21 +155,6 @@ module Pod
       end
 
       private
-
-      # Ensures the cache on disk was created with the same CocoaPods version as
-      # is currently running.
-      #
-      # @return [Void]
-      #
-      def ensure_matching_version
-        version_file = root + 'VERSION'
-        version = version_file.read.strip if version_file.file?
-
-        root.rmtree if version != Pod::VERSION && root.exist?
-        root.mkpath
-
-        version_file.open('w') { |f| f << Pod::VERSION }
-      end
 
       # @param  [Request] request
       #         the request to be downloaded.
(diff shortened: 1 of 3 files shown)


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* [Cache] Don't delete cache for other versions of CocoaPods.  
  [codiophile](https://github.com/codiophile)
```

</p>

codiophile:
By the way, I added

    gem 'ruby-debug-ide'
    gem 'debase'

to the `debugging` group in the `Gemfile` to be able to debug the tests in VS Code. I was wondering if I should commit that. I could also commit `.vscode/launch.json`, but I'd have to figure out how to configure it without absolute paths specific to my system. In a separate pull request, of course. :)

PR #10052, opened 2020-09-11
Allow app/test targets to not need empty .swift files when their deps use swift

segiddins:
* Allow building app & test spec targets that depend on a library that uses
  Swift without requiring an empty Swift file be present.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   produce warnings.  
   [Samuel Giddins](https://github.com/segiddins)
 
+* Allow building app & test spec targets that depend on a library that uses
+  Swift without requiring an empty Swift file be present.  
+  [Samuel Giddins](https://github.com/segiddins)
+
 ##### Bug Fixes
 
 * Ensure cache integrity on concurrent installations.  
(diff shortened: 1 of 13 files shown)


dnkoutso (approved):

amorde (approved):

justinseanmartin (approved):
The `ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES ` flag is now being passed explicitly for all targets and configurations (even ones where Xcode can figure it out), where before `-lswiftSwiftOnoneSupport` was only being sent to non-test configs. Seems fine, but thought it was worth calling out explicitly.

Overall LGTM and thanks for adding the integration test.

segiddins:
@justinseanmartin i think it's still only being added for app and test targets?

justinseanmartin:
I think it was only being passed to test bundles previously in Debug configuration. Seeing the flag in a bunch of targets in the xcconfigs that didn't previously have `-lswiftSwiftOnoneSupport`:

https://github.com/CocoaPods/cocoapods-integration-specs/compare/c70e7ceb2658f1c4640d9b73a5e3626e4e547948...97abe5626eed1ca6b3f2aa166ebd967ade4776bb#diff-178f0f1c7c141dc3e1c184c5586adb34

Like I said though, I don't think this is a big deal though (not a problem), I think Xcode was already inferring that it needed to embed Swift in those cases previously and now you're making it more explicit.

Issue #10053, opened 2020-09-11
Add command for validating the cache


codiophile:
* [x] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md). 🌈

# Feature Request

## Motivation

Once #10018 is released, cache corruption should get a lot less common, but different versions of CocoaPods can still corrupt the cache if run concurrently. That could be fixed with #10035, but even then, an older version of CocoaPods not containing that fix could corrupt the cache.

We are currently trialling #10018 in our environment. We've updated all major branches in all major projects, but there are still branches and projects using old versions of CocoaPods. Since those run much less frequently, the risk of cache corruption is greatly reduced, but still present.

Cache corruption normally manifests itself as a missing file error in the logs, which is easy to diagnose as corrupted cache, but sometimes it causes obscure build failures that are much harder to diagnose. It would be very helpful if there was a command to validate that the cache is in a good state, so that cache corruption can be quickly ruled out or confirmed.

Also, preemptively running cache validation before the build would make the risk of using a broken cache extremely low.

## Suggestion

I would suggest adding the following command:

    pod cache validate

The most basic use case would be to combine it with `pod cache clean --all`:

    pod cache validate || pod cache clean --all

Although, it shouldn't be too difficult to repair the cache, so having the following would be very convenient:

    pod cache validate || pod cache repair

Ideally that could be combined into:

    pod cache validate --repair # or maybe --fix

Ultimately, if the cache validation is performant enough, it could be part of `pod install`, so that there is no risk of using a corrupted cache.

## Questions

Do we have any checksums or file lists that could be used to validate the cache? Otherwise, we could calculate checksums for validation purposes while the files are still in the temp directory and not yet copied to the cache.

I suppose that this might also help in allowing multiple versions to use the same cache, as the checksums could potentially be used to validate that each version uses the same cache format. #10035 was just the easiest and safest way to do it, but not having duplicate caches for each CocoaPods version ever used on the system would be desirable.

codiophile:
@dnkoutso Thank you for adding labels to my issues and pull requests. :)

I've been experimenting with generating checksums for directories in the cache. Next step is to integrate it in CocoaPods. I still haven't received any feedback on #10035, so I'm not sure whether to build off of that or try to make them independent changes.

[...]

PR #10064, opened 2020-09-17
[Cache] Add integrity check using checksums.

codiophile:
Relates to #10053, but adds built in integrity checking instead of adding a separate command.

This is a first version intended to show the effect. Will probably have to do some refactoring, to make sure that we have a lock while performing integrity checks.

### DIFF
--- a/lib/cocoapods/downloader/cache.rb
+++ b/lib/cocoapods/downloader/cache.rb
@@ -1,5 +1,6 @@
 require 'fileutils'
 require 'tmpdir'
+require 'digest/md5'
 
 module Pod
   module Downloader
@@ -156,6 +157,34 @@ module Pod
 
       private
 
+      def checksum_for_directory(directory)
+        md5 = Digest::MD5.new
+        Dir["#{directory}/**/*"].reject { |f| File.directory?(f) }.sort.each do |path|
+          File.open(path, 'rb') do |f|
+            stat = f.stat
+            chunk_size = stat.blksize || 4096
+            md5 << stat.size.to_s
+            while chunk = f.read(chunk_size)
+              md5 << chunk
+            end
+          end
+        end
+        md5.hexdigest
+      end
+
+      def validate_cache(path)
+        checksum_file = "#{path}.md5"
+        return unless File.exist? checksum_file
+        expected_checksum = File.read(checksum_file).strip
+        actual_checksum = checksum_for_directory(path)
+        expected_checksum == actual_checksum
+      end
+
+      def save_checksum(checksum, destination)
+        checksum_file = "#{destination}.md5"
+        File.write(checksum_file, checksum)
+      end
+
       # Ensures the cache on disk was created with the same CocoaPods version as
       # is currently running.
       #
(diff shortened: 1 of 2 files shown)


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* [Cache] Add integrity check using checksums.  
  [codiophile](https://github.com/codiophile)
```

</p>

lib/cocoapods/downloader/cache.rb
```diff
@@ -156,6 +157,34 @@ def self.valid_lock?(file, filename)
 
       private
 
+      def checksum_for_directory(directory)
```
dnkoutso, line 160: how "expensive" (slow) can this get? should we provide a cache keyed by `directory`?

dnkoutso (approved):

dnkoutso:
@codiophile can you rebase? we actually converted to GH actions on latest master.

codiophile:
> @codiophile can you rebase? we actually converted to GH actions on latest master.

Hi @dnkoutso, I just saw this. I've been swamped with work lately. I will have a look when I find some time.

lib/cocoapods/downloader/cache.rb
codiophile: > how "expensive" (slow) can this get?

Well, it didn't have any noticeable impact on test execution time, so it seems fairly cheap. I would assume that the main bottleneck for MD5 is the I/O and not CPU. Considering that we are already doing other I/O operations (like downloading and copying) in connection with this, I think it is fair to assume that the files being checked are already cached in RAM. I suppose, in theory, you could have a problem with pods that are too big to fit in RAM, but pods are generally small and RAM is generally plentiful, so in practice, that should never be a problem.

> should we provide a cache keyed by `directory`?

I'm afraid I don't understand the question. :)

dnkoutso:
@codiophile friendly ping if you'd like to rebase this one just to ensure everything still passes.

dnkoutso:
@igor-makarov does this look good to you?

lib/cocoapods/downloader/cache.rb
```diff
@@ -156,6 +157,34 @@ def self.valid_lock?(file, filename)
...
+            md5 << stat.size.to_s
+            while chunk = f.read(chunk_size)
+              md5 << chunk
+            end
+          end
+        end
+        md5.hexdigest
+      end
+
+      def validate_cache(path)
```
dnkoutso, line 175: this seems to be returning a bool, perhaps rename to `cache_valid?` instead?

lib/cocoapods/downloader/cache.rb
```diff
@@ -281,12 +310,14 @@ def in_tmpdir(&blk)
       #
       def copy_and_clean(source, destination, spec)
         specs_by_platform = group_subspecs_by_platform(spec)
+        Pod::Installer::PodSourcePreparer.new(spec, source).prepare!
```
dnkoutso, line 313: what is the reason to move those out of the lock file block?

PR #10069, opened 2020-09-19
Support adding a git spec repo with a sparse checkout

igor-makarov:
* Support adding a git spec repo with a sparse checkout.
* Add the main spec repo git `CocoaPods/Specs` as sparse.


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Support adding a git spec repo with a sparse checkout  
```

</p>

lib/cocoapods/command/repo/add.rb
```diff
@@ -83,8 +85,25 @@ def clone_repo
...
               git!(command)
+
+              if @sparse
+                Dir.chdir(@name) do
+                  git!(%w[config core.sparseCheckout true])
+                  git!(%w[config core.sparseCheckoutCone true])
+                  File.open('.git/info/sparse-checkout', 'w') do |sparse|
+                    sparse.puts '/*'
+                    sparse.puts '!/*/'
+                    sparse.puts 'Specs/2/e/c/RxSwift'
```
dnkoutso, line 100: what is this needed for?

igor-makarov: Well, this is a funny thing, actually - the `Specs` dir has no direct files, only folders - so using the 2 filters above, git doesn't create it together with the rest of the repo root.   
Because there is no `Specs` dir at the time of creation, `Source#specs_dir` returns an incorrect value.

Adding this arbitrary pod dir causes `Specs` to get created and therefore the rest of the process to work correctly.

We could add a `.gitkeep` file into the spec repo and have a less esoteric sparse checkout init, what do you think?

lib/cocoapods/command/repo/add.rb
```diff
@@ -83,8 +85,25 @@ def clone_repo
             Dir.chdir(config.repos_dir) do
               command = ['clone', @url]
               command << '--progress' if @progress
+              command << %w[--filter=blob:none --no-checkout --depth 1] if @sparse
```
endocrimes, line 88: We should check with GitHub if adding `--depth 1` is going to be problematic here. (I expect it might be, given that they don't recommend that to folks here: https://github.blog/2020-01-17-bring-your-monorepo-down-to-size-with-sparse-checkout/)

lib/cocoapods/command/repo/add.rb
endocrimes: Adding a `.gitkeep` or similar (preferably with content that explains its use) would be preferred to an arbitrary pod I think

lib/cocoapods/command/repo/add.rb
```diff
@@ -83,8 +85,25 @@ def clone_repo
...
               command << '--' << @name
+
               git!(command)
+
+              if @sparse
+                Dir.chdir(@name) do
```
endocrimes, line 94: break this section into a new method and add some tests?

lib/cocoapods/command/repo/add.rb
igor-makarov: The `--depth 1` part cannot be removed, as cloning without it brings us back to the non-performant full clone scenario. 😭

The good news is that I think there's no need to worry about this. The cloning part doesn't even involve sparse checkouts at all, but separate git feature called [partial clones](https://about.gitlab.com/blog/2020/03/13/partial-clone-for-massive-repositories/). 

It basically allows you to perform a clone without bringing over all the objects. Additionally, it works great with shallow clones as it simply keeps the clone shallow and allows updating it forward without unshallowing.

That said, I'd love to run it by GH folks if you can point me to anyone, @endocrimes?

igor-makarov:
@endocrimes @dnkoutso I've added the `.gitkeep` to Specs, please merge [Specs#14473](https://github.com/CocoaPods/Specs/pull/14473) so that I can proceed with fixing the current PR.

dnkoutso:
@igor-makarov want to rebase just in case to build with GH actions? Also is this considered done?

igor-makarov:
@dnkoutso I still need to add some tests for it...  Got sidelined by the whole CDN shenanigans and the CI build flake shenanigans 😂

PR #10070, opened 2020-09-19
Resolve new Xcode 12 deployment target warning

### LINKED ISSUE #9884: Raise default macOS/iOS deployment target
* [X] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md). 🌈

# Report

## What did you do?

Built a macOS project that includes pods which do not have the deployment target set for macOS (for example, [`Argon2`](https://cocoapods.org/pods/Argon2)).

Compile without errors and warnings.

## What happened instead?

It seems that starting with Xcode 12, I get a warning when building those pods (I don't get them with Xcode 11):

`The macOS deployment target 'MACOSX_DEPLOYMENT_TARGET' is set to 10.6, but the range of supported deployment target versions is 10.9 to 10.16.99.`

Since 10.6 is ancient, it would be nice to raise the default deployment target for macOS to 10.9 to prevent this warning.

paulb777:
Fix #9884

Sets deployment_target to max of the Podfile and podspec specifications

Integration test PR at https://github.com/CocoaPods/cocoapods-integration-specs/pull/294

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
-* None.  
-
+* Resolve new Xcode 12 deployment target warning  
+  [Paul Beusterien](https://github.com/paulb777)
+  [#9392](https://github.com/CocoaPods/CocoaPods/issues/9884) 
 
 ## 1.10.0.rc.1 (2020-09-15)
 
--- a/lib/cocoapods/installer/analyzer.rb
+++ b/lib/cocoapods/installer/analyzer.rb
@@ -860,6 +860,8 @@ module Pod
           minimum = Version.new('8.0')
           deployment_target = [deployment_target, minimum].max
         end
+        # Set deployment_target to max of the Podfile and podspec specifications.
+        deployment_target = [deployment_target, target_definitions.first.platform.deployment_target].max
         Platform.new(platform_name, deployment_target)
       end
 
--- a/spec/unit/installer/analyzer_spec.rb
+++ b/spec/unit/installer/analyzer_spec.rb
@@ -2465,6 +2465,46 @@ module Pod
       end
     end
 
+    describe 'platform version' do
+      before do
+        @banana_spec = fixture_spec('banana-lib/BananaLib.podspec')
+        @podfile = Pod::Podfile.new do
+          target 'SampleProject' do
+            pod 'BananaLib', :git => 'example.com'
+          end
+        end
+        @analyzer = Pod::Installer::Analyzer.new(config.sandbox, @podfile)
+      end
+
+      it 'returns static library default ios deployment target of 4.3' do
+        target_definition = fixture_target_definition('App')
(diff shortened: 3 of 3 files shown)


aaronbrethorst:
🎉 thank you, Paul!

lib/cocoapods/installer/analyzer.rb
```diff
@@ -860,6 +860,8 @@ def determine_platform(specs, target_definitions, build_type)
           minimum = Version.new('8.0')
           deployment_target = [deployment_target, minimum].max
         end
+        # Set deployment_target to max of the Podfile and podspec specifications.
+        deployment_target = [deployment_target, target_definitions.first.platform.deployment_target].max
```
dnkoutso, line 864: hmmm why the `first` target definition? I guess it doesnt matter because if a pod is embedded between two target definitions with different deployment targets then we split the pod target into each deployment target?

segiddins:
I think this is a breaking change?? We intentionally use the deployment target specified in the podspec, and changing the deployment target can cause code to stop compiling

lib/cocoapods/installer/analyzer.rb
paulb777: Yes this function is already making that assumption.  See line 854.

paulb777:
@segiddins It seems that Apple generally supports code building for future deployment targets, but I suppose there could be breakage for API removals or deprecation warnings.

Would it be reasonable to make this change on a new Podfile option for the cases where it does work and to support those who want to get rid of the Xcode 12 warning?

Or maybe we should wait a bit to see if a future Xcode provides a better way to deal with libraries supporting earlier iOS versions than Xcode 12.0: https://forums.swift.org/t/minimum-ios-version-xcode-12-and-build-warnings/40224

paulb777:
https://github.com/firebase/firebase-ios-sdk/issues/6502 is an example of why this PR is potentially breaking and that doing it on option might be the best route.

CHANGELOG.md
```diff
@@ -12,8 +12,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
-* None.  
-
+* Resolve new Xcode 12 deployment target warning  
+  [Paul Beusterien](https://github.com/paulb777)
+  [#9392](https://github.com/CocoaPods/CocoaPods/issues/9884)
```
dnkoutso, line 17: The link doesn't match the issue number

dnkoutso:
hmmm this adds into a bigger discussion on allowing folks to override values of pod targets...

it also came up with `swift_version` so you can do:

`pod 'MyPod', :swift_version => '5.0'`.

I never implemented a generic override mechanism due to the fact that `post_install` hooks exist.

I am not sure which way to go tbh...

paulb777:
I'm getting convinced. Perhaps we should close this PR and close #9884 with instructions on how to use a `post_install` hook to change the deployment target.

PR #10092, opened 2020-09-25
Consider `swift_versions` attribute for detecting if a pod uses Swift - V2 🌈

gyfelton:
the original PR: https://github.com/CocoaPods/CocoaPods/pull/8762 is declined at the end because @dnkoutso raised a concern that "s.swift_versions DSL is currently a root attribute, This means a pod target will be considered as if its using Swift across all of its subspecs whereas some subspecs or tests specs or app specs might not." This is a totally legit concern.

What this PR in addition is that the use of 's.swift_version' strictly applies to a pod that has no source code for itself as well its subspecs (what this issue entails [#8649](https://github.com/CocoaPods/CocoaPods/issues/8649). `should_build` is therefore `false` and the whole pod target can be treated as a single "item". In this case the best way that we know to deduce whether it uses Swift or not is through the `swift_version` attribute. And it does not impact subspecs because there is nothing to be built for subspecs (if any).


CHANGELOG.md
```diff
@@ -26,7 +26,11 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Sebastian Shanus](https://github.com/sebastianv1)
   [#10016](https://github.com/CocoaPods/CocoaPods/pull/10016)
 
-
+* Correctly detect that a prebuilt pod uses Swift.
```
dnkoutso, line 29: nit: add two empty spaces at the end of this line for markdown formatting

spec/unit/target/pod_target_spec.rb
```diff
@@ -634,6 +634,11 @@ module Pod
           @pod_target.uses_swift?.should == true
         end
 
+        it 'returns that it uses swift with a pre built vendored framework and specifying swift_versions' do
```
dnkoutso, line 637: did you verify the test fails without your change and then goes green with your change? I often do that to avoid false positives!

lib/cocoapods/target/pod_target.rb
```diff
@@ -302,14 +302,14 @@ def app_spec_consumers
     end
 
     # @return [Boolean] Whether the target uses Swift code. This excludes source files from non library specs.
+    # Note that if a target contains no source code, we fallback to check if it indicates a swift version
```
dnkoutso, line 305: nit: I think the formatting in this line is off, also missing period.

dnkoutso (approved):

dnkoutso:
cc @paulb777 and maybe @amorde who were original reviewers in https://github.com/CocoaPods/CocoaPods/pull/8762

spec/unit/target/pod_target_spec.rb
gyfelton: yeah i usually do this but today I was being lazy :( let me break it then put this back

gyfelton: revert my change and verified that it failed

amorde (approved):

paulb777 (approved):

gyfelton:
failed on 
```
Pod::Informative: [!] Unable to add a source with url `https://cdn.cocoapods.org/` named `trunk-1`.
	
for spec - runs pod repo add-cdn when there is no matching source for CocoaPods Trunk [FAILED]
```
This happens to other PR jobs and i cannot reproduce locally (it passes lcoally)...

spec/unit/sources_manager_spec.rb
```diff
@@ -75,7 +75,7 @@ module Pod
 
           it 'runs `pod repo add-cdn` when there is no matching source for CocoaPods Trunk' do
             Command::Repo::AddCDN.any_instance.stubs(:run).once
-            stub_url_as_cdn('https://cdn.cocoapods.org/')
```
gyfelton: this fixed this test but then broke integration tests.... so strange

spec/unit/sources_manager_spec.rb
```diff
@@ -42,6 +42,8 @@ module Pod
       @sources_manager = Source::Manager.new(config.repos_dir)
       stub_url_as_cdn('https://cdn.cocoapods.org')
       stub_url_as_cdn('http://cdn.cocoapods.org')
+      stub_url_as_cdn('https://cdn.cocoapods.org/')
```
gyfelton, line 45: @dnkoutso For some reason the test case that uses `Pod::TrunkSource::TRUNK_REPO_URL` returns `https://cdn.cocoapods.org/` as the URL. This ending slash is fine for local machine but breaks for CI... so I end up adding this and it fixes the issue 🤦 wondering why...

dnkoutso: cc @igor-makarov is it possible related to the intermittent issue and the change we did? I wonder why we need to have this now.

igor-makarov: This is possible.

 @gyfelton can you try removing this change and see if itthe build still passes?

gyfelton: no it's there to fix an rspec failure on a test case in the first place, let's use the other PR I tagged you at to continue the discussion

PR #10098, opened 2020-09-28
Add Github Actions workflow

igor-makarov:
Travis is very slow, this replaces it.

It also replaces CircleCI as the performance of GH runners is comparable to CC and it's more convenient to have every test in one place.

Integration specs fixtures: https://github.com/CocoaPods/cocoapods-integration-specs/pull/296

The Travis/Circle checks are holding up the build but it can still be merged. After merging, the required checks for the protected branch will need changing (I can do it) and the CircleCI build needs to be disabled (@dnkoutso cause I don't have CircleCI admin).


dnkoutso:
Want to do it for 1-10-stable branch since we are still converging there and we wont be able to land changes/fixes?

dnkoutso (approved):

PR #10117, opened 2020-10-03
Fix badges in the README

igor-makarov:
* Replace old Travis badge with GH Actions badge
* Fix the gem version shield to use HTTPS
* Remove CodeClimate as it isn't being used

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,10 +2,8 @@
 
 ### CocoaPods: The Cocoa dependency manager
 
-[![Build Status](http://img.shields.io/travis/CocoaPods/CocoaPods/master.svg?style=flat)](https://travis-ci.org/CocoaPods/CocoaPods)
-[![Gem Version](http://img.shields.io/gem/v/cocoapods.svg?style=flat)](http://badge.fury.io/rb/cocoapods)
-[![Maintainability](https://api.codeclimate.com/v1/badges/8f0fe544baf2ae1acc2b/maintainability)](https://codeclimate.com/github/CocoaPods/CocoaPods/maintainability)
-[![Test Coverage](https://api.codeclimate.com/v1/badges/8f0fe544baf2ae1acc2b/test_coverage)](https://codeclimate.com/github/CocoaPods/CocoaPods/test_coverage)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,10 +2,8 @@
 
 ### CocoaPods: The Cocoa dependency manager
 
-[![Build Status](http://img.shields.io/travis/CocoaPods/CocoaPods/master.svg?style=flat)](https://travis-ci.org/CocoaPods/CocoaPods)
-[![Gem Version](http://img.shields.io/gem/v/cocoapods.svg?style=flat)](http://badge.fury.io/rb/cocoapods)
-[![Maintainability](https://api.codeclimate.com/v1/badges/8f0fe544baf2ae1acc2b/maintainability)](https://codeclimate.com/github/CocoaPods/CocoaPods/maintainability)
-[![Test Coverage](https://api.codeclimate.com/v1/badges/8f0fe544baf2ae1acc2b/test_coverage)](https://codeclimate.com/github/CocoaPods/CocoaPods/test_coverage)
```
dnkoutso, line 8: did e lose codeclimate as part of this?

dnkoutso (approved):
want to do 1-10-stable?

README.md
igor-makarov: Not as part of this - I assumed it wasn't working because I have never seen it in PR statuses.

igor-makarov:
What do you mean @dnkoutso? Add a badge in 1.10?

dnkoutso:
ah nm yes! that should be ok.

igor-makarov:
@dnkoutso can I merge this? I'm not missing anything with CodeClimate?

README.md
dnkoutso: I thought we got it to work btw, is the link not going to be visible anymore?

igor-makarov: I looked at the link... frankly I'm not sure what I'm seeing.  
Looks like some sort of scoring system + another linter?  
Seems pretty hard to determine what version of the code it checked and it doesn't actually report anything to the PRs.

Who was in charge of it?

amorde: I recently updated this, we don't check on PRs but the badge should be up to date

amorde: It updates on merges to `master` I believe

amorde:
I think we can keep CodeClimate - it doesn't have to be a check on PRs to be useful as a badge

Issue #10111, opened 2020-10-01
Add Xcode 12 compatible support for debug symbols in XCFrameworks


johntmcintosh:
* [🌈] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

With **Xcode 11**, there was no automated mechanism for Xcode to embed debug symbols (dSYMs and BCSymbolMaps) that were embedded in an XCFramework into the built product's archive. As a result, CocoaPods' support for XCFrameworks included automatically extracting debug symbols from each architecture's framework directory, and including them into the built app. The expectation was that the XCFramework would be structured like this:

```
MySDK.dSYMs/
- MySDK.framework.ios-arm64.dSYM
- MySDK.framework.ios-x86_64-simulator.dSYM
MySDK.xcframework/
- Info.plist
- ios-arm64/
    - MySDK.framework/
        - BCSymbolMaps/
- ios-x86_64-simulator/
    - MySDK.framework/
```

**EDIT (Oct 7): Originally I had described that the dSYMs were placing as siblings of the BCSymbolMaps in the Xcode 11 style, but after some additional testing, I realized that the Xcode 11 style that CocoaPods was expecting was for the dSYMs to be placed in a .dSYMs directory as a top level sibling of the .xcframework.**

This was great!

Now, **Xcode 12** has added support to `xcodebuild -create-xcframework` for passing a `-debug-symbols` argument. When creating an XCFramework this way with Xcode 12, the resulting directory structure does not match what we had presumed when initially building support for Xcode 11:

```
MySDK.xcframework
- Info.plist
- ios-arm64/
    - BCSymbolMaps/
    - dSYMs/
    - MySDK.framework/
- ios-arm64_x86_64-simulator/
    - dSYMs/
    - MySDK.framework/
```

The debug symbols are placed as siblings of each architecture's `.framework` rather than as children of it.

As a result, when integrating through CocoaPods, the debug symbols from this type of XCFramework are not being included in the final product archive.

## What did you expect to happen?

XCFrameworks that are compiled with Xcode 12 to include debug symbols should automatically support those debug symbols being included in the built app when integrating the XCFramework through CocoaPods.

## What happened instead?

XCFrameworks that are compiled with Xcode 12 to include debug symbols automatically support debug symbol inclusion in the built app when manually added to the parent project, but not when integrated through CocoaPods.

### Stack

[...]

PR #10114, opened 2020-10-02
Do not warn for missing arch if `$ONLY_ACTIVE_ARCH` is set to `NO`.

dnkoutso:
closes #10076

integration specs: https://github.com/CocoaPods/cocoapods-integration-specs/pull/299

PR #10122, opened 2020-10-05
#10111 - Update XCFramework installation to support Xcode 12 debug symbols

johntmcintosh:
#10111 - Update XCFramework installation to support Xcode 12 debug symbols placed as siblings of the .framework inside each architecture's directory

Related integration-specs PR: https://github.com/CocoaPods/cocoapods-integration-specs/pull/301

### DIFF
--- a/lib/cocoapods/generator/copy_xcframework_script.rb
+++ b/lib/cocoapods/generator/copy_xcframework_script.rb
@@ -194,12 +194,8 @@ install_xcframework() {
     mkdir -p "$destination"
   fi
 
-  if [[ "$package_type" == "library" ]]; then
-    # Libraries can contain headers, module maps, and a binary, so we'll copy everything in the folder over
-    copy_dir "$source/" "$destination"
-  elif [[ "$package_type" == "framework" ]]; then
-    copy_dir "$source" "$destination"
-  fi
+  copy_dir "$source/" "$destination"
+
   echo "Copied $source to $destination"
 }
 
@@ -226,12 +222,7 @@ install_xcframework() {
         is_framework = xcframework.build_type.framework?
         args << shell_escape(is_framework ? 'framework' : 'library')
         slices.each do |slice|
-          args << if is_framework
-                    shell_escape(slice.path.relative_path_from(root))
-                  else
-                    # We don't want the path to the library binary, we want the dir that contains it
-                    shell_escape(slice.path.dirname.relative_path_from(root))
-                  end
+          args << shell_escape(slice.path.dirname.relative_path_from(root))
         end
         args.join(' ')
       end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Wes Campaigne](https://github.com/Westacular)
   [#10071](https://github.com/CocoaPods/CocoaPods/issues/10071)
 
+* Add support for automatically embeddeding XCFramework debug symbols for XCFrameworks generated with Xcode 12  
+  [johntmcintosh](https://github.com/johntmcintosh)
+  [#10111](https://github.com/CocoaPods/CocoaPods/issues/10111)
 
 ## 1.10.0.rc.1 (2020-09-15)
 
--- a/spec/unit/generator/copy_xcframeworks_script_spec.rb
+++ b/spec/unit/generator/copy_xcframeworks_script_spec.rb
@@ -6,7 +6,7 @@ module Pod
       xcframework = Xcode::XCFramework.new(fixture('CoconutLib.xcframework'))
       generator = CopyXCFrameworksScript.new([xcframework], temporary_sandbox.root, Platform.ios)
       generator.send(:script).should.include <<-SH.strip_heredoc
-        install_xcframework "${PODS_ROOT}/../../spec/fixtures/CoconutLib.xcframework" "CoconutLib" "framework" "ios-armv7_arm64/CoconutLib.framework" "ios-i386_x86_64-simulator/CoconutLib.framework"
+        install_xcframework "${PODS_ROOT}/../../spec/fixtures/CoconutLib.xcframework" "CoconutLib" "framework" "ios-armv7_arm64" "ios-i386_x86_64-simulator"
       SH
     end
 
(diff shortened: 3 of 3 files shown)


dnkoutso:
@johntmcintosh can you rebase against `1-10-stable` branch so we can ship this with 1.10?

johntmcintosh:
@dnkoutso sure thing. This and the submodule PR are rebased and re-pointed to their respective `1-10-stable` branches

lib/cocoapods/generator/copy_xcframework_script.rb
```diff
@@ -226,12 +222,7 @@ def install_xcframework_args(xcframework, slices)
         is_framework = xcframework.build_type.framework?
         args << shell_escape(is_framework ? 'framework' : 'library')
         slices.each do |slice|
-          args << if is_framework
-                    shell_escape(slice.path.relative_path_from(root))
-                  else
-                    # We don't want the path to the library binary, we want the dir that contains it
-                    shell_escape(slice.path.dirname.relative_path_from(root))
-                  end
+          args << shell_escape(slice.path.dirname.relative_path_from(root))
```
amorde, line 225: hah, glad this was already supported in one code path :)

lib/cocoapods/generator/copy_xcframework_script.rb
```diff
@@ -194,12 +194,8 @@ def script
     mkdir -p "$destination"
   fi
 
-  if [[ "$package_type" == "library" ]]; then
-    # Libraries can contain headers, module maps, and a binary, so we'll copy everything in the folder over
-    copy_dir "$source/" "$destination"
-  elif [[ "$package_type" == "framework" ]]; then
-    copy_dir "$source" "$destination"
-  fi
+  copy_dir "$source/" "$destination"
```
amorde, line 197: We might run in to problems here if xcframeworks are vended using the new style of embedding dSYMs but user's are still on Xcode 11.x. In that case, we would not want to copy those files and instead rely on the existing logic here:

https://github.com/CocoaPods/CocoaPods/blob/a3632cc1ad06245e57ccb46a845a47874f55c4fb/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb#L814-L823

I believe there may be some exposed variables here that we can use to conditionally opt-in to the newer behavior, but I haven't thought this out thoroughly yet

lib/cocoapods/generator/copy_xcframework_script.rb
johntmcintosh: I was relieved to find it 😅

lib/cocoapods/generator/copy_xcframework_script.rb
johntmcintosh: I did some initial testing with that use-case today and it actually appeared to be working with this implementation using new style dSYMs and Xcode 11. My theory at the time was that the logic that's looking for dSYMs happened to be looking from new location down, rather than starting down one level. If that's the case, then we might be in the clear with this, but I can do some more focused investigation on that use-case tomorrow as well.

johntmcintosh: @amorde while investigating, I'm getting stuck tracing something... in the generated `Pods-MyApp-frameworks.sh` script, there's an `install_dsym()` function which I think is added to that script from [embed_frameworks_script.rb](https://github.com/CocoaPods/CocoaPods/blob/1-10-stable/lib/cocoapods/generator/embed_frameworks_script.rb#L134), but I haven't been able to find what calls that function. Is that function being used there and I'm overlooking it, or is it leftover from a previous iteration and no longer used?

dnkoutso: @johnmcintosh I have seen this code while inspecting the sources recently. I think its vestigial and can be removed. You can keep it there out of abundance of caution since its mostly irrelevant to this PR.

amorde: My concern is when archiving, we need to make sure we don't leave the dSYMs or bcsymbolmaps in the archive (in the incorrect place) or Apple will reject the archive. I haven't testing Xcode 12's archiving using the new dSYMs to see where they end up but I'd guess they use the previous location and don't leave it inside the framework

johntmcintosh: > My concern is when archiving, we need to make sure we don't leave the dSYMs or bcsymbolmaps in the archive (in the incorrect place) or Apple will reject the archive.

@amorde yep, definitely. To make sure I'm testing in the right direction, what is your understanding of the source location for dSYMs from Xcode 11 xcframeworks? Since Apple didn't have guidance (that I'm aware of) it was up to us where to put them. What had been your expectation originally for where the Xcode 11 xcframework dSYMs would be located?

I had originally been thinking that they would be inside each architecture's .framework next to the BCSymbolMaps directory, but using the [PSPDFKit xcframework podspec](https://customers.pspdfkit.com/pspdfkit-ios/latest.podspec) as a reference right now, they were placing the dSYMs as a sibling to the xcframework in the root of the zip download, rather than nested down somewhere inside of the xcframework. In that example since they aren't part of the .frameworks, I don't think there's any "removal" from a source location that would need to happen for that Xcode 11 style.

I polished up my test project a bit today. If you can help clarify where CocoaPods was expecting xcframework debug symbols to be placed previously, then I can spend some time tomorrow working on migrating the test project I've been using into the examples tests that will run with CI so we can have validation of both expectations.

johntmcintosh: @amorde after a little more investigation today, this approach is still seeming correct for everything I've been able to validate. I just pushed up a branch off of this branch as [`xcframeworkDebugSymbolsXcode12-example`](https://github.com/johntmcintosh/CocoaPods/tree/xcframeworkDebugSymbolsXcode12-example). (I created a separate branch since the example compiled xcframework is checked in and there were hundreds of files in the diff so I didn't want to clutter this diff right now).

If you pull that branch, I've added a new example **examples/Vendored-XCFramework12-Example** which has one example project that pulls in two libraries.

- BananaLib is setup in the "current" Xcode 11 recommended style:
  - dSYMs are placed in a .dSYMs directory as a sibling to the .xcframework [Ref #10114](https://github.com/CocoaPods/CocoaPods/pull/10114/files#r499044968). They do not need to be stripped from the built archive's copy of the .framework, because they were never inside of it to start.
  - BCSymbolMaps are placed per architecture at `BananaLib.xcframework/<architecture>/BananaLib.framework/BCSymbolMaps/` and do need to be from the built archive's copy of the .framework.
- CoconutLib is setup in the "new" Xcode 12 style:
  - dSYMs and BCSymbolMaps are both placed per architecture as a sibling to the architecture's .framework: 
    - `CoconutLib.xcframework/<architecture>/BCSymbolMaps/`
    - `CoconutLib.xcframework/<architecture>/dSYMs/`
    - `CoconutLib.xcframework/<architecture>/CoconutLib.framework/`
  - Because they are not children of the .framework they also don't need to be stripped from the built archive, they just need to be copied into it at the correct location.

If you open the **Vendored-XCFramework12-Example/Example.xcworkspace**, there's an **ArchiveAndValidate** aggregate target scheme that you can compile. There's a script build phase that compiles a release archive of the app, and then validates that:

- The expected Xcode 11 style BCSymbolMaps and dSYMs are installed from BananaLib
- The expected Xcode 12 style BCSymbolMaps and dSYMs are installed from CoconutLib
- The embedded BCSymbolMaps from the XCode 11 style BananaLib are stripped from the embedded BananaLib.framework

If any of those validations do not pass, the script will fail the build. This setup appears to have this test now be included when I run the example tests (`bundle exec rake examples`). 

The test example should be in a state that it could be PR'd on its own, I just didn't want to clutter this one up. Let me know if you'd prefer I go ahead and include it and I can update this PR to include that test.

One last note, when my test runs, the result also includes three warnings along the lines of

> [ CP] Vendored binary '< long-path >' contains architectures (armv7k arm64_32) none of which match the current build architectures (arm64).

which relates to #10114. @dnkoutso if helpful for that work, my test project here may also be able to be used for validating work related to only including the relevant debug symbols (when the debug symbols are being included in the source pod using the Xcode 11 style structure).

amorde: @johntmcintosh thank you so much for the attention to detail here. Looking over your example, I see the source of my own confusion. I was thinking the new Xcode 12-style had the dSYMs and BCSymbolmaps _inside_ the `.framework` bundle, but I see now they are parallel to the `.framework` in the directory structure.

The example you've setup is really great, and is a big improvement over our existing examples since it validates the structure of an archived app. This is a gap that we've had in our automated tests for a long time.

amorde (approved):

lib/cocoapods/generator/copy_xcframework_script.rb
johntmcintosh: Thanks @amorde ! I was glad the idea of a run script in the aggregate target was able to work with the current setup. The overall structure of how the example tests are setup is neat.

FYI - I saw you merged the integration-specs PR, so I've pushed an update to this one bumping the submodule pointer to the new, merged commit.

Let me know if you'd prefer me add the example as part of this PR, otherwise I'll plan to put up a second PR with it right after this one merges.

amorde: Second PR sounds great 👍

PR #10133, opened 2020-10-07
Do not add dependencies to 'Link Binary With Libraries' phase.

dnkoutso:
With the introduction of per-configuration dependencies, we cant add dependencies to 'Link Binary With Libraries' phase in Xcode because Xcode cannot differentiate between adding a dependency for only DEBUG or only RELEASE.

We rely on CocoaPods xcconfig to provide all linker flags and we maintain the part that we add dependencies in the 'Dependencies' phase as we expect them to be built so we can then embed them to targets.

@paulb777 I'd appreciate if you have the time to run this change against your Firebase CI as this code is kinda old...

This is for 1.11.0 so it will be a while before it ships.

Integration specs: https://github.com/CocoaPods/cocoapods-integration-specs/pull/302

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
+* Do not add dependencies to 'Link Binary With Libraries' phase.  
+  [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
+  [#10133](https://github.com/CocoaPods/CocoaPods/pull/10133)
+
 * Ensure cache integrity on concurrent installations.  
   [Erik Blomqvist](https://github.com/codiophile)
   [#10013](https://github.com/CocoaPods/CocoaPods/issues/10013)
--- a/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_dependency_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_dependency_installer.rb
@@ -36,25 +36,24 @@ module Pod
             pod_target = pod_target_installation_result.target
             native_target = pod_target_installation_result.native_target
             project = native_target.project
-            frameworks_group = project.frameworks_group
 
             # First, wire up all resource bundles.
             wire_resource_bundle_targets(pod_target_installation_result.resource_bundle_targets,
                                          native_target, pod_target)
             # Wire up all dependencies to this pod target, if any.
             wire_target_dependencies(pod_target, native_target, project, pod_target_installation_results,
-                                     metadata_cache, frameworks_group)
+                                     metadata_cache)
 
             # Wire up test native targets.
             unless pod_target_installation_result.test_native_targets.empty?
               wire_test_native_targets(pod_target, pod_target_installation_result, pod_target_installation_results,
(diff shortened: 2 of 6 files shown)


segiddins:
very suspect this doesn't break any integration tests -- are we missing an integration test case here?

dnkoutso:
I was waiting for it to fail! :D but yes I hope it breaks some and I will update.

segiddins (approved):

amorde (approved):

paulb777:
Thanks for the ping @dnkoutso 

This fails in the Firebase repo:

At https://github.com/firebase/firebase-ios-sdk/tree/pb-test-cp10133:

`bundle exec pod gen FirebaseCore.podspec --auto-open --local-sources=./ --platforms=ios`
Command-U

Screen Shot 2020-10-08 at 11 21 27 AM

amorde:
Thank you for checking @paulb777

dnkoutso:
Thanks @paulb777 investigating this further...

dnkoutso:
@paulb777 initial report seems to be related to `'OTHER_CFLAGS' => '-fno-autolink'` why is this set?

segiddins:
If this is failing with auto-linking disabled, that would imply we're missing linker flags

dnkoutso:
yeap we do it seems Firebase.debug.xcconfig is missing its dependencies ```OTHER_LDFLAGS = $(inherited) -framework "Foundation" -framework "UIKit"
```

I see them being counted for but for some reason it does not include them..investigating further

dnkoutso:
I may have a lead for this...

dnkoutso:
@paulb777 I pushed an update that adds the frameworks to link for the library target that used to be part of 'Link Library With Binaries' phase.

I was able to successfully build the branch you posted. Would appreciate another comprehensive CI run here.

paulb777:
Full Firebase CI running in https://github.com/firebase/firebase-ios-sdk/pull/6704

dnkoutso:
@paulb777 @segiddins @amorde should be fixed.

One thing to note...

With the build phase we used to add _direct_ dependencies to the build phase for _all_ spec types (libraries, app specs and test specs).

Now that the code was deleted I noticed:

- We also used to add ALL (direct and transitive) flags for test specs and app specs, static and dynamic.
- We never added anything for dynamic library specs. This PR changes that and we now add ALL (direct and transitive) dynamic dependencies. We do not add static ones because they will yield duplicate symbols.

segiddins:
Do we need to add the transitive ones? I think just adding the direct flags for dynamic deps should do the trick?

dnkoutso:
No strong preference here, I did it for consistency. I can update once @paulb777 verifies this one works anyway. I didnt want to squash yet as it breaks the verification.

amorde:
If we only add direct will symbols of transitive deps still be available?

dnkoutso:
@amorde I guess the question is how was it working before?

dnkoutso:
Switched to use direct dependencies to match what the build phase did. The internal bug we have is fixed and I can also build and CMD+U on the Firebase project. Will still want to verify 100% Firebase CI passes with this and then I will merge.

paulb777 (approved):
Firebase tests look good.

PR #10139, opened 2020-10-09
Fix badges in the README

igor-makarov:
* Replace old Travis badge with GH Actions badge
* Fix the gem version shield to use HTTPS

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,8 +2,8 @@
 
 ### CocoaPods: The Cocoa dependency manager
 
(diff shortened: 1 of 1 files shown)


igor-makarov:
@dnkoutso new PR because old one didn't update for some reason. Also, GH actions aren't running for some reason?

paulb777:
GH actions look like they're flaky right now - they didn't start on a recent Firebase PR either.

dnkoutso (approved):

PR #9441, opened 2019-12-27
When pod target is a static framework, save time by copying compiled resources

igor-makarov:
I've investigated an issue with pods as static frameworks and discovered the following problem:  

Because the pod target is of framework type, the `Copy Bundle Resources` phase actually gets executed (as opposed to static library). This means that processed resources (XIBs, storyboards etc.) are processed using the appropriate tool. Additionally, Xcode does this processing in parallel.

However, when the integration phase `[CP] Copy Pods Resources` runs, its inputs are the original resource files, like in the case of a static library. This results in the resource processing tools running again, this time serially.  
This results in duplicate work and performance penalty.

This PR solves this by converting the `[CP] Copy Pods Resources` input paths to be the framework output files (`.nib`, `.storyboardc` etc.). These get handled by plain `rsync` and run blazingly fast.

PR #10175, opened 2020-10-28
Fix processing of xcassets resources when pod target is static framework

### LINKED ISSUE #10170: Pods-<target>-resources.sh contains incorrect assets path
🌈

# Report

## What did you do?

Tried to build a project with a dev pod containing `.xcassets` file, using statically-linked frameworks.

Expected the project to build successfully.

## What happened instead?

Got an error during the `[CP] Copy Pods Resources` build phase:
```
error: Resource "/Users/federicotrimboli/Library/Developer/Xcode/DerivedData/AssetsBug-daxuxarxrtwbxadonvosjaawkieq/Build/Products/Debug-iphonesimulator/DevPod/DevPod.framework/Images.car" not found. Run 'pod install' to update the copy resources script.
```

The original `.xcassets` file included in the dev pod is indeed called `Images.xcassets`, but when navigating to the path shown in the error, I can see that the compiled `.car` file is actually called `Assets.car`. Looks like Xcode assigns `Assets.car` to every compiled `xcassets` regardless of the original name. However, the `Pods-<target>-resources.sh` script tries to access the file using the original name `Images` instead of `Assets`.

Additionally, if there were more than one `xcassets`, all of them would be merged+compiled into a single `Assets.car`, so the naming makes sense. In this case, the script should contain a single line but instead it's generating one line per `xcassets` file.

### Stack

```
   CocoaPods : 1.10.0
        Ruby : ruby 2.6.3p62 (2019-04-16 revision 67580) [x86_64-darwin18]
    RubyGems : 3.0.3
        Host : Mac OS X 10.15.7 (19H2)
       Xcode : 12.0.1 (12A7300)
         Git : git version 2.24.3 (Apple Git-128)
Ruby lib dir : /Users/federicotrimboli/.rvm/rubies/ruby-2.6.3/lib
Repositories : master - git - https://github.com/CocoaPods/Specs.git @ af326a1bd1b3bbe0188e494fc00d6355e455e836

               trunk - CDN - https://cdn.cocoapods.org/
```

### Installation Source

```
Executable Path: /Users/federicotrimboli/.rvm/gems/ruby-2.6.3/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.4
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-trunk       : 1.5.0
cocoapods-try         : 1.2.0
```

### Podfile

```ruby
unless defined?(Bundler)
  puts "\nPlease re-run using:".red
  puts "  bundle exec pod #{ARGV.join(' ')}\n"
  exit(1)
end

install! 'cocoapods'
platform :ios, '12.0'
use_frameworks! :linkage => :static

target 'AssetsBug' do
  pod 'DevPod', :path => 'DevPod'
end
```

I'm attaching a sample Xcode project that demonstrates the issue: [AssetsBug.zip](https://github.com/CocoaPods/CocoaPods/files/5438117/AssetsBug.zip)

fedetrim:
#9441 introduced an optimization for static framework pod targets by copying compiled resources (storyboards, xcdatamodels, etc) instead of the original resources and therefore avoiding duplicate work (once in the framework's `Copy Bundle Resources` phase, and another in the final target's integration phase `[CP] Copy Pods Resources`).

This optimization assumes that each `<name>.xcassets` has a compiled `<name>.car` counterpart, and that you can have many `.car` files in a single bundle.
However, in reality, all `<name>.xcassets` from the same framework are merged and compiled into a single `Assets.car`, meaning that the  `[CP] Copy Pods Resources` integration phase is failing because it's not able to find the compiled resources:
Screen Shot 2020-10-28 at 16 58 31

With the failing piece of code being within `Pods-<target>-resources.sh`:
```ruby
  install_resource "${BUILT_PRODUCTS_DIR}/DevPod/DevPod.framework/AnotherOne.car"
  install_resource "${BUILT_PRODUCTS_DIR}/DevPod/DevPod.framework/Images.car"
fi
```

In order for that to not fail, it should look like this:
```ruby
  install_resource "${BUILT_PRODUCTS_DIR}/DevPod/DevPod.framework/Assets.car"
fi
```

Even if we addressed that issue by correcting the path of the compiled resources, we still wouldn't be able to copy more than one `Assets.car` (one per each framework) to the main bundle.

The only way of handling xcassets is by merging+building them all at once, and then copying that single `Assets.car` to the main bundle (which is what was being done before this optimization).

This PR solves the issue by adding the original xcassets back to being processed by the `[CP] Copy Pods Resources` integration phase, and removes the resources from the frameworks' `Copy Bundle Resources` phase so that the compiling is only done once.

Closes #10170

### DIFF
--- a/spec/unit/installer/xcode/pods_project_generator/pod_target_installer_spec.rb
+++ b/spec/unit/installer/xcode/pods_project_generator/pod_target_installer_spec.rb
@@ -772,6 +772,24 @@ module Pod
               resource.should.be.nil
             end
 
+            it 'doesn\'t add xcassets resources to the static framework target' do
+              @pod_target.stubs(:build_type => BuildType.static_framework)
+              @installer.install!
+              resources = @project.targets.first.resources_build_phase.files
+              resources.count.should > 0
+              resource = resources.find { |res| res.file_ref.path.include?('Images.xcassets') }
+              resource.should.be.nil
+            end
+
+            it 'adds xcassets resources to the dynamic framework target' do
+              @pod_target.stubs(:build_type => BuildType.dynamic_framework)
+              @installer.install!
+              resources = @project.targets.first.resources_build_phase.files
+              resources.count.should > 0
+              resource = resources.find { |res| res.file_ref.path.include?('Images.xcassets') }
+              resource.should.be.not.nil
+            end
+
             it 'includes spec info_plist entries for dynamic frameworks' do
               @pod_target.stubs(:build_type => BuildType.dynamic_framework)
               expected_entries = {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
-* None.  
+* Fix processing of xcassets resources when pod target is static framework  
+  [Federico Trimboli](https://github.com/fedetrim)
+  [#10175](https://github.com/CocoaPods/CocoaPods/pull/10175)
+  [#10170](https://github.com/CocoaPods/CocoaPods/issues/10170)
 
 
 ## 1.10.0 (2020-10-20)
--- a/lib/cocoapods/target.rb
+++ b/lib/cocoapods/target.rb
@@ -350,7 +350,7 @@ module Pod
     end
 
     def self.resource_extension_compilable?(input_extension)
-      output_extension_for_resource(input_extension) != input_extension
+      output_extension_for_resource(input_extension) != input_extension && input_extension != '.xcassets'
     end
 
     #-------------------------------------------------------------------------#
--- a/spec/unit/target/aggregate_target_spec.rb
+++ b/spec/unit/target/aggregate_target_spec.rb
@@ -199,6 +199,17 @@ module Pod
           resource_paths_by_config['Release'].should == expected_files
         end
 
+        it 'checks xcassets resource paths are not converted for static frameworks' do
+          @pod_target.stubs(:should_build?).returns(true)
+          @pod_target.stubs(:build_type => BuildType.static_framework)
+          @pod_target.stubs(:resource_paths).returns('BananaLib' => ['/some/absolute/path/to/Images.xcassets'])
+          @target.stubs(:bridge_support_file).returns(nil)
+          resource_paths_by_config = @target.resource_paths_by_config
+          expected_files = ['/some/absolute/path/to/Images.xcassets']
+          resource_paths_by_config['Debug'].should == expected_files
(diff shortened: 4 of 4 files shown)


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Fix processing of xcassets resources when pod target is static framework  
  [fedetrim](https://github.com/fedetrim)
```

</p>

igor-makarov:
Hi @fedetrim!

I think the change you're proposing is valid - asset catalogs and static frameworks probably never worked right before.

However, I think the right place to introduce the check is in `Pod::Target#resource_extension_compilable?` - there are two places the function is called and both need this change.

fedetrim:
@igor-makarov thanks for reviewing!

> However, I think the right place to introduce the check is in Pod::Target#resource_extension_compilable? - there are two places the function is called and both need this change.

I had tried that first and it didn't work -- however I just tried again and it did! Probably did something differently the first time.
Anyway, just pushed the change :)

igor-makarov:
@fedetrim excellent - could you please add some tests for this case?

fedetrim:
@igor-makarov done

spec/unit/installer/xcode/pods_project_generator/pod_target_installer_spec.rb
```diff
@@ -762,6 +762,15 @@ class PodsProjectGenerator
...
 
+            it 'doesn\'t add xcassets resources to the static framework target' do
+              @pod_target.stubs(:build_type => BuildType.static_framework)
+              @installer.install!
+              resources = @project.targets.first.resources_build_phase.files
+              resources.count.should > 0
+              resource = resources.find { |res| res.file_ref.path.include?('Images.xcassets') }
+              resource.should.be.nil
+            end
+
```
igor-makarov, line 773: Could you add one more? For the dynamic framework, ensuring it does add the asset catalog?

fedetrim: good idea, done! 😄

dnkoutso:
@fedetrim can you rebase and point to the `1-10-stable` branch instead? We want to ship this with 1.10.1 version.

fedetrim:
@dnkoutso
>can you rebase and point to the 1-10-stable branch instead?

Done 👍
Thanks for reviewing!

dnkoutso (approved):

dnkoutso:
thanks!

Issue #10206, opened 2020-11-12
Cocoapods 1.10.0 generates invalid scripts for static framework


till0xff:
# Report

## What did you do?

I have local pod with static framework (`s.static_framework = true`) which contains resource (xib or storyboard) with dot-separated name (e.g. `View.Suffix.xib`). Cocoapods 1.10.0 generates script `Pods-Bug_Example-resources.sh` with invalid resource name: `View.nib` instead of `View.Suffix.nib`.

With cocoapods 1.9.3 the error is not reproduced.

## What did you expect to happen?

Build successful

## What happened instead?

Build failed with error

```
error: Resource "/Users/till0xff/Library/Developer/Xcode/DerivedData/Bug-cmbxgygdohzlgagwgynzdhhoicou/Build/Products/Debug-iphonesimulator/Bug/Bug.framework/View.nib" not found. Run 'pod install' to update the copy resources script.
```

## Project that demonstrates the issue

 https://github.com/till0xff/cocoapods-1.10.0-bug-example.git

dnkoutso:
@igor-makarov I believe this change https://github.com/CocoaPods/CocoaPods/commit/842dceabaa715df0082502520731c626ac8cf575 caused this.

I think specifically
```ruby
                  built_product_dir.join(File.basename(resource_path, extname)).sub_ext(output_extname).to_s
```

dnkoutso:
fixed for 1.10.1 which will ship soon.

PR #10320, opened 2020-12-31
Fix file extension replacement for resource paths when using static frameworks

dnkoutso:
closes https://github.com/CocoaPods/CocoaPods/issues/10206

### DIFF
--- a/lib/cocoapods/target/aggregate_target.rb
+++ b/lib/cocoapods/target/aggregate_target.rb
@@ -292,7 +292,7 @@ module Pod
                 extname = File.extname(resource_path)
                 if self.class.resource_extension_compilable?(extname)
                   output_extname = self.class.output_extension_for_resource(extname)
-                  built_product_dir.join(File.basename(resource_path, extname)).sub_ext(output_extname).to_s
+                  built_product_dir.join(File.basename(resource_path)).sub_ext(output_extname).to_s
                 else
                   resource_path
                 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
+* Fix file extension replacement for resource paths when using static frameworks.  
+  [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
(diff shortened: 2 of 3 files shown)


lib/cocoapods/target/aggregate_target.rb
```diff
@@ -292,7 +292,7 @@ def resource_paths_by_config
                 extname = File.extname(resource_path)
                 if self.class.resource_extension_compilable?(extname)
                   output_extname = self.class.output_extension_for_resource(extname)
-                  built_product_dir.join(File.basename(resource_path, extname)).sub_ext(output_extname).to_s
```
igor-makarov, line 295: @dnkoutso 
This is actually unnecessary. I realize I've used `sub_ext` incorrectly in my impl.  
The correct use would be: 
```
built_product_dir.join(resource_path).sub_ext(output_extname).to_s
```

lib/cocoapods/target/aggregate_target.rb
```diff
@@ -292,7 +292,7 @@ def resource_paths_by_config
...
+                  built_product_dir.join(change_ext(File.basename(resource_path), output_extname)).to_s
```
igor-makarov, line 295: @dnkoutso 
This is actually unnecessary. I realize I've used `sub_ext` incorrectly in my impl.  
The correct use would be: 
```
built_product_dir.join(resource_path).sub_ext(output_extname).to_s
```

dnkoutso: updating!

dnkoutso: @igor-makarov fixed! the test passes too with your recommendation.

dnkoutso: actually @igor-makarov this is not correct still:

Screen_Shot_2020-12-31_at_11_28_34_AM

dnkoutso: I think we need:

```ruby
                  built_product_dir.join(File.basename(resource_path)).sub_ext(output_extname).to_s
```

dnkoutso: also updated tests to catch this.

igor-makarov: Yes! This is the right one!

igor-makarov (approved):

PR #10346, opened 2021-01-11
Rewrite the only place dependent on `typhoeus`

jasl:
`typhoeus` is the only blocking that prevent CocoaPods running on M1 natively

`cdn_url?` is the only place that using `typhoeus` so I just rewrite it with Ruby's `OpenURI`

WebMock dependency was updated in the `Gemfile.lock` to support testing natively bundled `OpenURI` on newer Ruby versions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [dcvz](https://github.com/dcvz)
   [#9935](https://github.com/CocoaPods/CocoaPods/pull/9935)
 
+* Rewrite the only place dependent on `typhoeus`.  
+  [Jun Jiang](https://github.com/jasl), [Igor Makarov](https://github.com/igor-makarov)
+  [#10346](https://github.com/CocoaPods/CocoaPods/pull/10346)
+
 * Add a `--update-sources` option to `pod repo push` so one can ensure sources are up-to-date. 
   [Elton Gao](https://github.com/gyfelton)
   [Justin Martin](https://github.com/justinseanmartin)
--- a/lib/cocoapods/sources_manager.rb
+++ b/lib/cocoapods/sources_manager.rb
@@ -1,4 +1,5 @@
 require 'cocoapods-core/source'
+require 'cocoapods/open-uri'
 require 'netrc'
 require 'set'
 require 'rest'
@@ -70,15 +71,20 @@ module Pod
       #         The URL of the source.
       #
       def cdn_url?(url)
-        if url =~ %r{^https?:\/\/}
-          require 'typhoeus'
+        return unless url =~ %r{^https?:\/\/}
 
-          response = Typhoeus.get(url + '/CocoaPods-version.yml', :netrc_file => Netrc.default_path, :netrc => :optional)
-          response.code == 200 && begin
-            response_hash = YAML.load(response.body) # rubocop:disable Security/YAMLLoad
-            response_hash.is_a?(Hash) && !Source::Metadata.new(response_hash).latest_cocoapods_version.nil?
-          end
-        end
+        uri_options = {}
+
+        netrc_info = Netrc.read
+        netrc_host = URI.parse(url).host
+        credentials = netrc_info[netrc_host]
+        uri_options[:http_basic_authentication] = credentials if credentials
+
+        response = OpenURI.open_uri(url.chomp('/') + '/CocoaPods-version.yml', uri_options)
+        response_hash = YAML.load(response.read) # rubocop:disable Security/YAMLLoad
+        response_hash.is_a?(Hash) && !Source::Metadata.new(response_hash).latest_cocoapods_version.nil?
+      rescue ::OpenURI::HTTPError, SocketError
+        return false
       rescue => e
         raise Informative, "Couldn't determine repo type for URL: `#{url}`: #{e}"
       end
--- a/spec/unit/sources_manager_spec.rb
+++ b/spec/unit/sources_manager_spec.rb
@@ -1,5 +1,4 @@
 require File.expand_path('../../spec_helper', __FILE__)
-require 'webmock'
 
 def set_up_test_repo_for_update
   set_up_test_repo
@@ -24,12 +23,12 @@ CDN_REPO_RESPONSE = '---
             - 1'.freeze
 
 def stub_url_as_cdn(url)
-  WebMock.stub_request(:get, url + '/CocoaPods-version.yml').
+  WebMock.stub_request(:get, url.chomp('/') + '/CocoaPods-version.yml').
     to_return(:status => 200, :headers => {}, :body => CDN_REPO_RESPONSE)
 end
 
 def stub_as_404(url)
-  WebMock.stub_request(:get, url + '/CocoaPods-version.yml').
+  WebMock.stub_request(:get, url.chomp('/') + '/CocoaPods-version.yml').
     to_return(:status => 404, :headers => {}, :body => '')
 end
 
@@ -105,7 +104,7 @@ module Pod
           end
 
           it 'raises informative exception on network error' do
-            Typhoeus.stubs(:get).with do
+            OpenURI.stubs(:open_uri).with do
               raise StandardError, 'some network error'
             end
             @sources_manager.stubs(:source_with_url).returns(nil)
@@ -162,6 +161,19 @@ module Pod
           @sources_manager.cdn_url?('https://github.com/cocoapods/specs').should == false
         end
 
+        it 'netrc' do
+          WebMock.stub_request(:get, 'https://some_host.com/something/CocoaPods-version.yml').
+            with(:basic_auth => %w[user1 xxx]).
+            to_return(:status => 200, :body => CDN_REPO_RESPONSE)
+
+          netrc_file = temporary_directory + '.netrc'
+          File.open(netrc_file, 'w', 0o600) { |f| f.write("machine some_host.com\nlogin user1\npassword xxx\n") }
+
+          ENV['NETRC'] = temporary_directory.to_s
+          @sources_manager.cdn_url?('https://some_host.com/something').should == true
+          ENV.delete('NETRC')
+        end
+
         it 'fake 200 response' do
           HTML_RESPONSE = '<!doctype html>
           <html>
@@ -178,10 +190,26 @@ module Pod
           @sources_manager.cdn_url?('https://some_host.com/something').should == false
         end
 
-        it 'redirect' do
+        it 'handles trailing slash' do
+          stub_url_as_cdn('http://some_host.com/something')
+          @sources_manager.cdn_url?('http://some_host.com/something').should == true
+          @sources_manager.cdn_url?('http://some_host.com/something/').should == true
+        end
+
+        it 'handles redirects' do
           WebMock.stub_request(:get, 'http://some_host.com/something/CocoaPods-version.yml').
             to_return(:status => 301, :body => '', :headers => { 'Location' => ['http://some_host.com'] })
+          WebMock.stub_request(:get, 'http://some_host.com').
+            to_return(:status => 200, :body => '', :headers => {})
+
           @sources_manager.cdn_url?('http://some_host.com/something').should == false
+
+          WebMock.stub_request(:get, 'http://some_another_host.com/something/CocoaPods-version.yml').
+            to_return(:status => 301, :body => '', :headers => { 'Location' => ['http://some_another_host.com'] })
+          WebMock.stub_request(:get, 'http://some_another_host.com').
+            to_return(:status => 200, :body => CDN_REPO_RESPONSE)
+
+          @sources_manager.cdn_url?('http://some_another_host.com/something').should == true
         end
       end
     end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -155,8 +155,8 @@ GEM
     concurrent-ruby (1.1.7)
     cork (0.3.0)
       colored2 (~> 3.1)
-    crack (0.4.3)
-      safe_yaml (~> 1.0.0)
+    crack (0.4.5)
+      rexml
     danger (5.16.1)
       claide (~> 1.0)
       claide-plugins (>= 0.9.2)
@@ -184,7 +184,7 @@ GEM
     gh_inspector (1.1.3)
     git (1.7.0)
       rchardet (~> 1.8)
-    hashdiff (0.3.1)
+    hashdiff (1.0.1)
     httpclient (2.8.3)
     i18n (1.8.7)
       concurrent-ruby (~> 1.0)
@@ -254,7 +254,6 @@ GEM
     ruby-macho (1.4.0)
     ruby-prof (0.15.2)
     ruby-progressbar (1.10.1)
-    safe_yaml (1.0.4)
     sawyer (0.8.2)
       addressable (>= 2.3.5)
       faraday (> 0.8, < 2.0)
@@ -276,10 +275,10 @@ GEM
     tzinfo (1.2.9)
       thread_safe (~> 0.1)
     unicode-display_width (1.7.0)
-    webmock (2.3.1)
+    webmock (3.11.1)
       addressable (>= 2.3.6)
       crack (>= 0.3.2)
-      hashdiff
+      hashdiff (>= 0.4.0, < 2.0.0)
     yard (0.9.12)
 
 PLATFORMS
--- a/lib/cocoapods/command/spec/lint.rb
+++ b/lib/cocoapods/command/spec/lint.rb
@@ -122,7 +122,7 @@ module Pod
                 output_path = podspecs_tmp_dir + File.basename(path)
                 output_path.dirname.mkpath
                 begin
-                  open(path) do |io|
+                  OpenURI.open_uri(path) do |io|
                     output_path.open('w') { |f| f << io.read }
                   end
                 rescue => e


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* replace typhoeus with open-uri  
  [jasl](https://github.com/jasl)
```

</p>

jasl:
CI green now

jasl:
Add a comment about why `URI.send :open`

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
+* Remove dependency of `typhoeus`
```
igor-makarov, line 11: nit: please add 2 spaces at the end of line

lib/cocoapods/sources_manager.rb
```diff
@@ -70,15 +71,27 @@ def create_source_with_url(url)
...
-            response_hash.is_a?(Hash) && !Source::Metadata.new(response_hash).latest_cocoapods_version.nil?
-          end
-        end
+        return unless url =~ %r{^https?:\/\/}
+
+        opts = {
+          :redirect => false,
+        }
+
+        info = Netrc.read
```
igor-makarov, line 80: Perhaps name `netrc_info`?

lib/cocoapods/sources_manager.rb
```diff
@@ -70,15 +71,27 @@ def create_source_with_url(url)
...
+        netrc_host = URI.parse(url).host
+        creds = info[netrc_host]
```
igor-makarov, line 82: nit: `credentials`

lib/cocoapods/sources_manager.rb
```diff
@@ -70,15 +71,27 @@ def create_source_with_url(url)
...
+        opts[:http_basic_authentication] = creds if creds
```
igor-makarov, line 83: nit: `uri_options`

igor-makarov:
@jasl I've replaced the call to private method to `OpenURI::open_uri` - this is both portable and readable.

I've left some comments as part of the review. 

Also, I believe that if you remove the `Gemfile` and `gemspec` changes, we could merge this sooner.

Removing Typhoeus will require Core update first anyway, as it's a transitive dependency.

jasl:
> @jasl I've replaced the call to private method to `OpenURI::open_uri` - this is both portable and readable.
> 
> I've left some comments as part of the review.
> 
> Also, I believe that if you remove the `Gemfile` and `gemspec` changes, we could merge this sooner.
> 
> Removing Typhoeus will require Core update first anyway, as it's a transitive dependency.

You mean I need to keeping `typhoeus` as CocoaPods dependency? but if `cocoapods-core` dependent `typhoeus`, `cocoapods` should also dependent it indirectly?

in Gemfile.lock you can see

```
GIT
  revision: f4d4f16f0d92f422f15b1f27e2073276098f8f73
  branch: master
  specs:
    cocoapods-core (1.10.1)
      activesupport (> 5.0, < 6)
      addressable (~> 2.6)
      algoliasearch (~> 1.0)
      concurrent-ruby (~> 1.1)
      fuzzy_match (~> 2.0.4)
      nap (~> 1.0)
      netrc (~> 0.11)
      public_suffix
      typhoeus (~> 1.0)
```

so `typhoeus` still on the list

jasl:
@igor-makarov 
applied your suggests, thanks!

igor-makarov:
@jasl Yes - what I mean is that there's no benefit in removing the dependency now, as it will still get installed through Core.

CHANGELOG.md
```diff
@@ -8,10 +8,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
-* Remove dependency of `typhoeus`
+* Remove dependency of `typhoeus`.
   [Jun Jiang](https://github.com/jasl)
   [#10346](https://github.com/CocoaPods/CocoaPods/pull/10346)
-
+
```
igor-makarov, line 14: @jasl I'm sorry I wasn't clear - the two spaces are supposed to be at the end of line 11, not 14 😂

jasl:
> @jasl Yes - what I mean is that there's no benefit in removing the dependency now, as it will still get installed through Core.

I still don't understand why should keeping it in `CocoaPods`... but I just do as your sugguest.

I squash commits, added your credit in Changelog, thanks you for reviewing!

igor-makarov:
Looks good - however there's still a `Gemfile.lock` change.

jasl:
Because we don't remove dependency of `typhoeus`, I change the title and changelog more accurate

jasl:
> @jasl Looks good - however there's still a `Gemfile.lock` change.

I have to upgrade `webmock` because it doesn't support newer Rubies bundled `open-uri`

igor-makarov:
@jasl got it now - I updated the PR description to mention this.

igor-makarov (approved):

jasl:
> @jasl got it now - I updated the PR description to mention this.

Thank you!

lib/cocoapods/sources_manager.rb
```diff
@@ -70,15 +71,27 @@ def create_source_with_url(url)
...
-          response = Typhoeus.get(url + '/CocoaPods-version.yml', :netrc_file => Netrc.default_path, :netrc => :optional)
-          response.code == 200 && begin
-            response_hash = YAML.load(response.body) # rubocop:disable Security/YAMLLoad
-            response_hash.is_a?(Hash) && !Source::Metadata.new(response_hash).latest_cocoapods_version.nil?
-          end
-        end
+        return unless url =~ %r{^https?:\/\/}
+
+        uri_options = {
+          :redirect => false,
```
segiddins, line 77: out of curiosity, why disable redirects?

lib/cocoapods/sources_manager.rb
```diff
@@ -70,15 +71,27 @@ def create_source_with_url(url)
...
+        netrc_info = Netrc.read
+        netrc_host = URI.parse(url).host
+        credentials = netrc_info[netrc_host]
+        uri_options[:http_basic_authentication] = credentials if credentials
+
+        response = OpenURI.open_uri(URI.join(url, 'CocoaPods-version.yml'), uri_options)
+
+        response_hash = YAML.load(response.read) # rubocop:disable Security/YAMLLoad
+        response_hash.is_a?(Hash) && !Source::Metadata.new(response_hash).latest_cocoapods_version.nil?
+      rescue ::OpenURI::HTTPError
```
segiddins, line 89: can rescue the three error classes in the same rescue statement

segiddins (approved):

lib/cocoapods/sources_manager.rb
jasl: the previous implemention

```ruby
response.code == 200 && begin
            response_hash = YAML.load(response.body) # rubocop:disable Security/YAMLLoad
            response_hash.is_a?(Hash) && !Source::Metadata.new(response_hash).latest_cocoapods_version.nil?
          end
```

and there's a test https://github.com/CocoaPods/CocoaPods/blob/master/spec/unit/sources_manager_spec.rb#L181-L185

I just keeping the same behavior

lib/cocoapods/sources_manager.rb
jasl: will do

lib/cocoapods/sources_manager.rb
amorde: the linked test:
```
        it 'redirect' do
          WebMock.stub_request(:get, 'http://some_host.com/something/CocoaPods-version.yml').
            to_return(:status => 301, :body => '', :headers => { 'Location' => ['http://some_host.com'] })
          @sources_manager.cdn_url?('http://some_host.com/something').should == false
        end
```

Isn't testing that redirects are _not_ followed. It's testing that if `some_host.com/something/CocoaPods-version.yml` redirects to `some_host.com/`, then the source in question isn't a CDN. I might be misunderstanding this, but I don't think we want to disable redirects

jasl: > It's testing that if some_host.com/something/CocoaPods-version.yml redirects to some_host.com/, then the source in question isn't a CDN.

In my understanding, `cdn_url?` attempting to get `"#{given_url}/CocoaPods-version.yml"` and trying to parse it, then it returns true.

base on your explain, you mean `cdn_url?` getting `http://some_host.com/something/CocoaPods-version.yml` through redirection returns `http://some_host.com/something`, in the test, it's not mocked, so maybe returns `SocketError`, so it's not a `cdn_url` then returns `false`

is this what your means?

jasl: I'm not consider the logic before... my mistake, your explain sounds reasonable to me

jasl: @amorde @segiddins 
Sorry, that's my mistake, I misunderstanding the meaning of that test,

I've corrected it

amorde: Apologies on the unclear wording. I think that test is actually relying on a redirect, and expects redirects to be supported and not disabled. I will defer to @igor-makarov to confirm the intent behind the test

jasl: I enable redirects now, looks CI green.

jasl:
@igor-makarov

jasl:
@igor-makarov 
I pushed some changes:

- Re-enable redirects, see conversation above, that's my mistake
- Revert your change `URI.join(url, "CocoaPods-version.yml")` becasue
  ```
  URI.join("http://some_host.com/something", "CocoaPods-version.yml")
   => #<URI::HTTP http://some_host.com/CocoaPods-version.yml>
  ```
- Add an unit test about `netrc`

spec/unit/sources_manager_spec.rb
```diff
@@ -181,6 +193,8 @@ module Pod
         it 'redirect' do
           WebMock.stub_request(:get, 'http://some_host.com/something/CocoaPods-version.yml').
             to_return(:status => 301, :body => '', :headers => { 'Location' => ['http://some_host.com'] })
+          WebMock.stub_request(:get, 'http://some_host.com').
+            to_return(:status => 200, :body => '', :headers => {})
```
jasl, line 197: I think this is needed, or `open-uri` will trying to request to the Internet,
can proof this by `WebMock.disable_net_connect!`

lib/cocoapods/sources_manager.rb
igor-makarov: Yes, this is the test case. Perhaps it could be clearer worded. 🤔

spec/unit/sources_manager_spec.rb
igor-makarov: Agreed.

igor-makarov:
Regarding `URI#join`, you are correct about this and I've found the [path vs. endpoint](https://apidock.com/ruby/URI/join/class#1245-Be-careful-with-path-vs-endpoint) warning on APIdock comments.

Thus, I believe that if you use 
```
URI.join(url + '/', "CocoaPods-version.yml")
```

- then we'll get the behavior we want with the slashes on or off. Might wanna add a test for it though.

jasl:
I just add more tests for tailing `/` and `handle redirects`

I have to leave about 2 hours, when I back to office, I'll try `URI.join(url + '/', "CocoaPods-version.yml")`

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 * Add a post_integrate_hook API  
   [dcvz](https://github.com/dcvz)
   [#9935](https://github.com/CocoaPods/CocoaPods/pull/9935)
+* Rewrite the only place dependent on `typhoeus`.
```
dnkoutso, line 14: nit missing empty line above this.

jasl:
> Regarding `URI#join`, you are correct about this and I've found the [path vs. endpoint](https://apidock.com/ruby/URI/join/class#1245-Be-careful-with-path-vs-endpoint) warning on APIdock comments.
> 
> Thus, I believe that if you use
> 
> ```
> URI.join(url + '/', "CocoaPods-version.yml")
> ```
> 
> then we'll get the behavior we want with the slashes on or off. Might wanna add a test for it though.
> 
> The reason I'm hung up on this is because people tend to sometimes add `source "https://cdn.cocoapods.org"` and sometimes `source "https://cdn.cocoapods.org/"` and we don't want either one to fail.

Applied this suggest, I've add tests for with or without tailing `/`

jasl:
```
LoadError: dlopen(/Users/runner/work/CocoaPods/CocoaPods/vendor/bundle/gems/bigdecimal-1.3.5/lib/bigdecimal.bundle, 9): Library not loaded: /usr/local/opt/ruby/lib/libruby.2.7.dylib
  Referenced from: /Users/runner/work/CocoaPods/CocoaPods/vendor/bundle/gems/bigdecimal-1.3.5/lib/bigdecimal.bundle
```

Not the code problem, don't know how to restart it, but I believe it should no problem on Ruby 2.7

igor-makarov:
@jasl @dnkoutso I think I know what's up. I've decided to add macOS 11 to the system test matrix #10350 and while adding it, I realized that the bundler cache keys aren't differentiated by OS version 🤦🏻‍♂️ but not before probably clobbering the bundler cache with newer native gems.

I think this PR will need to be rebased on main after #10350 is merged. Also, this means that we'll need to upstream #10350 into stable branch in case there would be PRs there.

jasl:
> @jasl @dnkoutso I think I know what's up. I've decided to add macOS 11 to the system test matrix #10350 and while adding it, I realized that the bundler cache keys aren't differentiated by OS version 🤦🏻‍♂️ but not before probably clobbering the bundler cache with newer native gems.
> 
> I think this PR will need to be rebased on main after #10350 is merged. Also, this means that we'll need to upstream #10350 into stable branch in case there would be PRs there.



> @jasl @dnkoutso I think I know what's up. I've decided to add macOS 11 to the system test matrix #10350 and while adding it, I realized that the bundler cache keys aren't differentiated by OS version 🤦🏻‍♂️ but not before probably clobbering the bundler cache with newer native gems.
> 
> I think this PR will need to be rebased on main after #10350 is merged. Also, this means that we'll need to upstream #10350 into stable branch in case there would be PRs there.

Could you help to take a look https://github.com/CocoaPods/CocoaPods/pull/10347 ?

I'll waiting #10350 merge, now I need to find a safe way for URI appending, `URI.join(url + '/', "CocoaPods-version.yml")` this I believe it should work in real-word but not work for `webmock`

igor-makarov:
I've looked at  #10347 and it looks good to me, but the custom JSON gem happened before my time, so I am waiting for more relevant people to take a look at it.

Regarding  #10350, I'm no longer certain that it's the cause for the failures. Investigating.

igor-makarov:
We're gonna put adding Big Sur on hold for now, the PR that's blocking this PR is #10351 - which (hopefully) fixes the CI problems.

jasl:
Rebased, waiting CI

igor-makarov:
@igorudovika can you take a look at the changes to see if the NetRC stuff looks good?

lib/cocoapods/sources_manager.rb
```diff
@@ -80,8 +80,7 @@ def cdn_url?(url)
         credentials = netrc_info[netrc_host]
         uri_options[:http_basic_authentication] = credentials if credentials
 
-        response = OpenURI.open_uri(URI.join(url + '/', 'CocoaPods-version.yml'), uri_options)
-
+        response = OpenURI.open_uri(URI.join(url.chomp('/') + '/', 'CocoaPods-version.yml'), uri_options)
```
jasl, line 83: `URI.join(url.chomp('/') + '/', 'CocoaPods-version.yml')` why not just `url.chomp('/') + '/', 'CocoaPods-version.yml'`?

igor-makarov: @jasl you're right, using `URI.join` seems redundant and cumbersome.

jasl: changed

spec/unit/sources_manager_spec.rb
```diff
@@ -24,12 +23,12 @@ def set_up_test_repo_for_update
...
 
 def stub_url_as_cdn(url)
-  WebMock.stub_request(:get, url + '/CocoaPods-version.yml').
+  WebMock.stub_request(:get, url.chomp('/') + '/CocoaPods-version.yml').
     to_return(:status => 200, :headers => {}, :body => CDN_REPO_RESPONSE)
 end
 
 def stub_as_404(url)
-  WebMock.stub_request(:get, url + '/CocoaPods-version.yml').
+  WebMock.stub_request(:get, URI.join(url.chomp('/') + '/', 'CocoaPods-version.yml')).
```
igor-makarov, line 31: @jasl probably should be as above

khomchyk:
Hello!
Can we, please, speed up the review of this PR? 
Seems that [#10351](https://github.com/CocoaPods/CocoaPods/pull/10351) has been already merged.

igor-makarov:
@khomchyk CocoaPods is maintained by volunteers in their free time. There's no concept of "speeding up" when people volunteer.

That said, unblocking native M1 support might take a while as the current PR only covers one of the gems. There's a much bigger PR [here](https://github.com/CocoaPods/Core/pull/669) that will also need to be merged and it may take significant time. 

If you wanna use CocoaPods on M1 right now, please refer to #9907 for workarounds.

igorudovika:
> @igorudovika can you take a look at the changes to see if the NetRC stuff looks good?

Sorry for late response. Sure, will look nearest time.

igor-makarov:
@dnkoutso this can merged - I believe you've gone over it already?

igor-makarov:
@jasl could you rebase please?

jasl:
> @jasl could you rebase please?

Done.

dnkoutso (approved):

im-ttmskk:
I am very much looking forward to the release of this contribution.

PR #10350, opened 2021-01-14
Add macOS 11.0 Big Sur to the test suite

igor-makarov:


igor-makarov:
Apparently [Big Sur jobs stall a lot](https://github.com/actions/virtual-environments/issues/1814#issuecomment-759332761) so this will have to wait, closing it for now.

PR #10351, opened 2021-01-14
Fix cache keys to reflect OS version

igor-makarov:
* Fix cache keys for Linux builds
* Remove caching for macOS builds as the cached gems don't work on subsequent builds

### DIFF
--- a/.github/workflows/Lint.yml
+++ b/.github/workflows/Lint.yml
@@ -2,7 +2,14 @@ name: Lint
 
 jobs:
   lint:
-    runs-on: ubuntu-16.04
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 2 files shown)


jasl:
Should CocaPods/Core also need this?

igor-makarov:
@jasl possibly, but I'd want to make sure the issue is resolved before moving ahead with that.

amorde (approved):

segiddins (approved):

dnkoutso (approved):

igor-makarov:
BTW @jasl I've checked and Core doesn't have CI caching for gems - looks like `bundle install` takes much less time in that repo because there's much less of them - so no change needed there!

PR #10576, opened 2021-04-12
Bump packages required for Apple Silicon support

LinusU:
With these modifications I've successfully used CocoaPods natively on my M1 MacBook 🎉 

Do note that it doesn't work with the _built-in_ version of Ruby (which is deprecated by Apple). To solve this I simply installed Ruby from Homebrew. More info about this can be found here: [Ruby on Apple Silicon M1 Macs](https://betterprogramming.pub/ruby-on-apple-silicon-m1-macs-fb159849b2f5)


segiddins (approved):

orta:
This PR shouldn't affect the version of CocoaPods which someone consumes to my knowledge, only the dependencies for this repo, so I'm not sure the title makes sense in context

LinusU:
@orta I think that if you clone the repo/download a specific release tarball and run `gem install` (which is what e.g. [Homebrew does](https://github.com/Homebrew/homebrew-core/blob/d517a99802b27e9c3538c540c1a0c199f3b4d72e/Formula/cocoapods.rb#L29)) it will use this lock file? 🤔

orta:
I don' t _think it will_, no, the lockfile is used for bundler not Rubygems - they are connected but not the same. Segiddins would probably know and he did 👍🏻 this

That said, a quick look through the install tests seems to maybe imply it does: https://github.com/rubygems/rubygems/blob/a969c0f16aa56cdb69e2a051f2b093427010e1d9/test/rubygems/test_gem_request_set_lockfile_parser.rb

segiddins:
This only effects people using Bundler in the CocoaPods repo

igor-makarov:
@orta @segiddins  Additionally, it seems that system Ruby isn't recoverable at this point ([link](https://betterprogramming.pub/ruby-on-apple-silicon-m1-macs-fb159849b2f5)), and a lot of our users use system Ruby.

PR #10609, opened 2021-04-26
Add version option to pod spec cat and pod spec which

pietbrauer:
This PR adds a new option to `pod spec cat` and `pod spec which` if someone might not need the latest pod spec but the one for a specific version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [cltnschlosser](https://github.com/cltnschlosser)
   [#9916](https://github.com/CocoaPods/CocoaPods/pull/9916)
 
+* Add a `--version` option to `pod spec cat` for listing the podspec of a specific version  
+  [pietbrauer](https://github.com/pietbrauer)
+  [#10609](https://github.com/CocoaPods/CocoaPods/pull/10609)
+
 ##### Bug Fixes
 
 * Respect `--configuration` option when analyzing via `pod lib lint --analyze`. 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,7 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [cltnschlosser](https://github.com/cltnschlosser)
   [#9916](https://github.com/CocoaPods/CocoaPods/pull/9916)
 
-* Add a `--version` option to `pod spec cat` for listing the podspec of a specific version  
+* Add a `--version` option to `pod spec cat` and `pod spec which` for listing the podspec of a specific version  
   [pietbrauer](https://github.com/pietbrauer)
   [#10609](https://github.com/CocoaPods/CocoaPods/pull/10609)
 
--- a/lib/cocoapods/command/spec.rb
+++ b/lib/cocoapods/command/spec.rb
@@ -33,13 +33,14 @@ module Pod
       # @param  [String] spec
       #         The name of the specification.
       #
-      # @param  [Bool] show_all
-      #         Whether the paths for all the versions should be returned or
-      #         only the one for the last version.
+      # @param  [Bool,String] version_filter
+      #         - If set to false, will return only the spec path for the latest version (the default).
+      #         - If set to true, will return a list of all paths of all the versions of that spec.
+      #         - If set to a String, will return only the spec path for the version specified by that string.
       #
       # @return [Pathname] the absolute path or paths of the given podspec
       #
-      def get_path_of_spec(spec, show_all = false)
+      def get_path_of_spec(spec, version_filter = false)
         sets = config.sources_manager.search_by_name(spec)
 
         if sets.count == 1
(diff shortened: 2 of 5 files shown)


lib/cocoapods/command/spec/cat.rb
```diff
@@ -39,6 +41,11 @@ def run
                        specs = get_path_of_spec(query, @show_all).split(/\n/)
                        index = UI.choose_from_array(specs, "Which spec would you like to print [1-#{specs.count}]? ")
                        specs[index]
+                     elsif @version
+                       specs = get_path_of_spec(query, true).split(/\n/)
+                       matching_spec = specs.select { |p| %r{(#{@version}\/)} =~ p }.first
```
AliSoftware, line 46: This will also match version 12.3.4 if the requested version is 2.3.4…

Imho it would be better to adapt the get_path_for_spec method so that its 2nd parameter can accept a string as well as a boolean, and if it is a string, only return the path for that version.

Would feel safer than relying on a regex on the resulting path, especially since that makes assumptions on how the path is formed, and if the convention on this changes in a future version of CocoaPods (like it did when we sharded the spec repo a couple of years ago) that would break.

lib/cocoapods/command/spec.rb
```diff
@@ -37,9 +37,12 @@ def validate_regex!(query)
       #         Whether the paths for all the versions should be returned or
       #         only the one for the last version.
       #
+      # @param  [String] version
+      #         Optional version to search for in the specs.
+      #
       # @return [Pathname] the absolute path or paths of the given podspec
       #
-      def get_path_of_spec(spec, show_all = false)
+      def get_path_of_spec(spec, show_all = false, version = nil)
```
AliSoftware, line 45: This odd thing with this is that `show_all` and `version` are exclusive. If `version` is specified, `show_all` will be ignored, and vice-versa. Especially it's not clear just from the method signature and doc what would happen with `show_all = true, version = '1.2.3'` 

I see 2 approaches to solve this:
 - Either clearly document which one takes precedence. In other words, in the documentation for `show_all`, add a mention that "this parameter is ignored if a non-nil `version` is provided"
 - Or combine the two parameters into one, which imho is the better and more idiomatic to ruby. Since Ruby is dynamically typed, it's totally valid to allow a parameter to be either a bool or a String. This would give something like this:

```ruby
# @param. [String] spec
#         The name of the specification.
#
# @param  [Bool,String] version_filter
#         - If set to false, will return only the spec path for the latest version (the default).
#         - If set to true, will return a list of all paths of all the versions of that spec.
#         - If set to a String, will return only the spec path for the version specified by that string.
# 
# @return [Pathname, Array<Pathname>] the absolute path or paths of the given podspec
#
def get_path_of_spec(spec, version_filter)
  …
  if version_filter.is_a?(String)
    all_paths_from_set(set, version_filter).split(/\n/).first
  elsif version_filter == true
    all_paths_from_set(set) # all
  else # latest
    …
  end
end
```

AliSoftware: [Bonus] To be completely honest, an even nicer API (and more idiomatic of ruby) would be:
 - To have `version_filter` be either a `String` or a `Symbol` (rather than "`String` or `Bool`), and use the symbol `:all` instead of value `true` to mean "all versions", and the symbol `:latest` instead of value `false` to mean "latest version only".
 - I also wonder why `all_paths_from_set` returns a multi-line string that the caller (here `get_path_of_spec`) needs to `.split(/\n/)`, instead of have `all_paths_from_set` be the one returning an `Array<String>` itself

But both those changes, however nice they would be by making the API of this method way more idiomatic, would require updating _all_ the call sites as well, which means a broader change and larger PR. So if you don't feel up for this, I can totally understand, and that's why the intermediate solution above (have `version_filter` be a "`String` or `Bool`" and not a "`String` or `Symbol`") would still be fine, as it would still keep retro-compatibility with existing code without needing to change anything in the rest of the codebase.

lib/cocoapods/command/spec.rb
```diff
@@ -67,14 +72,19 @@ def pathname_from_spec(spec, _source)
...
         paths = ''
 
         sources = set.sources
 
         sources.each do |source|
           versions = source.versions(set.name)
 
+          if specific_version
+            versions = versions.select { |v| v.version == specific_version }
+            raise Informative, "No #{set.name} spec in #{specific_version} found." if versions.empty?
```
AliSoftware, line 85:
```suggestion
            raise Informative, "Can't find version #{specific_version} for spec #{set.name}." if versions.empty?
```

AliSoftware (commented):
That's way better 👍  Thanks for tackling this and adding this feature! 🙌 

Left a couple of other suggestions to improve it a bit more. 🙂

CHANGELOG.md
```diff
@@ -36,6 +36,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [cltnschlosser](https://github.com/cltnschlosser)
   [#9916](https://github.com/CocoaPods/CocoaPods/pull/9916)
 
+* Add a `--version` option to `pod spec cat` for listing the podspec of a specific version  
+  [pietbrauer](https://github.com/pietbrauer)
+
```
AliSoftware, line 41:
```suggestion
  [pietbrauer](https://github.com/pietbrauer)
  [#10609](https://github.com/CocoaPods/CocoaPods/pull/10609)

```

AliSoftware, line 41:
```suggestion
* Add a `--version` option to `pod spec cat` for listing the podspec of a specific version.  
  [pietbrauer](https://github.com/pietbrauer)
  [#10609](https://github.com/CocoaPods/CocoaPods/pull/10609)

```

lib/cocoapods/command/spec.rb
pietbrauer: Makes total sense, I had it like you suggested but then it was a huge mess and I reverted half of it and used the two arguments. Now that I have a clearer picture of what I am doing I chose the `version_filter` option.

I converted `all_paths_from_set` to return an array and modified the callers.

AliSoftware (approved):
LGTM 👍 thanks for your awesome contribution!

PR #10795, opened 2021-07-21
Integrate test specs and app specs of pre-built pods.

dnkoutso:
Previously, CocoaPods would not integrate test specs or app specs of pre-built targets. We used to return early. This updates the logic to still integrate test specs and app specs even if the target has no sources to compile.

### DIFF
--- a/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
@@ -43,18 +43,14 @@ module Pod
               test_file_accessors = target.file_accessors.select { |fa| fa.spec.test_specification? }
               app_file_accessors = target.file_accessors.select { |fa| fa.spec.app_specification? }
 
-              unless target.should_build?
-                # For targets that should not be built (e.g. pre-built vendored frameworks etc), we add a placeholder
-                # PBXAggregateTarget that will be used to wire up dependencies later.
-                native_target = add_placeholder_target
-                resource_bundle_targets = add_resources_bundle_targets(library_file_accessors).values.flatten
-                create_copy_dsyms_script
-                create_copy_xcframeworks_script unless target.xcframeworks.values.all?(&:empty?)
-                create_xcconfig_file(native_target, resource_bundle_targets)
-                return TargetInstallationResult.new(target, native_target, resource_bundle_targets)
-              end
+              native_target = if target.should_build?
+                                add_target
+                              else
+                                # For targets that should not be built (e.g. pre-built vendored frameworks etc), we add a placeholder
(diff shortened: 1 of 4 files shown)


lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -64,8 +60,10 @@ def install!
               app_native_targets = add_app_targets
               app_resource_bundle_targets = add_resources_bundle_targets(app_file_accessors)
 
-              add_files_to_build_phases(native_target, test_native_targets, app_native_targets)
-              validate_targets_contain_sources(test_native_targets + app_native_targets.values + [native_target])
+              if target.should_build?
+                add_files_to_build_phases(native_target, test_native_targets, app_native_targets)
```
dnkoutso, line 64: oops this is wrong. This does not add the source files of test targets and app targets. Will need to fix.

dnkoutso: fixed, ready now.

dnkoutso: also updated tests to catch this.

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -987,7 +990,7 @@ def module_map_additional_headers
           end
 
           def create_umbrella_header(native_target)
-            return super(native_target) unless custom_module_map
+            super(native_target) unless custom_module_map
```
dnkoutso, line 993: IDE warning fix.

lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -311,19 +310,23 @@ def filter_resource_file_references(resource_file_references)
           #
           # @return [void]
           #
-          def add_files_to_build_phases(native_target, test_native_targets, app_native_targets)
+          def add_files_to_build_phases(library_native_target, test_native_targets, app_native_targets)
```
dnkoutso, line 313: this is another IDE warning fix. the param is the same as a local variable below.

lib/cocoapods/user_interface.rb
```diff
@@ -137,6 +137,10 @@ def title(title, verbose_prefix = '', relative_indentation = 2)
       #        The indentation level relative to the current,
       #        when the message is printed.
       #
+      # @yield  The action, this block is always executed.
```
dnkoutso, line 140: IDE warning fix

orta (approved):
Sounds pretty solid to me

Issue #10373, opened 2021-01-21
"Multiple commands produce" error in "[CP] Copy dSYMs" phase


acecilia:
* [x] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?
Make a podspec with more than one binary framework, and try to build the resulting target

## What did you expect to happen?
The target builds without issues

## What happened instead?

```
Build system information
error: Multiple commands produce '/Users/andres.luque/Library/Developer/Xcode/DerivedData/TestProject-aherercskthrkyduwuqzcfkuhrig/Build/Products/Debug-maccatalyst/Binaries/1A3DADB5-A46D-3DB6-B372-9EC26653046F.bcsymbolmap':
1) That command depends on command in Target 'Binaries' (project 'Pods'): script phase “[CP] Copy dSYMs”
2) That command depends on command in Target 'Binaries' (project 'Pods'): script phase “[CP] Copy dSYMs”


Build system information
error: Multiple commands produce '/Users/andres.luque/Library/Developer/Xcode/DerivedData/TestProject-aherercskthrkyduwuqzcfkuhrig/Build/Products/Debug-maccatalyst/Binaries/79247F36-4886-3FFB-BAF9-89BCDCFC553F.bcsymbolmap':
1) That command depends on command in Target 'Binaries' (project 'Pods'): script phase “[CP] Copy dSYMs”
2) That command depends on command in Target 'Binaries' (project 'Pods'): script phase “[CP] Copy dSYMs”


Build system information
error: Multiple commands produce '/Users/andres.luque/Library/Developer/Xcode/DerivedData/TestProject-aherercskthrkyduwuqzcfkuhrig/Build/Products/Debug-maccatalyst/Binaries/77A04A36-9A16-39FD-B9C0-94FC2AFB0E41.bcsymbolmap':
1) That command depends on command in Target 'Binaries' (project 'Pods'): script phase “[CP] Copy dSYMs”
2) That command depends on command in Target 'Binaries' (project 'Pods'): script phase “[CP] Copy dSYMs”


Build system information
[...]

PR #10800, opened 2021-07-22
Dedup bcsymbolmap paths found from multiple vendored frameworks.

dnkoutso:
closes https://github.com/CocoaPods/CocoaPods/issues/10373

when a target provides multiple vendored frameworks under the same root folder then we always scan and include all bcsymbolmap paths for all frameworks. This ensures we dedup them before we install them in the copy dSYM script phase.

### DIFF
--- a/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
+++ b/lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
@@ -1199,7 +1199,7 @@ module Pod
             def bcsymbolmap_paths(target)
               target.framework_paths.values.flatten.reject do |fmwk_path|
                 fmwk_path.bcsymbolmap_paths.nil?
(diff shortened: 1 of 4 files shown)


lib/cocoapods/installer/xcode/pods_project_generator/pod_target_installer.rb
```diff
@@ -1199,7 +1199,7 @@ def dsym_paths(target)
             def bcsymbolmap_paths(target)
               target.framework_paths.values.flatten.reject do |fmwk_path|
                 fmwk_path.bcsymbolmap_paths.nil?
-              end.flat_map(&:bcsymbolmap_paths)
+              end.flat_map(&:bcsymbolmap_paths).uniq
```
dnkoutso, line 1202: actual fix.

orta (approved):

Issue #9606, opened 2020-03-11
[Feature Request] On-Demand Resource Support


Damonvvong:
* [X] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# On-Demand Resource

On-demand resources are app contents that are hosted on the App Store and are separate from the related app bundle that you download. They enable smaller app bundles, faster downloads, and richer app content. The app requests sets of on-demand resources, and the operating system manages downloading and storage. The app uses the resources and then releases the request. After downloading, the resources may stay on the device through multiple launch cycles, making access even faster.

More infomation: https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/On_Demand_Resources_Guide/index.html

**But It is not support by Cocoapods.** 

Here are same problem at stackoverflow: https://stackoverflow.com/questions/51901563/ios-on-demand-resources-from-a-framework-or-cocoapod

# *.xcodeproj File Diff

if we want use **ODR**, `*.xcodeproj ` file must set options like this: 

## FileRef

```diff

-   25C339XXXXXXXXXXXXXXX2FD9 /* Test_ODR.bundle in Resources */ = {isa = PBXBuildFile; fileRef = 25C339XXXXXXXXXXXXXXX2FD9 /* Test_ODR.bundle */; };
+   25C339XXXXXXXXXXXXXXX2FD9 /* Test_ODR.bundle in Resources */ = {isa = PBXBuildFile; fileRef = 25C339XXXXXXXXXXXXXXX2FD9 /* Test_ODR.bundle */; settings = {ASSET_TAGS = (Tag, ); }; };

```

## Section

```diff
/* Begin PBXProject section */
    25C339XXXXXXXXXXXXXXX2FD9 /* Project object */ = {
      isa = PBXProject;
      attributes = {
+       KnownAssetTags = (
+         Tag,
+         Tag1,
+         Tag2,
+       );
```

# Question

no matter 

```ruby

s.resources    = 'ODRPodDemo/Assets/**/*.bundle'

```
and

```ruby

s.resource_bundles = {
    'ODRPodDemo' => ['ODRPodDemo/Assets/*.png']
}

```

did not support

# Request

whether there are some possible support `Pods`'s resource can use On-Demand Resource when `pod install` or `pod update` or not? Like this:

```ruby

s.on_demand_resources  = {
[...]

PR #10812, opened 2021-07-27
Add support for integrating on demand resources.

dnkoutso:
closes https://github.com/CocoaPods/CocoaPods/issues/9606

requires: https://github.com/CocoaPods/Core/pull/690 and https://github.com/CocoaPods/Xcodeproj/pull/844


lib/cocoapods/installer/user_project_integrator/target_integrator.rb
```diff
@@ -631,6 +632,71 @@ def remove_obsolete_script_phases(removed_phase_names = REMOVED_SCRIPT_PHASE_NAM
...
+                target.user_targets.each do |user_target|
+                  user_target.resources_build_phase.remove_file_reference(ref)
+                end
+                ref.remove_from_project
+              end
+              target_odr_group.recursive_children_groups.each { |g| g.remove_from_project if g.empty? }
+            end
+          end
+          unless asset_tags_added == known_asset_tags
+            attributes["KnownAssetTags"] = asset_tags_added.to_a
```
dnkoutso, line 695: this is wrong. we should not set this here because its possible we might remove other asset tags managed by the user and outside of cocoapods.

We should just ensure the ones we add are present.

cocoapods.gemspec
```diff
@@ -37,7 +37,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'cocoapods-trunk',       '>= 1.4.0', '< 2.0'
   s.add_runtime_dependency 'cocoapods-try',         '>= 1.1.0', '< 2.0'
   s.add_runtime_dependency 'molinillo',             '~> 0.7.0'
-  s.add_runtime_dependency 'xcodeproj',             '>= 1.19.0', '< 2.0'
+  s.add_runtime_dependency 'xcodeproj',             '>= 1.20.0', '< 2.0'
```
dnkoutso, line 40: this will become 1.21.0 when we ship it.

lib/cocoapods/installer/user_project_integrator/target_integrator.rb
```diff
@@ -631,6 +632,71 @@ def remove_obsolete_script_phases(removed_phase_names = REMOVED_SCRIPT_PHASE_NAM
...
+                ref.remove_from_project
+              end
+              target_odr_group.recursive_children_groups.each { |g| g.remove_from_project if g.empty? }
+            end
+          end
+
+          unless asset_tags_added.empty?
+            attributes = user_project.root_object.attributes
+            attributes['KnownAssetTags'] = (attributes['KnownAssetTags'] ||= []) | asset_tags_added.to_a
+            target.user_project.root_object.attributes = attributes
```
dnkoutso, line 696: we never remove known asset tags because even though we might have added them we cant confirm that. We only add the new ones we know we added.

lib/cocoapods/installer/project_cache/target_cache_key.rb
```diff
@@ -161,8 +161,8 @@ def self.from_aggregate_target(sandbox, aggregate_target)
           contents = {
             'BUILD_SETTINGS_CHECKSUM' => build_settings,
           }
-          if aggregate_target.includes_resources?
-            relative_file_paths = aggregate_target.resource_paths_by_config.values.flatten.uniq
+          if aggregate_target.includes_resources? || aggregate_target.includes_on_demand_resources?
```
dnkoutso, line 164: this ensures incremental installation works by accounting for on demand resource file references.

If they change then the aggregate target is marked as dirty and is re-integrated, otherwise it is skipped.

dnkoutso: tests were added below.

lib/cocoapods/installer/user_project_integrator/target_integrator.rb
```diff
@@ -631,6 +632,71 @@ def remove_obsolete_script_phases(removed_phase_names = REMOVED_SCRIPT_PHASE_NAM
           end
         end
 
+        # Updates all user targets to include on demand resources specified by libraries. Note that currently,
+        # only app level targets are allowed to include on demand resources.
+        #
+        # @return [void]
+        #
+        def add_on_demand_resources
```
dnkoutso, line 640: this method looks a bit complicated but its not. Its trying relatively to avoid adding duplicate file references.

OnDemandResources of a target are added to the users target because they need to be part of the resources_build_phase and also included in the "Resource Tags".

Note as of right now ODRs can only be used by app targets so if a pod specifies ODR paths then those are integrated into the target the pod is integrated with.

paulb777 (approved):

Issue #10106, opened 2020-09-30
Multiple commands produce error with XCFramework 


TheCodeTalker:
* [x] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

Hello we started to use xcframework recently in our project . I am having an issue with Cocoapods where we have library `IVTNetworking` is generated with xcframework and trying to integrate with pod as dependency to other library after pod install while build we see and error like 

``` 
Multiple commands produce /Users/chitaranjans/Library/Developer/Xcode/DerivedData/VTNetworkEngine-divuxvvhxfzuclfkagrswscrlgpm/Build/Products/Debug-iphoneos/XCFrameworkIntermediates/IVTNetworking  :
1) That command depends on command in Target 'IVTNetworking_Debug' (project 'Pods'): script phase “[CP] Copy XCFrameworks”
2) That command depends on command in Target 'IVTNetworking_Release' (project 'Pods'): script phase “[CP] Copy XCFrameworks”
 ```
<img width="1063" alt="Screenshot 2020-09-30 at 10 17 31 PM" src="https://user-images.githubusercontent.com/3683673/94715201-c1563080-036a-11eb-84b2-211ea457ad7d.png">

## What did you expect to happen?
As you build it should copy respective xcframework like (debug and release ) xcframework

## What happened instead?
As you seen in above screenshot getting same error  

### Stack

```
   CocoaPods : 1.10.0.beta.2
        Ruby : ruby 2.6.3p62 (2019-04-16 revision 67580) [universal.x86_64-darwin19]
    RubyGems : 3.0.3
        Host : Mac OS X 10.15.6 (19G2021)
       Xcode : 11.3.1 (11C505)
         Git : git version 2.21.1 (Apple Git-122.3)
Ruby lib dir : /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib
Repositories : cocoapods - git - https://github.com/CocoaPods/Specs.git @ fb84a8148a93b4ea7997d40e7bb3107e7451ba94

               trunk - CDN - https://cdn.cocoapods.org/
```

### Installation Source

```
Executable Path: /usr/local/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.4
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
[...]

PR #10826, opened 2021-07-30
Place frameworks from xcframeworks into a unique folder name to avoid duplicate outputs.

dnkoutso:
closes https://github.com/CocoaPods/CocoaPods/issues/10106

Screen_Shot_2021-07-30_at_8_33_40_PM

integration specs: https://github.com/CocoaPods/cocoapods-integration-specs/pull/319

Verified project https://github.com/CocoaPods/CocoaPods/issues/10106 successfully compiled now.

### DIFF
--- a/lib/cocoapods/generator/embed_frameworks_script.rb
+++ b/lib/cocoapods/generator/embed_frameworks_script.rb
@@ -159,8 +159,9 @@ code_sign_if_enabled() {
         end
         xcframeworks_by_config.each do |config, xcframeworks|
           xcframeworks.select { |xcf| xcf.build_type.dynamic_framework? }.each do |xcframework|
+            target_name = xcframework.target_name
             name = xcframework.name
-            contents_by_config[config] << %(  install_framework "#{Target::BuildSettings::XCFRAMEWORKS_BUILD_DIR_VARIABLE}/#{name}/#{name}.framework"\n)
+            contents_by_config[config] << %(  install_framework "#{Target::BuildSettings::XCFRAMEWORKS_BUILD_DIR_VARIABLE}/#{target_name}/#{name}.framework"\n)
           end
         end
         script << "\n" unless contents_by_config.empty?
--- a/lib/cocoapods/xcode/xcframework.rb
+++ b/lib/cocoapods/xcode/xcframework.rb
@@ -5,6 +5,10 @@ require 'cocoapods/xcode/xcframework/xcframework_slice'
 module Pod
   module Xcode
     class XCFramework
+      # @return [String] target_name the target name this XCFramework belongs to
+      #
+      attr_reader :target_name
+
       # @return [Pathname] path the path to the .xcframework on disk
       #
       attr_reader :path
@@ -23,12 +27,13 @@ module Pod
 
       # Initializes an XCFramework instance with a path on disk
       #
-      # @param [Pathname, String] path
-      #        The path to the .xcframework on disk
+      # @param [String] target_name @see target_name
+      # @param [Pathname, String] path @see path
       #
       # @return [XCFramework] the xcframework at the given path
       #
-      def initialize(path)
+      def initialize(target_name, path)
+        @target_name = target_name
         @path = Pathname.new(path).tap do |p|
           raise 'Absolute path is required' unless p.absolute?
         end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
+* Place frameworks from xcframeworks into a unique folder name to avoid duplicate outputs.  
+  [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
+  [#10106](https://github.com/CocoaPods/CocoaPods/issues/10106)
+
 * Update pod in Pods folder when changing the pod from branch to version in Podfie.  
   [gonghonglou](https://github.com/gonghonglou)
   [#10825](https://github.com/CocoaPods/CocoaPods/pull/10825)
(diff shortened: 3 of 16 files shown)


lib/cocoapods/xcode/xcframework.rb
```diff
@@ -5,6 +5,10 @@
 module Pod
   module Xcode
     class XCFramework
+      # @return [String] target_name the target name this XCFramework belongs to
+      #
+      attr_reader :target_name
```
amorde, line 10: do we need to store this here? it's odd that what was previously agnostic to the context now depends on what target provides it. if we only need this for the embed frameworks script can be provide that info some other way?

lib/cocoapods/generator/embed_frameworks_script.rb
```diff
@@ -159,8 +159,9 @@ def script
         end
         xcframeworks_by_config.each do |config, xcframeworks|
           xcframeworks.select { |xcf| xcf.build_type.dynamic_framework? }.each do |xcframework|
+            target_name = xcframework.target_name
             name = xcframework.name
-            contents_by_config[config] << %(  install_framework "#{Target::BuildSettings::XCFRAMEWORKS_BUILD_DIR_VARIABLE}/#{name}/#{name}.framework"\n)
+            contents_by_config[config] << %(  install_framework "#{Target::BuildSettings::XCFRAMEWORKS_BUILD_DIR_VARIABLE}/#{target_name}/#{name}.framework"\n)
```
amorde, line 164: we already have `config` here, can we do `/#{name}_#{config}/#{name.xcframework}` instead?

dnkoutso: oh yeah we can also try this instead.

lib/cocoapods/xcode/xcframework.rb
dnkoutso: thats a good point let me see what I can do.

lib/cocoapods/generator/embed_frameworks_script.rb
dnkoutso: actually @amorde we would need to pass the `config` into the `CopyXCFrameworksScript` as well, I guess its still better than adding `target_name` right?

lib/cocoapods/xcode/xcframework.rb
dnkoutso: we also need it for the copy-xcframeworks script or as you suggested use the `config` there as well.

lib/cocoapods/generator/embed_frameworks_script.rb
amorde: oh hm.. do we have that available as an environment variable?

amorde: `$CONFIGURATION`

dnkoutso: I will see if that works. I think I tried it but I was still getting some duplicates.

dnkoutso: @amorde now I remember...using $CONFIGURATION does not work.

Xcode always buils targets despite whether you build for release or debug it just doesnt link them.

This means we will build both SampleLibrary_Debug (actual pod name) and SampleLibrary_Release (actual pod name) however because we use $CONFIGURATION the environment sets it to Debug for both and so the build fails the same way.

```

Showing All Errors Only
Multiple commands produce '/Users/dimitris/Library/Developer/Xcode/DerivedData/Demo_XcFrameworks_Integration-ghazscornxdxcxejwakulleviesc/Build/Products/Debug-iphonesimulator/XCFrameworkIntermediates/SampleLibrary_Debug':
1) That command depends on command in Target 'SampleLibrary_Debug' (project 'Pods'): script phase “[CP] Copy XCFrameworks”
2) That command depends on command in Target 'SampleLibrary_Release' (project 'Pods'): script phase “[CP] Copy XCFrameworks”
```

dnkoutso: I wonder if I can use target name without polluting the XCFramework class.

dnkoutso: nope its not easy. The problem is for embed frameworks script for an aggregate target, we do not know which xcframework comes from which target in order to provide the correct path...........

any other suggestions/recommendations?

amorde: A bit confused. Even if we are always building the target regardless of config doesn't `$CONFIGURATION` have the correct value? Why is `SampleLibrary_Release` modifying `XCFrameworkIntermediates/SampleLibrary_Debug`?

amorde: `Copy XCFrameworks` should be placing the framework into `XCFrameworkIntermediates/SampleLibrary_$CONFIGURATION/SampleLibrary.framework` right?

dnkoutso: because the output path of `SampleLibrary_Release` is `XCFrameworkIntermediates/SampleLibrary_$CONFIGURATION` so when I build DEBUG it will be `XCFrameworkIntermediates/SampleLibrary_Debug`. The same applies for `SampleLibrary_Debug` pod. so they both output the same framework.

amorde: I think the use of the config name was getting me confused. If the pod's name is `SampleLibrary_Release`, then I would expect `Copy XCFrameworks` to place the framework slice into `XCFrameworkIntermediates/SampleLibrary_Release_Release/SampleLibrary_Release.framework` (note the double `_Release`)

amorde: Synced offline - I understand why this doesn't work now. Both targets have different names but produce the same `SampleLibrary.framework` / `SampleLibrary.xcframework`, and both targets are built regardless of config. The path will be the same unless we incorporate the target's name or some other way to distinguish the two

amorde (approved):

PR #10130, opened 2020-10-07
Remove unused generate install_xcframework_library function

mokagio:
I was hunting down the source of a "`[CP] Unable to find matching .xcframework slice in...`" warning when I noticed that the string appears twice in the generated `-xcframework.sh` script. In particular, in the `install_xcframework_library()` and `install_xcframework()` functions.

The former doesn't have any usages in the script and removing it didn't compromise the script execution in my project.

I also run `git grep install_xcframework_*` and couldn't find any usage or any interpolation that might have resulted in it being called.

<details>
<summary>Click to reveal `git grep` output</summary>

```
lib/cocoapods/generator/copy_xcframework_script.rb:install_xcframework_library() {
lib/cocoapods/generator/copy_xcframework_script.rb:install_xcframework() {
lib/cocoapods/generator/copy_xcframework_script.rb:          args = install_xcframework_args(xcframework, slices)
lib/cocoapods/generator/copy_xcframework_script.rb:          script << "install_xcframework #{args}\n"
lib/cocoapods/generator/copy_xcframework_script.rb:      def install_xcframework_args(xcframework, slices)
spec/unit/generator/copy_xcframeworks_script_spec.rb:        install_xcframework "${PODS_ROOT}/../../spec/fixtures/CoconutLib.xcframework" "CoconutLib" "framework" "ios-armv7_arm64/CoconutLib.framework" "ios-i386_x86_64-simulator/CoconutLib.framework"
spec/unit/generator/copy_xcframeworks_script_spec.rb:        install_xcframework "${PODS_ROOT}/../../spec/fixtures/CoconutLib.xcframework" "CoconutLib" "framework" "macos-x86_64/CoconutLib.framework"
spec/unit/generator/copy_xcframeworks_script_spec.rb:        install_xcframework "${PODS_ROOT}/../../spec/fixtures/CoconutLib.xcframework" "CoconutLib" "framework" "ios-armv7_arm64/CoconutLib.framework" "ios-i386_x86_64-simulator/CoconutLib.framework"
spec/unit/generator/copy_xcframeworks_script_spec.rb:        install_xcframework "${PODS_ROOT}/../../spec/fixtures/CoconutLib.xcframework" "CoconutLib" "framework" "watchos-i386-simulator/CoconutLib.framework" "watchos-armv7k_arm64_32/CoconutLib.framework"
[...]

PR #10831, opened 2021-08-03
Remove unused `install_xcframework_library` code.

dnkoutso:
supersedes https://github.com/CocoaPods/CocoaPods/pull/10130

integration specs https://github.com/CocoaPods/cocoapods-integration-specs/pull/321

### DIFF
--- a/lib/cocoapods/generator/copy_xcframework_script.rb
+++ b/lib/cocoapods/generator/copy_xcframework_script.rb
@@ -157,22 +157,6 @@ install_framework()
   echo "Copied $source to $destination"
 }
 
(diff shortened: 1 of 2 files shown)


lib/cocoapods/generator/copy_xcframework_script.rb
```diff
@@ -157,22 +157,6 @@ def script
   echo "Copied $source to $destination"
 }
 
-install_xcframework_library() {
```
dnkoutso, line 160: code is never used. it was an old way of integrating with xcframeworks

paulb777 (approved):

Issue #10378, opened 2021-01-25
Only one vendored_frameworks containing multiple xcframework is correctly integrated into project


qchenqizhi:
* [x] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report
Only one vendored_frameworks containing multiple `xcframework` is correctly integrated into the project

## What did you do?

pod install

## What did you expect to happen?
Install all pod dependencies correctly. build success

## What happened instead?
pod install success
build with error:
```
No such module 'expat'
```

### Stack

```
   CocoaPods : 1.10.1
        Ruby : ruby 2.6.3p62 (2019-04-16 revision 67580) [universal.x86_64-darwin19]
    RubyGems : 3.0.3
        Host : Mac OS X 10.15.6 (19G2021)
       Xcode : 12.3 (12C33)
         Git : git version 2.15.0
Ruby lib dir : /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib
               trunk - CDN - https://cdn.cocoapods.org/
```

### Installation Source

```
Executable Path: /usr/local/bin/pod
```

### Plugins

```
cocoapods-deintegrate : 1.0.4
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.0
cocoapods-stats       : 1.1.0
cocoapods-trunk       : 1.4.1
cocoapods-try         : 1.1.0
```

### Podfile

```ruby
platform :ios, '11.0'

target 'Demo' do
  platform :ios, '11.0'
  use_frameworks!
  pod 'mobile-ffmpeg-full', '~> 4.4'
end
```

## Project that demonstrates the issue
It can be easily reproduced:
Add `mobile-ffmpeg-full` to the Podfile, and then import multiple frameworks provided by `mobile-ffmpeg-full` in project
![image](https://user-images.githubusercontent.com/973316/105673208-b0848880-5f20-11eb-9d07-f80eb8d6f663.png)

Maybe related: https://github.com/tanersener/mobile-ffmpeg/issues/663

dnkoutso:
Please provide a sample app demonstrating the issue.

dnkoutso:
oh nevermind you have some clear and easy instructions to do so.

dnkoutso:
yeap confirmed. I think I got a fix as well.

dnkoutso:
[...]

PR #10884, opened 2021-08-20
Correctly process multiple `xcframeworks` a pod provides.

dnkoutso:
closes https://github.com/CocoaPods/CocoaPods/issues/10378

integration specs https://github.com/CocoaPods/cocoapods-integration-specs/pull/326

### DIFF
--- a/lib/cocoapods/generator/copy_xcframework_script.rb
+++ b/lib/cocoapods/generator/copy_xcframework_script.rb
@@ -70,8 +70,8 @@ copy_dir()
   local destination="$2"
 
   # Use filter instead of exclude so missing patterns don't throw errors.
-  echo "rsync --delete -av "${RSYNC_PROTECT_TMP_FILES[@]}" --links --filter \\"- CVS/\\" --filter \\"- .svn/\\" --filter \\"- .git/\\" --filter \\"- .hg/\\" \\"${source}\\" \\"${destination}\\""
-  rsync --delete -av "${RSYNC_PROTECT_TMP_FILES[@]}" --links --filter "- CVS/" --filter "- .svn/" --filter "- .git/" --filter "- .hg/" "${source}" "${destination}"
+  echo "rsync --delete -av "${RSYNC_PROTECT_TMP_FILES[@]}" --links --filter \\"- CVS/\\" --filter \\"- .svn/\\" --filter \\"- .git/\\" --filter \\"- .hg/\\" \\"${source}*\\" \\"${destination}\\""
+  rsync --delete -av "${RSYNC_PROTECT_TMP_FILES[@]}" --links --filter "- CVS/" --filter "- .svn/" --filter "- .git/" --filter "- .hg/" "${source}"/* "${destination}"
 }
 
 SELECT_SLICE_RETVAL=""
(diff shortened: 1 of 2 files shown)


lib/cocoapods/generator/copy_xcframework_script.rb
```diff
@@ -130,33 +130,6 @@ def script
   done
 }
 
-install_library() {
```
dnkoutso, line 133: @amorde I noticed this isn't called anywhere along with the one below.

are they used?

lib/cocoapods/generator/copy_xcframework_script.rb
```diff
@@ -171,15 +144,16 @@ def script
     return
   fi
   local source="$basepath/$target_path"
-
-  local destination="#{Pod::Target::BuildSettings::XCFRAMEWORKS_BUILD_DIR_VARIABLE}/${name}"
+  local files=( ${source}/*.framework )
```
dnkoutso, line 147: this is the actual fix. we now. tell `rsync` to copy the `path/to/my.framework` into `dest/dir/my.framework` now instead.

This ensures the `--delete` option above does not accidentally delete folders (other frameworks) that are not part of the source.

@amorde one question: we always take`files[0]` as you can see, do we expect we can have multiple `.framework` for a given arch? Is that design allowed by `.xcframework`s?

dnkoutso: note: this also preserves the functionality of `--delete` in case a pod is bumped and the `.framework` has new files or delete some existing ones.

lib/cocoapods/generator/copy_xcframework_script.rb
```diff
@@ -130,33 +130,6 @@ def script
...
-
-  # Libraries can contain headers, module maps, and a binary, so we'll copy everything in the folder over
-
-  local source="$binary"
-  echo "rsync --delete -av "${RSYNC_PROTECT_TMP_FILES[@]}" --links --filter \\"- CVS/\\" --filter \\"- .svn/\\" --filter \\"- .git/\\" --filter \\"- .hg/\\" \\"${source}/*\\" \\"${destination}\\""
-  rsync --delete -av "${RSYNC_PROTECT_TMP_FILES[@]}" --links --filter "- CVS/" --filter "- .svn/" --filter "- .git/" --filter "- .hg/" "${source}/*" "${destination}"
-}
-
-# Copies a framework to derived data for use in later build phases
-install_framework()
```
dnkoutso, line 146: see https://github.com/CocoaPods/CocoaPods/pull/10884/files#r693123741 above.

lib/cocoapods/generator/copy_xcframework_script.rb
dnkoutso: actually thank god for examples this breaks a static library integration.

lib/cocoapods/generator/copy_xcframework_script.rb
```diff
@@ -70,8 +70,8 @@ def script
   local destination="$2"
 
   # Use filter instead of exclude so missing patterns don't throw errors.
-  echo "rsync --delete -av "${RSYNC_PROTECT_TMP_FILES[@]}" --links --filter \\"- CVS/\\" --filter \\"- .svn/\\" --filter \\"- .git/\\" --filter \\"- .hg/\\" \\"${source}\\" \\"${destination}\\""
-  rsync --delete -av "${RSYNC_PROTECT_TMP_FILES[@]}" --links --filter "- CVS/" --filter "- .svn/" --filter "- .git/" --filter "- .hg/" "${source}" "${destination}"
+  echo "rsync --delete -av "${RSYNC_PROTECT_TMP_FILES[@]}" --links --filter \\"- CVS/\\" --filter \\"- .svn/\\" --filter \\"- .git/\\" --filter \\"- .hg/\\" \\"${source}*\\" \\"${destination}\\""
+  rsync --delete -av "${RSYNC_PROTECT_TMP_FILES[@]}" --links --filter "- CVS/" --filter "- .svn/" --filter "- .git/" --filter "- .hg/" "${source}"/* "${destination}"
```
dnkoutso, line 74: So we now make `rsync` expand the file list from the directory rather than copying the directory itself. This ensures we copy everything and since we use `--delete` it also ensures we do not delete other folders or files as part of this process when multiple xcframeworks are installed by a single pod.

From SO:

```

  --delete
    This  tells  rsync to delete extraneous files from the RECEIVING SIDE (ones
    that AREN’T ON THE SENDING SIDE), but only for the directories that are
    being synchronized.  **You must have asked rsync to send the whole directory
    (e.g.  "dir" or "dir/") without using a wildcard for the directory’s contents
    (e.g. "dir/*") since the wildcard is expanded by the shell and rsync thus
    gets a request to transfer individual files, not the files’ parent directory.**
    Files  that  are  excluded  from  the transfer are also excluded from being
    deleted unless you use the --delete-excluded option or mark the rules as
    only matching on the sending side (see the include/exclude modifiers in the
    FILTER RULES section).  ...
```

lib/cocoapods/generator/copy_xcframework_script.rb
dnkoutso: these dont seem to be used, all tests pass and examples build now without it. I think its dead code

paulb777 (approved):

PR #10950, opened 2021-09-16
Use `${DEVELOPMENT_LANGUAGE}` as the default `CFBundleDevelopmentRegion` value in any generated `Info.plist`.

dnkoutso:
Xcode by default uses `${DEVELOPMENT_LANGUAGE}` for this value, however, there is no UI to update it. I put it behind `PODS_DEVELOPMENT_LANGUAGE` so folks can go to build settings to change it.

Screen Shot 2021-09-16 at 3 44 27 PM

Integration specs https://github.com/CocoaPods/cocoapods-integration-specs/pull/327

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -219,7 +219,7 @@ begin
         exit 1
       end
       title 'Running Integration tests'
-      rm_rf 'tmp'
+      FileUtils.rm_rf 'tmp'
       title 'Building all the fixtures'
       sh('bundle exec bacon spec/integration.rb') {}
       title 'Storing fixtures'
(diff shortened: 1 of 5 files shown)


Rakefile
```diff
@@ -219,7 +219,7 @@ begin
         exit 1
       end
       title 'Running Integration tests'
-      rm_rf 'tmp'
+      FileUtils.rm_rf 'tmp'
```
dnkoutso, line 222: fixes for Ruby 3.0 when rebuilding integration spec fixtures.

amorde (approved):

Issue #11040, opened 2021-11-04
'readme' and 'changelog' attributes unrecognized by `pod trunk push`


jrullman:
* [X] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

Run `pod trunk push` with a podspec having 'readme' and 'changelog' attributes, after installing cocoapods 1.11.

## What did you expect to happen?

Push should succeed, as these new attributes are recognized by cocoapods 1.11.

## What happened instead?

Push failed due to warnings of unrecognized attributes: 'readme' and 'changelog'.

### Stack

```
   CocoaPods : 1.11.2
        Ruby : ruby 2.6.3p62 (2019-04-16 revision 67580) [universal.x86_64-darwin20]
    RubyGems : 3.0.3
        Host : macOS 11.6.1 (20G224)
       Xcode : 13.0 (13A233)
         Git : git version 2.34.0.rc0.344.g81b53c2807-goog
Ruby lib dir : /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib
Repositories : trunk - CDN - https://cdn.cocoapods.org/
```

### Installation Source

```
Executable Path: /usr/local/bin/pod
```

### Plugins

```
cocoapods-deintegrate  : 1.0.4
cocoapods-dependencies : 1.3.0
cocoapods-plugins      : 1.0.0
cocoapods-search       : 1.0.0
cocoapods-stats        : 1.1.0
cocoapods-trunk        : 1.4.0
cocoapods-try          : 1.1.0
```

PR #11041, opened 2021-11-04
Update cocoapods-trunk gem minimum version requirement.

jrullman:
Fixes CocoaPods/CocoaPods#11040

### DIFF
--- a/cocoapods.gemspec
+++ b/cocoapods.gemspec
@@ -34,7 +34,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'cocoapods-downloader',  '>= 1.4.0', '< 2.0'
   s.add_runtime_dependency 'cocoapods-plugins',     '>= 1.0.0', '< 2.0'
   s.add_runtime_dependency 'cocoapods-search',      '>= 1.0.0', '< 2.0'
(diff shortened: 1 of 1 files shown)


dnkoutso:
This wont fix the issue I thnk we need this https://github.com/CocoaPods/trunk.cocoapods.org/pull/317 to land.

jrullman:
Are you saying the linting is happening on the server side using code from cocoapods-core?

dnkoutso:
The linting happens locally however the serialization of the podspec happens in the trunk service. 

If you are seeing a lint error locally you might need to do `bundle update cocoapods-trunk`. The version requirement you changed wont have an effect much the requirement accepts the latest version of cocoapods-trunk which 1.6.0.

However, I hope readme or changelog are properly serialized by trunk in the version its on when the pod is published.

jrullman:
The publishing works fine - we have been doing it for several releases now using --allow-warnings.

How do we force `gem install cocoapods` to update cocoapods-trunk to 1.6.0 and use the latest version?

dnkoutso:
You cannot but I think we accept the PR you made. This can be for cocoapods 1.12.

dnkoutso:
btw I think `gem install cocoapods` _should_ choose the 1.6.0 version. Its possible its just an outdated machine that had run `gem install cocoapods` when 1.4.0 was the only released version.

Issue #10390, opened 2021-01-29
`ruby-macho` dependency fails to read some dylibs


PeqNP:
One of our 3rd party xcframeworks can not be read by the version of `ruby-macho` (~1.4) that cocoapods depends on. When I `puts` the error that `MachO` provided it returned the following error:

```
Unrecognized Mach-O load command: 0x80000034
```

The effect is that the 3rd party xcframework does not get its respective `install_framework` line added in the targets `Pods-mytarget-frameworks.sh` file. This causes the app to crash as it can't load the library on app launch.

It produces the same error on cocoapods v1.8 and v1.10.1. I was able to determine a working version of the lib by copying the code from `def dynamic_library?` into a `test.rb` script, updating the version of `ruby-macho` to the latest (which is 2.5), and then running `test.rb`.

Bundler does not provide a way to override this dependency. Therefore, my options are to either fork cocoapods, and possibly other dependencies that require this version, and update the respective `Gemfile` (no thanks) or pray that some kind soul can at least help me determine which project's `Gemfile` needs to be updated. I've looked at the cocoapods's `Gemfile` and several other dependencies and could not find the declaration.

dnkoutso:
Thanks for the report! Cocoapods gemspec shows:
```ruby
s.add_runtime_dependency 'ruby-macho',    '~> 1.4'
```

Is there any version until 2.0 that fixes the issue or is it fixed only in 2.x and above?

PeqNP:
I have no idea how I missed that. I am so sorry. I'll see if I can find it and update it. Thank you so much!

dnkoutso:
@PeqNP looks like there is no other version above 1.4 fo ruby-macho. it goes straight to 2.0.

The only thing to do here is for us to update cocoapods to work with 2.x version or if you fork and change the gemspec.

PeqNP:
@dnkoutso thank you so much. I was able to fix the problem with the following procedure.

- Fork `cocoapods`
- Checkout the `1.10.1` tag
- Make the modification to switch `ruby-macho` to 2.5
- Created a new tag `1.10.1nb` ("nb" stands for "Nordstrom Best". Long story)
- Updated `Gemfile` and added line: `gem "cocoapods", :git => "https://github.com/PeqNP/CocoaPods", :tag => "1.10.1nb"`
- `pod install`

I now see the `install_framework` line in the respective bash script. Yay!

[...]

PR #11061, opened 2021-11-18
Change minimal required version of ruby-macho to 2.0

jerbob92:
:rainbow: 
In response to #10390

This is needed because gem does not automatically upgrade ruby-macho to a 2+ version when upgrading CocoaPods.
Since more and more people will upgrade their macOS to Big Sur, this issue will be more common.


xuzhongping:
Hello,the specific repair version of ruby-macho is 2.3.0, not 2.0, so I created a pr #11071

jerbob92:
@xuzhongping Hey, thanks, you are correct! I did not realize there were versions below 2.4.0 (I only checked GitHub releases.

Closed in favor of #11071

PR #11071, opened 2021-11-26
Change minimal required version of ruby-macho to 2.3.0

xuzhongping:
The problem comes from #10390 
The specific repair version of ruby-macho is 2.3.0 [ruby-macho-commit](https://github.com/Homebrew/ruby-macho/commit/f17ed3780ad15db85e5079eafd10b26f5b58c489)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,11 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 * Clean sandbox when a pod switches from remote to local.  
   [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
   [#11213](https://github.com/CocoaPods/CocoaPods/pull/11213)
+* Change minimal required version of ruby-macho to 2.3.0.
+  [xuzhongping](https://github.com/xuzhongping)
+  [#10390](https://github.com/CocoaPods/CocoaPods/issues/10390)
+
+
 
 * Run post install hooks when "skip Pods.xcodeproj generation" option is set
   [Elton Gao](https://github.com/gyfelton)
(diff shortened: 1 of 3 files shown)


jerbob92:
@dnkoutso could you perhaps look into this? This is causing quite some issues for our developers and also for our customers.

xuzhongping:
After my research and testing,in the following situations, when the target deployment platform version is greater than or equal to iOS iOS15 or Mac OS12 (not tested on other platforms), two different loading commands will be generated and cause problems.
Reference:[How iOS 15 makes your app launch faster](https://medium.com/geekculture/how-ios-15-makes-your-app-launch-faster-51cf0aa6c520)

jerbob92:
@xuzhongping will it also cause issues with later ruby-macho?
Because for me the issue has a different origin: adding a M1 (ARM) slice to xcframeworks

xuzhongping:
@jerbob92 The problem caused by the two different load commands I described is the defect that I mentioned earlier that ruby-macho-2.3.0 fixes.
I am also using the M1 mac. I think this problem should have nothing to do with the architecture. Is the origin of your problem also generated these load commands?
If it has nothing to do with the two load commands mentioned above, I also want to know what is the structure of your xcframework slice file. It would be better if you can describe how to generate or upload an instance.

jerbob92:
@xuzhongping Yes, it's because M1 binaries also contain new load commands. The xcframework contains armv7 and arm64 (x86_64 and arm64 for the simulator).
cocoapods uses ruby-macho to check whether binaries are static or dynamically linked (I think) and because the new load commands broke ruby-macho the command failed too, making cocoapods thinking they were not static and not including the framework slice properly in the build.

xuzhongping:
@jerbob92 You think is right!(cocoapods uses ruby-macho to check whether binaries are static or dynamically linked),ruby-macho is the source of this type of problem, so it is indeed necessary to correct this dependency error as soon as possible.

PR #11073, opened 2021-11-26
Fix post install hooks from Podfile not ran if skip_pods_project_generation

gyfelton:
Right now `run_podfile_post_install_hooks ` is only ran when `skip_pods_project_generation` option is set to false
I have a use case where I want to set skip_pods_project_generation to false and still run some post install hooks.

Unless there is explicit reason not to call it when skip_pods_project_generation is false, I see this as a bug? Just that few ppl ever run into my combo. (want skip_pods_project_generation + run post install hook)

TODO:
- [x] Change log
- [x]  Write/fix specs
- [x] Have an integration test case?

### DIFF
--- a/lib/cocoapods/installer.rb
+++ b/lib/cocoapods/installer.rb
@@ -163,6 +163,7 @@ module Pod
       validate_targets
       if installation_options.skip_pods_project_generation?
         show_skip_pods_project_generation_message
+        run_podfile_post_install_hooks
       else
         integrate
       end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
   [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
   [#11213](https://github.com/CocoaPods/CocoaPods/pull/11213)
 
+* Run post install hooks when "skip Pods.xcodeproj generation" option is set
+  [Elton Gao](https://github.com/gyfelton)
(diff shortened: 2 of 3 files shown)


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Fix post install hooks from Podfile not ran if skip_pods_project_generation  
  [gyfelton](https://github.com/gyfelton)
```

</p>

lib/cocoapods/installer.rb
```diff
@@ -163,6 +163,7 @@ def install!
       validate_targets
       if installation_options.skip_pods_project_generation?
         show_skip_pods_project_generation_message
+        run_podfile_post_install_hooks
```
amorde, line 166: Makes sense to me. Spec for this can be adapted from this existing one https://github.com/CocoaPods/CocoaPods/blob/6896aa979505035a6f2a61fa1583a89952388d99/spec/unit/installer_spec.rb#L95-L118

gyfelton: thx for looking into this! going to resume pushing on this by writing the spec to get it red then green with this change

gyfelton:
I think Dimitris need this change to ensure the new code is running with our current setup.
Cc @dnkoutso

amorde (approved):

PR #11099, opened 2021-12-13
fix missing require leading to uninitialized constant ActiveSupport::XmlMini::IsolatedExecutionState

dorianmariecom:
To use `active_support`, one is supposed to load active support first

this fixes support for rails 7 rc1

https://github.com/rails/rails/pull/43852#issuecomment-992674343

> It shouldn't be necessary to require this file manually, there's an autoload for it here:
> 
> https://github.com/rails/rails/blob/85f45350e6d3fb6284c744d36b77b79ce639dfee/activesupport/lib/active_support.rb#L51
> 
> I suspect the problem is that CocoaPods is cherry-picking this file without requiring `active_support` first, which is how Active Support is intended to be used:
> 
> https://guides.rubyonrails.org/active_support_core_extensions.html#stand-alone-active-support

see https://github.com/CocoaPods/CocoaPods/issues/11097, https://github.com/rails/rails/issues/43851 and https://github.com/rails/rails/pull/43852

### DIFF
--- a/cocoapods.gemspec
+++ b/cocoapods.gemspec
@@ -48,6 +48,8 @@ Gem::Specification.new do |s|
 
   s.add_runtime_dependency 'addressable', '~> 2.8'
 
+  s.add_runtime_dependency 'activesupport', '>= 5.0'
+
   s.add_development_dependency 'bacon', '~> 1.1'
   s.add_development_dependency 'bundler', '~> 2.0'
   s.add_development_dependency 'rake', '~> 10.0'


ghiculescu:
Should `activesupport` also be added to https://github.com/CocoaPods/CocoaPods/blob/master/cocoapods.gemspec ?

dorianmariefr:
`cocoapods-core` as activesupport as a dependency, but I added it as it is clearer that this gem requires activesupport. I added `>= 6`, like what is done in core except that there is no upper bound because active support is pretty stable

PR #11157, opened 2022-01-14
fix: turn a warning message to a error message

TyrantDante:
if it just cause a warning,  most of us maybe ignore it, then it will cost some  errors like 
- header file cannot be found 
- symbol not found

and developer will be puzzled by these errors,because the frameworks are copy to the build space(XCFRAMEWORKS_BUILD_DIR_VARIABLE), in another way, it cause a not visual error

### DIFF
--- a/lib/cocoapods/generator/copy_xcframework_script.rb
+++ b/lib/cocoapods/generator/copy_xcframework_script.rb
@@ -140,7 +140,7 @@ install_xcframework() {
   select_slice "${paths[@]}"
   local target_path="$SELECT_SLICE_RETVAL"
   if [[ -z "$target_path" ]]; then
-    echo "warning: [CP] Unable to find matching .xcframework slice in '${paths[@]}' for the current build architectures ($ARCHS)."
+    echo "error: [CP] Unable to find matching .xcframework slice in '${paths[@]}' for the current build architectures ($ARCHS)."
     return
   fi
   local source="$basepath/$target_path"


TyrantDante (commented):
yeap

amorde (approved):
This looks good to me. I cannot remember why I chose to make this a warning instead of an error, and I did check old PRs to see if I ever changed this but didn't find anything

lib/cocoapods/generator/copy_xcframework_script.rb
```diff
@@ -140,7 +140,7 @@ def script
   select_slice "${paths[@]}"
   local target_path="$SELECT_SLICE_RETVAL"
   if [[ -z "$target_path" ]]; then
-    echo "warning: [CP] Unable to find matching .xcframework slice in '${paths[@]}' for the current build architectures ($ARCHS)."
+    echo "error: [CP] Unable to find matching .xcframework slice in '${paths[@]}' for the current build architectures ($ARCHS)."
```
amorde, line 143: actually I think I remember why I made this a warning - there can be cases where a framework is only used in certain build variants (ex. a framework that is only used on either Simulator or device, but not both) and I wasn't confident that we should block the build 100% of the time

amorde (changes requested):
removing prior approval

lib/cocoapods/generator/copy_xcframework_script.rb
TyrantDante: That is a good idea for no-block, but if the framework do not contain the target arch,at last it will sumit a `undefine symbol` error.

amorde: Only if you are using that symbol in that architecture. What I am describing is something where you might have a framework that you only use on device, and then your code looks like this:

```
#if CAN_USE_DEVICE_FRAMEWORK_XYZ
import XYZ
...
#endif
```

TyrantDante: it will cause cannot found header file error
for example
i have a xcframework like this
- a.xcframework
 - arm64_armv7.framework
 - x86_64-simulartor
when run simulartor destination
it wil no copy any framework to PODS_XCFRAMEWORKS_BUILD_DIR and just echo a warning "unable to ...."
when compile a file include "a/xx.h" will cause cannot find `a/xx.h`,becase PODS_XCFRAMEWORKS_BUILD_DIR has no `a/xx.h`

PR #11158, opened 2022-01-14
Fix setting `LD_RUNTIME_SEARCH_PATHS` for aggregate targets that include dynamic xcframeworks.

dnkoutso:
If an app (and its aggregate target) only consume pre-built dynamic xcframeworks (xcframeworks with slices that are dynamically linked) then we never set the `LD_RUNTIME_SEARCH_PATHS` because we only do it if any of our dependencies are dynamic or whether vendored *frameworks* are dynamic.

This PR fixes this.

### DIFF
--- a/spec/unit/target/build_settings/aggregate_target_settings_spec.rb
+++ b/spec/unit/target/build_settings/aggregate_target_settings_spec.rb
@@ -342,6 +342,24 @@ module Pod
             end
           end
 
+          describe 'with a vendored dynamic xcframework pod' do
+            before do
+              coconut_pod_target = @pod_targets.first
+              coconut_pod_target.stubs(:build_type).returns(BuildType.static_library)
+            end
+
+            def specs
+              spec = fixture_spec('xcframeworks/xcframework-spec.podspec')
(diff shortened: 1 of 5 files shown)


spec/unit/target/build_settings/aggregate_target_settings_spec.rb
```diff
@@ -342,6 +342,24 @@ def specs
...
+
+            def specs
+              spec = fixture_spec('xcframeworks/xcframework-spec.podspec')
+              spec.vendored_frameworks = 'DynamicFramework/CoconutLib.xcframework'
+              [spec]
+            end
+
+            it 'includes default runpath search path list when linking vendored dynamic xcframework' do
+              @target.stubs(:build_type => BuildType.dynamic_framework)
+              @generator.generate.to_hash['LD_RUNPATH_SEARCH_PATHS'].should == "$(inherited) '@executable_path/Frameworks' '@loader_path/Frameworks'"
```
dnkoutso, line 359: verified this fails without my change.

amorde (approved):

PR #11213, opened 2022-02-18
Clean sandbox when a pod switches from remote to local.

dnkoutso:


lib/cocoapods/installer.rb
```diff
@@ -456,7 +456,22 @@ def clean_sandbox(pod_targets)
...
+        end
+      end
+
+      # Check any changed pods that became local pods and used to be remote pods and
+      # ensure the sandbox is cleaned up.
+      unless sandbox_state.changed.empty?
+        title_options = { :verbose_prefix => '-> '.red }
+        sandbox_state.changed.each do |pod_name|
+          UI.titled_section("Cleaning #{pod_name}".red, title_options) do
+            previous_spec_repo = sandbox.manifest.spec_repo(pod_name)
```
amorde, line 472: should we also be checking for other checkout options such as `git:` or does this only occur when going from spec repo -> local?

dnkoutso: form spec repo -> local, I think switching sources works correctly.

dnkoutso: The latter works because I think we decide to reinstall the pod which eventually cleans it up.

amorde (approved):

PR #11262, opened 2022-03-21
Try updating Danger to 8.x

dnkoutso:
This updates the gem `kramdown` to 2.3.x which resolves a high vulnerability issue in this repo.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -172,27 +172,49 @@ GEM
       colored2 (~> 3.1)
     crack (0.4.5)
       rexml
-    danger (5.16.1)
+    danger (8.5.0)
       claide (~> 1.0)
       claide-plugins (>= 0.9.2)
       colored2 (~> 3.1)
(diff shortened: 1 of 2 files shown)


Gemfile.lock
```diff
@@ -229,7 +254,7 @@ GEM
     netrc (0.11.0)
     no_proxy_fix (0.1.2)
     notify (0.5.2)
-    octokit (4.18.0)
+    octokit (4.22.0)
```
amorde, line 257: when I did this earlier it broke Danger after merging. I think there's something that changed with the way auth works between these two versions https://github.com/CocoaPods/CocoaPods/pull/11225

dnkoutso: ah let me fix that.

dnkoutso: updated.

orta:
Looks reasonable to me 👍🏻

PR #11093, opened 2021-12-09
Fix static Swift XCFramework import paths

igor-makarov:
Resolves #10058

Has an integration fixture update here: https://github.com/CocoaPods/cocoapods-integration-specs/pull/329

Issue #10058, opened 2020-09-15
1.10.0.beta.2 - XCFrameworks + static libs error


omarzl:
* [X] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

We are migrating some code to static libraries using xcframeworks and we want to distribute it via Cocoapods.
In the demo there is a StaticLib folder with an xcframework that has only one class and its podspec, also there is an Example project with its Podfile.

I added the xcframework to the `vendored_frameworks` and after running `pod install` I noticed that it is setup in the Library Search Paths

<img width="1166" alt="image" src="https://user-images.githubusercontent.com/6267487/93150607-44b32900-f6bf-11ea-811a-0fccbc72fcaf.png">

But if I try to compile it gives me a `No such module 'StaticLib'` error, I found out that Xcode can't find the `StaticLib.swiftmodule` so I modified the `SWIFT_INCLUDE_PATHS` using the `s.user_target_xcconfig` in the `StaticLib.podspec`

```
  s.user_target_xcconfig = {
          'SWIFT_INCLUDE_PATHS' => '"${PODS_XCFRAMEWORKS_BUILD_DIR}/'+"#{s.name}"+'"'
    }
```

I read this PR https://github.com/CocoaPods/CocoaPods/pull/9720 but it only says that the `${HEADER_SEARCH_PATHS}` adds the static library headers but mine is not Obj-C, its pure Swift.

Also its good to say that I built the xcframework using the xcodebuild command:
```
xcodebuild -create-xcframework \
  -library "./simulator/libStaticLib.a" \
  -library "./device/libStaticLib.a" \
  -output StaticLib.xcframework
```

I would like to ask if I am doing something wrong or its missing something at Cocoapods code to include the `SWIFT_INCLUDE_PATHS` part, I could leave it like that and it works but maybe its not the correct way.

--------

Also I noticed another issue, when I try to run the project in a device it gives me this error:

```
Showing All Messages
[...]

PR #11282, opened 2022-04-01
Fix `SWIFT_INCLUDE_PATHS` for swift library xcframework

xuzhongping:
## Issue
* https://github.com/CocoaPods/CocoaPods/issues/10058
* https://github.com/CocoaPods/CocoaPods/pull/11093

### Existing issues
1. If a swift library of type library exists in vendored xcframework, `SWIFT_INCLUDE_PATHS` is not set correctly.
2. The [PR](https://github.com/CocoaPods/CocoaPods/pull/11093) does not fully fix the issue.
### About the part I fixed `SWIFT_INCLUDE_PATHS`
1. when vendored_xcframework exists in the podspec and contains the swift library,
2. 1 and when test_spec or app_spec exists in podspec,
3. 1 and 2 and when source files exist in the podspec


dnkoutso:
Cc @igor-makarov

xuzhongping:
The unit test failed in the ubuntu system. I debugged and found that the file path is case-sensitive, while mac is case-insensitive. I found in the `xcframework_slice.rb` file that this method lowercases the first letter of the name. This caused my unit test to fail. Why is it downcase here? Does anyone know the reason?🤔
```
# @return [String] the name of the framework
        #
        def name
          @name ||= begin
            case package_type
            when :framework
              File.basename(path, '.framework')
            when :library
              result = File.basename(path, '.a').gsub(/^lib/, '')
              result[0] = result.downcase[0]
              result
            else
              raise Informative, "Invalid package type `#{package_type}`"
            end
          end
        end

xuzhongping:
I don't quite understand why 'config.spec -> exclude_from_backup ' fails, but it shouldn't be related to my modification.

Issue #11203, opened 2022-02-14
Development pod is linked against test target and app host target


itskoBits:
🌈
* [x] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

I created a development pod `MyLib`.  I added unit tests and a host app in the podspec using `test_spec` and `app_spec`. Here is what the podspec looks like.
```ruby
Pod::Spec.new do |s|
  s.name         = 'MyLib'
  s.version      = '1.0'
  s.homepage     = 'http://fake-home-page.com'
  s.authors      = 'My library authors'
  s.summary      = 'My library summery'
  s.description  = 'My library description'
  s.source       = { :git => 'http://fake-repo-lib.git', :tag => 'v1.0' }
  s.license      = { :type => 'MIT' }

  s.ios.deployment_target = '12.0'

  s.source_files = 'Classes/*.swift'

  s.test_spec do |test_spec|
    test_spec.source_files = 'Tests/*.swift'
    
    test_spec.dependency 'MyLib/Demo'
    
    test_spec.requires_app_host = true
    test_spec.app_host_name = 'MyLib/Demo'
  end
  
  s.app_spec 'Demo' do |app_spec|
    app_spec.source_files = 'App/*.swift'
  end
end
```
I integrated the `MyLib` pod and its unit tests in a new app called  `SampleApp`. I didn't use `use_frameworks` so `MyLib` is linked statically. I tried running the`MyLib` unit tests and they passed, however I noticed the following in the logs:
```
Class SomeClass is implemented in both /MyLib-Demo.app/MyLib-Demo (0x104a11370) and MyLib-Demo.app/PlugIns/MyLib-Unit-Tests.xctest/MyLib-Unit-Tests (0x105040390). One of the two will be used. Which one is undefined.
```
After looking into the generated `xcconfig` files by CocoaPods I noticed that both `MyLib-Unit-Tests` and `MyLib-Demo` were linked against `MyLib`. That's what I think the issue is. `MyLib` is being linked statically. Linking it against the unit tests and app targets leads to duplicate symbols when running unit tests.

## What did you expect to happen?

I expected that only `MyLib-Demo` will be linked against `MyLib`.

## What happened instead?

`MyLib` is being linked against  both `MyLib-Demo` and `MyLib-Unit-Tests`. This leads to different issues in more complex projects.

[...]

PR #11302, opened 2022-04-13
Don't add static library to OTHER_LDFLAGS in tests xcconfig, if the tests are using a host app

itskoBits:
A couple of weeks ago I reported an issue #11203. I think this change fixes the issue.

### DIFF
--- a/lib/cocoapods/target/build_settings.rb
+++ b/lib/cocoapods/target/build_settings.rb
@@ -800,7 +800,8 @@ module Pod
                                     uniq
             static_libraries_to_import.concat linker_names_from_libraries(xcframework_libraries)
           end
-          static_libraries_to_import << target.product_basename if target.should_build? && target.build_as_static_library?
+          tests_hosted_in_app = test_xcconfig && !target.test_app_hosts_by_spec[non_library_spec].nil?
+          static_libraries_to_import << target.product_basename if target.should_build? && target.build_as_static_library? && !tests_hosted_in_app
           static_libraries_to_import
         end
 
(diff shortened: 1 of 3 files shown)


lib/cocoapods/target/build_settings.rb
```diff
@@ -800,7 +800,8 @@ def linker_names_from_libraries(libraries)
                                     uniq
             static_libraries_to_import.concat linker_names_from_libraries(xcframework_libraries)
           end
-          static_libraries_to_import << target.product_basename if target.should_build? && target.build_as_static_library?
+          testsHostedInApp = test_xcconfig && !target.test_app_hosts_by_spec[non_library_spec].nil?
```
dnkoutso, line 803: nit use snake case for variable name `lib_hosted_in_app`

itskoBits: Sorry about that. I am fairly inexperienced in ruby.

dnkoutso: no worries at all!

Can you also add a spec (test) in `build_settings_spec.rb`?

itskoBits: I looked into the existing tests in `build_settings_spec.rb` and I am not sure what test to add over there. Do you have any particular scenario you have in mind?

itskoBits: @dnkoutso 
Hey, I don't know if you had the chance to see my last reply.

I spent some time looking into the existing `build_settings_spec.rb` and tried writing some myself. The test was testing the same thing I already added a tested in `pod_target_settings_spec.rb` so I was worried I was doing something wrong.

The test I wrote was something like:
```ruby
it 'does not add OTHER_SWIFT_FLAGS to the xcconfig if the target does not use swift' do
  target = fixture_pod_target('...')
  build_settings = pod(target)
  other_ld_flags = build_settings.xcconfig.to_hash['OTHER_LDFLAGS']
  other_swift_flags.should == '$(inherited) -ObjC'
end
```

Am I missing something here? Is there another test that can be written?

itskoBits: @dnkoutso 
Hey, 
I just added a spec containing 2 tests in `build_settings_spec.rb`. In these tests, instead of initializing a `PodTarget` with `fixture_pod_target` I stubbed most of the functionality. I hope it makes sense :).

itskoBits: @dnkoutso  I am not sure what's the status of this change. Should someone review it? Should I do something else?

PR #11417, opened 2022-06-17
Re-use the same path lists for pods that share the same root

dnkoutso:
Given a repo that has multiple podspecs at the root, it seems all podspecs retrieve a separate instance of `PathList` which in turn calls `read_file_system` which constantly globs the entire repository.

For large repositories this can be very slow.


watt (approved):
🙏

amorde (approved):

sexflixxxtest (commented):
Add saké par list

PR #11502, opened 2022-08-15
Update README.md

SkiingIsFun123:
Improved readability of the README.md file


orta:
Thanks for trying to contribute but these are sometimes wrong (version -> versions) and sometimes move things to be American spelling instead of the canonical British we use on CocoaPods. I think we're pretty good without these changes.

Issue #11509, opened 2022-08-22
Pod only install after second time


paulocoutinhox:
Hi,

My pod only install after second time.

In the first time, the directory is empty.

I made a project to show this:
https://github.com/paulocoutinhox/google-cast-bug

Thanks.

paulocoutinhox:
The full log is here:
https://github.com/paulocoutinhox/google-cast-bug/runs/7962291542?check_suite_focus=true

The is "First attempt" and "Second attempt" showing the difference between the two "pod install" calls.

xuzhongping:
Hi, using the git option to install the pod, I thought the repository should contain all the files needed and not point to another source; but it did work on the second install, and I'll track down what caused the problem.

paulocoutinhox:
I solve it:
https://github.com/ubook-editora/ub-google-cast-sdk-no-bluetooth/blob/main/ub-google-cast-sdk-no-bluetooth.podspec

And:
https://github.com/paulocoutinhox/google-cast-bug/blob/main/Podfile

PR #11544, opened 2022-09-17
Fix an external pod that downloaded the wrong content

xuzhongping:
## Issue
- https://github.com/CocoaPods/CocoaPods/issues/11509
### How to appear
1. Introduce a PodA in the `Podfile: pod 'A', :git => 'ARepo'`;
2. In ARepo, ` source = 'BRepo'` in the A.podspec file, and ` file_patterns` will not match any files

This causes an empty folder A to be created in the sandbox after the first pod install, but on the second pod install, the contents of BRepo will appear in folder A in the sandbox.
It is very puzzling that no changes were made between the two pod installs, but two different results are produced.
### Reason
In the second pod install, because in the process of processing the external pod, only the existence of the folder is judged, but not whether there is no content, so it is considered that it does not belong to the pre-download; in the later dependency download process it is judged as the sandbox state of the release pod, and the contents of the podspec are used to re-download, which leads to the different results of the two installs.


xuzhongping:
Sorry,there were some errors and I would look for exceptions in the commit code.

xuzhongping:
@dnkoutso @amorde @orta Hi, can you review this PR? thanks

PR #11762, opened 2023-02-10
Use TARGET_BUILD_DIR for actool output

akkyie:
🌈 

This PR changes the copy-resource script to use `$TARGET_BUILD_DIR` for the output location instead of `$BUILT_PRODUCTS_DIR` for actool. This aligns with the way other tools such as `ibtool` and `momc` operate.

### Background

I am using a unique setup with CocoaPods for my app where I call `xcodebuild` directly on CocoaPods-generated `Pods.xcodeproj` to prebuild Pods, by overriding some build settings such as `SYMROOT`. The prebuilt pods are then integrated by adding build phases manually to the app target to run `Pods-<target>-frameworks.sh` and `...-resources.sh`, with an overridden `$BUILT_PRODUCTS_DIR` that points to the directory containing prebuilt pods.

This setup has been working well with any pod, including ones containing XCFrameworks or resource bundles. However, I encountered a problem when I tried to add a pod that contains `xcassets`. The copy-frameworks script tries to run `actool` to write the assets into the app binary, but it fails because `${BUILT_PRODUCTS_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}` points to the prebuild directory instead of the app's build directory in my custom setup.

Other tools in the copy-resources script use `${TARGET_BUILD_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}` for their output (some excerpted below), and in the script it's `mkdir -p`-ing only `${TARGET_BUILD_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}`. Based on these observations, I believe it is safe to change `actool` to match with other tools. Also, with the standard setup, `$TARGET_BUILD_DIR` basically points the same path as `$BUILT_PRODUCTS_DIR` AFAIK.

Please let me know if any of my assumptions are incorrect. Thank you!

https://github.com/CocoaPods/CocoaPods/blob/2a501772e74e42d83dd68984513bb6c022309400/lib/cocoapods/generator/copy_resources_script.rb#L153-L156

https://github.com/CocoaPods/CocoaPods/blob/2a501772e74e42d83dd68984513bb6c022309400/lib/cocoapods/generator/copy_resources_script.rb#L167-L170


akkyie:
Rebased onto the latest master as it was failing due to Alamofire deprecation.
Also updated CHANGELOG (though I'm not sure if it's an enhancement) and [the associated cocoapods-integration-specs PR](https://github.com/CocoaPods/cocoapods-integration-specs/pull/339). I'll update this PR again when the spec change get merged.

Issue #11753, opened 2023-02-03
Stuck at "generating deterministic UUIDs", causing kernel panic


radex:
Hi, I don't have a reproducible project, and I know that React Native-related CocoaPods issues aren't covered here. I understand this, and I don't expect maintainers to do anything other than close this issue…

… still, I think someone else might find it helpful if they also encounter this issue, so I decided to post it anyway ;)

## What happened

in this project: https://github.com/Nozbe/WatermelonDB/commit/d9be088f1a44632c9433ed68fbc24d63cce0823e , running cocoapods would get stuck here for minutes:

```
Integrating target `hermes-engine`
    Adding Build Phase '[CP] Copy XCFrameworks' to project.
  - Generating deterministic UUIDs
```

This worked for me just yesterday, could not be replicated by a coworker, but I could reliably hit this.

If I `^C`, this is the traceback:

<details>

```
	28: from /Users/radex/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/xcodeproj-1.22.0/lib/xcodeproj/project.rb:413:in `predictabilize_uuids'
	27: from /Users/radex/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/xcodeproj-1.22.0/lib/xcodeproj/project/uuid_generator.rb:12:in `generate!'
	26: from /Users/radex/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/xcodeproj-1.22.0/lib/xcodeproj/project/uuid_generator.rb:64:in `generate_all_paths_by_objects'
	25: from /Users/radex/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/xcodeproj-1.22.0/lib/xcodeproj/project/uuid_generator.rb:64:in `each'
	24: from /Users/radex/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/xcodeproj-1.22.0/lib/xcodeproj/project/uuid_generator.rb:64:in `block in generate_all_paths_by_objects'
	23: from /Users/radex/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/xcodeproj-1.22.0/lib/xcodeproj/project/uuid_generator.rb:72:in `generate_paths'
[...]

PR #11766, opened 2023-02-11
Limit auto-detected docs in development pods to 100

radex:
See: https://github.com/CocoaPods/CocoaPods/issues/11753

TL;DR:

- In Development Pods, CP will try to auto-detect docs files, and add them to Pods/Development Pods/NAME/Pod/ group in Xcode
- This might be helpful if there's a few files, but trying to add an unlimited number of files (that are just there for a convenience and not required to build the app) to an Xcode project is not a good idea. In my case, I had a `docs-website` folder, containing 30k files in its node_modules.
- Besides bloating Xcodeproj, a very large number of files can cause Xcodeproj's UUID generator to exhaust all system resources and cause a kernel panic ;) Surely that could be improved, but I don't think we should be adding an arbitrary number of docs files to Xcodeproj anyway.

### DIFF
--- a/lib/cocoapods/sandbox/file_accessor.rb
+++ b/lib/cocoapods/sandbox/file_accessor.rb
@@ -11,7 +11,8 @@ module Pod
     class FileAccessor
       HEADER_EXTENSIONS = Xcodeproj::Constants::HEADER_FILES_EXTENSIONS
       SOURCE_FILE_EXTENSIONS = (%w(.m .mm .i .c .cc .cxx .cpp .c++ .swift) + HEADER_EXTENSIONS).uniq.freeze
-
+      MAX_AUTO_DETECTED_DOCS = 100
+      
       GLOB_PATTERNS = {
         :readme              => 'readme{*,.*}'.freeze,
         :license             => 'licen{c,s}e{*,.*}'.freeze,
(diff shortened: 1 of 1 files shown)


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Limit auto-detected docs in development pods to 100  
  [radex](https://github.com/radex)
```

</p>

dnkoutso:
This seems quite arbitrary in my opinion. I would perhaps provide an option in the podfile to globally ignore those or a per-pod configuration:

```ruby
pod 'MyPod', :skip_developer_files => true
```

or something like that.

radex:
I agree that the exact number is arbitrary.

My main counterpoint is that this implicit behavior can break `pod install` in a way that's very hard to debug, or could bloat the Xcode project to the point where it's very unwieldy - without the user even noticing.

For _some_ sufficiently large number of documentation files (100? 1000?), it will be very unlikely that the user actually intended this to happen, especially that it only happens for development pods, and has no impact on the build.

If you still disagree, then I would propose a compromise in which docs are not arbitrarily limited, but a warning is emitted above 100 auto-detected docs so the user knows to use `skip_developer_files: true`

philips77:
I just encountered the issue. Docs generated with DocC have thousands of files and the generated `Pods.xcodeproj/project.pbxproj` is 4.5 MB long, which makes Xcode's new Source Control navigator unusable. Waiting time to even display the changes is counted in minutes on MacBook Pro.

PR #11828, opened 2023-03-23
Xcode 14.3 fix: Pass the -f option when resolving the path to the symlinked source.

chrisvasselli:
🌈

Fixes the issue where archives fail when using Xcode 14.3. https://github.com/CocoaPods/CocoaPods/issues/11808

Xcode 14.3 is now using a relative path in its symlink for frameworks. Without the -f flag, this relative path would be evaluated relative to the working directory of the script being executed, instead of relative to the framework symlink itself. With the -f flag, it resolves that relative path and returns the full path to the source.

`bundle exec rake spec` failed, but it looks like maybe just because there were changes? Snapshot testing maybe? Also, it might make sense to add a test case to handle this scenario, but I don't know CocoaPods well enough to know what to add or where.

[rake.txt](https://github.com/CocoaPods/CocoaPods/files/11048739/rake.txt)

### DIFF
--- a/lib/cocoapods/generator/embed_frameworks_script.rb
+++ b/lib/cocoapods/generator/embed_frameworks_script.rb
@@ -83,7 +83,7 @@ install_framework()
 
   if [ -L "${source}" ]; then
     echo "Symlinked..."
-    source="$(readlink "${source}")"
+    source="$(readlink -f "${source}")"
   fi
 
   if [ -d "${source}/${BCSYMBOLMAP_DIR}" ]; then
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,10 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
-* None.  
+* Xcode 14.3 fix: Pass the -f option when resolving the path to the symlinked source.  
+  [Chris Vasselli](https://github.com/chrisvasselli)
+  [#11828](https://github.com/CocoaPods/CocoaPods/pull/11828)
+  [#11808](https://github.com/CocoaPods/CocoaPods/issues/11808)
 
 
 ## 1.12.0 (2023-02-27)


ehagerty (commented):
Chip: Apple M1 Max
Memory: 64 GB
macOS: 13.4 Beta (22F5027f)

Hi, forgive me if this is the wrong way to suggest it, but in embed_frameworks_script.rb I needed to change your use of readlink to the alternative greadlink which is installed via Brew with coreutils and which by default isn't aliased in place of readlink.  Also, when building for macos, the script Pods-Runner-frameworks.sh also needs to use greadlink but it required me to use the full path to the binary whereas in embed_frameworks_script.rb it only required the binary name.

gali8:
+1

holyavkin:
consider renaming the PR and adding a prefix:
`Xcode 14.3 fix:`

chrisvasselli:
> Chip: Apple M1 Max Memory: 64 GB macOS: 13.4 Beta (22F5027f)
> 
> Hi, forgive me if this is the wrong way to suggest it, but in embed_frameworks_script.rb I needed to change your use of readlink to the alternative greadlink which is installed via Brew with coreutils and which by default isn't aliased in place of readlink. Also, when building for macos, the script Pods-Runner-frameworks.sh also needs to use greadlink but it required me to use the full path to the binary whereas in embed_frameworks_script.rb it only required the binary name. By using the fully qualified binary in embed_frameworks_script.rb, it then generates that into Pods-Runner-frameworks.sh so I assume that is the only place that needs a change (if I am right at all in the first place).

Thanks for the suggestion! What did greadlink do that readlink -f didn't?

chrisvasselli:
Hey @dnkoutso sorry to bug you, but this seems like a pretty urgent fix now that Xcode 14.3, and you seem like the author of a bunch of recent commits. Are you able to take a look at this PR, or suggest someone I could reach out to?

ehagerty:
> > Chip: Apple M1 Max Memory: 64 GB macOS: 13.4 Beta (22F5027f)
> > Hi, forgive me if this is the wrong way to suggest it, but in embed_frameworks_script.rb I needed to change your use of readlink to the alternative greadlink which is installed via Brew with coreutils and which by default isn't aliased in place of readlink. Also, when building for macos, the script Pods-Runner-frameworks.sh also needs to use greadlink but it required me to use the full path to the binary whereas in embed_frameworks_script.rb it only required the binary name. By using the fully qualified binary in embed_frameworks_script.rb, it then generates that into Pods-Runner-frameworks.sh so I assume that is the only place that needs a change (if I am right at all in the first place).
> 
> Thanks for the suggestion! What did greadlink do that readlink -f didn't?

Hi @chrisvasselli i should first thank you for all your work to find this in the first place!  using readlink gives an error saying it doesn't know what to do with the -r option, although humorously it then says you could try the -r option...  I recalled reading somewhere about the old school readlink command being not quite the same as the greadlink command on the mac so tried it and it ran to completion.  I apologise if I've got the wrong end of the stick in making this suggestion, but did want to help others get past the error - as you did.

lib/cocoapods/generator/embed_frameworks_script.rb
```diff
@@ -83,7 +83,7 @@ def script
 
   if [ -L "${source}" ]; then
     echo "Symlinked..."
-    source="$(readlink "${source}")"
+    source="$(readlink -f "${source}")"
```
chamodanethra, line 86: This solution is working. I used this workaround for a Flutter project running on 3.7.7.
Thanks, @chrisvasselli

olmosbek: good job bro. Thanks!

fivegreenapples:
@chrisvasselli I appreciate your work on this. Saved me a ton of time. Thanks

dnkoutso:
Running CI might need integration specs update.

CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Xcode 14.3 fix: Pass the -f option when resolving the path to the symlinked source.  
  [chrisvasselli](https://github.com/chrisvasselli)
```

</p>

chrisvasselli:
@dnkoutso thanks for taking a look. Created a PR for the specs update, and added an entry to the changelog.

https://github.com/CocoaPods/cocoapods-integration-specs/pull/341

chrisvasselli:
@ehagerty the correct option to pass is `-f` not `-r`, did you try that?

lib/cocoapods/generator/embed_frameworks_script.rb
nickjun00: Hi, I cant fix it with my Flutter project. Could you help me please?

jaybhum: Hey nickjun03,

> Hi, I cant fix it with my Flutter project. Could you help me please?

I managed to solve the issue by doing the following:

1. Download the cocoapods repository and extract it in a folder
2. make the change in embed_frameworks_script.rb as shown above
3. Follow the steps for 'Set up a local dev copy of CocoaPods' here : https://guides.cocoapods.org/using/unreleased-features . The real world walk through is very helpful. I did not have to perform 'git checkout swift' by the way. If you are missing bundle, you can install it by: `sudo gem install bundler:2.3.26`
4. In your Flutter's ios folder, you can perform `pod deintegrate` , delete podfile.lock, then do pod install but with the local installation, like this example: `/Users/orta/spiel/ruby/CocoaPods/bin/pod install`
5. What i also had to do for the build to be successful is, I had to open Pods.xcodeproj, and go to 'ObjectMapper' in 'Targets', and set iOS Deployment Target to be 11.0 . 

I hope this helps!

geminiwen:
meet the same issues on Xcode 14.3 with flutter

maccman:
Just ran into this!

concernedrat:
Ran into this with Xcode 14.3 and Flutter today

esodot:
Running into the issue on latest xcode Version 14.3 (14E222b)

arnab:
I’m pretty sure anyone running XCode 14.3 would run into this problem. 

I suggest just subscribing to this issue instead of commenting on it (unless you have more insights into the problem). 
 
 
Then the emails/notifications from issue (which goes to all subscribers) will be more useful (and about actual progress/fix).

dnkoutso:
@chrisvasselli I merged the integration specs PR can you update the submodule here and then push? It should go green.

samueldeschamps:
Ran into the same issue on Xcode 14.3 with Flutter...
The bad luck was somehow Xcode just updated to 14.3 in my Mac without asking me, breaking with all my Flutter projects.

chrisvasselli:
@dnkoutso done.

saxahan (approved):

marco2704:
can we merge this already?

DennisAshford:
Adding the `-f` tag worked for me, however, now the Archive build is not indexing with the build version or number placed into Xcode. Is there a way to fix that so I can upload a build to TestFlight with a new build number and version with this `-f` tag?

knottx:
same issue on Xcode 14.3 with flutter

olegyablokov:
How do I update CocoaPods to pick up these fixes? Sorry for a lame question but I guess it may save time for some newcomers

dnkoutso:
Use bundler and point to latest master branch of CocoaPods https://bundler.io/

LFMAKER:
same issue on Xcode 14.3 with React Native 0.71.5.

@dnkoutso 
Can you prepare a new version of cocoapods with this fix? I'll flag React Native time on this.

TheAwndre:
A fix release will be done in the meantime or only when all of the planned 1.12.1 things are completed?

ShenYj:
1.12.1 has not released?

Coedice:
> 1.12.1 has not released?

Not yet. Still on `1.12.0`. Hopefully we can get this patch out soon, since it does seem to be causing a lot of people problems (including myself).

ShenYj:
Before the 1.12.1 release, I set the Gemfile with
 `gem 'cocoapods', :git => 'https://github.com/CocoaPods/CocoaPods.git', :branch => 'master'`

plkgq:
image

image


Fastlane can't package when I set minimum version， on xcode 14.3

plkgq:
xcode 14.3 archive failed 
image

gaoshanyu:
> Before the 1.12.1 release, I set the Gemfile with `gem 'cocoapods', :git => 'https://github.com/CocoaPods/CocoaPods.git', :branch => 'master'`

How to do this?

lib/cocoapods/generator/embed_frameworks_script.rb
pikaychahar: Hello @chamodanethra thanks

My issue has been fixed

gaoshanyu:
We use Xcode Cloud as CI for our Flutter project, I fixed this by:
- In shell file `ci_post_clone.sh`, Add command line `find ios/Pods -name 'Pods-Runner-frameworks.sh' -type f -exec sed -i '' 's/source="$(readlink "${source}")"/source="$(readlink -f "${source}")"/' {} \;` after command line `flutter build ipa`


<details><summary>The whole code of `ci_post_clone.sh`</summary>

```
#!/bin/sh

#  ci_post_clone.sh
#  Runner
#
#  Created by raniys on 11/19/22.
#

echo "/********** Start ci_post_clone.sh **********/"

# The default execution directory of this script is the ci_scripts directory.
cd $CI_WORKSPACE # change working directory to the root of your cloned repo.

# Install Flutter using git.
git clone https://github.com/flutter/flutter.git --depth 1 -b 3.7.8 $HOME/flutter
export PATH="$PATH:$HOME/flutter/bin"

# Check flutter version
flutter --version

# Install Flutter artifacts for iOS (--ios), or macOS (--macos) platforms.
flutter precache --ios

# Install Flutter dependencies.
flutter pub get
flutter pub global activate intl_utils 2.8.1
flutter --no-color pub global run intl_utils:generate
flutter pub run build_runner build --delete-conflicting-outputs

# Install CocoaPods using Homebrew.
HOMEBREW_NO_AUTO_UPDATE=1 # disable homebrew's automatic updates.
brew install cocoapods
pod --version

# Install CocoaPods dependencies.
# cd ios && pod install # run `pod install` in the `ios` directory.

flutter build ipa

find ios/Pods -name 'Pods-Runner-frameworks.sh' -type f -exec sed -i '' 's/source="$(readlink "${source}")"/source="$(readlink -f "${source}")"/' {} \;

exit 0
echo "/********** End ci_post_clone.sh **********/"



```
</details>

Wish for help someone, cheers!!!

ehagerty (commented):
Chip: Apple M1 Max
Memory: 64 GB
macOS: 13.4 Beta (22F5027f)

Additional change needed to fix the issue where archives fail when using Xcode 14.3. https://github.com/CocoaPods/CocoaPods/issues/11808

As referenced by @chrisvasselli in https://github.com/CocoaPods/CocoaPods/pull/11828, there is a problem with lib/cocoapods/generator/embed_frameworks_script.rb which manifests itself once migrating to Xcode 14.3. Xcode now uses a relative path in its symlink resolution for frameworks. This relative path, incorrectly, will now be evaluated relative to the working directory of the script being executed, not to the framework symlink itself.

In order to use relative path resolution, normally you would pass the -f option to the readlink binary - the same binary already referenced by @chrisvasselli. However, in some (if not many) environments, readlink is the legacy, bundled version from days of yore. If you pass it -f it will fail in a curious way (ironically by telling you to pass the -f option). If instead you use the coreutils version of readlink - named greadlink out of the box - then the -f option works as expected and this issue is resolved. NB, if you pass it OINLY greadlink, your flutter build will now work for ios but NOT macos. if you pass it your FULL path to your greadlink binary, your flutter build will work for both ios AND macos.

Having said all this, my concern is that by passing -f, there may be unintended breakage elsewhere in the cases where absolute path resolution previously worked. Is there a different approach, known to the core devs, which resolves this case of needing relative path resolution but does so without using -f? At a minimum perhaps I'd suggest adding an inline check/error message if the initial readlink call fails?

Can I also be an old man and point out that the recommendations in the comment stream of https://github.com/CocoaPods/CocoaPods/pull/11828 seem to be telling people to switch their cocoapods from a stable version to master so they can 'just get past this issue'. If true, and you do this in anything like a 'production' environment, you are probably going to bump into the law of unintended consequences...

I'm not a doctor, I only play one on television. If I have the wrong end of the stick here I apologise for the distraction.

danysz:
when this will be released ?

goldenmoon67:
when can we use this bug-fixed version

chrisvasselli:
@ehagerty thanks for this. To be honest, I'm out of my depth at this point and a CocoaPods maintainer should probably chime in!

maxonclaxon:
Hola! 

As an alternative and temporary solution, I can suggest downgrading xcode version to 14.2. You can download version 14.2 here: https://developer.apple.com/download/all/?q=Xcode%2014.2.

Save this to your Applications folder named like xcode_v_14_2.

Next, you need to tell Xcode to use the legacy command line tools. To do this: Xcode -> Settings -> Locations -> Command Line Tools and select previous.

sboudouk:
The fix is merged but is it live or do we have to wait for a new version release ?

iosdec:
> Hola!
> 
> As an alternative and temporary solution, I can suggest downgrading xcode version to 14.2. You can download version 14.2 here: https://developer.apple.com/download/all/?q=Xcode%2014.2.
> 
> Save this to your Applications folder named like xcode_v_14_2.
> 
> Next, you need to tell Xcode to use the legacy command line tools. To do this: Xcode -> Settings -> Locations -> Command Line Tools and select previous.

This is okay if you don't need to submit anything other than Ad-Hoc builds. Apple doesn't allow any other builds on versions that aren't current.

iosdec:
The only thing that worked for me - it works perfectly now:

1: Download Xcode 14.2
2: Once downloaded, extract it, then right click, show package contents
3: Navigate to `Contents/Developer/Toolchains`
4: Right click on `XcodeDefaults.xctoolchain`
5: Press `Show Package Contents`
6: Navigate to `usr/lib`
7: Copy the folder `arc`
8: Head to the same path in your Xcode 14.3 (`/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefaults.xctoolchain/usr/lib`)
9: Paste the `arc` folder there

Somebody is hosting these files on this repo: https://github.com/kamyarelyasi/Libarclite-Files

But, if you don't trust those - follow the steps above.

nuunoo1977:
> Hola!
> 
> As an alternative and temporary solution, I can suggest downgrading xcode version to 14.2. You can download version 14.2 here: https://developer.apple.com/download/all/?q=Xcode%2014.2.
> 
> Save this to your Applications folder named like xcode_v_14_2.
> 
> Next, you need to tell Xcode to use the legacy command line tools. To do this: Xcode -> Settings -> Locations -> Command Line Tools and select previous.

For me this worked and seems the easiest and non obstructive solution. Just download previous version, named with other name and on the 14.3 version just change one setting so that 14.3 uses the command line tools from the 14.2 version that we just installed. Perfect solution that worked. Thanks @maxonclaxon

ehagerty:
@dnkoutso sorry for my slow reply.  I did some further digging and I believe you are doing the right thing by pushing the '-f option' first.  The 'greadlink option' can be investigated separately for whomever still needs it.  NB It looks like this may be one of those things that spans multiple camps.  Flutter is impacted b/c they autogen their own scripting based off cocoapods, and it seems also there may be some interaction with ruby - when used via rbenv - as rbenv previously tried to use greadlink in lieu of readlink for the reasons I touched on https://github.com/rbenv/rbenv/pull/1484.  At a guess, the flutter and cocoapods build chain is using its own ruby environment and hence saw different behaviour to what I expected.

oinowe:
> > Hola!
> > As an alternative and temporary solution, I can suggest downgrading xcode version to 14.2. You can download version 14.2 here: https://developer.apple.com/download/all/?q=Xcode%2014.2.
> > Save this to your Applications folder named like xcode_v_14_2.
> > Next, you need to tell Xcode to use the legacy command line tools. To do this: Xcode -> Settings -> Locations -> Command Line Tools and select previous.
> 
> For me this worked and seems the easiest and non obstructive solution. Just download previous version, named with other name and on the 14.3 version just change one setting so that 14.3 uses the command line tools from the 14.2 version that we just installed. Perfect solution that worked. Thanks @maxonclaxon

For me, I just downgraded to version 14.2, no need to change the setting in the command line tools. Hope this can help!

gali8:
Temporary fix:

Into the "Pods-_APPNAME_-frameworks.sh" file:

Change:
```
  if [ -L "${source}" ]; then
    echo "Symlinked..."
    source="$(readlink "${source}")"
  fi
```
with:
```
  if [ -L "${source}" ]; then
    echo "Symlinked..."
    source="$(readlink -f "${source}")"
  fi
```

rudifa:
@gali8 It worked here too, thanks!

ghenry22:
Any chance of pushing out a quick release with this fix? It would be greatly appreciated as we are also currently stuck with a bunch of failing builds and rolling back a whole team of systems or manually editing a bunch of projects is not going to be fun when the fix is here already.

is there a way to install cocoapods from GitHub directly using brew? Like a nightly or bleeding edge version tag?

lib/cocoapods/generator/embed_frameworks_script.rb
leegaeul: This solution is working. You saved my life. Thanks!

lunching-patryk-konieczny:
for flutter users:
https://groups.google.com/g/flutter-announce/c/9qZL8GOVNb4

anuj-sitesurface:
upgrade your flutter version to 3.7.10

ondrejstasek:
> Any chance of pushing out a quick release with this fix? It would be greatly appreciated as we are also currently stuck with a bunch of failing builds and rolling back a whole team of systems or manually editing a bunch of projects is not going to be fun when the fix is here already.
> 
> is there a way to install cocoapods from GitHub directly using brew? Like a nightly or bleeding edge version tag?

If you are using `Gemfile` (`bundle`) to install CocoaPods, you can use something like this:
`gem 'cocoapods', :git => 'https://github.com/CocoaPods/CocoaPods.git', :branch => 'master'`
The fix is already merged to `master` but has not yet been released.

Revoola:
@gali8 It didn't work if you have two .sh files (one for iPhone one for Watch).

dhruv4567:
@gali8  It didn't work if you have two .sh files (one for iPhone one for Watch).

JonasESmith:
> 

This works for my iOS build but gives me the following error when I run `flutter run -d macos -v`

```[        ] ** BUILD FAILED **
[        ] The following build commands failed:
[        ]      /{directory}build/macos/Build/Products/Debug/nanopb/nanopb.framework/Versions/A/nanopb normal (in target 'nanopb' from project 'Pods')
[        ] (1 failure)```

**Comment by Fudoshiki** [comment]
```zsh
git clone https://github.com/CocoaPods/CocoaPods.git
cd CocoaPods
gem build cocoapods.gemspec
sudo gem install ./cocoapods-1.12.0.gem
```

jsare:
BE SURE  to download flutter to version 3.7.10 first. In the terminal, just do "flutter upgrade"

Next, check whether the terminal has access to the disk:
Go to System Preferences - Security & Privacy - Full Disk Access - Terminal

Turn on access if it is disabled.

Go to the terminal and enter the following:
`cd /Applications/Xcode.app/Contents/Developer/Toolchains/Xcode Default.xctoolchain/usr/lib/`

Then we enter this:

```
sudo mkdir var
cd  arc
sudo git clone https://github.com/kamyarelyasi/Libarclite-Files.git .
```

We give the necessary permissions:
`sudo chmod +x *`

This is all thanks to the person on StackOverflow:
https://stackoverflow.com/a/75924853/19880050

Then we go to the Finder, search for our application and run the Runner file.xcworkspace - optional if running on a physical device

lib/cocoapods/generator/embed_frameworks_script.rb
chamodanethra: Flutter 3.7.10 has fixed this: https://github.com/flutter/flutter/issues/123890

vbrunko-voc:
https://github.com/flutter/flutter/issues/123890 fixed the issue for me ✅

advancedngwarati:
> Hola!
> 
> As an alternative and temporary solution, I can suggest downgrading xcode version to 14.2. You can download version 14.2 here: https://developer.apple.com/download/all/?q=Xcode%2014.2.
> 
> Save this to your Applications folder named like xcode_v_14_2.
> 
> Next, you need to tell Xcode to use the legacy command line tools. To do this: Xcode -> Settings -> Locations -> Command Line Tools and select previous.

Downloading Xcode 14.2 and use that one instead of 14.3 worked for me. I did not change the Command Line Tools.

harmeetsingh-work:
> BE SURE to download flutter to version 3.7.10 first. In the terminal, just do "flutter upgrade"
> 
> Next, check whether the terminal has access to the disk: Go to System Preferences - Security & Privacy - Full Disk Access - Terminal
> 
> Turn on access if it is disabled.
> 
> Go to the terminal and enter the following: `cd /Applications/Xcode.app/Contents/Developer/Toolchains/Xcode Default.xctoolchain/usr/lib/`
> 
> Then we enter this:
> 
> ```
> sudo mkdir var
> cd  arc
> sudo git clone https://github.com/kamyarelyasi/Libarclite-Files.git .
> ```
> 
> We give the necessary permissions: `sudo chmod +x *`
> 
> This is all thanks to the person on StackOverflow: https://stackoverflow.com/a/75924853/19880050
> 
> Then we go to the Finder, search for our application and run the Runner file.xcworkspace - optional if running on a physical device

@jsare thanx for the fix:
Just one thing, I think that folder is `arc` for `mkdir` command

jsare:
> > BE SURE to download flutter to version 3.7.10 first. In the terminal, just do "flutter upgrade"
> > Next, check whether the terminal has access to the disk: Go to System Preferences - Security & Privacy - Full Disk Access - Terminal
> > Turn on access if it is disabled.
> > Go to the terminal and enter the following: `cd /Applications/Xcode.app/Contents/Developer/Toolchains/Xcode Default.xctoolchain/usr/lib/`
> > Then we enter this:
> > ```
> > sudo mkdir var
> > cd  arc
> > sudo git clone https://github.com/kamyarelyasi/Libarclite-Files.git .
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > We give the necessary permissions: `sudo chmod +x *`
> > This is all thanks to the person on StackOverflow: https://stackoverflow.com/a/75924853/19880050
> > Then we go to the Finder, search for our application and run the Runner file.xcworkspace - optional if running on a physical device
> 
> @jsare thanx for the fix: Just one thing, I think that folder is `arc` for `mkdir` command instead of `var`

Yeah man, you are right. It's my bad) I fix it in my comment

HoustonDuane:
Is there any possibility that we released 1.12.1 the following week? Our team is really stoked to experience Swift 5.8 in Xcode 14.3 however this issue is blocking us.

lexrus:
After removing all unnecessary `self` in our codebase... then I found this issue 😭

vdugnist:
If you've installed cocoapods with gem and want to install version from master:
```
gem install specific_install
gem specific_install -l https://github.com/CocoaPods/CocoaPods.git
```

Probably you will see a warning 
```
You don't have ~/.gem/ruby/2.6.0/bin in your PATH,
	  gem executables will not run.
```

In this case you can run cocoapods from that folder
```
~/.gem/ruby/2.6.0/bin/pod install
```

lexrus:
https://github.com/CocoaPods/CocoaPods/issues/11808#issuecomment-1500907375
This workaround worked for my project.
Remember to replace `File::exists?` with `File.exist?` if it failed with the undefined Ruby method error.

lib/cocoapods/generator/embed_frameworks_script.rb
appsowner: upgrade Flutter 3.7.10 solution for my

palancana:
For Flutter users, the latest hotfix in the stable channel ([3.7.11](https://github.com/flutter/flutter/wiki/Hotfixes-to-the-Stable-Channel)) should fix the building and running of Flutter apps with Xcode 14.3. 

It includes the following fixes:
https://github.com/flutter/flutter/issues/124529
https://github.com/flutter/flutter/issues/124340

alex-taffe:
Any timeline on 1.12.1? We're entirely blocked on Ticketmaster's new SDK because of this (requires Xcode 14.3)

bastecklein:
> Any timeline on 1.12.1? We're entirely blocked on Ticketmaster's new SDK because of this (requires Xcode 14.3)

Seems strange that they would let such a breaking bug in their core product linger for so long.

alex-taffe:
> > Any timeline on 1.12.1? We're entirely blocked on Ticketmaster's new SDK because of this (requires Xcode 14.3)
> 
> Seems strange that they would let such a breaking bug in their core product linger for so long.

I mean it's open source, it's not like they're losing revenue, it's still just people working on this because they enjoy it, technically no obligation

vdugnist:
> Any timeline on 1.12.1? We're entirely blocked on Ticketmaster's new SDK because of this (requires Xcode 14.3). @dnkoutso

I've posted a solution few answers above, you can use it before 1.12.1 release.

alex-taffe:
> > Any timeline on 1.12.1? We're entirely blocked on Ticketmaster's new SDK because of this (requires Xcode 14.3). @dnkoutso
> 
> I've posted a solution few answers above, you can use it before 1.12.1 release.

Our CI pipeline has non versioned releases disabled from a safety perspective. If we absolutely have to, we can disable those checks, but if the timeline was like this afternoon or Monday morning, then it wasnt really worth it

PratibhaSanath:
Hi 

I am using Xcode 14.3 for my iOS project. In order to fix archive issue, should I upgrade my cocoa pods version to 1.12.1? The archive is part of our CI CD pipeline process used for generating release builds.

saschpe:
While the linker issue also came with Xcode-14.3, this is better raised with those individual gems targeting outdated iOS versions.

chrisvasselli:
Hey @dnkoutso what do you think about releasing 1.12.1 with what's in there now, to get this fix out in a public release? This seems like an urgent fix to a bug making CocoaPods essentially unusable, and Xcode 14.3 has been publicly available for a couple weeks now.

Sorry to bug you again, but it seems like you're the only active maintainer as far as I can tell?

dnkoutso:
I would love to ship this but unfortunately I keep seeing various reports like [this one](https://github.com/CocoaPods/CocoaPods/issues/11808#issuecomment-1510351203) and [this one](https://github.com/CocoaPods/CocoaPods/issues/11808#issuecomment-1505285909) that perhaps its not the best fix?

I havent had time to look myself into the actual issue. I might end up shipping 1.12.1 by EOW unless any other maintainer does it before me.

chrisvasselli:
@dnkoutso Makes sense to be cautious! My sense though is that fixing this issue will unblock far more people than it will negatively impact.

And for what it's worth, neither of those reports you linked worry me, and I believe both should be resolved by the fix. I think what you're actually seeing is people having trouble applying the workaround.

I think part of what's going on is that the fix is in a script that generates a bunch of files. And the workaround I also gave was to just manually find/replace the line in question in the files that the script generated. So I think some people are trying the workaround, and then maybe not finding all the files, or reporting the name of the generated file that they needed to modify.

If there's one issue that I _do_ think is worth keeping an eye on, it's probably [this one](https://github.com/CocoaPods/CocoaPods/pull/11828#pullrequestreview-1363574262). This person had a non-standard version of the `readlink` binary on his machine that didn't support the flag I added. For them, my fix will not work. However, it won't make things worse for them either.

Their suggested solution is to use `greadlink`, but since that would be adding a new dependency to cocoapods, I'm not sure how easy/desirable that would be. In the meantime though, I don't think it should prevent pushing out this change.

ehagerty:
Hey @chrisvasselli, thanks for the link to my suggestion :-) I feel famous now!

@dnkoutso I feel comfortable that you can and should move ahead with the 'readlink' fix as is.

NB, it isn't that I have a non standard version of readlink.  I'm fairly sure that /anyone/ who tries to use the /default/ version of readlink on an M1 mac will likely hit the -f error.  I've seen this bug in places as disparate as builds of mongodb, fish shell etc.  The reason people may /not/ hit the error is because the standard install of coreutils via brew creates a symbolic link from 'readlink' to the 'greadlink' binary.  In other words, I'm hitting the error because I /am/ using the standard binary and everyone else is letting their system 'magically' use the 'non-standard' greadlink.

The other tickets are as @chrisvasselli suggests - a ton of noise because people crashed ahead with find and replace / HEAD branches etc instead of reading enough to understand what was going on.  Anyone who us using brew/coreutils is covered - it is probably only antique C developers like me that have a system configured without the extra 'magic'.

I'm not a doctor, but I do play one on television - ymmv.

fivegreenapples:
Hi @ehagerty re. your readlink/greadlink comments, I have done a little digging and I haven't found what you're suggesting.  In fact rather the opposite:

On a friend's Mac Studio, M1 Max, macOS 13.3.1 (22E261), with no `coreutils` installed by `brew`, his `readlink` binary is very happy with `readlink -f`. Also the same on my wife's M1 Macbook Air (admittedly running macOS 12.something). fwiw on my Intel based iMac, also with no coreutils, my readlink is also fine.

So I kinda wonder what `readlink` is on your mac? Can you run `which readlink` to see where it is and confirm it is in the standard location, and also confirm it isn't symlinked to something unexpected. Also `man readlink` to see what that shows, and lastly double check that `readlink -f` errors.

Sorry that is a bunch of things to look at but I'd like to get to the bottom of it if possible.

For reference, this is output from my friend's M1 Mac:

```
~$ which readlink
/usr/bin/readlink
~$ ls -l /usr/bin/ | grep readlink
-rwxr-xr-x   2 root   wheel    135008  1 Apr 17:46 readlink

~$ man readlink
STAT(1)                     General Commands Manual                    STAT(1)

NAME
     stat, readlink – display file status

SYNOPSIS
     stat [-FLnq] [-f format | -l | -r | -s | -x] [-t timefmt] [file ...]
     readlink [-fn] [file ...]
etc...
```

chrisvasselli:
@fivegreenapples @ehagerty As far as I can tell I also don't have coreutils or greadlink installed:

```
brew list | grep coreutils
# no output

which readlink
/usr/bin/readlink

which greadlink
greadlink not found
```

And /usr/bin/readlink is not aliased.

I'm on an M1 MacBookPro running macOS 13.2.1.

HoustonDuane:
Hi folk, I see that 1.12.1 has been released, and I've tried that with
```bash
sudo gem install cocoapods
sudo xcode-select -s /Applications/Xcode14_3.app/Contents/Developer
pod install
```
I make sure the Podfile.lock has updated the cocoapods version:
image
 
After hitting `Clean Build Folder` and rerunning the app, the problem persists.

image

Any idea?

fivegreenapples:
@HoustonDuane I'm pretty sure you're issue is the same as this one: https://github.com/CocoaPods/CocoaPods/issues/11808#issuecomment-1505471562

HoustonDuane:
@fivegreenapples Isn't my problem supposed be resolved once I upgrade to 1.12.1? Unfortunately I can't easily move up the deployment target to 11.0.

fivegreenapples:
@HoustonDuane 
> Did you mean that upgrading to 1.12.1 does not solve the `libarclite` problem? 

No - 1.12.1 fixes only the 3 issues mentioned here: https://github.com/CocoaPods/CocoaPods/releases/tag/1.12.1
One of those is the fix in this pull request but that is not related to `libarclite`

>Unfortunately I can't easily move up the deployment target to 11.0. Can anybody help?
From what I understand you just need to be above 8.0 (per this comment further down in that thread: https://github.com/CocoaPods/CocoaPods/issues/11808#issuecomment-1505581357)

hth

HoustonDuane:
@fivegreenapples A million thanks for your reply, I managed to solve the issue by adding the following to my Podfile:
```ruby
post_install do |installer|
        # Fix libarclite_xxx.a file not found.
            config.build_settings['IPHONEOS_DEPLOYMENT_TARGET'] = '9.0'
        end
    end
end
```
To anyone who's wondering, you only need to specify the deployment target greater than 8.0. I've not tried 8.0 itself, but 9.0 solves my problem (and I exactly need it to be 9.0).

Whew, what a month. I've noticed Xcode 14.3 is blazing fast compared to 14.1, don't know what magic the developers from Apple has done.

magsimillion:
hello! first of all thank you so much for your fix; adding the -f flag has helped in fixing ONE problem. the thing is, another has arisen and I have the vague suspicion that it has something to do with the new Xcode update:

In signing & capabilities I continuously get these two errors:
- Failed to register bundle identifier
The app identifier "{our identifier}" cannot be registered to your development team because it is not available. Change your bundle identifier to a unique string to try again.
- No profiles for '{our identifier}' were found
Xcode couldn't find any iOS App Development provisioning profiles matching '{our identifier}'.

wouldn't these two errors imply that we a) already have an identifier with the same name but also b) it can't find the identifier?
now I've tried every fix I could find with a simple google search, but it still doesn't work. these fixes that I found were pretty old though and I've wondered, if this could be due to Xcode having similar path finding issues as described above in order to find our profile?

just as an aside: we have registered this identifier over on developer.apple.com and have successfully uploaded and released our app multiple times (only for beta testing now), but this has now stopped working since the update came out (and forced us to update in order to continue releasing the app).

cheers!

lib/cocoapods/generator/embed_frameworks_script.rb
rafaellop: Doesn't work in some cases: 
https://github.com/flutter/flutter/issues/124081#issuecomment-1515864512

zwiya: this works after spending days Thanks !!!! for flutter edit find file called 
Pods-Runner-frameworks.sh
and then search for this 

source="$(readlink "${source}")"

and add the -f 

Enjoy <3

ibatuhanozdemir: Thank you soo much. Works like a charm.

usmanabid94:
For the people facing error in Xcode after update to 14.3.

In your xcode, Go to PODS folder :
Target Support Files => Pods-{Your Project} => Pods-{Your Project}-frameworks

Change

source="$(readlink "${source}")"
this :

source="$(readlink -f "${source}")"

imilakovic:
There's no need for workarounds any more, just upgrade CocoaPods to version 1.12.1

Please see this tag for more details:
https://github.com/CocoaPods/CocoaPods/releases/tag/1.12.1

PWhittle86:
Is anyone else still experiencing this issue? I am still seeing `source="$(readlink "${source}")"` and experiencing compile errors even though I'm on version 1.13.0. Any ideas?

Ccalary:
The same question，in Xcode15，I did this，solved it。Targets -> Build Setting -> Build Options -> User Script Sandboxing，set Yes To No。You can search "ENABLE_USER_SCRIPT_SANDBOXING"

PR #11840, opened 2023-04-03
Xcode 14.3 fix: Pass the -f option to GREADLINK when resolving the path to the symlinked source.

ehagerty:
Chip: Apple M1 Max
Memory: 64 GB
macOS: 13.4 Beta (22F5027f)

Additional change needed to fix the issue where archives fail when using Xcode 14.3. https://github.com/CocoaPods/CocoaPods/issues/11808

As referenced by @chrisvasselli in https://github.com/CocoaPods/CocoaPods/pull/11828, there is a problem with lib/cocoapods/generator/embed_frameworks_script.rb which manifests itself once migrating to Xcode 14.3.  Xcode now uses a relative path in its symlink resolution for frameworks. This relative path, incorrectly, will now be evaluated relative to the working directory of the script being executed, not to the framework symlink itself. 

In order to use relative path resolution, normally you would pass the -f option to the readlink binary - the same binary already referenced by @chrisvasselli.  However, in some (if not many) environments, readlink is the legacy, bundled version from days of yore.  If you pass it -f it will fail in a curious way (ironically by telling you to pass the -f option).  If instead you use the coreutils version of readlink - named greadlink out of the box - then the -f option works as expected and this issue is resolved.  NB, if you pass it OINLY greadlink, your flutter build will now work for ios but NOT macos.  if you pass it your FULL path to your greadlink binary, your flutter build will work for both ios AND macos.

Having said all this, my concern is that by passing -f, there may be unintended breakage elsewhere in the cases where absolute path resolution previously worked.  Is there a different approach, known to the core devs, which resolves this case of needing relative path resolution but does so without using -f?  At a minimum perhaps I'd suggest adding an inline check/error message if the initial readlink call fails?

Can I also be an old man and point out that the recommendations in the comment stream of https://github.com/CocoaPods/CocoaPods/pull/11828 seem to be telling people to switch their cocoapods from a stable version to master so they can 'just get past this issue'. If true, and you do this in anything like a 'production' environment, you are probably going to bump into the law of unintended consequences...

I'm not a doctor, I only play one on television.  If I have the wrong end of the stick here I apologise for the distraction.


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          3 Errors
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false"><code>xit</code> or <code>fit</code> left in tests (spec/functional/command/repo/update_spec.rb)</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false"><code>xit</code> or <code>fit</code> left in tests (spec/unit/config_spec.rb)</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

PR #11920, opened 2023-05-26
Do not remove the cache under 'Local Podspecs' if the Pod is a local  pod

CrazyFanFan:
The method `clean_pod` skips the deletion of the cache under 'Local Podspecs' if the Pod is a local pod.

### DIFF
--- a/lib/cocoapods/sandbox.rb
+++ b/lib/cocoapods/sandbox.rb
@@ -93,7 +93,7 @@ module Pod
     def clean_pod(name, pod_dir)
       pod_dir.rmtree if pod_dir&.exist?
       podspec_path = specification_path(name)
-      podspec_path.rmtree if podspec_path&.exist?
+      podspec_path.rmtree if podspec_path&.exist? && !local?(name)
       pod_target_project_path = pod_target_project_path(name)
       pod_target_project_path.rmtree if pod_target_project_path&.exist?
     end


orta:
I think you probably should give us an explanation of why you're requesting this change, it's not obvious why this is correct

CrazyFanFan:
In general, the cache of Local Pods under `Pods/Local Podspecs` is retained, except for the first time when a Pod is switched from a remote Pod to a local Pod, and only in this case is the cache file deleted. The cache file will be retained upon subsequent installations.

The reason for discovering this issue is that I did some additional work using the cache files under `Pods/Local Podspecs` after running `pod install`, which resulted in some anomalies.

The goals of this submission are two-fold:
1. To address the anomaly issue I encountered.
2. Unify the behavior of local Pods. 

Is there any specific design goal for not retaining the cache during the first switch to a local Pod?

The above text was translated into English by AI. If there are any unclear expressions, please forgive me and point them out for me. Thank you.

Liiiszt:
> I think you probably should give us an explanation of why you're requesting this change, it's not obvious why this is correct

When I tried to change the reference of a third-party library, ProgressHUD, from remote to local, I encountered the following error:
```
Errno::ENOENT - No such file or directory @ rb_sysopen - .../workspace/Pods/Local Podspecs/ProgressHUD.podspec.json
```
I also noticed that the `ProgressHUD.podspec.json` file was initially added to the 'Local Podspecs' folder and then removed.

This commit successfully fixes the issue, and I hope it could be merged. Thanks.

action121:
Yes, I also encountered this problem, and made local pod code modifications based on this code submission plan, which solved my problem.

Liiiszt (approved):

Issue #11942, opened 2023-06-08
[RFC] Declare SPM dependencies in podspec


mfazekas:
## Background

Swift Package Manager is becoming the de-facto standard for new libraries, due to its builtin support in Xcode. Currently Cocoapods libraries that want's to consume a SPM package have these choices:
- convert their lib to SPM and migrate off CocoaPods
- write a pod spec for the dependency
- ask users to install the SPM manually [?]

## Suggested solution

Introduce SPM dependencies as part of pod spec.

In ruby this could look like:

```ruby
  s.spm_dependency(
    url: 'https://github.com/CocoaLumberjack/CocoaLumberjack.git',
    requirement: {kind: 'upToNextMajorVersion', minimumVersion: '3.8.0'},
    product_name: 'CocoaLumberjack'
  )
```

This would perform the following modifications:
1.) Add the declared Swift Package Manger depenency to the Pods project 
2.) Add the CocoaLumberjack dependency to the pod target

## Prototype

Proof of concept implementation is available at:
https://github.com/mfazekas/pods-spm-dep-poc/

## Notes:

- Stale SPM dependencies can be handled by removing all non required SPM packages from the Pods project
- Multiple pod might depend on the same SPM package, in such case we'd need to merge requirements and handle potential conflicts
- in `spm_dependency` syntax above `product_name` might be better called as `target(s)`. I think a single SPM can declare multiple targets.

## References/related issues: 

#10049
#11618
#10463

https://developer.apple.com/documentation/packagedescription/package/dependency/requirement-swift.enum
https://www.swift.org/package-manager/
https://guides.cocoapods.org/syntax/podspec.html

Westacular:
I think it's becoming increasingly important for CocoaPods to address this need.

My initial understanding (which could be mistaken) is that for Swift Macros, the _only_ way to define and distribute them that works with Xcode is via SPM. And as you said, a variety of projects (notably the official Apple `swift-*` utility ones, like swift-algorithms and swift-collections) are SPM only.

dnkoutso:
Yes, I echo that sentiment for Swift macro support.

mfazekas:
Added a prs: https://github.com/CocoaPods/Core/pull/743 and https://github.com/CocoaPods/CocoaPods/pull/11953

yaochenfeng:
+1

zhu-xiaowei:
+1

Mioke:
[...]

Issue #743, opened 2013-01-16
Copy resources v2


alloy:
The outcome of a discussion @irrationalfab and I had today:

We’re going to:
- Perform processing on resources at `pod install` time. This includes:
  - Run pngcrush
  - Compile Core Data model.
- All resources (after processing, if needed) will be placed in Pods/Resources and the copy-resources script will reference them from there.
- Use rsync with options such as [these](http://superuser.com/questions/47986/copy-all-files-and-folders-excluding-subversion-files-and-folders-on-os-x/48003#48003) to copy new/updated resources into the product.

**Except** in the case of a local `:path` pod:
- No processing. The `:path` option is a **DURING DEVELOPMENT ONLY** feature and should not be used for release products.
- Symlink resources into Pods/Resources, instead of copying them.

_Updates:_
- mention the updated `:path` instead of `:local` directive, which is going to be deprecated

PR #11953, opened 2023-06-15
SPM: allow podspecs to declare Swift Package Manager dependencies with spm_dependency

mfazekas:
Implements: https://github.com/CocoaPods/CocoaPods/issues/11942

First draft.

See also: [#743](https://github.com/CocoaPods/Core/pull/743)


lib/cocoapods/installer/xcode/pods_project_generator.rb
```diff
@@ -193,6 +193,67 @@ def wire_target_dependencies(target_installation_results)
...
+            build_settings = native_target.build_settings(config.name)
+            build_settings['SWIFT_INCLUDE_PATHS'] ||= ['$(inherited)']
+            unless build_settings['SWIFT_INCLUDE_PATHS'].include?(search_path)
+              build_settings['SWIFT_INCLUDE_PATHS'].push(search_path)
+            end
+          end
+        end
+
+        def wire_target_spm_dependencies(target_installation_results, spm_dependencies_by_url)
+          target_installation_results.pod_target_installation_results.each_value do |_, pod_target_installation_result|
```
Dragna, line 237: Hi, 
Thank you for this PR, 
I tested it and found out that it wasn't working in my case, I had to change  this line to make it work 
```suggestion
          target_installation_results.pod_target_installation_results.each_value do |pod_target_installation_result|
```
I had error telling me that `pod_target_installation_result` was nil without this change.

### Error

```
NoMethodError - undefined method `native_target' for nil:NilClass
/Users/sdrode/Downloads/CocoaPods/lib/cocoapods/installer/xcode/pods_project_generator.rb:246:in `block (2 levels) in wire_target_spm_dependencies'
/Users/sdrode/Downloads/CocoaPods/lib/cocoapods/installer/xcode/pods_project_generator.rb:245:in `each'
/Users/sdrode/Downloads/CocoaPods/lib/cocoapods/installer/xcode/pods_project_generator.rb:245:in `block in wire_target_spm_dependencies'
/Users/sdrode/Downloads/CocoaPods/lib/cocoapods/installer/xcode/pods_project_generator.rb:237:in `each_value'
/Users/sdrode/Downloads/CocoaPods/lib/cocoapods/installer/xcode/pods_project_generator.rb:237:in `wire_target_spm_dependencies'
/Users/sdrode/Downloads/CocoaPods/lib/cocoapods/installer/xcode/single_pods_project_generator.rb:26:in `generate!'
/Users/sdrode/Downloads/CocoaPods/lib/cocoapods/installer.rb:320:in `block in create_and_save_projects'
/Users/sdrode/Downloads/CocoaPods/lib/cocoapods/user_interface.rb:64:in `section'
/Users/sdrode/Downloads/CocoaPods/lib/cocoapods/installer.rb:315:in `create_and_save_projects'
/Users/sdrode/Downloads/CocoaPods/lib/cocoapods/installer.rb:307:in `generate_pods_project'
/Users/sdrode/Downloads/CocoaPods/lib/cocoapods/installer.rb:183:in `integrate'
/Users/sdrode/Downloads/CocoaPods/lib/cocoapods/installer.rb:170:in `install!'
/Users/sdrode/Downloads/CocoaPods/lib/cocoapods/command/install.rb:52:in `run'
/Users/sdrode/.rvm/gems/ruby-3.0.0/bundler/gems/CLAide-97b765ec88aa/lib/claide/command.rb:334:in `run'
/Users/sdrode/Downloads/CocoaPods/lib/cocoapods/command.rb:52:in `run'
/Users/sdrode/Downloads/CocoaPods/bin/pod:55:in `<main>'
```

I'm no Ruby expert so maybe I didn't understand something 🙂

mfazekas: @Dragna thanks much, sorry for the issue and thanks for checking out and the patch. I've now updated my PR.

.gitmodules
```diff
@@ -6,7 +6,8 @@
 	url = https://github.com/tonymillion/Reachability.git
 [submodule "spec/cocoapods-integration-specs"]
 	path = spec/cocoapods-integration-specs
-	url = https://github.com/CocoaPods/cocoapods-integration-specs.git
+	url = https://github.com/mfazekas/cocoapods-integration-specs.git
```
mfazekas, line 9: should be reverted before merge

Gemfile
```diff
@@ -20,7 +20,8 @@ gemspec
 
 group :development do
   cp_gem 'claide',                'CLAide'
-  cp_gem 'cocoapods-core',        'Core'
+  # cp_gem 'cocoapods-core',        'Core'
+  gem 'cocoapods-core', :git => 'https://github.com/mfazekas/Core', branch: 'podspec-spm'
```
mfazekas, line 24: should be reverted before merge

osrufung:
Thank you for this! 👏

mfazekas:
@dnkoutso how can we move forward with this?

Dragna:
is there anything we can do to help in shipping this ? 🙂

orta:
A lot of it depends on folks' time to do OSS. As CocoaPods, if there's not a critical need for a release during betas we tend to really only ship after the RC version of Xcode - so I think this is likely to get some eyes over it in prep for making a release for the next Xcode major 

We're happy conceptually with the idea, just needs some eyes over implementation ideally around when we are looking at making the next release

Dragna:
@mfazekas I tested it (after rebasing on cocoapods 1.13.0), and it seems that if I define a dependency in a subspec, I see the library declared in the pod target where I defined the SPM dependency, but the spm dependency isn't added to the Pods xcodeproj.

I want to look into this to make it work, would you have some insight for me on where to look ? That would be really helpful🙂 🙏

lib/cocoapods/installer/xcode/pods_project_generator.rb
```diff
@@ -193,6 +193,67 @@ def wire_target_dependencies(target_installation_results)
...
+              end
+              acc
+            end
+          else
+            requirements.first
+          end
+        end
+
+        def install_spm_dedpendecies(project, pod_targets)
+          spm_dependencies = pod_targets.flat_map { |target| target.root_spec.consumer(target.platform).spm_dependencies }
```
Dragna, line 210: updating this line to get the dependencies on all the specs lets us define spm dependencies at the subspec level 🙂 

```suggestion
          spm_dependencies = pod_targets.flat_map { |target|
            target.specs.flat_map{ |spec| spec.consumer(target.platform).spm_dependencies }
          }
```

Dragna:
I'm getting here step by step, I added the possibility to add local packages to a podspec.

one problem I have now is that for reasons (unknown to me for now) I get a `undefined symbols` when building the project.

it means that if I have a Pod module A that depends on a swift package B, when I build my project, A compiles, but I get a compile error on the project stating that it doesn't know about B.

I may be missing a piece here (I'm lacking understanding of the build system) and any help would be great 🙂.

mfazekas:
> I'm getting here step by step, I added the possibility to add local packages to a podspec.
> 
> one problem I have now is that for reasons (unknown to me for now) I get a `undefined symbols` when building the project.
> 
> it means that if I have a Pod module A that depends on a swift package B, when I build my project, A compiles, but I get a compile error on the project stating that it doesn't know about B.
> 
> I may be missing a piece here (I'm lacking understanding of the build system) and any help would be great 🙂.

Can you check `point 3.)` here? 
https://github.com/mfazekas/pods-spm-dep-poc/#how-it-works

Dragna:
> Can you check `3.)` here?
> https://github.com/mfazekas/pods-spm-dep-poc/#how-it-works
 
 Thank you for your quick answer 🙏

It seems that I have everything set here and the paths were added. 

my error is not a `No Such Module X` on the pod importing my swiftpackage, it's more at the level of the target I'm trying to use which depends on the pod.

my target does not use the swift package code, only the pod.

if I manually add my package to the target, everything compiles.

I'm trying with a big project, I'll try making an example

Dragna:
Ok I have a way of reproducing this, if the podspec that import the swift package is defined as a `static_framework` 
`s.static_framework = true`, it's in this case that I have my error. 
I must be missing a build setting in this case.
here is a small example project

[SPMIntegrationInPodsExample.zip](https://github.com/CocoaPods/CocoaPods/files/13442489/SPMIntegrationInPodsExample.zip)

you can install the env by doing : 

1. `make` // to install my fork of cocoapods where are my currents modifications (based on what you did)
2. `make install` or `bundle exec pod install` // instead of a pod install (there is still some debug logging sorry for that 🙏)

I would totally understand that you don't have time for this, but if you think of anything that could be of help to me on how to better understand / solve this that would be great 🙂

contfedorov:
@mfazekas 
Could you please clarify if your PR deals with dependency resources correctly? 

Imagine the situation. I have an OSS project on Github that defines SPM package. I am able to consume it via Cocoapods by manually creating `.podspec` with `s.source` and `s.source_files` defined. I have just one issue with this. Project has resources (strings, images, etc.) that are packed into `.bundle` file. XCode generates `resource_bundle_accessor.swift` that finds that bundle and makes it acessable to Swift code. In case dependency is consumed by SPM, that file is generated. In case dependency is consumed by Cocoapods, that file is not generated. 

So, I am wondering if your PR can handle this.

mfazekas:
@contfedorov so this PR does what's described in: https://github.com/mfazekas/pods-spm-dep-poc/#how-it-works 

1.) Adds the Package Dependency to the Pods project.
2.) Adds the targets dependencies to Frameworks and Libraries of the pod
3.) (Adds to Swift Import Paths)

contfedorov:
@mfazekas 
Can I do the same by modifying my `.podspec` in any way?

Issue #11895, opened 2023-05-08
Xcode 14.3 - Cannot publish new Pod version having deps with deployment target < iOS 11


adrian-niculescu:
Publishing pods to our private specs repo works on Xcode 14.2 and is broken on Xcode 14.3.

* [ x] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

ℹ 
```
pod repo push <my-specs-repo> ./<MyPod>.podspec --private --sources=https://<my-specs-repo>.git,https://cdn.cocoapods.org/ --allow-warnings
```

results using Xcode 14.3 in 
```
Validating spec
 -> <MyPod> (7.1.1)
    - ERROR | [iOS] xcodebuild: Returned an unsuccessful exit code. You can use `--verbose` for more information.
    - NOTE  | xcodebuild:  note: Using codesigning identity override: -
    - NOTE  | [iOS] xcodebuild:  note: Building targets in dependency order
    - NOTE  | [iOS] xcodebuild:  Pods.xcodeproj: warning: The iOS Simulator deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 9.0, but the range of supported deployment target versions is 11.0 to 16.4.99. (in target 'nanopb' from project 'Pods')
    - NOTE  | [iOS] xcodebuild:  Pods.xcodeproj: warning: The iOS Simulator deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 8.0, but the range of supported deployment target versions is 11.0 to 16.4.99. (in target 'Starscream' from project 'Pods')
    - NOTE  | [iOS] xcodebuild:  Pods.xcodeproj: warning: The iOS Simulator deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 10.0, but the range of supported deployment target versions is 11.0 to 16.4.99. (in target 'FirebaseInstallations' from project 'Pods')
    - NOTE  | [iOS] xcodebuild:  clang: error: linker command failed with exit code 1 (use -v to see invocation)
    - NOTE  | [iOS] xcodebuild:  Pods.xcodeproj: warning: The iOS Simulator deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 9.0, but the range of supported deployment target versions is 11.0 to 16.4.99. (in target 'YYText' from project 'Pods')
[...]

Issue #11839, opened 2023-04-01
Xcode 14.3 has `pod lib lint` fail


groue:
* [X] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

Hello,

I ran `pod lib lint GRDB.swift.podspec --allow-warnings --verbose` from a fresh clone of https://github.com/groue/GRDB.swift/tree/v6.10.1, with Xcode 14.3 (14E222b)

## What did you expect to happen?

A success, as with Xcode 14.2.

## What happened instead?

An error, unlike Xcode 14.2. Relevant lines from the output of the command:

```
...
    ld: file not found: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/arc/libarclite_macosx.a
    clang: error: linker command failed with exit code 1 (use -v to see invocation)
    
    /var/folders/dp/2xq7_lgj7k5fdnl5xm0s126r0000gn/T/CocoaPods-Lint-20230401-1982-y8jwp9-GRDB.swift/Pods/Pods.xcodeproj: warning: The macOS deployment target 'MACOSX_DEPLOYMENT_TARGET' is set to 10.9, but the range of supported deployment target versions is 10.13 to 13.3.99. (in target 'SQLCipher' from project 'Pods')
    ** BUILD FAILED **
...
   Testing with `xcodebuild`. 
 -> GRDB.swift (6.10.1)
    - NOTE  | [GRDB.swift/standard,GRDB.swift/SQLCipher] xcodebuild:  note: Using codesigning identity override: -
    - NOTE  | [GRDB.swift/standard,GRDB.swift/SQLCipher] xcodebuild:  note: Building targets in dependency order
    - NOTE  | [GRDB.swift/standard,GRDB.swift/SQLCipher] xcodebuild:  note: Metadata extraction skipped. No AppIntents.framework dependency found. (in target 'GRDB.swift' from project 'Pods')
    - NOTE  | [GRDB.swift/standard,GRDB.swift/SQLCipher] xcodebuild:  note: Metadata extraction skipped. No AppIntents.framework dependency found. (in target 'App' from project 'App')
    - NOTE  | [GRDB.swift/standard,GRDB.swift/SQLCipher] xcodebuild:  note: Using codesigning identity override: 
[...]

PR #12151, opened 2023-11-16
Update analyzer.rb

tkgka:
fix validation fail when Publishing pods to private specs repo using Xcode 14.3. or upper
- validation check on iOS 12 or upper because Xcode 14.3. or upper doesn't support iOS 11 or lower
- Related Issues: https://github.com/CocoaPods/CocoaPods/issues/11895

### DIFF
--- a/lib/cocoapods/installer/analyzer.rb
+++ b/lib/cocoapods/installer/analyzer.rb
@@ -865,7 +865,7 @@ module Pod
           Version.new(library_spec.deployment_target(platform_name) || default)
         end.max
         if platform_name == :ios && build_type.framework?
-          minimum = Version.new('8.0')
+          minimum = Version.new('12.0')
           deployment_target = [deployment_target, minimum].max
         end
         Platform.new(platform_name, deployment_target)


paulb777 (commented):
Thanks for the PR. More is needed here: 

- Xcode 14 still supports iOS 11
- Other platforms
- Overriding dependency minimum options should be an option.

More discussion at https://github.com/CocoaPods/CocoaPods/issues/11839

CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Fix validation fail when Publishing pods to private specs repo using Xcode 14.3. or upper  
  [tkgka](https://github.com/tkgka)
```

</p>

tkgka:
> ```
> * Fix validation fail when Publishing pods to private specs repo using Xcode 14.3. or upper  
>   [tkgka](https://github.com/tkgka)
> ```
1. change validation target to iOS 11 or upper
2. what platforms do you mean?
3. can you give me more information about `Overriding dependency minimum options should be an option.` because it happen even I set  `s.ios.deployment_target = '13.0'`

tkgka:
and I think people who use Xcode 13 or lower can still use lower version of the CocoaPods to deploy their library even if we update the minimum deployment target

lib/cocoapods/installer/analyzer.rb
```diff
@@ -865,7 +865,7 @@ def determine_platform(specs, target_definitions, build_type)
           Version.new(library_spec.deployment_target(platform_name) || default)
         end.max
         if platform_name == :ios && build_type.framework?
-          minimum = Version.new('8.0')
+          minimum = Version.new('11.0')
```
paulb777, line 868: This is probably the wrong place in the code for this change, since this is specifying 8.0 is when Apple introduced framework support for iOS. The change should also impact tvOS, macOS, and tvOS

tkgka: oh now I understood

tkgka: @paulb777 
how does it possible to impact other OS(tvOS, macOS and tvOS) the change only run when the `platform_name` is `ios`because of the top line code 
`if platform_name == :ios && build_type.framework?`

tkgka: and I think it right place to change cause I change every singe code that determine the version `8.0` to `12.0` one by one and only place that can help was the `lib/cocoapods/installer/analyzer.rb`

paulb777: macos and tvos have the same issue as ios and should have a similar fix.

tkgka: ah now I see, but currently this fix is for iOS, so what about fix other platform's issue on the other commit? (I cannot find the source for other platform on this file)

paulb777: As described, this is the wrong place for managing minimum version. This code is for testing if frameworks are supported or not on iOS.

GNUGradyn:
This should be emergency priority. Can't build pods with xCode 14.3 and newer because of this, and can't build with anything older because of iOS 16!

PR #12154, opened 2023-11-19
Optimize performance during uncached pod installation.

dnkoutso:
When installing pods from a given repo, we are currently over-doing a bunch of unnecessary work by copying the original repo from within a temp dir N times as many pods we are installing from that repo. We then trim (clean) the copied repo to the files that are only relevant to the pod we are installing.

This can be very slow if a repo is large and offers multiple podspecs.

We also need to do this work _only_ if there is a prepare command to run which currently operates or can operate against the initial cloned repo. 

The PR here optimizes by splitting pods that require a `prepare_command` and ones that do not. For the ones that need one we execute the existing behavior but for the ones that do not, we trim the cloned repo and then copy exactly the files we need.

This can speed up installation dramatically for repos that publish multiple podspecs.

### DIFF
--- a/lib/cocoapods/downloader.rb
+++ b/lib/cocoapods/downloader.rb
@@ -51,8 +51,7 @@ module Pod
       if target && result.location && target != result.location
         UI.message "Copying #{request.name} from `#{result.location}` to #{UI.path target}", '> ' do
           Cache.read_lock(result.location) do
-            FileUtils.rm_rf target
-            FileUtils.cp_r(result.location, target)
+            FileUtils.cp_r(result.location, target, :remove_destination => true)
           end
         end
       end
--- a/lib/cocoapods/downloader/cache.rb
+++ b/lib/cocoapods/downloader/cache.rb
@@ -235,13 +235,38 @@ module Pod
       #         was not found in the download cache.
       #
       def uncached_pod(request)
-        in_tmpdir do |target|
-          result, podspecs = download(request, target)
+        in_tmpdir do |tmp_dir|
+          result, podspecs = download(request, tmp_dir)
           result.location = nil
 
-          podspecs.each do |name, spec|
+          # Split by pods that require a prepare command or not to speed up installation.
+          no_prep_cmd_specs, prep_cmd_specs = podspecs.partition { |_, spec| spec.prepare_command.nil? }.map(&:to_h)
+
+          # Pods with a prepare command currently copy the entire repo, run the prepare command against the whole
+          # repo and then clean it up. We configure those first to ensure the repo is pristine.
+          prep_cmd_specs.each do |name, spec|
             destination = path_for_pod(request, :name => name, :params => result.checkout_options)
-            copy_and_clean(target, destination, spec)
+            copy_source_and_clean(tmp_dir, destination, spec)
+            write_spec(spec, path_for_spec(request, :name => name, :params => result.checkout_options))
+            if request.name == name
+              result.location = destination
+            end
+          end
+
+          specs_by_platform = group_subspecs_by_platform(no_prep_cmd_specs.values)
+
+          # Remaining pods without a prepare command can be optimized by cleaning the repo first
+          # and then copying only the files needed.
+          pod_dir_cleaner = Sandbox::PodDirCleaner.new(tmp_dir, specs_by_platform)
+          Cache.write_lock(tmp_dir) do
+            pod_dir_cleaner.clean!
+          end
+
+          no_prep_cmd_specs.each do |name, spec|
+            destination = path_for_pod(request, :name => name, :params => result.checkout_options)
+            file_accessors = pod_dir_cleaner.file_accessors.select { |fa| fa.spec.root.name == spec.name }
+            files = Pod::Sandbox::FileAccessor.all_files(file_accessors).map(&:to_s)
+            copy_files(files, tmp_dir, destination)
             write_spec(spec, path_for_spec(request, :name => name, :params => result.checkout_options))
             if request.name == name
               result.location = destination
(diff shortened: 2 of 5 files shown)


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Optimize pod cache during installation  
```

</p>

lib/cocoapods/sandbox/pod_dir_cleaner.rb
```diff
@@ -15,7 +15,7 @@ def initialize(root, specs_by_platform)
       # @return [void]
       #
       def clean!
-        clean_paths.each { |path| FileUtils.rm_rf(path) } if root.exist?
+        FileUtils.rm_rf(clean_paths) if root.exist?
```
dnkoutso, line 18: optimization

lib/cocoapods/downloader.rb
```diff
@@ -51,8 +51,7 @@ def self.download(
       if target && result.location && target != result.location
         UI.message "Copying #{request.name} from `#{result.location}` to #{UI.path target}", '> ' do
           Cache.read_lock(result.location) do
-            FileUtils.rm_rf target
-            FileUtils.cp_r(result.location, target)
+            FileUtils.cp_r(result.location, target, :remove_destination => true)
```
dnkoutso, line 54: minor optimization

dnkoutso:
@paulb777 I checked Firebase can be installed which includes a prepare_command with this change.

lib/cocoapods/downloader/cache.rb
```diff
@@ -279,23 +304,58 @@ def in_tmpdir(&blk)
       #
       # @return [Void]
       #
-      def copy_and_clean(source, destination, spec)
-        specs_by_platform = group_subspecs_by_platform(spec)
+      def copy_source_and_clean(source, destination, spec)
+        specs_by_platform = group_subspecs_by_platform([spec])
         destination.parent.mkpath
         Cache.write_lock(destination) do
-          FileUtils.rm_rf(destination)
```
dnkoutso, line 286: I do not understand how we grabbed a lock and then deleted the entire folder here? Was it obviating the lock?

dnkoutso: Either way I preserved this logic below too.

justinseanmartin: I assume the write lock is just to make mutations exclusive to avoid invalidating other parallel work in the cache.

dnkoutso: yes but we are actually deleting the folder the lock resides in?

justinseanmartin: Oh, good point, I think you're right, not sure this is doing what we'd expect.

justinseanmartin: Actually, isn't the "#{destination}.lock" sit next to the folder, as opposed to within it?

lib/cocoapods/downloader/cache.rb
```diff
@@ -279,23 +304,58 @@ def in_tmpdir(&blk)
...
+            destination_path = File.join(destination, relative_path)
+            File.dirname(destination_path)
+          end
+
+          files_by_dir.each do |dir, files_to_copy|
+            FileUtils.mkdir_p(dir)
+            FileUtils.cp_r(files_to_copy, dir)
+          end
+
+          Pod::Installer::PodSourcePreparer.new(spec, destination).prepare!
```
justinseanmartin, line 347: Isn't this always a no-op?

justinseanmartin (approved):
Any specs for the new code path? Or are there already cases that cover this?

lib/cocoapods/downloader/cache.rb
dnkoutso: good point will delete.

lib/cocoapods/downloader/cache.rb
dnkoutso: ah yes thats what happens.

dnkoutso: now that this is figured out I updated the code to work for parallel execution.

dnkoutso:
> Any specs for the new code path? Or are there already cases that cover this? 

we could add Firebase with a prepare command as an example to build and test.

lib/cocoapods/downloader/cache.rb
```diff
@@ -279,23 +304,55 @@ def in_tmpdir(&blk)
...
+          FileUtils.cp_r(source, destination, :remove_destination => true)
           Pod::Installer::PodSourcePreparer.new(spec, destination).prepare!
           Sandbox::PodDirCleaner.new(destination, specs_by_platform).clean!
         end
       end
 
-      def group_subspecs_by_platform(spec)
+      # Copies the `files` from the `source` directory to `destination` _without_ cleaning the
+      # `destination` directory of any files unused by `spec`. This is a faster version used when
+      # installing pods without a prepare command in which has already happened prior.
```
paulb777, line 319: nit: delete "in"

paulb777 (approved):
Thanks! Any measurements of the speedup that you can share?

lib/cocoapods/downloader/cache.rb
dnkoutso: fixing

jerrymarino:
@dnkoutso @justinseanmartin this change is busting cocoapods for me in some apps. Here's a gist of what I ran into :
1. checkout a repo with a podfile.lock on version N of a pod
2. update the pod to version N + 1
3. update cocopoads to this version
4. run pod install
5. the sources for the pod at hand are missing in the `Pods` directory

dnkoutso:
Will take a look.

PR #12253, opened 2024-02-05
Fix an issue attempting to re-copy the same file when the destination does not have write permissions

amorde:
Fixes https://github.com/CocoaPods/CocoaPods/issues/12226

The underlying issue was attempting to copy a file when it already exists at the destination and does not have the write permission set, causing a write permission error.

The copy is duplicative because a previous pass had already copied a parent directory including the files. The fix is to update the logic to only copy individual files instead of also copying folders.

Note: this logic could be optimized a bit further, but I was looking for a reliable fast solution to the bug for now.

### DIFF
--- a/lib/cocoapods/downloader/cache.rb
+++ b/lib/cocoapods/downloader/cache.rb
@@ -327,13 +327,15 @@ module Pod
       # @return [Void]
       #
       def copy_files(files, source, destination)
-        files = files.select { |f| File.exist?(f) }
+        files = files.
+                map { |f| Pathname(f) }.
+                select { |p| p.exist? && p.file? }
         destination.parent.mkpath
         Cache.write_lock(destination) do
           FileUtils.rm_rf(destination)
           destination.mkpath
           files_by_dir = files.group_by do |file|
-            relative_path = Pathname(file).relative_path_from(Pathname(source)).to_s
+            relative_path = file.relative_path_from(source)
             destination_path = File.join(destination, relative_path)
             File.dirname(destination_path)
           end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
-* None.  
+* Fix an issue when caching downloads of certain Pods.  
+  [Eric Amorde](https://github.com/amorde)
+  [#12226](https://github.com/CocoaPods/CocoaPods/issues/12226)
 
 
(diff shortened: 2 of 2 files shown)


lib/cocoapods/downloader/cache.rb
```diff
@@ -327,13 +327,15 @@ def copy_source_and_clean(source, destination, spec)
       # @return [Void]
       #
       def copy_files(files, source, destination)
-        files = files.select { |f| File.exist?(f) }
+        files = files
```
dnkoutso, line 330: nit. do we need to map first and then select or just select and convert to Pathname within the `select`?

dnkoutso: I guess we do because we use them below as pathnames

dnkoutso (approved):

paulb777 (approved):
Confirmed this fixes my repro

Chaseshaw:
Didn't fix my ruby 3.3.0 / macOS 13.6.4

amorde:
@Chaseshaw got an example I could try?

CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Fix an issue attempting to re-copy the same file when the destination does not have write permissions  
```

</p>

dnkoutso (approved):

amorde:
Going to go ahead and merge + ship since we know it fixes at least certain cases of this issue.

nikolaykasyanov:
@amorde the following Podfile is causing `pod install` to fail with 1.15.1:
```ruby
platform :ios, '17.0'
inhibit_all_warnings!
use_frameworks!
install! 'cocoapods',
         :parallel_pod_downloads => true

target 'CocoaPodsReproducer' do
	pod 'Instabug', '12.6.0', :source => 'https://cdn.cocoapods.org/'
end
```

The error:
```
$ pod install
Analyzing dependencies
Downloading dependencies
Downloading Instabug (12.6.0)
Installing Instabug (12.6.0)
Generating Pods project
[!] The plist file at path `/Users/User/Desktop/CocoaPodsReproducer/Pods/Instabug/Instabug.xcframework/Info.plist` doesn't exist.
```

foxware00:
Can confirm, I can't use 1.15.1 when 1.15.0 is fine

amorde:
We'll push out an update that reverts this PR as well as the original change that this PR was attempting to fix.

PR #12331, opened 2024-04-02
[Feature][DerivedData] Add an environment variable to customize the derived data path

PatShen:
I made this PR to add an environment variable `COCOAPODS_DERIVED_DATA_PATH ` to customize the `-derivedDataPath` of `xcodebuild` when user attempts to run the CocoaPods command, such as `pod lib lint`, `pod repo push`, etc.

### DIFF
--- a/lib/cocoapods/validator.rb
+++ b/lib/cocoapods/validator.rb
@@ -1115,6 +1115,10 @@ module Pod
         command += %w(CLANG_ANALYZER_OUTPUT=html CLANG_ANALYZER_OUTPUT_DIR=analyzer)
       end
 
+      if !ENV['COCOAPODS_DERIVED_DATA_PATH'].nil? && !ENV['COCOAPODS_DERIVED_DATA_PATH'].empty?
+        command += %W(-derivedDataPath #{ENV['COCOAPODS_DERIVED_DATA_PATH']})
+      end
+
       begin
         _xcodebuild(command, true)
       rescue => e


PatShen:
When I set the environment variable like `COCOAPODS_DERIVED_DATA_PATH=~/Desktop/DerivedData`, the script will run with a `-derivedDataPath` parameter

Screenshot 2024-04-03 at 09 23 51

stidch:
I like this idea and we're missing this feature. There is no other way to force using another DerivedData folder, is there?

PatShen:
> I like this idea and we're missing this feature. There is no other way to force using another DerivedData folder, is there?

Thank you for loving this idea. And yes, at present.

stidch:
We're currently using this approach:

1. Add a custom `xcodebuild` file, like

       if [ -n "$XCODE_DERIVED_DATA_PATH" ]; then
           /usr/bin/xcodebuild -derivedDataPath "$XCODE_DERIVED_DATA_PATH" "$@"
       else
            /usr/bin/xcodebuild "$@"
       fi
       
    the path of `xcodebuild` can be different in your case.

2. Make sure the script is executable:

       chmod +x /path/to/folder/of/the/xcodebuild_wrapper/xcodebuild

3. Make sure the folder of this script is listed in your `$PATH` before `/usr/bin`.

       export PATH=/path/to/folder/of/the/xcodebuild_wrapper:$PATH

4. Set the environment variable `XCODE_DERIVED_DATA_PATH` to your desired path.

       export XCODE_DERIVED_DATA_PATH=/path/to/desired/derivedData

5. Use `xcodebuild` as you normally would. The wrapper will automatically add the `-derivedDataPath` parameter if the variable is set.

Issue #12664, opened 2024-10-28
pod lib lint is broken


kstenerud:
* [x ] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

## What did you do?

`pod lib lint` on a fresh, very basic framework project.

## What did you expect to happen?

Successful linting

## What happened instead?

It fails with various errors, such as:

* `//privateTarget Support Files/MyFramework/MyFramework-umbrella.h:13:9: error: double-quoted include "MyFramework.h" in framework header, expected angle-bracketed instead`
* `Sandbox: rsync.samba(44452) deny(1) file-write-create /Users/karl.stenerud/Library/Developer/Xcode/DerivedData/App-gnevzkrtpocjzeduklkdbvflsxod/Build/Products/Release-iphonesimulator/App.app/Frameworks/MyFramework.framework/_CodeSignature (in target 'App' from project 'App')`

### Stack

```
   CocoaPods : 1.15.2
        Ruby : ruby 3.0.6p216 (2023-03-30 revision 23a532679b) [arm64-darwin22]
    RubyGems : 3.2.33
        Host : macOS 14.7 (23H124)
       Xcode : 15.4 (15F31d)
         Git : git version 2.39.3 (Apple Git-146)
Ruby lib dir : /Users/karl.stenerud/.rbenv/versions/3.0.6/lib
Repositories : ariadnext-idcheckio-axt-podspecs - git - https://git-externe.rennes.ariadnext.com/idcheckio/axt-podspecs.git @ 2ebd182962d6b2812d1a82909d413b80b6b56867

               cocoapods - git - https://github.com/CocoaPods/Specs.git @ 0ad68fd861cfd5cc53d5e5061cf7f6faedfd637f

               trunk - CDN - https://cdn.cocoapods.org/
```

### Installation Source

```
Executable Path: /Users/karl.stenerud/.rbenv/versions/3.0.6/bin/pod
```

### Plugins

```
claide-plugins        : 0.9.2
cocoapods-deintegrate : 1.0.5
cocoapods-plugins     : 1.0.0
cocoapods-search      : 1.0.1
cocoapods-trunk       : 1.6.0
cocoapods-try         : 1.2.0
```

## Project that demonstrates the issue

https://github.com/kstenerud/cocoapods-framework-issue

Saimoness:
[...]

PR #12669, opened 2024-10-29
Fix validation failure from new xcodeproj settings

amorde:
Fixes https://github.com/CocoaPods/CocoaPods/issues/12664

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
-* None.  
+* Fix an issue breaking `pod lib lint` / `pod spec lint` due to new `ENABLE_USER_SCRIPT_SANDBOXING` setting.  
+  [Eric Amorde](https://github.com/amorde)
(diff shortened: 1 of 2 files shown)


paulb777 (approved):

dnkoutso (approved):

paulb777:
Confirmed the fix fixes my repro case and running full Firebase CI at https://github.com/firebase/firebase-ios-sdk/pull/13989

paulb777:
Hmm, while my Gemfile patch works locally, there's a ruby error in GitHub Actions, so I think I'll need to wait for release to do a full test.

amorde:
> `require': cannot load such file -- cocoapods (LoadError)

huh, weird.. does that repro if you retry the job?

paulb777:
> > `require': cannot load such file -- cocoapods (LoadError)
> 
> huh, weird.. does that repro if you retry the job?

It seems to be happening on the pod lib lint runs for all the Firebase products https://github.com/firebase/firebase-ios-sdk/pull/13989

amorde:
Ok, well I'll go ahead and merge as it seems unrelated.

Issue #12388, opened 2024-05-20
Cocoapods misleadingly detects Mergable Libraries as static frameworks


abarisain:
* [X] I've read and understood the [*CONTRIBUTING* guidelines and have done my best effort to follow](https://github.com/CocoaPods/CocoaPods/blob/master/CONTRIBUTING.md).

# Report

We distribute our framework via a precompiled, vendored XCFramework.
Starting with out latest release, we ship as a dynamic XCFramework with support for [Mergable Libraries](https://developer.apple.com/videos/play/wwdc2023/10268/)

However, on installation, CocoaPods fails to detect that the XCFramework is dynamic, which means that it's not included in `Pods-<target>-frameworks-Release-input-files.xcfilelist` and not copied during build. The app crashes on start as dyld can't find the framework.

I pinpointed this to this code: https://github.com/CocoaPods/CocoaPods/blob/master/lib/cocoapods/xcode/linkage_analyzer.rb

When `ruby-macho` fails to open a dylib, CocoaPods assumes that it is a static library.
Problem is, mergable libraries come with a new Load command: `LC_ATOM_INFO`.
`ruby-macho` does not support this (even in its latest version, 4.0.1).

The following ruby program:
```
require 'macho'

@libpath = '/Users/arnaud/dev/playground/cocoapods/tests/Pods/Batch/XCFrameworks/Batch.xcframework/ios-arm64/Batch.framework/Batch'

MachO.open(@libpath).dylib
```

prints
```
/Users/arnaud/dev/playground/machotest/vendor/bundle/ruby/2.6.0/gems/ruby-macho-4.0.1/lib/macho/macho_file.rb:598:in `block in populate_load_commands': Unrecognized Mach-O load command: 0x36 (MachO::LoadCommandError)
```

`0x36` being the code for `LC_ATOM_INFO`

As Podspecs do not allow us to bypass the auto detection, we have no choice but to make a specific build for CocoaPods without mergable library support.

## What did you do?

- Make an empty project
- Add a pod that is a vendored xcframework that is a mergable library (`pod 'Batch', '2.0.0'`)
- Build and run: The app crashes as the framework is missing at runtime 
```
dyld[64690]: Library not loaded: @rpath/Batch.framework/Batch
```

## What did you expect to happen?

CocoaPods properly detects our dynamic mergable library and copies it in the app bundle.

[...]

PR #12691, opened 2024-11-10
chore: bump ruby macho to 4.1.0 to enable mergable libraries

TheRogue76:
## Description
To address https://github.com/CocoaPods/CocoaPods/issues/12388, this PR aims to update Ruby macho to 4.1.0, which now supports Mergable libraries. The local tests i ran ran ok, but i will leave it up to the CI to check as well since bundle exec rake spec was having some issue with my ruby 3.3.0. As far as i saw comparing 2.5.1 and 4.1.0, i couldn't see a breaking change in the APIs that cocoapods uses (it is used in `linkage_analyzer.rb` and the `linkage_analyzer_spec.rb`), so no further changes needed as far as i know
## Test Results (before it threw an unrelated error regarding my Ruby 3.3.0)
Screenshot 2024-11-10 at 15 32 06

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Bug Fixes
 
-* None.  
+* Update ruby-macho to 4.1.0 to address new mergable libraries not beind detected correctly.  
+    [Parsa Nasirimehr](https://github.com/TheRogue76)
+    [#12691](https://github.com/CocoaPods/CocoaPods/pull/12691)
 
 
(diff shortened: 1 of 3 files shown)


amorde (approved):
Thank you! Could you please add an entry to `CHANGELOG.md`?

TheRogue76:
> Thank you! Could you please add an entry to `CHANGELOG.md`?

Will do

TheRogue76:
Added change log and resolved the conflict. @amorde is the place i put the changelog good or should i be putting it elsewhere?

nicolasrostan:
@amorde hey, this is blocking us and I'd like to avoid using master, could you release a patch?

PR #12706, opened 2024-11-22
Introducing `--no-lint` option for `pod repo push` command

gmazzo:
Adds `--no-lint` option to `pod repo push` command.

This flag will skip `validate_podspec_files` method from being invoked as part of the `Push` command.

### Use case
To enable publishing an iOS Pod from non-Mac environments like UNIX systems.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,9 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
 
 ##### Enhancements
 
-* None.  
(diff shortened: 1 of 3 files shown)


amorde (approved):

gmazzo:
2 years later 😂, but thanks. We have moved to SPM a month ago

amorde:
Yeah, apologies, we aren't really maintaining this anymore. I just happen to be looking on a Sunday afternoon 😅 thank you regardless!

PR #12899, opened 2026-04-02
fix: remove implicit current-directory search from resolution

snoopysecurity:
## Summary
`Pod::Executable.which` previously prepended `./` to its search paths, causing binaries in the current working directory (for example, `./git`) to be resolved before trusted system binaries. This PR fixes this issue in executable resolution that could allow local executable hijacking when running CocoaPods commands in an untrusted directory.

## Root Cause
- `which` forced current directory precedence with `paths.unshift('./')`.
- Empty/unsafe `PATH` entries could also implicitly include current directory.

## What Changed
- Removed implicit current-directory search from `PATH` resolution.
- Added explicit path handling:
  - If the requested program contains a path separator (for example `./tool`, `../tool`, `/usr/bin/tool`), CocoaPods resolves that exact path.
- Added defensive filtering of unsafe `PATH` entries:
  - Skip `''` and `'.'`.

## Impact
- Prevents accidental execution of attacker-controlled binaries from project directories.
- Preserves intentional execution of explicit local/absolute executables.

## Behavioral Notes
- `which('git')` now follows `PATH` without forced `./` precedence.
- `which('./git')` and absolute-path invocations continue to work.
- Windows `.exe` behavior is preserved.

## Tests Added
- `does not search the current working directory before PATH locations`
- `supports explicit relative executable paths`
- `ignores unsafe PATH entries for current directory`

## Validation
- `bundle exec ruby -I spec spec/unit/executable_spec.rb`
- Result: `16 specifications (30 requirements), 0 failures, 0 errors`

## Files Changed
- `lib/cocoapods/executable.rb`
- `spec/unit/executable_spec.rb`
- `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,8 @@ To install release candidates run `[sudo] gem install cocoapods --pre`
     [#12691](https://github.com/CocoaPods/CocoaPods/pull/12691)
 
 * Harden executable lookup in `Pod::Executable.which` by removing implicit current-directory search, honoring explicit path executables only, and skipping unsafe PATH entries (`''` and `.`) to prevent local executable hijacking.  
-  [CocoaPods Team](https://github.com/CocoaPods)
-  [#TBD](https://github.com/CocoaPods/CocoaPods/pull/TBD)
+  [Sam Sanoop](https://github.com/snoopysecurity)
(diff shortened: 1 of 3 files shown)


snoopysecurity:
@amorde any chance you could take a look at this PR? thank you

